PR #601, opened 2020-12-08
[WIP] Fix GitHub workflows

plant99:


plant99:
@ReimarBauer any idea how I'd run them as workflows without merging this PR with `develop`? since `develop` branch has 
```
on: [push]
```

ReimarBauer:
> @ReimarBauer any idea how I'd run them as workflows without merging this PR with `develop`? since `develop` branch has
> 
> ```
> on: [push]
> ```

 we should have on: pull_request so that after the action works we can see the result on the new merge request. (That will work than with new PR, but not with the action itselfs)

https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows

I think about creating a action test repo with just a hello world example, after it works there we can move the action.

ReimarBauer:
The borg ci code had the missing pieces we can do it like this

https://github.com/ReimarBauer/example/blob/main/.github/workflows/python-app.yml

PR #614, opened 2021-01-07
Add tool for roundtrip on topview

### LINKED ISSUE #610: navigation feature in topview, return to start point
We may be should add a button on topview to automaticly connect the last point with the origin point.
We should not run in the situation a user has to drop a point on an other.

detail_zoom

Marilyth:
Adds a text button, which can be used to connect the last and first waypoint, essentially creating a roundtrip, which was requested here: Open-MSS/MSS#610

A fitting icon should probably be added instead of the text.
Testing ran fine.


mslib/msui/_tests/test_topview.py
```diff
@@ -203,6 +203,32 @@ def test_move_point(self, mockbox):
...
 
+    @mock.patch("mslib.msui.mss_qt.QtWidgets.QMessageBox")
+    def test_connect_point(self, mockbox):
+        """
+        Test connecting the last and first point
+        Test connecting the first point to itself
+        """
+        count = len(self.window.waypoints_model.waypoints)
+
+        #Test if the last waypoint connects to the first
```
ReimarBauer: I'm currently not happy with the flake8 checker, usually there should be a a blank after a comment sign

https://www.python.org/dev/peps/pep-0008/#comments

mslib/msui/mpl_qtwidget.py
```diff
@@ -335,6 +339,27 @@ def delete_wp(self, *args):
...
 
+    def connect_wp(self, *args):
+        """
+        activate connect_wp tool, which copies the first waypoint and inserts it at the back of the list again
+        essentially creating a roundtrip
+        """
+        wpm = self.canvas.waypoints_interactor.waypoints_model
+
+        #Don't roundtrip with less than 2 waypoints
+        if len(wpm.all_waypoint_data()) < 2:
```
ReimarBauer: we may be should disable the button if we have less waypoints

mslib/msui/mpl_qtwidget.py
```diff
@@ -335,6 +339,27 @@ def delete_wp(self, *args):
...
+            return
+
+        #Also don't roundtrip if the first and last are the same position
+        first_waypoint = wpm.waypoint_data(0)
+        last_waypoint = wpm.waypoint_data(wpm.rowCount() - 1)
+        if first_waypoint.lat == last_waypoint.lat and first_waypoint.lon == last_waypoint.lon:
```
ReimarBauer: also here if this condition is given the button could be disabled

joernu76:
Hi May, great change request!

Still, I would move the Button to the TableView, where we have collected "similar" features that manipulate the flight path (e.g. a "reverse" fligthpath button. I think it is a bit out-of-place in the toolbar (which due to API changes in matplotlib is causing us troubles often; but we must place "modes" there that change how mouse-clicks  interact with the map).

Marilyth:
One part of me is unhappy with the redundancy of adding the same feature to two views, another part of me doesn't want to mess around with the views on a higher level though.

On another note, should SideView also have this feature?

joernu76:
Mmmh. The problems become ever more apparent. Collecting it at one point (The TableView(!)) would still be the better solution, especially as some of the flightpath tools in TableView require a selected row.

I would not put it into the SideView; no one actually uses this for moving points.

mslib/msui/qt5/ui_tableview_window.py
```diff
@@ -88,6 +91,17 @@ def retranslateUi(self, TableViewWindow):
         self.btCloneWaypoint.setText(_translate("TableViewWindow", "clone"))
         self.btDeleteWayPoint.setText(_translate("TableViewWindow", "delete selected"))
         self.btInvertDirection.setText(_translate("TableViewWindow", "reverse"))
+        self.btRoundtrip.setText(_translate("TableViewWindow", "make roundtrip"))
         self.actionCloseWindow.setText(_translate("TableViewWindow", "Close Window"))
         self.actionCloseWindow.setShortcut(_translate("TableViewWindow", "Ctrl+W"))
 
+
```
ReimarBauer: if this part is not needed please remove it. I guess it is manually added?

mslib/msui/qt5/ui_topview_window.py
```diff
@@ -104,3 +110,13 @@ def retranslateUi(self, TopViewWindow):
         self.actionCloseWindow.setShortcut(_translate("TopViewWindow", "Ctrl+W"))
 
 from mslib.msui.mpl_qtwidget import MplTopViewWidget
+
```
ReimarBauer: please remove this

mslib/msui/topview.py
```diff
@@ -312,3 +320,37 @@ def load_settings(self):
...
+            first_waypoint = self.waypoints_model.waypoint_data(0)
+            last_waypoint = self.waypoints_model.waypoint_data(self.waypoints_model.rowCount() - 1)
+
+            condition = first_waypoint.lat != last_waypoint.lat or first_waypoint.lon != last_waypoint.lon or \
+                        first_waypoint.flightlevel != last_waypoint.flightlevel
+
+        return condition
+
+    def update_roundtrip_enabled(self):
+        self.btRoundtrip.setEnabled(self.is_roundtrip_possible())
```
ReimarBauer: newline missing

mslib/msui/topview.py
```diff
@@ -312,3 +320,37 @@ def load_settings(self):
...
+        """
+        condition = self.waypoints_model.rowCount() > 1
+
+        if condition:
+            first_waypoint = self.waypoints_model.waypoint_data(0)
+            last_waypoint = self.waypoints_model.waypoint_data(self.waypoints_model.rowCount() - 1)
+
+            condition = first_waypoint.lat != last_waypoint.lat or first_waypoint.lon != last_waypoint.lon or \
+                        first_waypoint.flightlevel != last_waypoint.flightlevel
+
```
ReimarBauer, line 352: please use blank lines only if that improves readability. ~7% of our codebase are empty lines


https://www.openhub.net/p/mss/analyses/latest/languages_summary

Marilyth: Oh. I was actually taught highly to always group my code into logical segments to make understanding it easier on first sight.
I guess I overdid in this case.

You mention this specifically on this line, is it only an issue here, or are you unhappy with my blank lines in general?

mslib/msui/qt5/ui_tableview_window.py
joernu76: May, this may not be apparent. These files in the qt5 directory are "automatically" generated from the ui files in the "ui" directory. We use the qt-creator thingie to modify the .ui files and use "pyuic5" to convert the .ui file to a .py file (e.g. `pyuic5 mslib/msui/ui/ui_mainwindow.ui > mslib/msui/qt5/ui_mainwindow.py`)

joernu76: It might be sensible to actually generate them automatically with python setup.py build, but we are not doing that, currently.

joernu76: If you did use the creator, you should also add the changes to the ui files to the merge request.

mslib/msui/topview.py
joernu76: I see nothing wrong with your code. It is pep8 compliant and I maybe would have used maybe an empty line less than you. I agree with `please use blank lines only if that improves readability.` but I think that applies here. To some degree it is a matter of taste obviously.
We might talk about style on next Thursday to come to a common point of view.

mslib/msui/qt5/ui_tableview_window.py
Marilyth: These were created via pyuic5, I seem to have accidentally added an option which generated more code than necessary.
The changed ui files are in this commit as well, already.

mslib/msui/topview.py
ReimarBauer: Hi, it was not meant only for that location. It is more in general, we are below 10%. That's good but in all >5K empty lines sounds much. If we compare with XML code, there is not any empty line needed. Usually many empty lines means better readablity is done by using a helper function/method. Everything is still a kind of balance. It was a hint by me too.

ReimarBauer:
> One part of me is unhappy with the redundancy of adding the same feature to two views, another part of me doesn't want to mess around with the views on a higher level though.

Also the hexagon pattern may be used on the top view. Currently it is only vailable for tableview in the pull down "select to open control". 
If we have for example bounderies of a no go flight area overlayed on topview I think it is easier to add it on that view. 
In top view from "inkscape influenced" I would want to drag, move, resize it.
We should identify tools we want to have on all views and group them afterwards. 

Further thoughts on that:
There can be much more helper if we know to which base we have to return. 
Based on performance data we may be could show a range on the last way point how far we can go based on max fuel. 
A feature for mscolab projects could be to initialize the aircraft performance data too by selecting the aircraft for the flightpath.
If the data is always available we could use it for sinking, climbing, curves.



> 
> On another note, should SideView also have this feature?

PR #618, opened 2021-01-07
Replace mss_qt with PyQt5 where possible

### LINKED ISSUE #575: refactor  mss_qt move module imports to modules
**[Original report](https://bitbucket.org/wxmetvis/mss/issue/575) by Reimar Bauer (Bitbucket: [ReimarBauer](https://bitbucket.org/ReimarBauer), GitHub: [ReimarBauer](https://github.com/ReimarBauer)).**

We introdruced mss\_qt because of a short time we had pyqt4 and pyqt5 together.

The module import should become moved back where they are used, so that we as developer get our IDEs to use all help of autocompletion or documentation.

[https://bitbucket.org/wxmetvis/mss/src/develop/mslib/msui/mss\_qt.py#lines-138](https://bitbucket.org/wxmetvis/mss/src/develop/mslib/msui/mss_qt.py#lines-138)

Marilyth:
Import all components of PyQt from PyQt5 instead of mss_qt, fixes Open-MSS/MSS#575

### DIFF
--- a/mslib/msui/_tests/test_editor.py
+++ b/mslib/msui/_tests/test_editor.py
@@ -28,7 +28,7 @@
 import mock
 import os
 import sys
-from mslib.msui.mss_qt import QtWidgets
+from PyQt5 import QtWidgets
 from mslib.msui import editor
 from mslib._tests.constants import ROOT_DIR
 
@@ -39,7 +39,7 @@ class Test_Editor(object):
 
     save_file_name = os.path.join(ROOT_DIR, "testeditor_save.json")
 
-    @mock.patch("mslib.msui.mss_qt.QtWidgets.QMessageBox.warning", return_value=QtWidgets.QMessageBox.Yes)
+    @mock.patch("PyQt5.QtWidgets.QMessageBox.warning", return_value=QtWidgets.QMessageBox.Yes)
     def setup(self, mockmessage):
         self.application = QtWidgets.QApplication(sys.argv)
 
--- a/mslib/msui/kmloverlay_dockwidget.py
+++ b/mslib/msui/kmloverlay_dockwidget.py
@@ -32,9 +32,10 @@ from lxml import etree as et, objectify
 import os
 from matplotlib import patheffects
 
-from mslib.msui.mss_qt import QtGui, QtWidgets, QtCore, get_open_filenames, get_save_filename
+from mslib.msui.mss_qt import get_open_filenames, get_save_filename
 from mslib.msui.mss_qt import ui_kmloverlay_dockwidget as ui
 from mslib.msui.mss_qt import ui_customize_kml
+from PyQt5 import QtGui, QtWidgets, QtCore
 from mslib.utils import save_settings_qsettings, load_settings_qsettings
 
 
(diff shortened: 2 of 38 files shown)


mslib/msui/kmloverlay_dockwidget.py
```diff
@@ -32,9 +32,10 @@
 import os
 from matplotlib import patheffects
 
-from mslib.msui.mss_qt import QtGui, QtWidgets, QtCore, get_open_filenames, get_save_filename
+from mslib.msui.mss_qt import get_open_filenames, get_save_filename
 from mslib.msui.mss_qt import ui_kmloverlay_dockwidget as ui
 from mslib.msui.mss_qt import ui_customize_kml
+from PyQt5 import QtGui, QtWidgets, QtCore
```
joernu76, line 38: Stylistically imports should be ordered from standard libraries/foreign libraries/our library

joernu76:
In mss_qt.py, there are some imports that are not necessary anymore, i.e.
mslib/msui/mss_qt.py:36:1: F401 'PyQt5.QtGui' imported but unused
mslib/msui/mss_qt.py:36:1: F401 'PyQt5.QtTest' imported but unused
mslib/msui/mss_qt.py:36:1: F401 'PyQt5.Qt' imported but unused

mslib/msui/_tests/test_editor.py
```diff
@@ -28,7 +28,7 @@
 import mock
```
ReimarBauer, line 28: I do miss the general comment field as on bitbucket

please also look into the mss_qt.py for unused imports, see pylint or pycharm

mslib/msui/kmloverlay_dockwidget.py
Marilyth: @joernu76 I could run Pycharms "Optimize Imports" on the MSS project. 
This will automatically group them by standard/foreign/local libraries, which seems to have been done incorrectly in a lot of files actually.

Is this something you want me to do?

joernu76: Yes and no. Doing this manually sucks (and I normally only ask to observe this when doing changes oneself, not fixing it everywhere where it isn't the case), but I have yet to find a tool that doesn't sometime cause more trouble than its worth.
Certainly do not use this tool here. The order of imports is also not that important that I wouldn't merge it (there are some special cases where one cannot follow this rule, for which reason I also disable the corresponding flake8 check often).

Also autoformatting is something that works only 95% of the time, even though I'd like to like, e.g. 'black'.

But you could prepare a separate pull request, where we could review the proposed changes. I haven't used PyCharmed in two years, so they might have gotten smarter... Largely it should be fine, currently, but a tool will probably do a LOT of resorting.

joernu76 (approved):

joernu76:
Fine for me now. Do you want to merge Reimar?

mslib/msui/kmloverlay_dockwidget.py
ReimarBauer: The cleanup of order for imports on other modules should be an refactoring issue. It is not so important yet. I would prefer a better coverage first, 85 - 90%.

joernu76: Obviously, it is rather unimportant, but probably *much* easier to do (if the PyCharmed tool does it).

Issue #619, opened 2021-01-08
set up github actions for building/testing/containers


joernu76:
I think as we have the means, the next, most important step would be to set up the github actions to automatically execute and check the syntax and test cases. Second would be to visualize coverage using the Pages feature. 
It is very wasteful that we have not automated that.

I propose this course of action in this order

1) set up an "action" to install a conda environment and execute tests in a debian (or similar) based docker container. (I do not use the conda build mechanism, normally; typically installing the proper pacakges and executing pytest should work).

2) create a container with everything neede preinstalled, and use this one instead of creating a new environment everytime

3) Replicate the conda build process, install the package and test this one.

4)  build a container based on 2) including mss

2/4 would be new projects within the Open-MSS group.

What do you think? @ReimarBauer @Marilyth

joernu76:
Ancillarily, 2) allows us to run tests against a known-good environment as well as running (e.g. weekly) against a container employing the most recent packages, which will allow us to determine quickly if new packages break MSS.
I have a similar setup for juwave, which could serve as a pattern.

Marilyth:
I didn't yet get to learn much about GitHub actions and how it works with containers.
I will play around with it during the weekend and see if I can make an informed opinion afterwards.

joernu76:
Perfect. Maybe Reimar can chime in with some ideas of his. But this should be prioritized IMO, because having that set up properly will save us quite a lot of work. I have become so used to that from my other projects that checking manually MSS tests fells like walking through a quagmire.

ReimarBauer:
I think we should manage the installation for the tests by a 
`mamba install --file requiremtents.d/base.txt` and an additional `mamba install --file requiremtents.d/development.txt` here.
This is related to https://github.com/Open-MSS/MSS/issues/612 
On future gsoc projects I do prefer in this namespace an additional requirements file, which we can remove after merge. 

I have a container on the dockerhub, but I would prefer at least a weekly rebuild of the one used for tests. Because of the good idea not to fixate to much packages, we should quite fast see that we can't use "the newest" package of something. 

@plant99 wanted to look at a containerd definition https://github.com/Open-MSS/MSS/issues/600 

I'm not sure about the effort to create the package same way as on conda-forge, I think that this is too much.

Marilyth:
GitHub runners, as far as I am aware, just run on a console environment.
How would we go about testing the UI segments of MSS? The tests containing UI cause pytest to abort on the GitHub runners.
[...]

PR #623, opened 2021-01-12
Add simple testing action

Marilyth:
Tests the code on each push and pull request. This partly covers what was talked about in this issue Open-MSS/MSS#619
If any test fails, this action will fail. This also doesn't cover flake8 as that is covered in a parallel other action.

Building the environment takes about 1:30 to 2 minutes, testing the code takes about 14 minutes.
So in total this action takes around 16 minutes to complete.
The environment building time might be slightly reduced by using docker images.

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -1,0 +1,35 @@
+name: Pytest MSS
+
+on: [ push, pull_request ]
+
+jobs:
+  Test-MSS:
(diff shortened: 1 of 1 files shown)


Marilyth:
How interesting. test_mscolab.py failed here, but didn't in my fork.

I just ran it again, and it didn't fail this time.. I will just brush it under the "odd memory things" carpet for now.

PR #629, opened 2021-01-22
Testing via docker

### LINKED ISSUE #619: set up github actions for building/testing/containers

Marilyth:
This will replace the current action used for testing. fixes Open-MSS/MSS#619

This action 
1. pulls and uses the latest image at yourmss/mss-test
2. updates dependencies if the push or pull request changed the meta.yaml
3. tests the push or pull request
4. requests the docker images to be updated in case of success of a push

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -1,35 +1,51 @@
 name: Pytest MSS
 
 on: [ push, pull_request ]
+env:
+  PAT: ${{ secrets.PAT }}
 
 jobs:
   Test-MSS:
     runs-on: ubuntu-latest
-    strategy:
-      max-parallel: 5
+    
+    defaults:
+      run:
+        shell: bash
+
+    container:
+      image: yourmss/mss-test:latest
 
     steps:
     - uses: actions/checkout@v2
-
-    - name: Add conda to system path
+        
+    - name: Update dependencies if changed
       run: |
-        # $CONDA is an environment variable pointing to the root of the miniconda directory, action
-        echo $CONDA/bin >> $GITHUB_PATH
+        cmp -s /meta.yaml localbuild/meta.yaml || (echo Dependencies differ \
+        && cat localbuild/meta.yaml \
+        | sed -n '/^requirements:/,/^test:/p' \
+        | sed -e "s/.*- //" \
+        | sed -e "s/menuinst.*//" \
+        | sed -e "s/.*://" > reqs.txt \
+        && conda install -n mssenv --file reqs.txt --force-reinstall)
 
-    - name: Setup Conda
+    - name: Run tests
+      timeout-minutes: 25
       run: |
-        conda config --add channels conda-forge
-        conda config --add channels defaults
-        conda create -n mssenv python=3
-
-    - name: Install MSS
-      run: |
-        source ${CONDA}/bin/activate mssenv
-        conda install mamba
-        mamba install --only-deps mss
-        conda install --file requirements.d/development.txt
-
-    - name: Test with pytest
-      run: |
-        source ${CONDA}/bin/activate mssenv
+        source /opt/conda/bin/activate mssenv
+        cd $GITHUB_WORKSPACE
         xvfb-run pytest mslib
+    
+  Update-Images:
+    runs-on: ubuntu-latest
+    if: github.event_name == 'push'
+    needs: [Test-MSS]
+    
+    steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/testing.yml
```diff
@@ -1,35 +1,51 @@
...
-        conda config --add channels conda-forge
-        conda config --add channels defaults
-        conda create -n mssenv python=3
+        cmp -s /meta.yaml localbuild/meta.yaml || (echo Dependencies differ \
+        && cat localbuild/meta.yaml \
+        | sed -n '/^requirements:/,/^test:/p' \
+        | sed -e "s/.*- //" \
+        | sed -e "s/menuinst.*//" \
+        | sed -e "s/.*://" > reqs.txt \
+        && conda install -n mssenv --file reqs.txt --force-reinstall)
```
joernu76, line 29: Mmh. This is nifty, but I am not sure I like it. In case of changed dependencies I'd expect us to necessarily and potentially manually update the docker image before we can proceed.
It is typically only necessary that current builds work within the image. At least for juwave we so far have only added stuff to it and try that the code works with both the current stable and unstable image.

.github/workflows/testing.yml
```diff
@@ -1,35 +1,51 @@
 name: Pytest MSS
 
 on: [ push, pull_request ]
+env:
+  PAT: ${{ secrets.PAT }}
 
 jobs:
   Test-MSS:
     runs-on: ubuntu-latest
```
joernu76, line 9: what is the meaning of this? Are we not using our own container? Is this the environment within which our docker is executed?

.github/workflows/testing.yml
```diff
@@ -1,35 +1,51 @@
...
-        conda install --file requirements.d/development.txt
-
-    - name: Test with pytest
-      run: |
-        source ${CONDA}/bin/activate mssenv
+        source /opt/conda/bin/activate mssenv
+        cd $GITHUB_WORKSPACE
         xvfb-run pytest mslib
+    
+  Update-Images:
```
joernu76, line 38: No idea whats going on here...

joernu76 (approved):
Certainly a step forward. I do not fully understand everything though. But that is not necessary at the current point in time.

.github/workflows/testing.yml
Marilyth: We need an environment first, before we can execute a docker container in it.
So your second assumption is correct.

.github/workflows/testing.yml
Marilyth: This part is triggering this action here to run:
https://github.com/Open-MSS/dockerhub/blob/stable/.github/workflows/update-images.yml

That action on the other hand triggers 3 other actions to run, which update the corresponding mss, mss-test and mss-dev images.
It would probably make sense to only update the necessary images, but for now this is easier and shorter.

.github/workflows/testing.yml
ReimarBauer: this enables to see that an ongoing build based on the meta.yaml rules does not give a working state. We need to do a fresh build to figure that there are newer packages with changed dependencies for us.

Marilyth: I am also not sure I like it.

This is automating something that might be better off done by hand, yes.
However having something done automatically is always easier, and if it somehow breaks stuff, nothing happens to the stable images anyway.

joernu76: Again, I am not against merging this now. Let's how it works in practice.

ReimarBauer: We are not alone with this idea.
https://github.com/conda/conda/issues/5126

When it makes trouble we can look if always build alpha is an option and to install also from that  --deps-only

ReimarBauer (approved):
looks good

PR #647, opened 2021-01-27
Added fix for #642

### LINKED ISSUE #640: fix typo of word "collaboration" in documents
We should use the word "collaboration" instead of "colaboration" in all descriptions.

We keep the word mscolab for the module etc.

### LINKED ISSUE #642: only if a mss_settings.json is defined it should be loaded
The last change to the editor introduced a mandatory mss_settings.json file which we don't have by default.

It should only be loaded when it exists. 

While we don't have automated tests on stable this was not shown up.

ahelii16:
Added to code to check first if file exists and load it only if it exists.


mslib/msui/editor.py
```diff
@@ -35,6 +35,7 @@
 from mslib.msui import constants
 from mslib.msui.constants import MSS_CONFIG_PATH
 from mslib.msui.icons import icons
+import os.path
```
ReimarBauer: please use fs (this is pyfilesystem2)
there are plenty examples in the code base.

By this we for example can use a webdav server like nextcloud/owncloud  and are not limited to local files

ReimarBauer:
please use fs instead of os, see comment on the import

ahelii16:
Okay I will do it.

ahelii16:
I have fixed it. Will keep in mind from next time :)

mslib/msui/editor.py
```diff
@@ -50,9 +50,14 @@ def __init__(self, parent=None):
         # Could also use a QTextEdit and set self.editor.setAcceptRichText(False)
         self.editor = QtWidgets.QPlainTextEdit()
 
-        # Load existing mss_settings.json
-        readMe = open(MSS_CONFIG_PATH + "/mss_settings.json", 'r').read()
-        self.editor.insertPlainText(readMe)
+        # Load mss_settings.json (if already exists)
+        self.mss_path = fs.path.join(MSS_CONFIG_PATH, "mss_settings" + ".json")
+        if self.mss_path is not None:
```
Marilyth: This if will never be false. self.mss_path is just the path name joined together.
You are thus still running into the issue you tried to fix with this PR.

Marilyth (changes requested):

mslib/msui/editor.py
ahelii16: Okay I see, thanks. Will change it.

ReimarBauer: pyfilesystems main idea is to unify all these file operations independent the kind of "storage" 
The API is excatly the same if you open a container or a file or a network storage, some more infos 
https://reposhub.com/python/miscellaneous/PyFilesystem-pyfilesystem2.html

I guess this is the missing part ;)

>>> _test_dir = fs.open_fs(MSS_CONFIG_PATH)
>>> _test_dir.exists('mss_settings.json')

on normal operations we import constants from mslib.msui but for tests we do from mslib._tests.

That makes than for tests the changes needed to have an independent path.

ahelii16: Yes I did that in the latest commit :)

mslib/msui/editor.py
```diff
@@ -215,12 +220,12 @@ def edit_toggle_wrap(self):
...
+            self, self.tr("Save changes to mss_settings.json?"),
             self.tr("If you changed the mss_settings.json please restart the gui"),
             QtWidgets.QMessageBox.Save | QtWidgets.QMessageBox.Ignore, QtWidgets.QMessageBox.Save)
 
         if ret == QtWidgets.QMessageBox.Save:
             self.file_save()
             event.accept()
         else:
-            event.accept()
+            event.accept()
```
open-mss-build: @Ahelii16 the linter when it is finished will find this newline thing.

Marilyth (approved):

PR #648, opened 2021-01-27
Make MSS support 1.3.0

### LINKED ISSUE #201: refactor OGC WMS from 1.1.1 to 1.3.0
**[Original report](https://bitbucket.org/wxmetvis/mss/issue/201) by Reimar Bauer (Bitbucket: [ReimarBauer](https://bitbucket.org/ReimarBauer), GitHub: [ReimarBauer](https://github.com/ReimarBauer)).**

Version 1.1.1 is deprecated

http://www.opengeospatial.org/standards/wms

We have to move on to 1.3.0

Marilyth:
This is a bit of a handful, so I will try and summarise what I have done. If done correctly, this fixes #201 

For reference, this is a list of all differences between 1.1.1 and 1.3.0 I ran into, and tackled with this PR
- SRS is called CRS in 1.3.0
- lat/lon Axis instead of lon/lat for EPSG
- x, y are now called i, j in getfeatureinfo (I think this is irrelevant for mss)
- application/vnd.ogc.se_xml is deprecated, XML is the new default
- application/vnd.ogc.wms_xml is deprecated, text/xml is the new default
- LatLonBoundingBox is deprecated, use EX_GeographicBoundingBox instead
- WMT_MS_Capabilities is now called WMS_Capabilities
- the capabilities XML is now required to have namespaces
- Extents are now within dimensions in the XML
- The XML can not contain a DocType segment anymore

And as a summary of what I have done, or still need to do:
- [x] Make Server support 1.3.0
- [x] Make Server backwards compatible with 1.1.1
- [x] Make Client support 1.3.0
- [x] Make Client backwards compatible with 1.1.1
- [x] Make Client pick the default version of the Server, or 1.3.0 if that version is not supported
- [x] Handle namespaces in requested GetCapabilities documents to not fail 1.3.0
- [x] Rewrite tests to make sure 1.3.0 produces images and a working capabilities xml
- [x] Update ogcwms.py from owslib
- [ ] Incorporate Conformance Test Suite
- [x] Test if Client works with 2.0.0 Server (manually it does)

There are a few things I could not test properly, or am unsure if they are a good approach.
I will go ahead and mention them here for your input.

### DIFF
--- a/mslib/msui/mss_pyui.py
+++ b/mslib/msui/mss_pyui.py
@@ -243,7 +243,7 @@ class MSSMainWindow(QtWidgets.QMainWindow, ui.Ui_MSSMainWindow):
                 request = requests.get(base_url)
                 if pdlg.wasCanceled():
                     break
-                wms = wms_control.MSSWebMapService(request.url, version='1.1.1',
+                wms = wms_control.MSSWebMapService(request.url, version='1.3.0',
                                                    username=username, password=password)
                 wms_control.WMS_SERVICE_CACHE[wms.url] = wms
                 logging.info("Stored WMS info for '%s'", wms.url)
--- a/mslib/msui/mss_pyui.py
+++ b/mslib/msui/mss_pyui.py
@@ -170,7 +170,7 @@ class MSSMainWindow(QtWidgets.QMainWindow, ui.Ui_MSSMainWindow):
         # Reference to the flight track that is currently displayed in the
         # views.
         self.active_flight_track = None
-        self.last_save_directory = config_loader(dataset="data_dir", default=mss_default.data_dir)
+        self.last_save_directory = config_loader(dataset="data_dir")
         self.mscolab_window = None
         self.config_editor = None
 
@@ -206,13 +206,13 @@ class MSSMainWindow(QtWidgets.QMainWindow, ui.Ui_MSSMainWindow):
         # Views.
         self.listViews.itemActivated.connect(self.activate_sub_window)
 
-        self.add_import_filter("CSV", "csv", load_from_csv, pickertag="filepicker_flightrack")
-        self.add_export_filter("CSV", "csv", save_to_csv, pickertag="filepicker_flightrack")
+        self.add_import_filter("CSV", "csv", load_from_csv, pickertag="filepicker_default")
+        self.add_export_filter("CSV", "csv", save_to_csv, pickertag="filepicker_default")
 
         self._imported_plugins, self._exported_plugins = {}, {}
         self.add_plugins()
 
-        preload_urls = config_loader(dataset="WMS_preload", default=[])
+        preload_urls = config_loader(dataset="WMS_preload")
         self.preload_wms(preload_urls)
 
         # Status Bar
@@ -234,7 +234,7 @@ class MSSMainWindow(QtWidgets.QMainWindow, ui.Ui_MSSMainWindow):
             pdlg.setValue(i)
             QtWidgets.QApplication.processEvents()
             # initialize login cache from config file, but do not overwrite existing keys
-            for key, value in config_loader(dataset="WMS_login", default={}).items():
+            for key, value in config_loader(dataset="WMS_login").items():
                 if key not in constants.WMS_LOGIN_CACHE:
                     constants.WMS_LOGIN_CACHE[key] = value
             username, password = constants.WMS_LOGIN_CACHE.get(base_url, (None, None))
@@ -257,9 +257,9 @@ class MSSMainWindow(QtWidgets.QMainWindow, ui.Ui_MSSMainWindow):
 
     def add_plugins(self):
         picker_default = config_loader(
-            dataset="filepicker_default", default=mss_default.filepicker_default)
+            dataset="filepicker_default")
 
-        self._imported_plugins = config_loader(dataset="import_plugins", default={})
+        self._imported_plugins = config_loader(dataset="import_plugins")
         for name in self._imported_plugins:
             extension, module, function = self._imported_plugins[name][:3]
             picker_type = picker_default
@@ -286,7 +286,7 @@ class MSSMainWindow(QtWidgets.QMainWindow, ui.Ui_MSSMainWindow):
                         self._imported_plugins, type(ex), ex)))
                 continue
 
-        self._exported_plugins = config_loader(dataset="export_plugins", default={})
+        self._exported_plugins = config_loader(dataset="export_plugins")
         for name in self._exported_plugins:
             extension, module, function = self._exported_plugins[name][:3]
             picker_type = picker_default
@@ -426,7 +426,7 @@ class MSSMainWindow(QtWidgets.QMainWindow, ui.Ui_MSSMainWindow):
            a new instance of the view and adds a QActiveViewsListWidgetItem to
            the list of open views (self.listViews).
         """
-        layout = config_loader(dataset="layout", default=mss_default.layout)
+        layout = config_loader(dataset="layout")
         view_window = None
         if self.sender() == self.actionTopView:
             # Top view.
@@ -496,9 +496,8 @@ class MSSMainWindow(QtWidgets.QMainWindow, ui.Ui_MSSMainWindow):
         """
         if template is None:
             template = []
-            waypoints = config_loader(dataset="new_flighttrack_template", default=mss_default.new_flighttrack_template)
-            default_flightlevel = config_loader(dataset="new_flighttrack_flightlevel",
-                                                default=mss_default.new_flighttrack_flightlevel)
+            waypoints = config_loader(dataset="new_flighttrack_template")
+            default_flightlevel = config_loader(dataset="new_flighttrack_flightlevel")
             for wp in waypoints:
                 template.append(ft.Waypoint(flightlevel=default_flightlevel, location=wp))
             if len(template) < 2:
@@ -540,7 +539,7 @@ class MSSMainWindow(QtWidgets.QMainWindow, ui.Ui_MSSMainWindow):
         """
         filename = get_open_filename(
             self, "Open Flight Track", self.last_save_directory, "Flight Track Files (*.ftml)",
-            pickertag="filepicker_flightrack")
+            pickertag="filepicker_default")
         if filename is not None:
             try:
                 if filename.endswith('.ftml'):
@@ -604,7 +603,7 @@ class MSSMainWindow(QtWidgets.QMainWindow, ui.Ui_MSSMainWindow):
         """
         default_filename = os.path.join(self.last_save_directory, self.active_flight_track.name + ".ftml")
         filename = get_save_filename(
-            self, "Save Flight Track", default_filename, "Flight Track (*.ftml)", pickertag="filepicker_flightrack")
+            self, "Save Flight Track", default_filename, "Flight Track (*.ftml)", pickertag="filepicker_default")
         logging.debug("filename : '%s'", filename)
         if filename:
             self.last_save_directory = fs.path.dirname(filename)
--- a/mslib/msui/mss_pyui.py
+++ b/mslib/msui/mss_pyui.py
@@ -44,8 +44,6 @@ import sys
 import types
 import fs
 
-from owslib.map.common import WMSCapabilitiesReader
-
 from mslib import __version__
 from mslib.msui.mss_qt import ui_mainwindow as ui
 from mslib.msui.mss_qt import ui_about_dialog as ui_ab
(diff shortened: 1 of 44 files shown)


mslib/msui/wms_control.py
```diff
@@ -660,9 +664,15 @@ def get_capabilities(self):
         else:
             # url shortener url translated
             url = request.url
-            url = url.replace('?service=WMS&request=GetCapabilities&version=1.1.1', '')
+
+            # Take the default version of the server, or 1.3.0 if not supported
+            version = request.text.split("MS_Capabilities version=\"")[-1].split("\"")[0]
+            if version not in ["1.1.1", "1.3.0"]:
+                version = "1.3.0"
```
Marilyth: Instead of hardcoding the version for the client, the client scrapes the version from the default GetCapabilities document of the server.
Is this a good or a bad approach?

mslib/ogcwms.py
```diff
@@ -621,3 +641,9 @@ def readString(self, st):
         if not isinstance(st, str):
             raise ValueError("String must be of type string, not %s" % type(st))
         return etree.fromstring(st)
+
+
+def removeXMLNamespace(tree):
+    for elem in tree.iter():
+        if elem.tag.startswith("{"):
+            elem.tag = elem.tag.split("}")[-1]
```
Marilyth, line 390: I ran into an issue when testing a third party OGC WMS Server with MSS.
Namely when requesting their GetCapabilities, the elements weren't just called "Capabilities" but "{Namespace}Capabilities" causing MSS to fail finding the element.
By calling this function, the {Namespace} part is first removed before searching in it.
Is this a good or a bad thing?

mslib/retriever.py
```diff
@@ -246,9 +246,9 @@ def main():
...
                 url, auth=tuple(config["WMS_login"][url]),
-                params={"version": "1.1.1", "request": "GetMap", "format": "image/png",
-                        "exceptions": "application/vnd.ogc.se_xml",
-                        "srs": "VERT:LOGP", "layers": layer, "styles": style,
+                params={"version": "1.3.0", "request": "GetMap", "format": "image/png",
+                        "exceptions": "XML",
+                        "crs": "VERT:LOGP", "layers": layer, "styles": style,
                         "dim_init_time": init_time, "time": time,
                         "width": width, "height": height,
                         "path": ",".join(f"{wp[0]:.2f},{wp[1]:.2f}" for wp in wps),
```
Marilyth, line 254: I wasn't sure how to use the retriever, and ultimately unsure how to test it.
It would be very dandy if this class could somehow be integrated into the test suite, in order not to run into issues with it.
Anyway, this still needs to be tested and I don't know how, so I would love to hear your input.

Marilyth (commented):
These in particular are the things I am unsure about.
If you could give me your input, that would be great.

ReimarBauer:
until we merged https://github.com/Open-MSS/MSS/pull/647 please skip the test_editor, currently it stops there.

mslib/msui/tableview.py
```diff
@@ -230,8 +229,10 @@ def is_roundtrip_possible(self):
             first_waypoint = self.waypoints_model.waypoint_data(0)
             last_waypoint = self.waypoints_model.waypoint_data(self.waypoints_model.rowCount() - 1)
 
-            condition = first_waypoint.lat != last_waypoint.lat or first_waypoint.lon != last_waypoint.lon or \
```
joernu76: This seems to be unrelated to the 1.3.0 activity. Can this be separated into another merge request?

mslib/msui/wms_control.py
joernu76: The old approach seems very inflexible/wrong.
But the new one...
Isn't there a standard way to parse the URL for the parameters and use that?

mslib/msui/wms_control.py
```diff
@@ -660,9 +664,15 @@ def get_capabilities(self):
...
+
+            url = url.replace('?service=WMS&request=GetCapabilities', '')
```
joernu76: mmh.

mslib/mswms/wms.py
```diff
@@ -517,9 +526,10 @@ def application():
         server_url = urllib.parse.urljoin(url, urllib.parse.urlparse(url).path)
 
         if (request_type in ('getcapabilities', 'capabilities') and
-                request_service == 'wms' and request_version in ('1.1.1', '')):
-            return_data, return_format = server.get_capabilities(server_url)
-        elif request_type in ('getmap', 'getvsec') and request_version in ('1.1.1', ''):
+                request_service == 'wms' and request_version in ('1.1.1', '1.3.0', '')):
+            return_data, return_format = server.get_capabilities(server_url,
+                                                                 request_version if request_version != "" else "1.3.0")
```
joernu76: would it be fine to have 1.3.0 as default in request_version above?

mslib/ogcwms.py
```diff
@@ -54,7 +54,7 @@
 """
```
joernu76, line 54: This file has been copied form 
https://pypi.org/project/OWSLib/
version 0.14.
Would this be a good time to rebase to a more recent version? Maybe they support 1.3.0 in the meantime?

mslib/ogcwms.py
joernu76: That seems strange. Is this standard behaviour? What third party server was that?
Those that I tested for 1.1.1 didn't behave in that fashion (I discovered many problems in MSS, but this seems an unreasonable answer with {} in the reply?

Marilyth: I seem to have missed that part, but yes, namespaces, specifically {http://www.opengis.net/wms}, are now required per the standard in 1.3.0.
This is something I still have to add for the server then.

mslib/ogcwms.py
Marilyth: I will take a look at that.
From what I can see, they do support 1.3.0, but hardcoded in a seperate class from 1.1.1, so it is not backwards compatible out of the box.
Also all the modifications in ogcwms for MSS need to stay. So this is a bit of a fickle thing to rebase.

mslib/msui/tableview.py
Marilyth: Yes. I removed it from this PR

mslib/mswms/wms.py
Marilyth: Thinking more about it, I changed it to default to 1.1.1. Most other WMS Servers default to 1.1.1, too.
This will cause the least complications.

mslib/msui/wms_control.py
Marilyth: I changed it from scraping the response, to parsing the response and checking the version through that.
This should be more robust now.

mslib/ogcwms.py
Marilyth: The MSS Server now also contains namespaces for 1.3.0.
It is indeed required and not optional, I will quote from the 1.3.0 standard for getcapabilities:

> This  element  shall  contain   an   XML   Schema   instance   schemaLocation   attribute   that   binds   the   “http://www.opengis.net/wms” namespace  to  the  schema  in  E.1. 

I also went ahead and did the same for exceptions. While that isn't required per the document, their examples and other Servers still do that.

mslib/msui/wms_control.py
```diff
@@ -210,6 +215,7 @@ def getmap(self, layers=None, styles=None, srs=None, bbox=None,
         if hasattr(u, "info") and 'application/vnd.ogc.se_xml' in u.info()['Content-Type']:
             se_xml = u.read()
             se_tree = etree.fromstring(se_xml)
+            removeXMLNamespace(se_tree)
```
ReimarBauer, line 218: We should have a comment, explaining this remove. something like 
\# removed because we don't support this until it is mandatory 

I assume without removing this we get some errors?

mslib/retriever.py
ReimarBauer: @joernu76 please try this version.

mslib/msui/mss_pyui.py
```diff
@@ -243,7 +243,7 @@ def preload_wms(urls):
                 request = requests.get(base_url)
                 if pdlg.wasCanceled():
                     break
-                wms = wms_control.MSSWebMapService(request.url, version='1.1.1',
+                wms = wms_control.MSSWebMapService(request.url, version='1.3.0',
```
ReimarBauer: We have manually to verify that this version of msui can read a mss server based on MSS-2.0, old version '1.1.1'

We should have a tests for this.

ReimarBauer (changes requested):
if possible a test which shows that the mss_pyui still can read old servers, e.g. capabilties.

mslib/msui/mss_pyui.py
Marilyth: I will have to ponder on that. Not sure how this could be incorporated into a test yet.
But manually I can and already did verify the current ui works with an old stable server.

ReimarBauer: We could have a xml result for mocking the server request and feed this into the gui methods. The mock just replaces the request to the server and has a return_value with the V1.1.1 XML data.

here is an example how we used this return_value 
https://github.com/Open-MSS/MSS/blob/develop/mslib/msui/_tests/test_mss_pyui.py#L105

On this return_value data the other operations are done.

mslib/ogcwms.py
Marilyth: Done

mslib/msui/mss_pyui.py
Marilyth: I made a test for this, and it worked fine.
Ultimately I have decided against incorporating this test, because that test would not be consistent and will eventually fail. Maybe not next month, maybe not next year, but eventually.
MSS is constantly updating its packages, and at some point we will not be able to run MSS-2.0 anymore because of this. The DLR Server will soon be updated and afterwards the test might fail also.

I have also tested both MSS-2.0 and the DLR server by hand, using the new Client.
I tested each layer, made sure everything is drawing correctly, and made sure the communication works without errors.

mslib/msui/wms_control.py
Marilyth: Done

mslib/msui/wms_control.py
```diff
@@ -64,6 +66,20 @@ def add_wms_urls(combo_box, url_list):
...
+    """
+    Parses the version out of a GetCapabilities response
+    If parsing causes an error, return None for further investigation
+    """
+    try:
+        version = WMSCapabilitiesReader().readString(content).attrib["version"]
+        if version not in ["1.1.1", "1.3.0"]:
+            version = "1.1.1"
+        return version
+    except Exception as ex:
```
joernu76: Avoid needless wildcard exceptions. Specify the stuff you expect/know. Maybe throw in a log message as you already have ex available. I have been burnt by wildcards before.

mslib/msui/wms_control.py
```diff
@@ -556,7 +577,13 @@ def initialise_wms(self, base_url):
             str(base_url)  # to provoke early Unicode Error
             while wms is None:
                 try:
-                    wms = MSSWebMapService(base_url, version='1.1.1',
+                    # Version could not be detected without authentication
+                    if not version:
+                        response = openURL(base_url, "service=WMS&request=GetCapabilities",
```
joernu76: Does this generate a third request to the server? Cannot we fold that into the class instantiation below? We will use the version the server offers, won't we?

mslib/msui/wms_control.py
```diff
@@ -660,9 +686,13 @@ def get_capabilities(self):
         else:
             # url shortener url translated
             url = request.url
-            url = url.replace('?service=WMS&request=GetCapabilities&version=1.1.1', '')
+
+            # Take the default version of the server, 1.1.1 if not supported, None to determine on initialise_wms
+            version = determine_wms_version(request.content)
+
+            url = url.replace('?service=WMS&request=GetCapabilities', '')
```
joernu76: (Not that this was good before, but) is this robust? Do we assemble the URL here such that we know that this will work?

mslib/retriever.py
joernu76: The retriever has not really grown beyond being a prototype. If it breaks, it breaks. We have a major working package for this and I might tackle that during the next campaign if it hasn't until then.

joernu76 (changes requested):

mslib/msui/wms_control.py
Marilyth: You mean if the url will always contain `?service=WMS&request=GetCapabilities`?
Short answer is no. The safest way to remove the appended params would be something like
```
url.replace("?service=WMS", "") \
   .replace("&service=WMS", "") \
   .replace("?request=GetCapabilities", "") \
   .replace("&request=GetCapabilities", "")
```
I did that in the latest commit, in case that's what you meant

mslib/msui/wms_control.py
Marilyth: I changed it so the class handles dynamic versions, not wms_control. This gets rid of the third request.

mslib/msui/wms_control.py
Marilyth: I removed that function entirely with the latest commit

joernu76 (dismissed):

ReimarBauer (approved):
sounds good :+1:

joernu76 (approved):

PR #655, opened 2021-01-31
Merge Stable into develop

ReimarBauer:


Marilyth (approved):
Looks fine.

ReimarBauer:
Don't click on, Don't Don't Don't !!

never

joernu76:
This requires a IMO a rebase to be mergable. As we cannot rebase stable, we need to fork stable, rebase it onto develop and then merge it.
merging develop into stable should work similarily.
I need to think about how much sense a semi-linear history has for a two-branch development, though.
Also, the update branch button is not-clickable due to my restriction on semi-linear history.

ReimarBauer:
has been done by rebasing as suggested

PR #657, opened 2021-02-01
Stable2

joernu76:


ReimarBauer (approved):
rebasing that way works, great ;)

PR #661, opened 2021-02-02
Editing mscolab Url must be blocked on connect #637

### LINKED ISSUE #637: Editing mscolab Url must be blocked on connect
We need to disable editing of the mcolab url once connected. And enable it on disconnect.

block_editing_of_url

https://github.com/Open-MSS/MSS/blob/develop/mslib/msui/mscolab.py

ahelii16:
Disabled editing of the mscolab url on connecting, also disabled drop down. Enabled it again on clicking disconnect.

### DIFF
--- a/mslib/msui/mscolab.py
+++ b/mslib/msui/mscolab.py
@@ -164,7 +164,6 @@ class MSSMscolabWindow(QtWidgets.QMainWindow, ui.Ui_MSSMscolabWindow):
         self.addUser.setEnabled(False)
         self.connectMscolab.setEnabled(True)
         self.url.setEnabled(True)
-        self.url.setEditable(True)
         # set mscolab_server_url to None
         self.mscolab_server_url = None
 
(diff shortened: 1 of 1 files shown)


mslib/msui/mscolab.py
```diff
@@ -163,6 +163,8 @@ def disconnect_handler(self):
         self.loginButton.setEnabled(False)
         self.addUser.setEnabled(False)
         self.connectMscolab.setEnabled(True)
+        self.url.setEnabled(True)
+        self.url.setEditable(True)
```
Tanish0019: Not sure but do we need to use setEditable as well? Won't just using setEnabled(False) be enough as it completely disabled the input?

ahelii16: Yes you're right. I did not think of that. Wouldn't need line 167 as well if setEnabled is True!

Tanish0019 (approved):

Tanish0019:
@ReimarBauer Approved this, you can check and merge it

ReimarBauer (changes requested):
wait, I see it is not the wrong number it is the wrong text, I fix it

ReimarBauer (approved):

PR #663, opened 2021-02-03
new optional parameter for pytest --mss_settings / improving config_loader

### LINKED ISSUE #649: add a test configuration which uses the "filepicker_default": "fs"
the default of MSS is to use the QT filepicker. But the code should always be able to use the fs_filepicker too. 

Also checks for files need to use the fs library, otherwise we can't change the default to some external storage e.g. webdavfs

The syntax of  a fs url can be different to the one handled by the os module.
e.g.
https://docs.pyfilesystem.org/en/latest/openers.html

The test config may be just use a different syntax like osfs: 
The mem: syntax may also be interesting

### LINKED ISSUE #660: config_loader needs to return a default value when this is missing in a config
This test fails in the moment a config is given and num_labels not set
```
self = <mslib._tests.test_utils.TestConfigLoader object at 0x7fb4453c00a0>

    def test_default_config(self):
        data = utils.config_loader()
        assert isinstance(data, dict)
>       assert data["num_labels"] == 10
E       KeyError: 'num_labels'

mslib/_tests/test_utils.py:68: KeyError
```

currently this give

in config_loader

```
        if dataset:
            try:
>               return data[dataset]
E               KeyError: 'num_labels'

mslib/utils.py:130: KeyError
```
```
        if dataset:
            try:
                return data[dataset]
            except KeyError:
                logging.debug("Config File used: '%s'", config_file)
                logging.debug("Key not defined in config_file! '%s'", dataset)
                if default is not None:
                    return default
>               raise KeyError("default value for key not set")
E               KeyError: 'default value for key not set'
```
In that case we should lookup if the default dict has a value and return that. It there is nothing given we have to raise KeyError

ReimarBauer:
by this parameter pytest can use a mss_settings.json
saved in the test tmp directory
example:
pytest -k test_existing_empty_config_file --mss_settings '{ }'
pytest -k test_existing_config_file_different_parameters --mss_settings '{"num_interpolation_points": 201 }'

After we agreed to this kind of configuration we can pick tests which can be verified by an individual "mss_settings.json" for tests. I guess we will do this by another test action.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -40,6 +40,19 @@ import fs
 from mslib.mswms.demodata import DataFiles
 import mslib._tests.constants as constants
 
+
+def pytest_addoption(parser):
+    parser.addoption("--mss_settings", action="store")
+
+
+def pytest_generate_tests(metafunc):
+    option_value = metafunc.config.option.mss_settings
+    if option_value is not None:
+        mss_settings_file_fs = fs.open_fs(constants.MSS_CONFIG_PATH)
+        mss_settings_file_fs.writetext("mss_settings.json", option_value)
+        mss_settings_file_fs.close()
+
+
 if os.getenv("TESTS_VISIBLE") == "TRUE":
     Display = None
 else:
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -213,15 +213,8 @@ facilitating collaboration on campaigns. To enable this feature apply
     "filepicker_default": "fs",
 
 to your configuration file. The allowed values are "qt" for QT-based dialogues, "fs" for
-fsfile_picker-based dialogues supporting remote locations, or "default" for the default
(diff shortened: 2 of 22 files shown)


ReimarBauer:
This is not the solution for the issue https://github.com/Open-MSS/MSS/issues/660 but the first step to reproduce it by a test

ReimarBauer:
- [x] move config_loader test files inside the tests
- [x] refactor config_loader

ReimarBauer:
I'm still not sure if I have catched all cases how we use this function.

If I could assume that we only can use names which are defined in dict(MissionSupportSystemDefaultConfig.__dict__) and that the values there are always the defaults we get a lower complexity.

joernu76 (approved):

PR #670, opened 2021-02-08
merged stable into develop

ReimarBauer:


joernu76:
This doesn't look properly rebased.

ReimarBauer:
that rebase thing is not that easy as it should be. We discussed on slack an idea to automate that.

Issue #279, opened 2017-09-07
faster timeout for configured but not reachable servers


ReimarBauer:
**[Original report](https://bitbucket.org/wxmetvis/mss/issue/279) by Reimar Bauer (Bitbucket: [ReimarBauer](https://bitbucket.org/ReimarBauer), GitHub: [ReimarBauer](https://github.com/ReimarBauer)).**

----------------------------------------

If a server is behind a firewall and we don't have access, we should get that pretty fast.

Currently a getcapabilities request needs to run into a timeout to stop this request.

ReimarBauer:
**Original comment by Reimar Bauer (Bitbucket: [ReimarBauer](https://bitbucket.org/ReimarBauer), GitHub: [ReimarBauer](https://github.com/ReimarBauer)).**

----------------------------------------

needs to become verified after 1.7.0 release

ReimarBauer:
**Original comment by Reimar Bauer (Bitbucket: [ReimarBauer](https://bitbucket.org/ReimarBauer), GitHub: [ReimarBauer](https://github.com/ReimarBauer)).**

----------------------------------------

all wms urls from configurations can be checked if they exists. The user interface should only show which can be used. Missing urls should become logged.

ReimarBauer:
**Original comment by Reimar Bauer (Bitbucket: [ReimarBauer](https://bitbucket.org/ReimarBauer), GitHub: [ReimarBauer](https://github.com/ReimarBauer)).**

----------------------------------------

I think it is better to have a mss\_pyui command line option to check defined urls. We should cleanup from the beginning urls which can’t be accessed.

ReimarBauer:
We could verify availability of the host. 
```
_s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
    try:
        _s.bind(("IP", port))
    except (socket.error, IOError):
       -> mark not available

```

Marilyth:
@ReimarBauer 
I am very confused here. What are we trying to find out? Why would binding a port solve anything?

So we have preconfigured servers, they may or may not be "available", and we want to know that before [get capabilities] is pressed?
Do we just want to automatically "press the button" before the user gets to, to make sure every server is replying correctly?

ReimarBauer:
You are right, that is not so simple as it looks like. 

a) user has an old server which does not provide data but seems alive this won't be recognized before a request.
b) also we can have the situation the valid server is just in an maintainance state.

We can not autmoate this. Therefore I close it.

PR #676, opened 2021-02-18
Multilayering

### LINKED ISSUE #3: Integrate support for multiple simultaneous WMS layers into msui client
**[Original report](https://bitbucket.org/wxmetvis/mss/issue/3) by Marc R. (Bitbucket: [wxmetvis](https://bitbucket.org/wxmetvis), ).**

Currently the msui can only display one WMS image at a time. Similar to other WMS clients, an overlay of multiple (transparent) maps should be possible.

This could be realized by implementing:

* support for multiple WMSControlWidgets
* these WMSControlWidgets should (optionally) synchronise level, init time, valid time with a "master" control
* multiple images may need to be combined into one image for display with matplotlib -- maybe this has changed, otherwise PIL (Python Imaging Library) should provide functions

### LINKED ISSUE #159: server side simultaneously plot two layers
**[Original report](https://bitbucket.org/wxmetvis/mss/issue/159) by Reimar Bauer (Bitbucket: [ReimarBauer](https://bitbucket.org/ReimarBauer), GitHub: [ReimarBauer](https://github.com/ReimarBauer)).**

The "WMS-correct" way of doing this would be add a layer on the server side that returns a map displaying the desired data. Two options: Extend the server so that it can simultaneously plot two layers (that is defined in the WMS standard)

We can work on this after we migrated the owslib to a more recent version

### LINKED ISSUE #480: mark and remove URLs from history
**[Original report](https://bitbucket.org/wxmetvis/mss/issue/480) by Reimar Bauer (Bitbucket: [ReimarBauer](https://bitbucket.org/ReimarBauer), GitHub: [ReimarBauer](https://github.com/ReimarBauer)).**

The URL fields have a drop down list of possible URLs to be filled from both the configuration file and possibly history of used URLs comparable to the WMS control. This is an important feature as it simplifies the preconfiguration of MSS for campaign users using the JSON file.

‌

Possibly multiple URLS should be retained in recent history, if a mechanism could be implemented to mark and remove those URLs again.

‌

‌

‌

Marilyth:
Contains the current WIP of the multilayer support for the Server and Client.
There is still a lot of work to be done, especially design wise, so this will stay a draft until it is ready for deployment.
This PR primarily exists to test and talk about features that should exist or aren't satisfactory as is.

Implemented and planned features to date include:
- [x] Server supports requests for multiple layers
- [x] Client supports retrieving and displaying multiple layers
- [x] Layers have a selectable priority in which they will be displayed
- [x] Make multilayer list searchable/filterable
- [x] Ability to synchronise layer parameters
- [x] Disallow synchronisation of layers with no common parameters
- [x] Show intersection parameters of synchronised multilayers
- [x] Remove redundant parts of the interface
- [x] Overall improve the UI to be less click intensive (up for debate, UI is very replaceable now however)
- [x] Find a solution for displaying legends and general information about all layers
- [x] Ability to draw single layer without deactivating all others
- [x] Additional tests
- [x] Retrieve should always be visible
- [x] Refactor wms_control.py to remove now unnecessary parts

Please let me know if I am missing something so I can keep this list up to date.


Marilyth:
While the design is not currently finished, I would like the functionality to be reviewed already.
As it stands now, the design is very independent from the layers and can be changed without much hassle.

mslib/msui/_tests/test_wms_control.py
```diff
@@ -252,6 +252,41 @@ def test_server_service_cache(self, mockbox):
...
+        self.window.multilayers.multilayer_clicked(server.child(0))
+        assert previous_layer != self.window.lLayerName.text()
+        assert self.window.listLayers.itemWidget(server.child(0), 1) is not None
+        assert self.window.listLayers.itemWidget(server.child(2), 1) is not None
+        assert self.window.listLayers.itemWidget(server.child(1), 1) is None
+        QtTest.QTest.mouseClick(self.window.btGetMap, QtCore.Qt.LeftButton)
+        QtWidgets.QApplication.processEvents()
+        QtTest.QTest.qWait(1000)
+        QtWidgets.QApplication.processEvents()
+        QtTest.QTest.qWait(6000)
```
ReimarBauer, line 331: usually an action has to be done before QtWidgets.QApplication.processEvents() is invoked. I guess one can be removed.

mslib/msui/multilayers.py
```diff
@@ -0,0 +1,470 @@
+from PyQt5 import QtWidgets, QtCore
```
ReimarBauer, line 27: add the standard header of the project. Add your name: into the line :copyright: Copyright 2021 May Bär

mslib/msui/multilayers.py
```diff
@@ -0,0 +1,470 @@
...
+        self.parent.leMultiFilter.textChanged.connect(self.filter_multilayers)
+        self.parent.listLayers.setColumnWidth(1, 50)
+        self.parent.listLayers.header().setSectionResizeMode(0, QtWidgets.QHeaderView.Stretch)
+        self.parent.listLayers.setContextMenuPolicy(QtCore.Qt.CustomContextMenu)
+        self.parent.listLayers.customContextMenuRequested.connect(self.right_clicked)
+
+    def get_current_wms(self):
+        if self.current_layer:
+            return self.layers[self.current_layer.wms_name]["wms"]
+        return None
```
ReimarBauer: return None is default. If you don't add a return None also None is returned :)

Any method or function returns always None if there is nothing else defined.

mslib/msui/multilayers.py
```diff
@@ -0,0 +1,470 @@
...
+        layers = self.get_active_layers(only_synced=True)
+        if additional_layer:
+            layers.append(additional_layer)
+
+        elevation_values = []
+        init_time_values = []
+        valid_time_values = []
+        crs_values = []
+
+        wms_before = self.parent.wms
```
ReimarBauer: additional line 104 
what is the intend for this mapping?

mslib/msui/multilayers.py
```diff
@@ -0,0 +1,470 @@
...
+        for values in elevation_values:
+            elevation_values[0] = list(set(elevation_values[0]).intersection(values))
+        for values in init_time_values:
+            init_time_values[0] = list(set(init_time_values[0]).intersection(values))
+        for values in valid_time_values:
+            valid_time_values[0] = list(set(valid_time_values[0]).intersection(values))
+        for values in crs_values:
+            crs_values[0] = list(set(crs_values[0]).intersection(values))
+
+        self.parent.wms = wms_before
```
ReimarBauer: see comment line 87

mslib/msui/multilayers.py
```diff
@@ -0,0 +1,470 @@
...
+
+        return sorted(elevation_values[0], key=lambda x: float(x.split()[0])) if len(elevation_values) > 0 else [], \
+               sorted(init_time_values[0]) if len(init_time_values) > 0 else [], \
+               sorted(valid_time_values[0]) if len(valid_time_values) > 0 else [], \
+               sorted(crs_values[0]) if len(crs_values) > 0 else []
+
+    def get_multilayer_priority(self, layer_widget):
+        """
+        Returns the priority of a layer, or None if it isn't checked
```
ReimarBauer: this can't return None. The "isn't checked" is not verified here.

mslib/msui/multilayers.py
```diff
@@ -0,0 +1,470 @@
...
+                             (len(itimes) > 0 or len(itimes_before) == 0) and
+                             (len(vtimes) > 0 or len(vtimes_before) == 0))
+
+        sync_menu.exec(self.parent.listLayers.viewport().mapToGlobal(pointer))
+        self.multilayer_clicked(widget)
+
+
+class Layer(QtWidgets.QTreeWidgetItem):
+    def __init__(self, header, parent, layerobj, is_empty=False):
+        if not is_empty:
```
ReimarBauer: is this separation twice time if not is_empty: needed?

mslib/msui/wms_control.py
```diff
@@ -762,7 +759,6 @@ def activate_wms(self, wms):
             self.valid_time_changed()
 
         # Reconnect layerChanged.
```
ReimarBauer: the comment belongs to the removed line

mslib/msui/wms_control.py
```diff
@@ -1391,15 +1281,14 @@ def normalize_crs(crs):
                         try:
                             value = int(prefetch_config[_x])
                         except ValueError as ex:
-                            logging.error("ERROR: %s %s", type(ex), ex)
+                            logging.error("ERRORasd: %s %s", type(ex), ex)
```
ReimarBauer: meaning of ERRORasd?

ReimarBauer (changes requested):
see my comments.

mslib/msui/wms_control.py
Marilyth: Cats can be sneaky creatures

ReimarBauer:
If a login is required and the credentional are not set in the mss_settings.json we need a popup dialog 

login_required_popup

Marilyth:
> If a login is required and the credentional are not set in the mss_settings.json we need a popup dialog

As far as I know I did nothing to change this behaviour, and it still works for me.
Does it not work for you? May I ask for which WMS? @ReimarBauer

ReimarBauer:
This may not capabable for sideview and shows a problem
http://maps.six.nsw.gov.au/arcgis/services/public/NSW_Imagery/MapServer/WMSServer

auth_delegate

ReimarBauer:
> > If a login is required and the credentional are not set in the mss_settings.json we need a popup dialog
> 
> As far as I know I did nothing to change this behaviour, and it still works for me.
> Does it not work for you? May I ask for which WMS? @ReimarBauer

I answered this on slack, I'm not sure  if this is a public server.

Now repeated, and this url gives a popup. May be it was for a moment not available?

ReimarBauer:
retrieve button should always been visible. I recognize that I too often click view instead.

I guess we could also enable auto update on default.

mslib/msui/multilayers.py
```diff
@@ -0,0 +1,519 @@
...
+    ~~~~~~~~~~~~~~~~~~~
+
+    This module contains classes for object oriented managing of WMS layers.
+    Improves upon the old method of loading each layer on UI changes,
+    the layers are all persistent and fully functional without requiring user input.
+
+    This file is part of mss.
+
+    :copyright: Copyright 2021 May Bär
+    :copyright: Copyright 2019-2021 by the mss team, see AUTHORS.
```
ReimarBauer: 2021 
because 2019 it wasn't there

mslib/msui/multilayers.py
```diff
@@ -0,0 +1,519 @@
...
+            if wms_hits == 0:
+                header.setHidden(True)
+            else:
+                header.setHidden(False)
+
+    def get_multilayer_common_options(self, additional_layer=None):
+        """
+        Return the common option for levels, init_times, valid_times and CRS
+        for all synchronised layers and the additional provided one
+        """
```
ReimarBauer, line 121: I'm currently not sure if we should synchronize per service or per layer. Per Layer we maybe need an indicator which are synced. I found a problem with the demo dataset.

synchronize_problem

If I start with Temperature selected only, synced enabled and retrieve and I add afterwards the Equivalent, synced and click review again it works.

It is indeed only an order problem by this example.

order_problem_shows_it

Of course, likly missing data, needs to be catched ;)

mslib/msui/multilayers.py
```diff
@@ -0,0 +1,519 @@
...
+
+        info = f"Common levels: {len(levels)} {f'({lost_levels})' if lost_levels < 0 else ''}\n"
+        info += f"Common init times: {len(itimes)} {f'({lost_itimes})' if lost_itimes < 0 else ''}\n"
+        info += f"Common valid times: {len(vtimes)} {f'({lost_vtimes})' if lost_vtimes < 0 else ''}"
+        info_label.setText(info)
+
+        def check_changed(state):
+            widget.is_synced = not widget.is_synced
+            if widget.is_synced:
+                widget.setText(0, widget.text(0) + " (synced)")
```
ReimarBauer: Where should I see this info?

ReimarBauer (changes requested):
We need likly a feature to clear all selection. 
If I do clear the selection, the info in "Layer Options": "Layer" keeps the old choosed layer names.
Perhaps if nothing is selected the Layer Options Tab could also be not clickable.
Found only a small problem by playing with the demodata set.

mslib/msui/multilayers.py
Marilyth: This is an issue with the demodata server. It is throwing an error when the client requests a transparent layer, but it has no white pixels.
The server is already fixed in my branch.

mslib/msui/multilayers.py
Marilyth: Directly in the list. After a layer is synced, the name of the layer is appended by `(synced)`
This makes it also possible to search synced layers by filtering for `(synced)`

mslib/msui/multilayers.py
Marilyth: Fixed already with eca9dff

mslib/msui/multilayers.py
ReimarBauer: ok, I recognized today with another PR that I should setup a develop server ;)

ReimarBauer:
Probably inactive / not reachable servers should be checked and marked as "try this one later again" or kicked from the list by the user.

https://github.com/Open-MSS/MSS/issues/279

mslib/mswms/mpl_hsec.py
```diff
@@ -391,10 +391,13 @@ def plot_hsection(self, data, lats, lons, bbox=(-180, -90, 180, 90),
...
-            facecolor_index = lut.index(tuple(facecolor_rgb))
-
-            logging.debug("saving figure as transparent PNG with transparency index %s.", facecolor_index)
-            palette_img.save(output, format="PNG", transparency=facecolor_index)
+            try:
+                facecolor_index = lut.index(tuple(facecolor_rgb))
+                logging.debug("saving figure as transparent PNG with transparency index %s.", facecolor_index)
+                palette_img.save(output, format="PNG", transparency=facecolor_index)
+            except ValueError:
+                logging.debug("transparency requested but not possible, saving non-transparent instead")
```
ReimarBauer, line 399: how does this look like for a user? I assume this can cover other data and a user trusts the transparency result. Should we message this to the user somehow?

mslib/mswms/wms.py
```diff
@@ -329,209 +344,212 @@ def produce_plot(self, query, mode):
...
+                    return self.create_service_exception(
+                        code="LayerNotDefined",
+                        text=f"Invalid LAYER '{dataset}.{layer}' requested",
+                        version=version)
 
-            # Check if the layer requires time information and if they are given.
-            if self.vsec_layer_registry[dataset][layer].uses_inittime_dimension():
-                if init_time is None:
+                # Check if the layer requires time information and if they are given.
+                if self.hsec_layer_registry[dataset][layer].uses_inittime_dimension() and init_time is None:
```
ReimarBauer: i can't assess this right now,
that which occurs most often should be asked first or the faster one ;)

This chek init_time is None will be less expensive than the other, perhaps we can save the time for the other check.

mslib/mswms/wms.py
```diff
@@ -329,209 +344,212 @@ def produce_plot(self, query, mode):
...
+                # Bounding box.
+                try:
+                    if is_yx:
+                        bbox = [float(v) for v in query.get('BBOX', '-90,-180,90,180').split(',')]
+                        bbox = (bbox[1], bbox[0], bbox[3], bbox[2])
+                    else:
+                        bbox = [float(v) for v in query.get('BBOX', '-180,-90,180,90').split(',')]
+
+                except ValueError:
+                    return self.create_service_exception(text=f"Invalid BBOX: {query.get('BBOX')}", version=version)
```
ReimarBauer, line 459: in the past there was also an IOError option and a logging, IOError won't happen without IO ;)
logging.error("ERROR: %s %s", type(ex), ex)

mslib/msui/multilayers.py
```diff
@@ -0,0 +1,555 @@
...
+        """
+        if self.level:
+            return self.get_level().split(" (")[0]
+
+    def get_legend_url(self):
+        style = self.get_style()
+        urlstr = None
+        if style and "legend" in self.layerobj.styles[style]:
+            urlstr = self.layerobj.styles[style]["legend"]
+
```
ReimarBauer: cosmetics ;) I would not have an empty line here.

mslib/msui/wms_control.py
```diff
@@ -1598,48 +1483,54 @@ class HSecWMSControlWidget(WMSControlWidget):
...
-        self.view.draw_metadata(title=self.get_layer_object(layer).title,
+        style_title = self.multilayers.get_current_layer().get_style()
+        self.view.draw_metadata(title=self.multilayers.get_current_layer().layerobj.title,
                                 init_time=init_time,
                                 valid_time=valid_time,
                                 level=level,
                                 style=style_title)
         self.view.draw_image(img)
-        self.view.draw_legend(legend_img)
+        self.view.draw_legend(legend_imgs[-1])
```
ReimarBauer: this shows the last legend, we need to know that the range of all legends -> data is the same.

ReimarBauer (changes requested):
added some comments.

mslib/mswms/mpl_hsec.py
Marilyth: From the WMS specifications:

> If the picture format does not support transparency, then the server shall
respond with a non-transparent image (in other words, it is not an error for the client to always request transparent
maps regardless of format).

I assume this also applies to formats that do support transparency, but the map contains no data-less area.
Data covering area will always cover other underlying data. Even if transparent. This is just a more hardcore case of it.

mslib/msui/wms_control.py
Marilyth: All legends are shown at once now. Though too many legends may make it quite hard to read, I suppose.

joernu76: Fine for now. I doubt that we we will have more than one in practise, so this doesn't terribly matter. In a second step one might add a feature to select one of multiple legends with a drop-down field. 
We also might think about actually supporting and using the legend feature in our MSS Server, BTW...

ReimarBauer: Hmm, if all legends have the same scale than this does not matter. We need to have checked this,

joernu76 (approved):
Very nice! Too much to test it in my head, though. I have some immediate feedback, that should be resolved after this has been merged. We also need to add some useful overlays to the server for meaningful testing, which would imply updating the MSS instance...

joernu76:
I added, for testing, two new layers to the SouthTrac server, HS_PVC and HS_Barbs_C, which are contours for putting on top of other things.
Works fine :1st_place_medal:

ReimarBauer (approved):

PR #687, opened 2021-02-24
Resolves Issue #650

### LINKED ISSUE #650: Remove email field from admin window in mscolab
We don't want to show the emails of all the users in the admin window for privacy reason so we need to remove that field from the admin window in mscolab.

**Opening Admin Window**
 - Start mscolab server and the main mss ui
 - Go to mscolab window from the toolbar at the top
 - Connect to mscolab server, login as user, select a project then click on manage users

image

nakul-shahdadpuri:
Fixes #650 
@Tanish0019 @ReimarBauer 
Removed Email field from admin window in mscolab.
## Screenshot of the current look
Screenshot from 2021-02-24 13-54-52


ReimarBauer (changes requested):
comments can be used for commenting something, removing lines should be removing them ;)

Tanish0019:
@nakul-shahdadpuri I see you're populating table by removing emails from the response sent back in the frontend. It will be better if you instead update the mscolab api for this and not do the filtering on the frontend. Also can you play around with the window/table sizes in the UI so that username column is not so stretched. It looks a bit weird right now. Thanks!

Tanish0019:
Also we never change the ui files by hand. Always use the Qt designer to update the .ui files. [This](https://realpython.com/qt-designer-python/) is a helpful article on getting started with qt designer

mslib/msui/mscolab_admin_window.py
```diff
@@ -83,6 +83,9 @@ def __init__(self, token, p_id, user, project_name, projects, conn, parent=None,
         self.populate_import_permission_cb()
 
     def populate_table(self, table, users):
+        for item in users:
+            item.pop(1)
+        print(users)
```
ReimarBauer: this will users confuse. Instead you can use logging.debug

mslib/msui/qt5/ui_mscolab_admin_window.py
```diff
@@ -55,12 +55,12 @@ def setupUi(self, MscolabAdminWindow):
...
         self.addUsersTable.setSelectionBehavior(QtWidgets.QAbstractItemView.SelectRows)
         self.addUsersTable.setObjectName("addUsersTable")
-        self.addUsersTable.setColumnCount(2)
+        self.addUsersTable.setColumnCount(1)
         self.addUsersTable.setRowCount(0)
         item = QtWidgets.QTableWidgetItem()
         self.addUsersTable.setHorizontalHeaderItem(0, item)
-        item = QtWidgets.QTableWidgetItem()
-        self.addUsersTable.setHorizontalHeaderItem(1, item)
+        # item = QtWidgets.QTableWidgetItem()
```
ReimarBauer: if this is not done in the ui file by the designer it will return by pyuic

nakul-shahdadpuri:
@ReimarBauer yup I'll do that. 
@Tanish0019 I initially had started by changing the [models.py](https://github.com/Open-MSS/MSS/blob/develop/mslib/mscolab/models.py) file, but I wasn't sure if the API was being called anywhere else, so I just removed it from the admin window frontend. I'll try tinkering with the models now. Yup, the UI looked a little stretched I'll work on that too. :)

Tanish0019:
@nakul-shahdadpuri no need to update models.py that defines the schema. We need to update 2 apis which are used to populate these two tables. You can just remove email from the response in those apis 😀

nakul-shahdadpuri:
@Tanish0019 @ReimarBauer 
This is how the admin tab looks now. I removed the emailid attribute from the response by changing the API. 
Screenshot from 2021-02-27 15-14-42

ReimarBauer (changes requested):
You have to use designer ui_mscolab_admin_window.ui to remove elements there and convert it by pyuic5 to  ui_mscolab_admin_window.py

The ui_mscolab_admin_window.py is never edited manually.

https://pythonbasics.org/qt-designer-python/

https://realpython.com/qt-designer-python/

ReimarBauer (changes requested):
look at pytest -k test_mscolab_admin_window.py
The tests needs a change too.

take care on test output on github. (of course tests should fit)

I tried a clone of your repositiory, it looks to me like in this peek video. On the right side the permission is not shown.

admin_window

ReimarBauer:
I can't reproduce the proposed solution. 
permissions_missing

ReimarBauer:
@Tanish0019 or/and @Marilyth please can you try out the proposed solution. Perhaps my setup can have issues

nakul-shahdadpuri:
@ReimarBauer @Tanish0019 @Marilyth 
I am still getting the following output.
I had removed the email field from [file_manger.py](https://github.com/Open-MSS/MSS/blob/develop/mslib/mscolab/file_manager.py) in fetch_users_with_permission and fetch_users_without_permission which was being called in [server.py](https://github.com/Open-MSS/MSS/blob/54a0439b47ab0ae164122e6d6a35a59dc313ddbf/mslib/mscolab/server.py#L405)
I am not sure why the email field is still present in your version.
current look

ReimarBauer:
I get a new clone tom. and try again.

ReimarBauer:
I'm not sure if you are on our slack channel. If you want an invite send me a mail to mss-info@fz-juelich.de If you are there ping me. (https://open-mss.github.io/contributing/)

We should find what makes this difference. I tried to debug the behaviour with the yourmss server.

My server sends as second attribute the email address.

mssdata

users

ReimarBauer:
Got it. My failure. yourmss is on the 2.0.1 server version ;)

ReimarBauer (approved):

PR #690, opened 2021-02-26
Install with mamba, check development.txt

### LINKED ISSUE #669: Docker images don't always pick up new packages
As we could see with #665 our docker images don't seem to be installing the latest versions of packages reliably.
Issues that arise from updated packages thus don't always get caught.

It seems like the issue might come from using conda instead of mamba for installation, as these two don't behave the same way.
We might also have to run `conda/mamba update --all` to make sure new packages are installed.

### LINKED ISSUE #682: Testing should update development.txt modules if they differ
Currently only meta.yaml is checked for differences with the testing image.
This should also be done with development.txt to ensure the tests run on potential new or downgraded modules needed for testing.

This issue recently came to light in #679

Marilyth:
This PR makes sure the packages get removed and reinstalled when either the local meta.yaml or development.txt differ from the image. 
There is no simple way to remove all packages of a specific file, so instead all mssenv packages are removed and installed again either way.
This will add roughly 2.5 minutes to the total testing should differences arise. This could be optimised in the future by only removing either development.txt or meta.yaml packages.

This fixes #682 and fixes #669

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -18,18 +18,22 @@ jobs:
     steps:
     - uses: actions/checkout@v2
         
-    - name: Update dependencies if changed
+    - name: Reinstall dependencies if changed
       run: |
-        cmp -s /meta.yaml localbuild/meta.yaml || (echo Dependencies differ \
+        cmp -s /meta.yaml localbuild/meta.yaml && cmp -s /development.txt requirements.d/development.txt || (echo Dependencies differ \
         && cat localbuild/meta.yaml \
         | sed -n '/^requirements:/,/^test:/p' \
         | sed -e "s/.*- //" \
         | sed -e "s/menuinst.*//" \
         | sed -e "s/.*://" > reqs.txt \
+        && source /opt/conda/bin/activate mssenv \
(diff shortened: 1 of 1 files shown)


.github/workflows/testing.yml
```diff
@@ -18,18 +18,22 @@ jobs:
...
+    - name: Reinstall dependencies if changed
       run: |
-        cmp -s /meta.yaml localbuild/meta.yaml || (echo Dependencies differ \
+        cmp -s /meta.yaml localbuild/meta.yaml && cmp -s /development.txt requirements.d/development.txt || (echo Dependencies differ \
         && cat localbuild/meta.yaml \
         | sed -n '/^requirements:/,/^test:/p' \
         | sed -e "s/.*- //" \
         | sed -e "s/menuinst.*//" \
         | sed -e "s/.*://" > reqs.txt \
+        && source /opt/conda/bin/activate mssenv \
```
ReimarBauer, line 29: I've no idea if it makes a difference to have first the env activated or the channels updated

I think that it has to be done before we install something, which in both cases is done

ReimarBauer (approved):
thx for solving this :)

PR #691, opened 2021-02-26
Install with mamba, check development.txt (#690)

Marilyth:
Merges stable into develop


ReimarBauer (approved):
thx for this improvement.

PR #692, opened 2021-02-26
Refactoring mscolab tests

### LINKED ISSUE #530: review tests and maybe refactor for pytest parallel
**[Original report](https://bitbucket.org/wxmetvis/mss/issue/530) by Reimar Bauer (Bitbucket: [ReimarBauer](https://bitbucket.org/ReimarBauer), GitHub: [ReimarBauer](https://github.com/ReimarBauer)).**

We may be can use multiple workers for pytest

[https://pypi.org/project/pytest-parallel/](https://pypi.org/project/pytest-parallel/)

We have to carefully check how we have to change our tests.  

multi threading likly won’t be possible.

‌

ReimarBauer:


.github/workflows/testing.yml
```diff
@@ -35,7 +35,7 @@ jobs:
         && mamba install --file requirements.d/development.txt)
         
     - name: Run tests
-      timeout-minutes: 35
+      timeout-minutes: 60
```
Marilyth: The reason the test is timing out is not because it needs more time.
It seems like pytest is not giving control back after all tests are finished, maybe there is some unclosed thread still hanging around somewhere.
I had that on my PC once, pytest finished, I saw all results but I had to ctrl+C out to get control again. This is likely what is happening here.

ReimarBauer:
later the xdist test should be the required default. the other can persist but should not been required

.github/workflows/testing.yml
ReimarBauer: the finalizer in conftest is broken, did not terminate the processes

ReimarBauer:
This will be restarted from a fresh fork

Issue #681, opened 2021-02-20
refactoring of mscolab tests


ReimarBauer:
the mscolab tests needs a refactoring for faster execution, if possible executed by pytest-xdist.
This is not mandatory. A first step is to setup a framework that generates users and data on the fly and start fresh for each test to remove dependency.

By my (Reimar) current refactoring based on the existing tools I'm not sure if it ever will work perfect with xdist. But trying with that also shows parts to improve.

ReimarBauer:
We have a lot of dependencies that have to be taken into account

For the scope loadfiles or in general

Tests must not have any dependencies to other tests especially:

- use new sockets for each test 
- use new ports
- reinitialize databases, or teardown added data

PR #707, opened 2021-03-05
initial refactoring of mscolab tests for xdist, solves #681

ReimarBauer:


ReimarBauer:
Hopefully I have now all blocking connections fixed. 
On my system it is amazing fast. 

pytest -n auto --dist loadfile --max-worker-restart 0 mslib
 346 passed, 17 skipped, 3767 warnings in 280.17s (0:04:40)

pytest -n auto --dist loadscope --max-worker-restart 0 mslib
346 passed, 17 skipped, 3725 warnings in 266.82s (0:04:26)

conftest.py
```diff
@@ -173,43 +163,3 @@ def configure_testsetup(request):
         VIRT_DISPLAY.stop()
     else:
         yield
-
-
-process = None
-
-
-@pytest.fixture(scope="session", autouse=True)
```
joernu76, line 181: Wouldn't it be better to use a *test-specific* fixture? Your solution by calling a function does it similarly, but "fixtures" might be stylistically better when using pytest?
Location of the code in _tests/utils.py is fine, I think. We should not overload the conftest.py thingie.

Both setting up the server, as well as tearing it down should be brought to own place to remove doubling of code.
Something like a decorator might be able to do that, but as we seem to be using classes for testing, inheritance might be the solution here?

mslib/_tests/utils.py
```diff
@@ -153,6 +158,39 @@ def mscolab_get_project_id(app, msc_url, email, password, username, path):
             return p['p_id']
 
 
+def mscolab_check_free_port(all_ports, port):
```
joernu76, line 161: I have severe doubts about this function. It is recurse and opens(?) a lot of sockets?
The port returned is always the first one tried, or the first one from the list, never a later one. (port = mscolab_check_free_port(...) would likely fix that).
In case multiple threads call this function, I am not sure, it will work reliably. There needs to be a semaphore or something...?

mslib/_tests/utils.py
```diff
@@ -153,6 +158,39 @@ def mscolab_get_project_id(app, msc_url, email, password, username, path):
...
+    except (socket.error, IOError):
+        port = all_ports.pop()
+        mscolab_check_free_port(all_ports, port)
+    _s.close()
+    return port
+
+
+def mscolab_start_server(all_ports, mscolab_settings=mscolab_settings):
+    handle_db_seed()
+    port = mscolab_check_free_port(all_ports, all_ports.pop())
```
joernu76, line 175: all_ports.pop() modifies the all_ports object supplied from above. Intended?

mslib/_tests/utils.py
```diff
@@ -153,6 +158,39 @@ def mscolab_get_project_id(app, msc_url, email, password, username, path):
             return p['p_id']
 
 
+def mscolab_check_free_port(all_ports, port):
+    _s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+    try:
+        _s.bind(("127.0.0.1", port))
+    except (socket.error, IOError):
+        port = all_ports.pop()
```
joernu76, line 166: Modifies the all_ports object suppllied. Intended?

mslib/mscolab/_tests/test_chat_manager.py
```diff
@@ -23,35 +23,37 @@
...
 
 class Test_Chat_Manager(object):
     def setup(self):
-        self.sockets = []
-        self.app = APP
-        self.app.config['SQLALCHEMY_DATABASE_URI'] = mscolab_settings.SQLALCHEMY_DB_URI
-        self.app.config['MSCOLAB_DATA_DIR'] = mscolab_settings.MSCOLAB_DATA_DIR
-        self.app, _, _, fm = initialize_managers(self.app)
-        self.fm = fm
+        self.process, self.url, self.app, _, self.cm, self.fm = mscolab_start_server(PORTS)
```
joernu76, line 45: Copy paste from above.This is complicated and we should have just one implementation...?

mslib/mscolab/_tests/test_file_manager.py
```diff
@@ -25,43 +25,50 @@
...
+from mslib._tests.utils import mscolab_start_server
+
+PORTS = list(range(19341, 19390))
 
 
 class Test_FileManager(object):
     def setup(self):
-        assert 'tmp' in mscolab_settings.MSCOLAB_DATA_DIR
-        handle_db_seed()
+        self.process, self.url, self.app, _, self.cm, self.fm = mscolab_start_server(PORTS)
```
joernu76, line 44: Again, copy paste.

joernu76 (changes requested):

conftest.py
ReimarBauer: I had that with a fixture before and that did not work. And autouse is also a bad idea. But also with a pytest marker it made at that time trouble. The biggest Problem is the multiprocessing that needs definitly always very fast a new unused port.

joernu76: Obviously, the old approach using one session thing and autouse was also wrong. Never used fixtures in this way.
However, your new use is exactly what a "standard" by-test fixture would do.

mslib/_tests/utils.py
ReimarBauer: the current solution is that each module provides its individuell port range by all_ports
The function uses pop() that reduces the list by the used port. So it is only a matter of having enough in the list.

mslib/_tests/utils.py
ReimarBauer: yes, this is a fast solution to get a port for checking and kick it from the list.

the parallel tests can be run by 

--dist loadscope: Tests are grouped by module for test functions and by class for test methods. Groups are distributed to available workers as whole units. This guarantees that all tests in a group run in the same process. This can be useful if you have expensive module-level or class-level fixtures. Grouping by class takes priority over grouping by module.
--dist loadfile: Tests are grouped by their containing file. Groups are distributed to available workers as whole units. This guarantees that all tests in a file run in the same worker.

mslib/_tests/utils.py
ReimarBauer: yes, it gets a port, it can return that port. Anyway the amount of possible free ports have to become reduced

mslib/mscolab/_tests/test_chat_manager.py
ReimarBauer: that is wanted, we need PORTS from that module

mslib/mscolab/_tests/test_file_manager.py
ReimarBauer: that is wanted, we need PORTS from that module

conftest.py
ReimarBauer: This way it definitly works with multiprocessing. Once we have a similiar fast way to get always unique unused ports into each test we can use fixtures again. if you have 6 cores conftest is executed 6 times and these processes did not know which port the others had already. Sometimes they are so fast that some find a free port and assign it for two tests.

I do later a test if a fixture uses the input of a module. I don't know currently.

ReimarBauer: there can be also more improved after the wms tests are refactored too. But that should be a different step, not related to that issue.

mslib/mscolab/_tests/test_chat_manager.py
ReimarBauer: At the moment I have learned how I can use a fixture in the class this will be moved to conftest. I want only one place where I have to add self.window.show().

mslib/mscolab/_tests/test_file_manager.py
ReimarBauer: yeah best place is conftest, but I don't know currently how to use it from there

ReimarBauer: the needed fixtures could be done by a factory. But this raises some new questions of refactoring I do need more time to find a good solution. 

We should merge this already because further steps could also done with the wms part.

joernu76 (approved):

PR #709, opened 2021-03-07
Solved issue #644

### LINKED ISSUE #644: kml overlay should show the selected color close to the filename
It is difficult with multiple kml files in a list to find the right one. A double click opens the navigation to updates,

kml_colorpicker

We want right off the filename a box with the color used and optional that it can be changed by a click
Same for the line style.

Something like: 

kml

Sources:
- https://github.com/Open-MSS/MSS/blob/develop/mslib/msui/ui/ui_kmloverlay_dockwidget.ui
- https://github.com/Open-MSS/MSS/blob/develop/mslib/msui/kmloverlay_dockwidget.py

risehr:
**kml overlay should show the selected color close to the filename #644.** 

Added functionality to edit linewidth directly from KML overlay dockwidget.
Added a change color button.
Set icons of color of the KML files just before them in the listWidget whose heights corresponds to the selected file's linewidth.
Added message boxes to display messages to guide them for this functionality.
Run flake8 tests


mslib/msui/kmloverlay_dockwidget.py
```diff
@@ -319,59 +321,72 @@ def __init__(self, parent=None, view=None):
...
         }
         save_settings_qsettings(self.settings_tag, settings)
 
+    def set_color_icons(self):
+        """
+        Shows the color of individual KML files set as icons before them whose heights
+        are in correspondence to the linewidth(thickness) of the KML file.
+        """
+
+        if(self.listWidget.count() != 0):
```
joernu76: This looks like C. Please follow the PEP8 style guidelines (https://www.python.org/dev/peps/pep-0008/). You can use the flake8 tool to test your code for style and syntax errors. The merge request cannot be accepted until all tests pass, one of which is the flake8 tool.

mslib/msui/qt5/ui_kmloverlay_dockwidget.py
```diff
@@ -2,16 +2,24 @@
```
joernu76: I just tested your code. The idea was to have one lw/color field per KML file. Not global buttons. The buttons are not too shabby, but one still cannot see, with one glance, which color belongs to which KML file.

joernu76 (changes requested):

mslib/msui/kmloverlay_dockwidget.py
ReimarBauer: flake8 was successful. but the style ( ) is not needed here

risehr: Sorry for this style ( ) . I will remove those.

mslib/msui/qt5/ui_kmloverlay_dockwidget.py
risehr: @joernu76 can you please guide me through the changes that you want here? or exactly what to add more?

risehr:
I have modified the test_kmloverlay_dockwidget.py and all the tests have been successfully passed.

mslib/msui/kmloverlay_dockwidget.py
risehr: The style ( ) for if has been removed.

mslib/msui/kmloverlay_dockwidget.py
```diff
@@ -290,10 +290,13 @@ def __init__(self, parent=None, view=None):
...
 
-        self.dialog = CustomizeKMLWidget(self)  # create object of dialog UI Box
-        self.listWidget.itemDoubleClicked.connect(self.open_customize_kml_dialog)
-        self.dialog.pushButton_colour.clicked.connect(self.select_color)
+        # Connect slots and signals (For showing color and linewidth of clicked file).
+        self.listWidget.itemClicked.connect(self.show_linewidth)
+        self.pushButton_color.clicked.connect(self.select_color)
+        self.dsbx_linewidth.valueChanged.connect(self.select_linewidth)
 
         self.listWidget.itemChanged.connect(self.load_file)  # list of files in ListWidget
```
Marilyth: This triggers also when you change the icons for the items.
So whenever you change the colour or linewidth, self.load_file is executed twice in the process which is very slow.
This should first be disconnected before doing any changes to the items, any changes whatsoever.

There may be more such traps you ran into which cause the whole thing to slow down significantly.
Check for any redundant method calls.

Marilyth (dismissed):
I like how it looks.
There are a few things I would like to bring to your attention so you can check them out.

1. When I press up or down in the linewidth editor, it changes 2 steps instead of 1 when many kml files are active. This is partly because of 3., the delay in triggering the button and giving the user control again is causing the button to be triggered again.

2. It would maybe make sense to disable the colour and linewidth tools until a file is selected, instead of warning the user when they click on them.

3. Changing the linewidth or colour is very slow, which is causing other issues like 1. Digging around a little in the code I see some redundant method calls. It would be great if you could try and make this whole thing more performant. If you need help with that, give me a yell :)

mslib/msui/kmloverlay_dockwidget.py
risehr: Okay @Marilyth I would try to work on the points that you mentioned.

risehr: Thanks for the review.

mslib/msui/qt5/ui_kmloverlay_dockwidget.py
joernu76: I will have another look this afternoon and provide some graphical input to you.

joernu76: The current state looks quite improved.
There is a nice "preview" on the left between the tickbox and the filename showcasing the style. Keep that. The change-colour button changes its color depending on the last selected colour, which does not make sense. Remove that. If a new KML file is added, it should use the defaul color black.

The line width widget is fine.

This can be merged soon.

risehr: Thankyou! @joernu76 Yeah, I will change the colour to black. (I have thought upon this before also.) And I am working on the changes that you and @Marilyth proposed.

risehr:
Hello @Marilyth @joernu76 I have addressed all the changes requested by you  both. All the points.
@Marilyth Your points of change have been covered well and working all fine.

My flake8 and pytest have been passed successfully in my system. I don't understand why it has been failing here. 

Please make a review of the changes I have made.

risehr:
The tests have been working fine now. Both locally and here(github). Thanks @ReimarBauer

ReimarBauer (changes requested):
Please remove the file .coverage from the PR

I think it is better to have the background color of the Change Color the default as the other buttons. Not changing the color.
This solves also the problem that it now has the last set color, while I have clicked on a blue line it shows green.  The used color is not in the sub widget, If I won't be careful all lines will switch to green. 

max thickness should be 10 and minimum 0.1. 0 would mean deselected, which we have as function.

risehr:
@ReimarBauer I have made the changes requested.

joernu76 (approved):
Very nice!

ReimarBauer (approved):

PR #710, opened 2021-03-07
Ahelii16 mscolab disconnect fix

ahelii16:
Added _connect to test functions and removed it from _login function.


ahelii16:
@ReimarBauer the tests in test_mscolab.py are now working.

ReimarBauer (changes requested):
there are new empty files in the PR .coverage / .coverage.ah...

please remove those. 

please look at the failing test

mslib/msui/_tests/test_mscolab_version_history.py
```diff
@@ -133,8 +133,8 @@ def test_refresh(self):
         assert new_changes_count == changes_count + 2
 
     def _connect_to_mscolab(self):
-        self.window.url.setEditText("http://localhost:8084")
-        QtTest.QTest.mouseClick(self.window.connectMscolab, QtCore.Qt.LeftButton)
+        self.window.url.setEditText("http://localhost:8083")
```
ReimarBauer: the port is wrong, see for comparison 
https://github.com/Open-MSS/MSS/blob/develop/mslib/msui/_tests/test_mscolab_version_history.py#L136

We try to avoid 8083 because this can get easily in conflict with a developer server

ReimarBauer (changes requested):
see comment about port problem

mslib/msui/_tests/test_mscolab_version_history.py
ahelii16: Alright, I will resolve this

ReimarBauer (changes requested):
it seems that the ui file has newer/different changes than the resulting py file.

was it converted by pyuic5 ?

mslib/msui/qt5/ui_mscolab_window.py
```diff
@@ -254,12 +251,10 @@ def retranslateUi(self, MSSMscolabWindow):
         _translate = QtCore.QCoreApplication.translate
         MSSMscolabWindow.setWindowTitle(_translate("MSSMscolabWindow", "Mission Support Collaboration"))
         self.label_2.setText(_translate("MSSMscolabWindow", "MSColab URL:"))
-        self.connectMscolab.setToolTip(_translate("MSSMscolabWindow", "Connect to Mscolab server"))
-        self.connectMscolab.setWhatsThis(_translate("MSSMscolabWindow", "Connect to mscolab server"))
-        self.connectMscolab.setText(_translate("MSSMscolabWindow", "Connect"))
+        self.toggleConnectionBtn.setToolTip(_translate("MSSMscolabWindow", "Connect/Disconnect to Mscolab server"))
```
ReimarBauer: there is different text in the ui file

ReimarBauer (changes requested):
After a conversion of the ui file by pyuic5 the layout has changes and also text is different.

mslib/msui/qt5/ui_mscolab_window.py
ahelii16: Since a single button is handling both connect and disconnect, I had changed the text a bit. I removed the old disconnect button from qt designer and regenerated the py file so that it gets properly removed. Is that alright?

ahelii16: @ReimarBauer

ReimarBauer: That there the string is changed is correct. But it is not the same as when we generate this file from the ui file.
This file needs to be created by pyuic5 from the ui_mscolab_window.ui file. It looks like an older state.
Also the layout change is not shown in the diff.

ReimarBauer (changes requested):
QVBoxLayout and toggleConnectionBtn differs from a file created by pyuic5
please create it again by pyuic5

the py file is derivated from the ui file. it never gets manual edits, because they are removed on a further conversion.

mslib/msui/qt5/ui_mscolab_window.py
```diff
@@ -297,3 +293,13 @@ def retranslateUi(self, MSSMscolabWindow):
         self.actionCloseWindow.setText(_translate("MSSMscolabWindow", "CloseWindow"))
         self.actionCloseWindow.setShortcut(_translate("MSSMscolabWindow", "Ctrl+W"))
 
+
```
ReimarBauer: None of the others  ui py files has this. Was this made by an additional option? Do you mind to create it without these lines?

ReimarBauer (changes requested):

mslib/msui/qt5/ui_mscolab_window.py
ahelii16: Yes sure I will remove it.

ahelii16: Just realized, I didn't add this. Let me check.

ahelii16: @ReimarBauer please check if it looks fine now. Looks like I was using an unnecessary parameter earlier while generating it.

ReimarBauer (dismissed):
Thx :)

ReimarBauer:
Great, thx, for merging this, I do at the moment the other test is finished.

ReimarBauer (approved):
Thx for this contribution

ahelii16:
You are welcome :)

PR #715, opened 2021-03-12
makes conftest useable for windows devs , fixes #714

### LINKED ISSUE #714: conftest on windows has problems with backslashes
Also our conftest is not optimal currently for windows we found a dependency we asked for help on pyfilesystem2
 
https://github.com/PyFilesystem/pyfilesystem2/issues/454

ReimarBauer:
This PR enables testing for windows users.
Major Test problems are:
 - AttributeError: Can't pickle local object 'app_loader.<locals>.file'
 - PermissionError: [WinError 32] Der Prozess kann nicht auf die Datei zugreifen
 -  fs.errors.InvalidCharsInPath:  

This needs someone with windows python knowledge or iterative work by us.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -101,6 +101,7 @@ class mscolab_settings(object):
     import logging
     import fs
     import secrets
+    from werkzeug.urls import url_join
 
     ROOT_DIR = '{constants.ROOT_DIR}'
     # directory where mss output files are stored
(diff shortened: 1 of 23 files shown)


ReimarBauer:
AttributeError: Can't pickle local object 'app_loader..file'

 Python 3.8 changes the default mode of multiprocessing on MacOS to spawn instead of fork
 
windows uses also spawn.

https://docs.python.org/3/library/multiprocessing.html


fork is default on unix

ReimarBauer:
for windows skipped all test which are based on multiprocessing. We need currently the fork method. windows knows only spawn.    

in a virtualized system with 6 cores used
220 passed, 143 skipped, 2981 warnings in 97.82s (0:01:37)

Marilyth (approved):
The tests run fine on Windows.
Though it is very unfortunate so many tests are skipped on Windows due to no fork support.
Perhaps there is a way to rewrite the tests to work on spawned processes entirely?

joernu76 (approved):

joernu76:
I assume the problem is that spawn processes would not have everything "import"'ed already?
Maybe switch everything to QThreads where we're at it? We need QThreads for the background prefetching of maps anyway and QT offers nice interprocess-communication...?

Marilyth:
Even just switching multiprocessing.Process to threading.Thread seems to have fixed a few test cases.
This is definitely fixable.

ReimarBauer:
> I assume the problem is that spawn processes would not have everything "import"'ed already?
> Maybe switch everything to QThreads where we're at it? We need QThreads for the background prefetching of maps anyway and QT offers nice interprocess-communication...?

we need the flask server code refactored to solve various "AttributeError: Can't pickle local object" for spawn.
Tanish looks on the mscolab code. This is a bit work. 

related to macOS
https://bugs.python.org/issue33725


We meed multiprocessing for our webservices.

PR #718, opened 2021-03-14
Fix for issue #638

### LINKED ISSUE #638: mscolab chat upload/download  did not use configured "filepicker_default": "fs"
This is similar to
https://github.com/Open-MSS/MSS/issues/636

if we configure in mss_settings.json the "fs" filepicker this is currently for mscolab chat upload/download ignored

To verify this you need a config file mss_settings.json to be stored in $HOME/.config/mss

{
 "filepicker_default": "fs"
}

mss_qt.get_open_filename needs to be used

source to look at
https://github.com/Open-MSS/MSS/blob/develop/mslib/msui/mscolab_project.py

please check if the test needs a fix too

aravindm711:


aravindm711:
The tests seems to be working fine with the changes.

mslib/msui/mscolab_project.py
```diff
@@ -231,14 +231,12 @@ def toggle_preview(self):
             self.previewBtn.setText("Write")
 
     def handle_upload(self):
-        img_type = "Image (*.png *.gif *.jpg *jpeg *.bmp)"
-        doc_type = "Document (*.*)"
-        file_filter = f'{img_type};;{doc_type}'
-        file_path, file_type = QtWidgets.QFileDialog.getOpenFileName(self, "Select a file", "", file_filter)
-        if file_path == "":
+        file_filter = ["Image (*.png *.gif *.jpg *jpeg *.bmp)", "Document (*.*)"]
```
ReimarBauer: All these examples use ";;" check this with the docs what is the difference to a ","

https://stackoverflow.com/questions/1604440/how-to-set-selected-filter-on-qfiledialog

ReimarBauer (changes requested):
please check meaning of delimiter change from ";;" to ","

mslib/msui/mscolab_project.py
aravindm711: the file_filter required for QtWidgets.QFileDialog.getOpenFileName is in the format that involves separating filters with ';;' (https://doc.qt.io/archives/qt-4.8/qfiledialog.html#getOpenFileName), but fs.file_picker takes file_filter as a list, that is why the new file_filter is a list of filters.

aravindm711: there was an error with qt filepicker after the changes. I have fixed them now for both fs and qt.

ReimarBauer (approved):
thx

PR #721, opened 2021-03-17
Fix sql versions

Marilyth:
The new versions sqlalchemy 1.4.0 and sqlite 3.35.1 are causing all tests to fail.
Fixing the versions to something before fixes that for the time being.

### DIFF
--- a/localbuild/meta.yaml
+++ b/localbuild/meta.yaml
@@ -72,6 +72,8 @@ requirements:
     - xstatic-bootstrap
     - pyperclip
     - geos <3.9.0
+    - sqlalchemy < 1.4.0
+    - sqlite < 3.35.1
 
 test:
   imports:


localbuild/meta.yaml
```diff
@@ -72,6 +72,8 @@ requirements:
     - xstatic-bootstrap
     - pyperclip
     - geos <3.9.0
+    - sqlalchemy < 1.4.0
```
ReimarBauer, line 75: the syntax is 

- sqlalchemy  <1.4.0

we have no linter here, but conda-forge has. After the module two blanks. Others in the list may have also this problem. No blank between < and 1.4.0

localbuild/meta.yaml
```diff
@@ -72,6 +72,8 @@ requirements:
...
+    - sqlite < 3.35.1
```
ReimarBauer, line 77: My local setup works with this too:

sqlite                    3.35.1               hdfb4753_0

ReimarBauer (approved):

PR #725, opened 2021-03-18
Fix for issue #200

### LINKED ISSUE #200: table view  add ons
**[Original report](https://bitbucket.org/wxmetvis/mss/issue/200) by Reimar Bauer (Bitbucket: [ReimarBauer](https://bitbucket.org/ReimarBauer), GitHub: [ReimarBauer](https://github.com/ReimarBauer)).**

May be the performance dialog could be moved into a docking widget similar as the hexagon flight pattern?

aravindm711:


mslib/msui/qt5/ui_performance_settings.py
```diff
@@ -1,130 +0,0 @@
...
-        self.label_2.setObjectName("label_2")
-        self.horizontalLayout_4.addWidget(self.label_2)
-        self.dsbTakeoffWeight = QtWidgets.QDoubleSpinBox(PerformanceSettingsDialog)
-        sizePolicy = QtWidgets.QSizePolicy(QtWidgets.QSizePolicy.Minimum, QtWidgets.QSizePolicy.Minimum)
-        sizePolicy.setHorizontalStretch(0)
-        sizePolicy.setVerticalStretch(0)
-        sizePolicy.setHeightForWidth(self.dsbTakeoffWeight.sizePolicy().hasHeightForWidth())
-        self.dsbTakeoffWeight.setSizePolicy(sizePolicy)
-        self.dsbTakeoffWeight.setDecimals(0)
-        self.dsbTakeoffWeight.setMaximum(9999999.0)
```
joernu76, line 83: Maximum/Decimals/Step is not set accordingly in new controls. They are upward limited to 99.99, which is wrong.

mslib/msui/qt5/ui_performance_dockwidget.py
```diff
@@ -0,0 +1,88 @@
...
+        self.horizontalLayout_3.addWidget(self.dteTakeoffTime)
+        self.gridLayout.addLayout(self.horizontalLayout_3, 0, 0, 1, 1)
+        self.horizontalLayout_5 = QtWidgets.QHBoxLayout()
+        self.horizontalLayout_5.setContentsMargins(-1, 4, -1, -1)
+        self.horizontalLayout_5.setObjectName("horizontalLayout_5")
+        self.label_5 = QtWidgets.QLabel(PerformanceDockWidget)
+        self.label_5.setIndent(1)
+        self.label_5.setObjectName("label_5")
+        self.horizontalLayout_5.addWidget(self.label_5)
+        self.dsbTakeoffWeight = QtWidgets.QDoubleSpinBox(PerformanceDockWidget)
```
joernu76, line 65: Modifying this does not update the Table.

mslib/msui/qt5/ui_performance_dockwidget.py
```diff
@@ -0,0 +1,88 @@
...
+        self.horizontalLayout_2.addWidget(self.pbLoadPerformance, 0, QtCore.Qt.AlignRight)
+        self.gridLayout.addLayout(self.horizontalLayout_2, 1, 1, 1, 1)
+        self.horizontalLayout_4 = QtWidgets.QHBoxLayout()
+        self.horizontalLayout_4.setContentsMargins(-1, -1, 0, -1)
+        self.horizontalLayout_4.setObjectName("horizontalLayout_4")
+        self.label_4 = QtWidgets.QLabel(PerformanceDockWidget)
+        self.label_4.setIndent(1)
+        self.label_4.setObjectName("label_4")
+        self.horizontalLayout_4.addWidget(self.label_4)
+        self.dsbFuel = QtWidgets.QDoubleSpinBox(PerformanceDockWidget)
```
joernu76, line 41: Modifying this spin box does not update the Table.

joernu76 (changes requested):
Looks already pretty good!

The Fuel/TakeOff weight double-spin-boxes are not yet correctly implemented. Changing the numbers should also update the table (similar to how it works for the takeoff time).
The numerical limits7steps/etc of the weight and fuel should be the same as those of the old spin boxes.

mslib/msui/qt5/ui_performance_settings.py
aravindm711: I have fixed this in the recent commit. Also the fuel value had the same constrains, fixed that as well.

mslib/msui/qt5/ui_performance_dockwidget.py
aravindm711: Changing the values of the spinbox actually does change the values in the waypoints model. Even though the AbstractTableModel is triggered with a dataChanged.emit, the table does not reflect the changes. Same goes for the other spinbox. It seems that i'm stuck at this point. could please provide some assistance @ReimarBauer ?

joernu76: I will try to have a look at this.

joernu76: OK, this here seems to fix the issue:
```
 (e) base  ~  src  mss  git diff                                                                                                                 table-view-issue-200 
diff --git a/mslib/msui/tableview.py b/mslib/msui/tableview.py
index 9293cd84..3745bbb8 100644
--- a/mslib/msui/tableview.py
+++ b/mslib/msui/tableview.py
@@ -94,6 +94,7 @@ class MSSTableViewWindow(MSSViewWindow, ui.Ui_TableViewWindow):
         self.waypoints_model.update_distances(0)
         self.waypoints_model.save_settings()
         self.resizeColumns()
+        self.tableWayPoints.viewport().repaint()
 
     def openTool(self, index):
         """Slot that handles requests to open tool windows.
```
I got the solution from here
https://stackoverflow.com/questions/20256493/qtableview-doesnt-respond-to-datachanged-when-not-focused
From that link, I suspect the issue occurs more often than just. I would suggest to connect the dataChanged signal with a viewport-update-function-slot as suggested. A bit more googling might be insightful.

aravindm711: Okay, I'll look into it.

aravindm711: I have fixed the update issue and also added required test for the new dock widget.

mslib/msui/qt5/ui_performance_dockwidget.py
aravindm711: Fixed this problem in the latest commit.

joernu76 (approved):
Code looks good!

PR #728, opened 2021-03-22
Multilayering favourites and fixes

### LINKED ISSUE #724: Iteration of multilayering
1) When selecting the multiplayering tickbox, the priority column should always be visible (less confusing context changes)
2) The priority column and the styles column overlap. Perhaps move the priority column to the front?
3) The Layer-Selection Dialogue should not "retrieve" maps, just "select" them....? I am unsure about this.
4) Switching from one layer to another should keep the previous init_time/valid_time/altitudes as good as possible. (One is typically interested in, say 12km, and wants to look at different maps of that altitude/time combination. 
  *) In case no match is possible, pick a similar value? Convert pressure to km and vice versa? Keep the last value for this general  type of layer and reuse that? This was heavily annoying in the past, but is probably difficult to do "right".
5) Favorites! Could one add a filled/unfilled "Star" column in the front and add another tickbox "show only favorites"? (low priority for now)
6) Add a button to clear the Filter. (possibly some broom like icon at the rightmost place of the box, or beside it.
7) readd a feature to view the layer "abstract". maybe by hovering over the feature? I actually plan to provide better captions here. Maybe readd a box/viewbutton for that to the main display?
8) The minimal horizontal extent for the WMS control can be reduced. It is currently quite a bit wider than it needs to be and drives the minimal width of the MSS TopView Window.
9) Can we generally get rid of the "title" of the dockwidgets if it is docked? We have now "WMS (TopView)" on top of the widget and on the tab below. When docked, the title is superfluous, when undocked it is necessary.
10) Accessing this server: http://osmwms.itc-halle.de/maps/osmfree produces an error.
11) The "view" button is tied to the selected layer, not the WMS server. This is unfortunate. Better tie it to the displayed server.

Marilyth:
Handles the points mentioned in #724

- [x] 1) Priority column always visible on multilayering mode
- [x] 2) Hopefully fix priority and style overlapping
- [x] 3) Remove "retrieve" in layers window
- [x] 4) Keep the level, itime and/or vtime from the previous selection if possible
- [x] 5) Favourite functionality has been added
- [x] 6) Button to clear the filter has been added
- [x] 7) Abstracts were always visible via hovering, just mentioning it here again
- [x] 8) Dockwidget and Multilayers window have been reduced in width a bit, and can be further reduced by the user
- [ ] 9) It is possible to hide the title bar while docked, or to hide the title itself while docked. Both cause big issues however so I would rather let it be.
- [x] 10) Fix error on e.g. http://osmwms.itc-halle.de/maps/osmfree
- [x] 11) Tie "view" button to the WMS in the box instead of the current layer.


joernu76 (approved):
Nice!

mslib/msui/multilayers.py
```diff
@@ -114,6 +159,8 @@ def filter_multilayers(self, filter_string=None):
             else:
                 header.setHidden(False)
 
+        self.filterRemoveAction.setVisible(self.filter_favourite or len(filter_string) > 0)
```
ReimarBauer, line 162: I'm not sure if we should strip() blanks too.

Issue #327, opened 2018-03-01
refactor os.path to fs.path


ReimarBauer:
**[Original report](https://bitbucket.org/wxmetvis/mss/issue/327) by Reimar Bauer (Bitbucket: [ReimarBauer](https://bitbucket.org/ReimarBauer), GitHub: [ReimarBauer](https://github.com/ReimarBauer)).**

----------------------------------------

Only on a few OS and filesystems os.path behaves similar to fs.path. 
We have to review the usage of os.path

ReimarBauer:
**Original comment by Reimar Bauer (Bitbucket: [ReimarBauer](https://bitbucket.org/ReimarBauer), GitHub: [ReimarBauer](https://github.com/ReimarBauer)).**

----------------------------------------

due that we have no failing test we move this to the next major

ReimarBauer:
if that is needed create a specific issue.

PR #732, opened 2021-03-24
Config

### LINKED ISSUE #626: configuration file update did not initialize msui
After saving a new configuration the msui is not restarted as in the past and the new data is not used.

restart is one option, without a restart and using the new configuration would be good too.

### LINKED ISSUE #730: MSS Config Editor does not properly close
When the config editor was opened, a reference to the editor is stored in the main GUI, which is not removed upon destruction of the editor. When the main GUI is closed, the Editor is thus closed again, in some, fashion, which causes another "are you sure..." dialog to be presented to the, confused, user.

The Configuration Editor should be blocking and provide an option to restart the GUI when the configuration was changed.

In my version, there also seems to be some issues with saving the configuration.

The Editor should have an obvious way to be closed in addition to the window manager "x" in the title bar like "save and exit" or "apply and exit".

joernu76:
Because the configuration load capability is fundamentally broken, this needs to go into stable.


joernu76:
Also fixes #626

mslib/msui/editor.py
```diff
@@ -51,22 +56,20 @@ def __init__(self, parent=None):
         self.editor = QtWidgets.QPlainTextEdit()
 
         # Load mss_settings.json (if already exists)
-        self.mss_path = fs.path.join(MSS_CONFIG_PATH, "mss_settings" + ".json")
-        if fs.open_fs(MSS_CONFIG_PATH).exists("mss_settings.json"):
-            file_name = fs.path.basename(self.mss_path)
-            with fs.open_fs(fs.path.dirname(self.mss_path)) as file_dir:
-                self.file_content = file_dir.readtext(file_name)
+        self.path = constants.CACHED_CONFIG_FILE
+        if self.path and os.path.exists(self.path):
```
ReimarBauer: with using os self.path can't be a FS Url. This should be replaced by a https://docs.pyfilesystem.org/en/latest/reference/base.html#fs.base.FS.exists

ReimarBauer (changes requested):
I have not tried it yet, seen some os. stuff used. os.  needs to become replaced.

mslib/msui/editor.py
joernu76: So I have to first split the name in directory, and filename, open the directory as file system and then for the existance of the filename in the file system? That seemed a bit convoluted to me.

joernu76: I have implemented an fs-based replacement, but the new code is way less stable than the previous one. I do not know fs well enough to make it better.
E.g. can the "split" command fail? What is when the "dirname" cannot be opened?

ReimarBauer: If you have selected a file name in a browser just before someone changes the dir or the rights of that, this gives other problems. 
I will later do a review of all fs used lines, that can be done in this #327

ReimarBauer (approved):
did a few manual tests, thx.

PR #735, opened 2021-03-24
Reworked add project dialogue to properly show file name.

### LINKED ISSUE #734: MSCOLAB Add project dialogue has strange location for selected filename
Adding an existing file as a project shows the name of the file in the description box.

joernu76:
Fix issue #734

### DIFF
--- a/mslib/msui/qt5/ui_add_project_dialog.py
+++ b/mslib/msui/qt5/ui_add_project_dialog.py
@@ -2,43 +2,46 @@
 
 # Form implementation generated from reading ui file 'mslib/msui/ui/ui_add_project.ui'
 #
-# Created by: PyQt5 UI code generator 5.9.2
+# Created by: PyQt5 UI code generator 5.12.3
 #
 # WARNING! All changes made in this file will be lost!
 
+
 from PyQt5 import QtCore, QtGui, QtWidgets
+
 
 class Ui_addProjectDialog(object):
     def setupUi(self, addProjectDialog):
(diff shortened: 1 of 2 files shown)


mslib/msui/qt5/ui_add_project_dialog.py
```diff
@@ -2,43 +2,45 @@
 
 # Form implementation generated from reading ui file 'mslib/msui/ui/ui_add_project.ui'
 #
-# Created by: PyQt5 UI code generator 5.9.2
+# Created by: PyQt5 UI code generator 5.12.3
```
ReimarBauer, line 5: this is also the version of qt we currently can't have on a fresh installation on linux, I check this tom. with the old version. Anyway after we changed the channel order we will all have 5.12.9

ReimarBauer (changes requested):
old_new

on the left site is the old on the right the new one.
I agree to the explaination text, but it looks like an input field because of the white color. 
Can it become a label or a changed background?

just for description,  uploading a file is not mandatory, it will overwrite the provided server code.

joernu76:
Mmh. The "label" had the problem that it is invisible and has no frame  or similar. I can try to "disable" the element, which might to the visual tricks. 
Any suggestions to a widget showing a text that is *not* editable? @Marilyth

joernu76:
I "disabled" the dialogue such that it is now grey.
We can certainly redesign the flow, but that is out-of-scope for my fix.

ReimarBauer (approved):
Thx

PR #741, opened 2021-03-27
Stable to develop

ReimarBauer:
updating meta.yaml and other fixes


.github/workflows/testing.yml
```diff
@@ -31,16 +31,16 @@ jobs:
         && conda config --add channels defaults \
         && conda install --revision 0 \
         && conda install mamba \
-        && mamba install --file reqs.txt \
+        && conda install --file reqs.txt --force-reinstall \
```
ReimarBauer: mamba crashed with force-reinstall

.github/workflows/testing.yml
```diff
@@ -17,7 +17,7 @@ jobs:
 
     steps:
     - uses: actions/checkout@v2
-        
+
     - name: Reinstall dependencies if changed
       run: |
         cmp -s /meta.yaml localbuild/meta.yaml && cmp -s /development.txt requirements.d/development.txt || (echo Dependencies differ \
```
ReimarBauer, line 23: this seems to be the origin meta.yaml not the new one

.github/workflows/testing.yml
ReimarBauer: and at all force-reinstall is not needed, because of conda install --revision 0

ReimarBauer:
reproduced with the docker image. 
now lets see if the reason is the order change

ReimarBauer:
testing changes need afterwards a backport to stable

joernu76 (approved):

PR #744, opened 2021-03-29
Preparations for 2.0.2

ReimarBauer:


joernu76 (approved):
Is the change in the lBlChanges really suitable for stable?

ReimarBauer:
> Is the change in the lBlChanges really suitable for stable?

the Changes file usually is mostly read by developers. We could also add text. At all we should add meanigful titles to the issue reports.

Issue #736, opened 2021-03-24
Dock-widgets in TableView prevent TableView from closing upon application exit


joernu76:
Steps to reproduce the problem:
1) Open a TableView
2) Open a dock widget
3) "Quit" the MainWindow
Expected Outcome:
4) All windows close
Outcome;
4) TableView window remains visible.

The TopView has dockwidgets that work in this regard. One should be able to transfer the proper handling from TopView to TableView.

risehr:
Hi @joernu76  @ReimarBauer ! I have solved this issue. Tests also passed. I have send a PR to develop.

PR #753, opened 2021-04-05
(Solved)Dock-widgets in TableView prevent TableView from closing upon application exit #736

risehr:
Solved the issue that prevented tableview to be closed automatically when MSS Main Window is closed. Previously, the tableview with its dock-widget continued to stick around on the screen even when the main window has been exited. The tableview then,along with the hexagon dockwidget had to be closed manually after which the MSS application finally quits. But now this problem has been solved.

On closing main MSS window, all the child window that it had opened, gets closed.


mslib/msui/mss_pyui.py
```diff
@@ -218,6 +218,8 @@ def __init__(self, *args):
 
         # Status Bar
         self.labelStatusbar.setText(self.status())
+        # Instantiates the tableview window object as None when started.
+        self.tv = None
```
ReimarBauer: self.tv_window may be later better understandable

mslib/msui/mss_pyui.py
```diff
@@ -416,6 +418,15 @@ def closeEvent(self, event):
                 self.mscolab_window.close()
             if self.config_editor is not None:
                 self.config_editor.close()
+            """
```
ReimarBauer, line 421: I think with self.tv_window it is clear what it is, also later the history will connect to the PR. Please compress this to a single comment line

mslib/msui/viewwindows.py
```diff
@@ -59,6 +59,8 @@ def __init__(self, parent=None, model=None, _id=None):
         self._id = _id
         # Used to force close window without the dialog popping up
         self.force_close = False
+        # Flag variable help in force closing main window with tableview still open.
+        self.tvflag = 0
```
ReimarBauer: self.exists = True

mslib/msui/viewwindows.py
```diff
@@ -85,10 +87,20 @@ def closeEvent(self, event):
             if self._id is not None:
                 self.viewClosesId.emit(self._id)
             logging.debug(self._id)
+            # sets a flag which assists during MSS main window closure.
+            self.tvflag = 1
```
ReimarBauer: self.exists = False

mslib/msui/viewwindows.py
```diff
@@ -85,10 +87,20 @@ def closeEvent(self, event):
...
             event.accept()
         else:
             event.ignore()
 
+    def flag_tv(self):
+        """
+        Returns the flag 1 if self.closeEvent() is triggered else returns 0.
```
ReimarBauer, line 98: this is boolean behaviour, or?

window_exists may be as name?

mslib/msui/mss_pyui.py
```diff
@@ -416,6 +418,15 @@ def closeEvent(self, event):
...
+            tv = tableview ; it is an object of MSS Table View Window
+            Tableview was not closing even when main window closes
+            But after this, it closes fine.
+            """
+            if self.tv is not None:
+                if self.tv.flag_tv() == 0:  # checks whether or not tableview window has closed before
```
ReimarBauer: see comments below, a better name for this and I think we can use a boolean.

ReimarBauer (changes requested):
please send a PR to stable.

good catch :)

mslib/msui/mss_pyui.py
risehr: Yeah, we can use a boolean. No problem in that. Also, tv_window_exists sounds like a better name for this boolean behaviour. I have noted all points. I will send a PR to stable. Thanks.

risehr:
I have already created a new PR into stable, that is, PR #758 
Hence, I am closing this pull request.

PR #758, opened 2021-04-06
(Solved issue #736 for stable)Dock-widgets in TableView prevent TableView from closing upon application exit

risehr:
Solved the issue that prevented tableview to be closed automatically when MSS Main Window is closed. Previously, the tableview with its dock-widget continued to stick around on the screen even when the main window has been exited. The tableview then,along with the hexagon dockwidget had to be closed manually after which the MSS application finally quits. But now this problem has been solved.

On closing main MSS window, all the child window that it had opened, gets closed.


risehr:
Hey @ReimarBauer, as you said I have created this new PR into stable. I am closing the previous pull request(PR #753 into develop)

Solved issue #736 !

mslib/msui/mss_pyui.py
```diff
@@ -419,6 +421,11 @@ def closeEvent(self, event):
                 self.mscolab_window.close()
             if self.config_editor is not None:
                 self.config_editor.close()
+            # tv = tableview ; it is an object of MSS Tableview window. Used for force closing it.
+            if self.tv_window is not None:
+                if self.tv_window.tvwindow_exists() is True:  # checks whether or not tableview window has closed before
```
ReimarBauer: below it is written as self.tv_window.tv_window_exists

only exists() would be sufficient.

see for example os.path.exists()

best practise it would be written for booleans if self.tv_window.exists(): see PEP8 too or,
https://docs.quantifiedcode.com/python-anti-patterns/readability/comparison_to_true.html

ReimarBauer (changes requested):
we do not to have a test for this functionality. found a typo ;)

mslib/msui/mss_pyui.py
risehr: Okay, I am rectifying that exists() part and PEP8 style. But do I need to write the tests?

ReimarBauer: you can try if you want. For bugs it is with starting a test sometimes easier to find the correct fix. So one makes first a test which fails by the bahaviour than the bug fix and the before failing test succeeds. It can also save a lot of time (that time one needs to try out different states). 

It is not needed now. We could later add tests which look on states of different windows when expected closed.

risehr: Alright then. I will try the tests later when we increase test coverage for other codes. For now, its done. I have already updated the changes that you had requested.

ReimarBauer (approved):
verified manually.

thx.

risehr:
Thankyou for the motivation and help ;)

PR #759, opened 2021-04-07
Hotfixes

### LINKED ISSUE #761: dataaccess units problems
we found issues with convertion of units in our generic plotstyles

joernu76:


mslib/mswms/dataaccess.py
```diff
@@ -217,7 +217,7 @@ def _parse_file(self, filename):
             init_time = netCDF4tools.num2date(0, time_var.units)
             if not self.uses_inittime_dimension():
                 init_time = None
-            valid_times = netCDF4tools.num2date(time_var[:], time_var.units)
+            valid_times = netCDF4tools.num2date(time_var[:], getattr(time_var, "units", "1"))
```
ReimarBauer: https://unidata.github.io/netcdf4-python/#num2date

units: a string of the form since describing the time units. can be days, hours, minutes, seconds, milliseconds or microseconds. is the time origin. months_since is allowed only for the 360_day calendar.

on some other place it is described as  units: A string with the format "{unit} since {epoch}",

I assume "1" is not fail safe. If exactly that is needed, please add a comment too

ReimarBauer (changes requested):
see my comment about units please

mslib/mswms/dataaccess.py
joernu76: Nice catch. Mistake in applying the hotfixes from campaign2019 to stable. Should be better now.

joernu76:
@gisi90 Some of these changes might make your life easier as well...

ReimarBauer (approved):

PR #766, opened 2021-04-08
merge stable to develop

ReimarBauer:


ReimarBauer:
first make a local copy of the MAIN 

merge stable into develop
pull this into your local fork
push 
and PR  to develop

joernu76 (dismissed):

Marilyth (commented):
There are two things I am not sure might cause issues lateron, depending on what this PR is for exactly.
1. c380449a58ee5ca90ad6e246d490582b897b8068 is not yet in stable.
2. You merged stable into develop, which Jörn told us not to do.

Are these done on purpose?

ReimarBauer:
> There are two things I am not sure might cause issues lateron, depending on what this PR is for exactly.
> 
> 1. [c380449](https://github.com/Open-MSS/MSS/commit/c380449a58ee5ca90ad6e246d490582b897b8068) is not yet in stable.
> 2. You merged stable into develop, which Jörn told us not to do.
> 
> Are these done on purpose?

We discussed the merge on the meeting. We do this from a released version to get all this multiple rebase questions done. We did this 5 years that way and changed the workflow on github. 

Yes I know, this was the conflict solution to get this mergeable. in develop the setup of tests are different and I found a general problem, see https://github.com/Open-MSS/MSS/issues/768 In this PR it is too much to investigate this further.
. [c380449](https://github.com/Open-MSS/MSS/commit/c380449a58ee5ca90ad6e246d490582b897b8068)

The next major is also a merge from develop to stable.

joernu76:
I am not sure about this. I think we agree that we need to merge in *one* direction to maintain sanity. Have we settled already on the need for merging in two directions? Probably. Rebasing only one branch causes issues down the line as it messes with the automatic conflict resolution.
So I think we have just two workable options:
1) merge in both directions, which also causes issues from time to time, but should be manageable
2) never merge from develop to stable, but have one branch for each release version and merge (rebased) hotfixes back to develop.

As 2) would require and generate quite a lot of branches (in contrast to tags), which we do not desired, we currently may need to settle for 1).

joernu76 (approved):

ReimarBauer:
> I am not sure about this. I think we agree that we need to merge in _one_ direction to maintain sanity. Have we settled already on the need for merging in two directions? Probably. Rebasing only one branch causes issues down the line as it messes with the automatic conflict resolution.
> So I think we have just two workable options:
> 
> 1. merge in both directions, which also causes issues from time to time, but should be manageable
> 2. never merge from develop to stable, but have one branch for each release version and merge (rebased) hotfixes back to develop.
> 
> As 2) would require and generate quite a lot of branches (in contrast to tags), which we do not desired, we currently may need to settle for 1).

I also looked again how others proceed with this situation, bigbluebutton for example keeps branches of released versions. so there is a 2.3.x .. 1.1.x .. to the beginning and they have tags also. I would do this similiar if we get into a situation where we have to do long term support for older releases.  As long as there is no force on us I do prefer to have only two branches for development. With these  multiple branches we would also have backports, which we currently don't. We fix in stable and move that forward. 

I've not completly understand  why there are so many times the same merge conflicts when doing the rebase to develop from a stable commit. Probably that can be improved.

PR #767, opened 2021-04-08
Show loading bar on get capabilities

### LINKED ISSUE #222: proceedin bar missing  on get capabilities
**[Original report](https://bitbucket.org/wxmetvis/mss/issue/222) by Reimar Bauer (Bitbucket: [ReimarBauer](https://bitbucket.org/ReimarBauer), GitHub: [ReimarBauer](https://github.com/ReimarBauer)).**

we should change the cursor to a  proceeding bar at the moment we click  get capabilities.

### LINKED ISSUE #764: get capabilities does not visibly block the interface
Using get capabilities currently blocks the UI. This is not apparent. There should be a blocking message box or similar (potentially including a Cancel button) indicating this. Currently, one may click around without effect until the capability document was retrieved.

Marilyth:
fixes #764
1. When get capabilities is pressed, a modal progress dialog pops up.
2. Moves the client server communication to a background thread and roughly update the user on the progress.
3. Create "mslib.utils.Worker" class, to encourage MSS to go async and stay responsive for long running tasks in the future.


ReimarBauer (changes requested):
mslib.worker looks rather small, are there reasons for a standalone module besides adding it to mslib.utils?

mslib/msui/_tests/test_sideview.py
```diff
@@ -170,7 +170,8 @@ def query_server(self, url):
         QtWidgets.QApplication.processEvents()
         QtTest.QTest.mouseClick(self.wms_control.multilayers.btGetCapabilities, QtCore.Qt.LeftButton)
         QtWidgets.QApplication.processEvents()
-        QtTest.QTest.qWait(2000)
+        while self.wms_control.cpdlg.isVisible():
+            QtTest.QTest.qWait(100)
```
ReimarBauer, line 174: shorter wait :))

mslib/msui/wms_control.py
```diff
@@ -547,51 +554,83 @@ def initialise_wms(self, base_url, version="1.3.0"):
...
+
+                # update the combo box, if entry requires change/insertion
+                found = False
+                current_url = self.multilayers.cbWMS_URL.currentText()
+                for count in range(self.multilayers.cbWMS_URL.count()):
+                    if self.multilayers.cbWMS_URL.itemText(count) == current_url:
+                        self.multilayers.cbWMS_URL.setItemText(count, base_url)
+                        self.multilayers.cbWMS_URL.setCurrentIndex(count)
+                        found = True
+                        break
```
ReimarBauer, line 575: Why is it needed to do this twice?

```
if self.multilayers.cbWMS_URL.itemText(count) == current_url:
    self.multilayers.cbWMS_URL.setCurrentIndex(count)
```

mslib/msui/wms_control.py
```diff
@@ -547,51 +554,83 @@ def initialise_wms(self, base_url, version="1.3.0"):
...
+                    logging.debug("inserting URL: %s", base_url)
+                    add_wms_urls(self.multilayers.cbWMS_URL, [base_url])
+                    self.multilayers.cbWMS_URL.setEditText(base_url)
+                    save_settings_qsettings('wms', {'recent_wms_url': base_url})
+
+                self.activate_wms(wms)
+                WMS_SERVICE_CACHE[wms.url] = wms
+                self.cpdlg.close()
+
+        def on_failure(e):
```
ReimarBauer, line 589: I'm not sure  if it gets better readable later for me if instead of (e) we have here (wms) as used for the on_sucess() too.

Am I correct that it is derivated from Exception?

mslib/msui/wms_control.py
```diff
@@ -613,6 +652,21 @@ def display_progress_dialog(self):
...
+        self.cpdlg.setModal(True)
+        self.cpdlg.show()
+
+    def stop_capabilities_retrieval(self):
+        logging.debug("stopping capabilities retrieval")
+        try:
+            self.capabilities_worker.quit()
+            self.capabilities_worker.disconnect()
+        except TypeError:
+            pass
```
ReimarBauer, line 668: is it worth to add a logging here?

ReimarBauer (changes requested):
a few questions / comments.

mslib/msui/wms_control.py
Marilyth: The TypeError just means that there are no slots connected to the worker.
Basically I expected that case to be outright ignored but QT still throws an exception for it, which is completely irrelevant.

mslib/msui/wms_control.py
Marilyth: The failed signal returns an Exception. I thought e would further signify this.

mslib/msui/wms_control.py
Marilyth: An error on my part. It is fixed now.

joernu76 (approved):

mslib/msui/wms_control.py
ReimarBauer: yes it does

I compared the two functions (top down reading) 

I should review my comments than bottom up.

ReimarBauer:
nice that update feature works as expected.

mslib/msui/wms_control.py
ReimarBauer: ts ts ts QT

ReimarBauer (approved):

Issue #658, opened 2021-02-01
remove the dockerfiles directory from the repository


ReimarBauer:
We have on the organisation Open-MSS a dedicated repository for builds on dockerhub

https://github.com/Open-MSS/dockerhub

We should remove the old outdated dockerfiles directory from the development branch.

ReimarBauer:
We moved to https://hub.docker.com/orgs/openmss/repositories

so we should only have one place where we manage/develop Dockerfiles 
This is https://github.com/Open-MSS/dockerhub/

Aut0R3V:
\assign

Aut0R3V:
Hi I'd like to work on this. What I've understood is that the files on https://github.com/Open-MSS/dockerhub need to be removed since they are no longer required. @ReimarBauer Please tell if this is right.

ReimarBauer:
> Hi I'd like to work on this. What I've understood is that the files on https://github.com/Open-MSS/dockerhub need to be removed since they are no longer required. @ReimarBauer Please tell if this is right.

This is a misunderstanding. We have an old version in the repository, see https://github.com/Open-MSS/MSS/tree/develop/dockerfiles

This was moved to the new project but not removed from the repository.

PR #786, opened 2021-04-15
Remove dockerfiles, issue #658

ReimarBauer:
Since we have an own repository the files in this directory are not used anymore


ReimarBauer:
something with that PR wents wrong

PR #788, opened 2021-04-15
replaces pyvirtualdisplay by xvfb-run, fixes issue #743

### LINKED ISSUE #743: remove pyvirtualdisplay
We want to use only xvfb-run  on *nix systems, because this is already with xvfb installed.
This needs changes in docs, requirements.d/development.txt and test actions

ReimarBauer:


joernu76 (approved):

joernu76:
Mmmh, rebasing it on HEAD broke it...

ReimarBauer:
Without pyvirtualdisplay the current github configuration with our modules creates for the xdist tests

```
_pytest.config.ConftestImportFailure: ImportError: Cannot load backend 'Qt5Agg' which requires the 'qt5' interactive framework, as 'headless' is currently running (from /__w/MSS/MSS/conftest.py)
```

this comes from
lib/python3.9/site-packages/matplotlib/pyplot.py +2469

```
# If rcParams['backend_fallback'] is true, and an interactive backend is
# requested, ignore rcParams['backend'] and force selection of a backend that
# is compatible with the current running interactive framework.

if (rcParams["backend_fallback"]
        and dict.__getitem__(rcParams, "backend") in (
            set(_interactive_bk) - {'WebAgg', 'nbAgg'})
        and cbook._get_running_interactive_framework()):
    dict.__setitem__(rcParams, "backend", rcsetup._auto_backend_sentinel)
# Set up the backend.
switch_backend(rcParams["backend"])
```
when we use pyvirtualdisplay in conftest this seems to make a difference for pytest-xdist.
But if we have to use pyvirtualdisplay than we have no need for xvfb-run.

Issue #726, opened 2021-03-19
Set semi-transparent background to colorbars


bkirbus:
For **horizontal views**, a fully white or semi-transparent white **background behind the colorbar** would be nice.
* Thus, it is easier to read all labels, title, etc. on the colorbar
  * Especially in cases where the maps in the background become very "noisy"

joernu76:
The HS_GenericStyle in mslib/mswms/mpl_hsec_styles.py provides a white background behind the colorbar labeling, which solves this issue in an elegant way (by adding a path effect to the labels).

This solution should be applied to all plotting classes in mslib/mswms/mpl_hsec_styles.py

PR #791, opened 2021-04-15
Consistently apply white background for in-set colorbar labels.

joernu76:
Also fixed a flake8 line length warning.

Fix issue #726


Marilyth (approved):
I think this does what it is supposed to. Is there some quick and easy way to make the white background (and colorbar) also go over contours like this?
Screenshot at 2021-04-15 17-18-06

ReimarBauer:
The other contour lines are covered by the legend.

joernu76:
> The other contour lines are covered by the legend.

What you see are probably the countries drawn by the client on top of the WMS plot. This is why the black line also covers the colorbar. This feature should be disabled by default. Most of *our* layers have countries already within and with thee multilayer feature the server could do that as well.

PR #793, opened 2021-04-15
version string changed to canonincal name on an installation.

### LINKED ISSUE #731: display canonical version string on installed version
Sometimes it is useful not only to talk about MSS versions, seeing which build it is can be better

We should find an easy way to show a string similar to this in the help info of the gui

```
mamba list -c | grep mss

conda-forge/linux-64::mss-2.0.1-py39hf3d152e_0
```

Some problems may occure because the package was installed in the little time frame some library was updated and we have not created a new build.

If I would see this string and together a bug report with sqalchemy problems I could immediatly tell, take the _1 we fixed it :)

ReimarBauer:
if mss is installed, now the canonincal name is displayed by mss -v

on a local build:
- mss-alpha-py39_1001

if mss-2.0.4 had that feature: 
- 'mss-2.0.4-py39hf3d152e_0'

if we have a developer version:
- 2.0.4.


ReimarBauer:
I have added this now to the command mss -v 

I can also add it to the gui too.

joernu76 (dismissed):
I am unsure how robust this solution is. I strongly dislike it, but couldn't quickly come up with a better one.

ReimarBauer:
we can postpone it, I've no hurry for this. We can wait for a better solution.

ReimarBauer:
I keep this idea for later. We can try this in some smaller tool first on all platforms. Probably someone had a better idea without that file i/O request

joernu76:
Is there an issue for this? The currently "best" solution I have is along the lines of parsing the output of `conda list`, which is equivalent to what you offered. 
We should keep this discussion in an issue.

ReimarBauer:
see linked issue, in the upper right navigation
display canonical version string on installed version #731

PR #809, opened 2021-04-20
Reload wms on difference

### LINKED ISSUE #805: Capability document not updated?
Steps to reproduce the problem.
develop branch. 
1. getCapabilties of a server
2. change server setup
3. getCapabilties of the same server

"view" should show the new capabilities.
Currently, it shows the first ones.

Marilyth:
Fixes #805 
If the capabilities document has changed, deletes the current entries and reloads them.
For now this also resets all multilayer selections corresponding to the changed document, which could be fixed in a later update.


joernu76 (commented):
I have not understood *how* it solves the problem, but I verified that it *does* solve it :-)

joernu76 (approved):

PR #821, opened 2021-04-21
Close on doubleclick, auto update on layer change

### LINKED ISSUE #816: Simplify use of Server/Layer window
In the Server/Layer window the available parameters to display can be chosen.
If one parameter is clicked (and a color scale has been chosen) the information is transferred to the main window.
(1) depending on the avaiable work screen, the  Server/Layer window is likely overlapping the main (top view) window and such blocks the sight to the main window. There should be an easy possibility to close this Server/Layer window after the parameter choice, either by double-klick on the parameter or by introducing a "dismiss" button. Or both.
(2) If the parameter has been chosen and the "Auto-Update" is chosen in the main window, the parameter should be displayed immedeately to avoid another klick of "retrieve"

Marilyth:
Fixes #816 
Double clicking on a layer now closes the layer window.
Switching layers now triggers auto-updating of the map.

### DIFF
--- a/mslib/msui/multilayers.py
+++ b/mslib/msui/multilayers.py
@@ -54,6 +54,7 @@ class Multilayers(QtWidgets.QDialog, ui.Ui_MultilayersDialog):
         self.listLayers.itemChanged.connect(self.multilayer_changed)
         self.listLayers.itemClicked.connect(self.multilayer_clicked)
         self.listLayers.itemClicked.connect(self.check_icon_clicked)
+        self.listLayers.itemDoubleClicked.connect(self.multilayer_doubleclicked)
         self.listLayers.setVisible(True)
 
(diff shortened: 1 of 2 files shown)


joernu76 (changes requested):
When I double-click a layer, it is selected and downloaded, but the dialogue does not close...?

It might also be nice not to download a new map in case the *same* layer was clicked, but only if it does not complicate the code too much.

Marilyth:
Double clicking works, so long as auto-update is disabled.
Making double clicking work during a modal dialog popup is a really ugly ordeal.. I am very close to throwing the towel here.

The only way to fix it for me was to make the progress dialog non modal, and instead disable all widgets for the duration except for the multilayer window.
Then again having the multilayers not blocked means people can change their selection during a get_map call which is horrible as well.

PR #827, opened 2021-04-22
Default to newest init time

### LINKED ISSUE #824: Displayed newest initialisation time as default
When on a MSS server multiple forecast data sets with different initialisation times are present, the default setting should be to disply the newest initialisation time. This is the most likely to be looked upon.
This is currently not the case. Please fix it.

Marilyth:
Fixes #824 
1. When a layer gets constructed, default to its newest initialisation time.
2. Improve carry-over logic. 
    Only carry over parameters that have been explicitly changed by the user, otherwise keep the default.
    Save carry-over parameters as to not lose them after a layer didn't support them.


Marilyth:
@joernu76 I tried a best of both worlds approach. Please let me know if this is a good solution or not.

mslib/msui/multilayers.py
```diff
@@ -530,7 +530,7 @@ def _parse_itimes(self):
                     self.parent.dock_widget, self.parent.dock_widget.tr("Web Map Service"),
                     self.parent.dock_widget.tr("ERROR: {}".format(msg)))
             else:
-                self.itime = self.itimes[0]
+                self.itime = self.itimes[-1]
```
joernu76, line 533: is this sorted?

Marilyth: yes

Marilyth:
This should go to develop, not stable, now that it isn't classified a bug anymore

PR #839, opened 2021-04-28
Fix issue #830 (KML updates)

### LINKED ISSUE #830: patch from kml overlay did not keep position when stereo map is moved
Example
[j.kml.txt](https://github.com/Open-MSS/MSS/files/6366349/j.kml.txt)

I found an issue by this. I don't get it to the right location in northern stereo 

it should be here
there

I get it at 

moved

I could move it around but it should have a fixed position

moved_map

joernu76:


ReimarBauer (changes requested):
This example video describes a missing clear of the old patch. The new one is on the correct position.

old_patch_needs_to_be_removed

With further panning or zooming additional patches on the wrong position are added in *stereo*

joernu76:
Strange, did not occur when I tested it; at least at some point in time. 
Nice catch, will have a look!

joernu76:
Problem should be fixed now.

ReimarBauer (approved):

PR #845, opened 2021-04-28
Fix issue #844

### LINKED ISSUE #844: add feature to allow horizontal sections without countries/graticules
Necessary for proper contour layers.

joernu76:


joernu76:
@Marilyth Hi May, could you try to reproduce this bug (the failing test is yours)? I cannot reproduce this or understand how the change impact the execution in any way.

mslib/mswms/mpl_hsec.py
```diff
@@ -330,17 +331,18 @@ def plot_hsection(self, data, lats, lons, bbox=(-180, -90, 180, 90),
...
-        # scatter() for drawing the flight tracks and trajectories.
-        # Curiously, plot() works fine without this setting, but scatter()
-        # doesn't.
-        bm.fillcontinents(color='0.98', lake_color='white', zorder=0)
-        self._draw_auto_graticule(bm)
+        if self._plot_countries:
+            # Set up the map appearance.
+            bm.drawcoastlines(color='0.25')
+            bm.drawcountries(color='0.5')
+            bm.drawmapboundary(fill_color='white')
```
ReimarBauer, line 338: 'white' is now a constant, does this mean that the option from the Map Appeareance is disabled?

ReimarBauer (changes requested):
see my question to the UI colors

mslib/mswms/mpl_hsec.py
joernu76: It was a constant before as well. No change here, I only made this code section optional in derived classes.

I think you confuse the client and server. The options affect only the client, this is server code.

ReimarBauer (approved):

PR #846, opened 2021-04-29
Fix issue #843

### LINKED ISSUE #843: units in mswms/utils not parseable by pint

joernu76:
* replaced some nice, but unparseable unit strings
* fixed wrong conversion ratios
* added test


ReimarBauer (approved):
I do wonder about about the previous scaling value, e.g. 1e6. Does that mean that older files needs an update?

jgrooss:
The old data files do not need an update if the scaling works.

joernu76:
The unit conversion replaces the old scaling. Currently they are scaled twice and thus wrongly.
Old files might not work due to unparseable units, but they will not work regardless of this change.

PR #859, opened 2021-04-29
Disable mscolab login without credentials

### LINKED ISSUE #852: Disable mscolab login without credentials
The login button should be disabled if the mail and password has not yet been provided.

Jatin2020-24:
Fixes: #852


ReimarBauer:
Please send this PR to stable branch

ReimarBauer (changes requested):
we need this PR to the stable branch

in principle. Bug Fixes -> stable, Enhancements -> develop

ReimarBauer:
That's not a good idea, 

we merge only on a new major release develop into stable. 

You need to cherry pick your changes e.g. save it to a patch file. checkout a stable brach (git checkout -b fixlogin) and apply your patch by e.g. patch -p1 < my.patch  and send from that a PR

Marilyth (changes requested):
A few things

1. After connecting to a server, the login button is enabled even without credentials
2. Removing the email after login is enabled doesn't disable it, as it is only tied to the password
3. I think there is no need to disable the password field, just connect both email and password to the login enabler

issues

PR #866, opened 2021-04-29
addresses various login problems

### LINKED ISSUE #849: Wrong credentials in mscolab result in fatal error
When logging in with wrong credentials, you receive a fatal JSON error as seen in the picture
login

### LINKED ISSUE #851: HTTP Authentication not cancelable
When needing to provide authentication for a mscolab server, clicking "cancel" just reopens the dialog.

### LINKED ISSUE #864: After mscolab logout, default password is cleared
when a user has a config for username and password than this is prefilled indo the login fields. When the user logs out the username persists and the password is cleared.

We want that the user can easily login again.

ReimarBauer:


mslib/msui/mscolab.py
```diff
@@ -549,6 +547,9 @@ def authorize(self):
             if r.status_code == 200:
                 constants.MSC_LOGIN_CACHE[self.mscolab_server_url] = (auth[0], auth[1])
                 self.after_authorize(emailid, r)
+            else:
+                self.error_dialog = QtWidgets.QErrorMessage()
+                self.error_dialog.showMessage('Oh no, server authentication were incorrect.')
```
Marilyth, line 551: This PR does not fix #849
I made a PR before I noticed you already wanted to tackle this issue. Perhaps just copy that part.
https://github.com/Open-MSS/MSS/pull/867

Marilyth (changes requested):

mslib/msui/mscolab.py
ReimarBauer: done, thx :)

Marilyth (approved):

PR #869, opened 2021-04-30
fixed mscolab permission upgrade, #853

### LINKED ISSUE #853: mscolab permission upgrade results in duplicate project
When the permissions for a project changes from e.g. collaborator to admin, the project gets added twice for the user.

ReimarBauer:
in the server code a line reverted from 3db9e60708657dae4c414ad95ad702259a0e6b26
because of SQLite objects created in a thread can only be used in that
same thread also the API for submitting the access_level extended.


Marilyth (dismissed):
Can confirm this fixes the issue

joernu76 (approved):
Looks good, but I cannot really evaluate that because I am unfamiliar with the APIs.

Tanish0019:
Looks good 😃

PR #877, opened 2021-04-30
Check path doesn't contain illegal characters

### LINKED ISSUE #875: mscolab project name with blanks
It seems, that a mscolab project needs to be named without spaces or "special characters". This is displayed in the selection box when choosing "Add Project". However, if accitently a project name including a blank is typed in, there comse the error message "The path already exists"
If project names need to be without blanks, please drop a meaningful error message

Marilyth:
Adds a regex check before adding the project to make sure the path contains no spaces or special characters other than _ and - as a substitute for spaces.
fixes #875

### DIFF
--- a/mslib/msui/mscolab.py
+++ b/mslib/msui/mscolab.py
@@ -33,6 +33,7 @@ import logging
 import types
 import fs
 import requests
+import re
 from fs import open_fs
 from werkzeug.urls import url_join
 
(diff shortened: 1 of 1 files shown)


ReimarBauer (changes requested):
if bool()

https://stackoverflow.com/questions/6576962/python-regular-expressions-return-true-false

ReimarBauer:
We should convert this to a bool so True/False instead different types can be used.

Later for tests that makes that simpler

Marilyth:
We use None for ifs all over MSS. I am not sure why it should be different here but okay.

ReimarBauer:
ok, then we do that later when we want to refactor that, agreed.

ReimarBauer (dismissed):
Thx

ReimarBauer (approved):
:)

PR #884, opened 2021-04-30
Fix text changed leak

### LINKED ISSUE #873: mscolab connect/disconnect signal leak
On each connection or disconnection to a server, the signal for the buttons is switched between connect_handler and disconnect_handler.
However the old signal is never cleaned up. So after a while a single button click will connect and disconnect an arbitrary amount of times, causing the whole application to slow down.

Marilyth:
Another leak bypassed my attention with Jatins PR.
This properly disconnects the signals.
fixes #873


ReimarBauer (approved):
Thx

Issue #887, opened 2021-05-02
gui About: New Features and Changes Link incomplete


ReimarBauer:
the milestone link did not work, version is missing

https://github.com/Open-MSS/MSS/issues?q=is%3Aclosed+milestone%3A

joernu76:
Fixed by #891

PR #888, opened 2021-05-03
Fixes gui About: New features and Changes link updated

Jatin2020-24:
Fixes: #887


ReimarBauer (changes requested):
we need a solution which is based on __version__ 

We don't want to update mss_pyui.py always with a new release.

ReimarBauer:
look at the test of https://github.com/Open-MSS/MSS/pull/891

PR #891, opened 2021-05-04
about new features link fixed

### LINKED ISSUE #887: gui About: New Features and Changes Link incomplete

ReimarBauer:


mslib/msui/_tests/test_mss_pyui.py
```diff
@@ -29,13 +29,34 @@
...
+class Test_MSS_AboutDialog():
+    def setup(self):
+        self.application = QtWidgets.QApplication(sys.argv)
+        self.window = mss_pyui.MSS_AboutDialog()
+
+    def test_milestone_url(self):
+        with urlopen(self.window.milestone_url) as f:
+            text = f.read()
+        pattern = f'value="is:closed milestone:{__version__[:-1]}"'
+        assert re.search(pattern, text.decode('utf-8'))
```
Marilyth: The pattern doesn't seem to contain any special regex.
In which case, is there a reason not to use 
`assert pattern in text.decode('utf-8')`
instead of
`assert re.search(pattern, text.decode('utf-8'))`
?

Marilyth: Also, `.` is interpreted as "any character" unless escaped. 
So while it does work for our case, it would also match versions like "31022" instead of "3.0.2"

ReimarBauer: oh, the first approach is much better, and I completly forgot about the "." feature.

Marilyth (approved):

joernu76:
to stable?

PR #903, opened 2021-05-05
Refactored config_loader.

joernu76:
* use builtin functions to simplify code
* fail early for missing parameters
* remove inaccessible return statement


ReimarBauer (approved):
good :)

PR #907, opened 2021-05-05
WIP: xdg

### LINKED ISSUE #908: better handling of default storage directories
Follow XDG convention and use may 
```
In [1]: import xdg

In [2]: xdg.XDG_CONFIG_HOME
Out[2]: PosixPath('/home/icgXXX/.config')

In [3]: xdg.XDG_CACHE_HOME
Out[3]: PosixPath('/home/icgXXX/.cache')

In [4]: xdg.XDG_DATA_HOME
Out[4]: PosixPath('/home/icgXXX/.local/share')
```
for default paths.

* .config/mss is already used and correct.
* FTML files should go in ~/Documents/mss
* MSCOLAB local stuff should go to .cache/mss or .local/share/mss

joernu76:


joernu76:
Comments? next steps would, maybe, be to use the xdg python package...

Marilyth:
I am not sure how OS friendly XDG is. I have certainly never had a Windows application using the paths suggested by XDG.
Instead things like %APPDATA% are used on Windows, which XDG ignores.
https://github.com/ActiveState/appdirs might be friendlier?

joernu76:
didn't test xdg on Windows but was fearing lacking support.
The appdirs one seems perfect for our use case!

mslib/msui/__init__.py
```diff
@@ -54,7 +54,7 @@ class MissionSupportSystemDefaultConfig(object):
     filepicker_default = "default"
 
     # dir where mss output files are stored
-    data_dir = "~/mssdata"
+    data_dir = os.path.join(os.path.expanduser("~/"), "Documents", "mss")
```
ReimarBauer: this should be an fs path.

and we don't have "Documents" on all systems. For example this depends if you do a desktop or server installation. I don't expect it in a docker installation.

joernu76:
Mmmh. appdirs is lacking non-app-specific locations like the Documents folder.
QT includes also such a functionality:
https://doc.qt.io/qt-5/qstandardpaths.html

my "problem": one is potentially offered multiple paths... grrr.

mslib/msui/__init__.py
```diff
@@ -180,7 +180,7 @@ class MissionSupportSystemDefaultConfig(object):
     mscolab_server_url = "http://localhost:8083"
     # ToDo refactor to rename this to data_dir/mss_data_dir
     # mss dir
-    mss_dir = "~/mss"
+    mscolab_data_dir = os.path.join(os.path.expanduser("~/"), ".local", "share", "mss")
```
ReimarBauer: see above, the dirs have to be verified to exists on all platforms

ReimarBauer (changes requested):

mslib/msui/__init__.py
joernu76: I do not understand how to properly use fs. See current issue.

This particular path should point to a sensible location to store ftml files, which would be the Documents folder on Windows and XDG compliant systems. This path is irrelevant for Servers.

mslib/msui/__init__.py
joernu76: I do not know how to verify the existance of directories with fs without using os.path.

mslib/msui/__init__.py
joernu76: I can remove all fs stuff and implement the checks, though.

I have serious doubts that anything works in case we would start to use fs for anything not on the local filesystem.

joernu76:
And if we use, e.g., QT for file discovery the paths would have the local filesystems syntax and might not work with fs ???

joernu76:
So, this moves the (constant?) config from __init__ to constants, where it fits better.

This demonstrates how to use QT standard path feature (using os).
Probably breaks a lot of things in this state.

ReimarBauer:
Have a look on this too
https://docs.pyfilesystem.org/en/latest/reference/appfs.html

joernu76:
will be realized after the os/fs refactoring.

PR #911, opened 2021-05-05
Default to newest init time

### LINKED ISSUE #824: Displayed newest initialisation time as default

Marilyth:
Fixes #824 
1. When a layer gets constructed, default to its newest initialisation time.
2. Improve carry-over logic. 
    Only carry over parameters that have been explicitly changed by the user, otherwise keep the default.
    Save carry-over parameters as to not lose them after a layer didn't support them.
3. Filter valid times to only show times after the selected init time

### DIFF
--- a/mslib/msui/_tests/test_wms_control.py
+++ b/mslib/msui/_tests/test_wms_control.py
@@ -367,6 +367,7 @@ class Test_HSecWMSControlWidget(WMSControlWidgetSetup):
         assert layer_a.get_vtime() == self.window.cbValidTime.currentText()
         assert layer_a.get_level() == layer_b.get_level()
         assert layer_a.get_vtime() == layer_b.get_vtime()
+        assert layer_a.get_itime() == layer_a.get_itimes()[-1]
         assert mockbox.critical.call_count == 0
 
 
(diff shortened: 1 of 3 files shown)


joernu76:
Mmmh.
I "run immediately into the wall" that the init time was updated to the most recent one, but the forecast time is still the old one and thus invalid.
I feel this is getting out of hand...

joernu76:
Ran into this error:
```
CRITICAL: Platform: Linux-5.10.0-6-amd64-x86_64-with-glibc2.10 (('64bit', 'ELF'))
  File "/home/icg173/src/mss/mslib/msui/wms_control.py", line 1071, in init_time_changed
    self.multilayers.get_current_layer().set_itime(self.cbInitTime.currentText())
  File "/home/icg173/src/mss/mslib/msui/multilayers.py", line 642, in set_itime
    self.set_vtime()
TypeError: set_vtime() missing 1 required positional argument: 'vtime'

```

To reproduce:
init time on 2021-05-09T00 valid time on 2021-05-09T12
choose init time 2021-05-10T00 -> error

Marilyth:
Careless mistake. Should be fixed now. Sorry!

joernu76 (approved):
Seems to work now.
I am not sure if the behaviour is really better than a persistance-less, much simpler solution. We need to try this logic out a bit.

PR #913, opened 2021-05-05
Remove qsettings store of used URLs.

### LINKED ISSUE #906: better management of MSCOLAB servers in drop down list
I have a long list of local/remote servers and the server I need is way down the list.
I'd like to remove some servers and I'd like MSS to have the last used server selected.

joernu76:
Fix issue #906

### DIFF
--- a/mslib/msui/mscolab.py
+++ b/mslib/msui/mscolab.py
@@ -144,18 +144,14 @@ class MSSMscolabWindow(QtWidgets.QMainWindow, ui.Ui_MSSMscolabWindow):
         self.url.setEditable(True)
         self.url.setModel(MSCOLAB_URL_LIST)
         # fill value of mscolab url from config
-        default_MSCOLAB = config_loader(
-            dataset="default_MSCOLAB")
+        default_MSCOLAB = config_loader(dataset="default_MSCOLAB")
         add_mscolab_urls(self.url, default_MSCOLAB)
         self.emailid.setEnabled(False)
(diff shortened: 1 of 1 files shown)


ReimarBauer (changes requested):
If we have there only 1 entry then we don't need a pull down list in the mscolab window.
I'm not sure if only developers have this problem, because we don't use for tests a different location of the qsettings files.

Another option would be to store only 3 and have the default_MSCOLAB on top as now, when it is defined.
A short list of 3 would do the same, without the change of the UI. 
When the UI gets changed, there we should also add a hint where to store this server name. 

As developer this will create a problem to me, because I don't want to test mscolab on the default server and I don't want to enter always the yourmss, staging server address or switch that in the file.

ReimarBauer:
> If we have there only 1 entry then we don't need a pull down list in the mscolab window.
> I'm not sure if only developers have this problem, because we don't use for tests a different location of the qsettings files.
> 
> Another option would be to store only 3 and have the default_MSCOLAB on top as now, when it is defined.
> A short list of 3 would do the same, without the change of the UI.
> When the UI gets changed, there we should also add a hint where to store this server name.
> 
> As developer this will create a problem to me, because I don't want to test mscolab on the default server and I don't want to enter always the yourmss, staging server address or switch that in the file.

Answering myself, the default_MSCOLAB is a list, so three or more in the navigation is not an issue

ReimarBauer (approved):

PR #923, opened 2021-05-06
Reimplemented parts of matplotlibs history mechanism for TopViews

### LINKED ISSUE #919: Topview Home Button return to initial coordinates
The topview Home Button does not load the initial coordinates. 
So there is no option to get back to the origin without restarting topview.

joernu76:
Works now also for stereographic projections and projection changes.

Fix issue #919

### DIFF
--- a/mslib/msui/mpl_qtwidget.py
+++ b/mslib/msui/mpl_qtwidget.py
@@ -273,6 +273,7 @@ class NavigationToolbar(NavigationToolbar2QT):
         self.setIconSize(QtCore.QSize(24, 24))
         self.layout().setSpacing(12)
         self.canvas = canvas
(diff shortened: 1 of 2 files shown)


ReimarBauer (dismissed):
Currently the home button is defined to the projection selected on the first view. Also when the map is changed it gets back to this map, but the control shows the old selection.

home_button


just change the projection selection list to the starting value

For a moment I thought to set the new projection than as initial for the home button, but then likly one will ask how to get back to the full information

joernu76:
Changed behaviour. Projection change now resets history. Home button goes back to defined projection. Consistent and probably what people want.

ReimarBauer (approved):
great :)

PR #925, opened 2021-05-06
Mscolab: collaborator can access version history

### LINKED ISSUE #920: set mscolab collaborator role to access version history
currently only project admins have access to the version history.
In the version history it is detailed who changes waypoints.
This is important to all users able to change waypoints.
So we should change the role for this feature from admin to collaborator.

Jatin2020-24:
Fixes: #920


mslib/msui/mscolab.py
```diff
@@ -680,10 +680,11 @@ def set_active_pid(self, item):
         if self.access_level == "viewer" or self.access_level == "collaborator":
             if self.access_level == "viewer":
                 self.chatWindowBtn.setEnabled(False)
+                self.versionHistoryBtn.setEnabled(False)
```
ReimarBauer: Why do you disable the self.versionHistoryBtn for the collaborator ?

ReimarBauer (changes requested):
the documentation will also need an update e.g. https://mss.readthedocs.io/en/stable/mscolab.html?highlight=collaborator#project-permissions

and see my question above.

mslib/msui/mscolab.py
```diff
@@ -680,10 +680,11 @@ def set_active_pid(self, item):
...
             else:
                 self.chatWindowBtn.setEnabled(True)
+                self.versionHistoryBtn.setEnabled(True)
```
Jatin2020-24: self.versionHistoryBtn enabled for collaborator.

mslib/msui/mscolab.py
Jatin2020-24: It is disabled for viewer only. [See](https://github.com/Open-MSS/MSS/pull/925/files/07ba45540d0ceaa17b8cb261ca7febf102c61c2c#diff-8bb1b53508f9fe252a80649575f92b7aa4756be8dda54c217a7c1d9063d6b0c5L683-R686)

ReimarBauer: This is currently not good readable, have you thought about a change based on if -- eilf ?

```
if self.access_level == "collaborator":

elif self.access_level == "viewer":
```

Jatin2020-24: Good idea, I will implement it using if -- elif

mslib/msui/mscolab.py
```diff
@@ -533,7 +533,7 @@ def handle_work_locally_toggle(self):
                         "Turn on work locally to work on local flight track file"))
             self.save_ft.setEnabled(False)
             self.fetch_ft.setEnabled(False)
-            if self.access_level == "admin" or self.access_level == "creator":
+            if self.access_level == "admin" or self.access_level == "creator" or self.access_level == "collaborator":
```
ReimarBauer: if self.access_level in ["admin", "creator", "collaborator"]:

ReimarBauer (changes requested):
Please verify that "Name Version" for a "collaborator" also works.

I do get
"Some error occurred!"

mslib/mscolab/file_manager.py
```diff
@@ -279,7 +292,7 @@ def get_change_content(self, ch_id):
         return change_content
 
     def set_version_name(self, ch_id, p_id, u_id, version_name):
-        if not self.is_admin(u_id, p_id):
+        if not self.is_admin(u_id, p_id) and self.is_collaborator(u_id, p_id):
```
ReimarBauer: you probably want to set this

if not (self.is_admin(u_id, p_id) or self.is_collaborator(u_id, p_id)):

ReimarBauer (changes requested):
see comment

mslib/mscolab/file_manager.py
```diff
@@ -279,7 +292,7 @@ def get_change_content(self, ch_id):
         return change_content
 
     def set_version_name(self, ch_id, p_id, u_id, version_name):
-        if not self.is_admin(u_id, p_id):
+        if not self.is_admin(u_id, p_id) or self.is_collaborator(u_id, p_id):
```
ReimarBauer: this must be 

not admin
or
not collaborator

this should reject a "viewer"

Jatin2020-24: we can use    
```
if self.is_viewer(u_id, p_id):
          return False
```
 it will reject viewer.

Jatin2020-24:
Now it rejects viewer, I have tested it on my laptop also.

mslib/mscolab/file_manager.py
ReimarBauer: For these kind of rules, I do prefer:

Explicit is better than implicit.

mslib/msui/mscolab.py
```diff
@@ -990,11 +989,14 @@ def handle_update_permission(self, p_id, u_id, access_level):
...
+                    self.versionHistoryBtn.setEnabled(False)
+                    if self.version_window is not None:
+                        self.version_window.close()
                 if self.admin_window is not None:
                     self.admin_window.close()
-                if self.version_window is not None:
-                    self.version_window.close()
             else:
                 self.adminWindowBtn.setEnabled(True)
                 self.versionHistoryBtn.setEnabled(True)
```
joernu76: The chat window button below seems inconsistent.

The buttons are enabled/disabled at different points in the code. This should be refactored to a single location.
I would prefer also more readable code.

perhaps a standardized code block like
```
allow_version_access = self.access_level in ["..."]
self.versionHistoryBtn.setEnabled(allow_version_access)
if not allow_version_access and self.version_window is not None:
    self.version_window.close()
```
for each button/window ?

joernu76 (changes requested):
See comment.
(rights seem inconsistent and should be refactored to allow less room for errors)

Jatin2020-24:
@joernu76 @ReimarBauer please review this PR

mslib/msui/mscolab.py
```diff
@@ -988,23 +987,18 @@ def handle_update_permission(self, p_id, u_id, access_level):
...
+            if not allow_version_access and self.chat_window is not None:
+                self.chat_window.close()
 
-            if self.access_level == "viewer":
-                self.chatWindowBtn.setEnabled(False)
-                if self.chat_window is not None:
-                    self.chat_window.close()
-            else:
-                self.chatWindowBtn.setEnabled(True)
+            allow_version_access1 = self.access_level in ["creator", "admin"]
```
ReimarBauer: what is the reason to introduce the allow_version_access**1** and not keep with a allow_version_access

ReimarBauer (changes requested):
you'r repo is not on recent develop version. Have a look on the change for the viewer in this change. The viewer does not need the "Save to Server" button, because he can't store.

https://github.com/Open-MSS/MSS/pull/944/files

ReimarBauer:
try to resolve this conflict

resolve_conflict

mslib/msui/mscolab.py
Jatin2020-24: Previously I am getting error with `allow_version_access` that's why I used `allow_version_access1` but now it is resolved.

Jatin2020-24:
I have also refactored permissions in `def set_active_pid(self):` as suggested by @joernu76

ReimarBauer (dismissed):
thx

joernu76 (dismissed):

mslib/msui/mscolab.py
```diff
@@ -1001,23 +995,24 @@ def handle_update_permission(self, p_id, u_id, access_level):
...
+                self.chat_window.close()
+            if not allow_version_access and self.handle_work_locally_toggle is not None:
+                self.handle_work_locally_toggle.close()
+            if not allow_version_access and self.handle_import is not None:
+                self.handle_import.close()
+
+            allow_version_access = self.access_level in ["creator", "admin"]
+            self.adminWindowBtn.setEnabled(allow_version_access)
+            if not allow_version_access and self.admin_window is not None:
+                self.admin_window.close()
```
joernu76: Much better!
It seems to me that in the second piece of code an update of "self.deleteProjectBtn" is missing?
If it is not missing, it is fine this way.
If the code needs to be touched again due to this, I would strongly suggest merging lines 731-741 and lines 998-1015 to a single function. The window closing may be superfluous in the first section, but the maintainability gain due to a single piece of code is huge.

Jatin2020-24: Great suggestion, I have merged both blocks of code to new function `def handle_mscolab_buttons(self):`

mslib/msui/mscolab.py
```diff
@@ -628,7 +654,7 @@ def after_authorize(self, emailid, r):
         self.conn = sc.ConnectionManager(self.token, user=self.user, mscolab_server_url=self.mscolab_server_url)
         self.conn.signal_reload.connect(self.reload_window)
         self.conn.signal_new_permission.connect(self.render_new_permission)
-        self.conn.signal_update_permission.connect(self.handle_update_permission)
+        self.conn.signal_update_permission.connect(self.handle_update_permission, self.handle_mscolab_buttons)
```
joernu76: I would call the handle_mscolab_button function manually within the other functions. I am not sure how timing and race conditions work with QT and do not want to, actually :-)

Jatin2020-24:
@joernu76  I've called `handle_mscolab_buttons` inside `set_active_pid` and `handle_update_permission`,  now it is completely working.

joernu76 (approved):
nice

Issue #901, opened 2021-05-05
Fatal error in config loader because of invalid character


ReimarBauer:
On OS Windows 10 a user sees a problem loading his settings file

![Clipboard - 5  Mai 2021 10 17](https://user-images.githubusercontent.com/278728/117115897-01144900-ad8e-11eb-95b5-03d1099b241e.png)

joernu76:
People are using the config-thingie a lot...
The path does not look like it contains invalid characters...?

Marilyth:
Might be an issue with fs no supporting `\` in paths
fs.path.split() also doesn't work on Windows, as it only splits on `/`, not on `\`

from fs.path.split
```
if "/" not in path:
    return ("", path)
split = path.rsplit("/", 1)
return (split[0] or "/", split[1])
```

joernu76:
fs is causing us a lot of issues.

Do you have a access to a Windows system for testing?

joernu76:
@ReimarBauer I have the feeling that we can use either os or fs but not both. (e.g. due to "/" "\").
Can you replace all os.path stuff with fs replacements? grep should discover a lot.

Marilyth:
I have a Windows and Linux system for troubleshooting. We had multiple instances of issues with fs on Windows already.

ReimarBauer:
On a old 2.0.1 without own configuration this does not happen on the guacamole windows 10 system.
on a recent 3.0.2 config editor opens without an error too.

ReimarBauer:
The problem happens with saving 

just
```
{
}
```
can't be saved.

Marilyth:
The issue arises when mss_settings.py was automatically loaded.
It does not happen when no settings were loaded, and it does not happen when the setting were manually loaded.

ReimarBauer:
I found some more problems and working on a test while I already have an idea how to solve it.
https://github.com/Open-MSS/MSS/pull/936

We have also a problem when there is not any mss_settings.json defined. I struggeled with windows over a None path and seen afterwards that I had this also on linux.

I added @Marilyth hotfix, but we can later merge both in a row.

ReimarBauer:
With current stable I have this without a mss_settings.json
```
Fatal error in MSS 3.0.2. on Linux-4.15.0-142-generic-x86_64-with-glibc2.27
Python 3.9.2 | packaged by conda-forge | (default, Feb 21 2021, 05:02:46) 
[GCC 9.3.0]

[...]

PR #931, opened 2021-05-07
mscolab working locally without created ~/mss folder, fixes #900

### LINKED ISSUE #900: Fatal error when working locally without ~/mss folder
When ticking the "Work Locally" checkbox in mscolab without an existing ~/mss folder, this fatal error pops up.
Screenshot at 2021-05-05 10-18-10

ReimarBauer:


mslib/msui/mscolab.py
```diff
@@ -74,7 +76,8 @@ class MSSMscolabWindow(QtWidgets.QMainWindow, ui.Ui_MSSMscolabWindow):
     identifier = None
     viewCloses = QtCore.pyqtSignal(name="viewCloses")
 
-    def __init__(self, parent=None, data_dir=mss_default.mss_dir, mscolab_server_url=mss_default.mscolab_server_url):
+    def __init__(self, parent=None, data_dir=config_loader(dataset="mss_dir"),
```
joernu76: Does this work? Timing wise? Always? I would prefer None here with access to config_loader in the function if it is None, similar to how one should handle dicts/lists.

mslib/msui/mscolab.py
```diff
@@ -530,8 +552,10 @@ def handle_work_locally_toggle(self):
             if self.version_window is not None:
                 self.version_window.close()
             self.create_local_project_file()
-            self.local_ftml_file = fs.path.join(self.data_dir, 'local_mscolab_data',
```
joernu76, line 533: I do not understand, from the documentation of fs, why this change is necessary. It seems to be the same, but more complicated...?

mslib/_tests/test_utils.py
```diff
@@ -121,7 +121,7 @@ def test_existing_empty_config_file(self):
         with fs.open_fs(MSS_CONFIG_PATH) as file_dir:
             file_content = file_dir.readtext("mss_settings.json")
         assert ":" not in file_content
-        config_file = fs.path.join(MSS_CONFIG_PATH, "mss_settings.json")
+        config_file = fs.path.combine(MSS_CONFIG_PATH, "mss_settings.json")
```
joernu76, line 124: I understand that this is more efficient?

joernu76 (commented):
Some questions to understand fs better. Seems fine for a hot fix.

mslib/_tests/test_utils.py
ReimarBauer: this keeps the double slash, while the other removes one.

joernu76: There is nothing in the documentation for that.

This method only supports fs-paths.

fs can only deal with PyFilesystem paths.
Conversion works using

unsyspath and getsyspath methods.

https://pyfilesystem.readthedocs.io/en/latest/osfs.html

This should solve our "problem" with QT. But for the JSON-config we need some speculative conversion routine, because we may have either an OS or a fs-path.

I am getting doubts if this is even fine for a hotfix.

ReimarBauer:
issue #901 will become a seperate PR, if it is not possible to fix in stable this configeditor I probably will deactivate it for windows until we have it 

This PR is about the "work locally" worflow in the msco window.

joernu76 (approved):

PR #932, opened 2021-05-07
Hotfix for #901

### LINKED ISSUE #901: Fatal error in config loader because of invalid character

Marilyth:
Fixes #901 
in fs, \\ is an illegal character for `_fs.exists` and can't be used for splitting dirname and filename either.
Replacing \\ with / fixes this particular error, at this particular place.

### DIFF
--- a/mslib/msui/editor.py
+++ b/mslib/msui/editor.py
@@ -52,8 +52,9 @@ class EditorMainWindow(QtWidgets.QMainWindow):
         self.editor = QtWidgets.QPlainTextEdit()
 
         # Load mss_settings.json (if already exists), change \\ to / so fs can work with it
-        self.path = constants.CACHED_CONFIG_FILE.replace("\\", "/")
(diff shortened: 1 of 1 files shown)


ReimarBauer:
I review currently the whole editor code. many times on windows the filename is None. I guess we can also change this line in constants.py 
```
 index 695cd5e2..34cd1963 100644
--- a/mslib/msui/constants.py
+++ b/mslib/msui/constants.py
@@ -31,12 +31,13 @@
 import os
 import fs

-HOME = os.path.expanduser("~/")
+HOME = os.path.expanduser(f"~{os.path.sep}")
```

Marilyth:
I noticed that as well. That would make sense to make things more uniform.
However it doesn't help with fs not working with os.sep

ReimarBauer (approved):
good catch

PR #934, opened 2021-05-07
914: add gpx export

### LINKED ISSUE #914: add a gpx export function
Hi all, 
there are already some export formats for the flight track. Another useful one would be [.gpx](https://en.wikipedia.org/wiki/GPS_Exchange_Format#References). It is a simple XML file that can be read by many other useful tools such as windy. Especially having the abilitiy to work with waypoints makes this format useful.

I think this would be a nice addition to MSS. I don't quite know where to start for implementing this. But with a few hints I would give it a go.

Here is an example file: https://speicherwolke.uni-leipzig.de/index.php/s/TyJKCdRykaDS9NM

jroettenbacher:
Hi all, 
I included a simple export function to gpx.
Maybe we could add something in the meta data still, but other than that I think it's doing the job. I also left out the height information, but if needed we can easily include it as well.
However, since the main purpose of this was to import it to windy.com the heigth info is not needed.

One thing to mention is that I introduce a further dependency in the form of the module `gpxpy`. we can discuss, if we want this or if I should try to avoid it. 

Best Johannes

### DIFF
--- a/mslib/plugins/io/gpx.py
+++ b/mslib/plugins/io/gpx.py
@@ -4,12 +4,13 @@
     mslib.plugins.io.csv
     ~~~~~~~~~~~~~~~~~~~~
 
-    plugin for csv format flight track export
+    plugin for gpx format flight track export
 
     This file is part of mss.
 
     :copyright: Copyright 2008-2014 Deutsches Zentrum fuer Luft- und Raumfahrt e.V.
     :copyright: Copyright 2011-2014 Marc Rautenhaus (mr)
+    :copyright: Copyright 2021 Johannes Roettenbacher
     :copyright: Copyright 2016-2021 by the mss team, see AUTHORS.
     :license: APACHE-2.0, see LICENSE for details.
 
@@ -51,8 +52,9 @@ def save_to_gpx(filename, name, waypoints):
 
     # create waypoints
     for i, wp in enumerate(waypoints):
-        # loc = str(wp.location)
-        gpx_segment.points.append(gpxpy.gpx.GPXTrackPoint(wp.lat, wp.lon, name=f"{i}"))
+        loc = str(wp.location)
+        wp_name = loc if loc else str(i)
+        gpx_segment.points.append(gpxpy.gpx.GPXTrackPoint(wp.lat, wp.lon, name=wp_name))
 
-    with _fs.open(_name, "w") as file:
-        file.write(gpx.to_xml())
+    with _fs.open(_name, "w") as fh:
+        fh.write(gpx.to_xml())
--- a/mslib/plugins/io/gpx.py
+++ b/mslib/plugins/io/gpx.py
@@ -29,14 +29,9 @@
 from __future__ import absolute_import
 from __future__ import division
 
-from builtins import str
-
-import unicodecsv as csv
 import os
-
-import mslib.msui.flighttrack as ft
 from fs import open_fs
-import xml.etree.ElementTree as ET
+import gpxpy
 
 
 def save_to_gpx(filename, name, waypoints):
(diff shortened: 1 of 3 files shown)


mslib/plugins/io/gpx.py
```diff
@@ -0,0 +1,58 @@
+# -*- coding: utf-8 -*-
+"""
+
+    mslib.plugins.io.csv
+    ~~~~~~~~~~~~~~~~~~~~
+
+    plugin for csv format flight track export
```
joernu76: csv -> gpx

mslib/plugins/io/gpx.py
```diff
@@ -0,0 +1,58 @@
...
+    See the License for the specific language governing permissions and
+    limitations under the License.
+"""
+
+from __future__ import absolute_import
+from __future__ import division
+
+import os
+from fs import open_fs
+import gpxpy
```
joernu76, line 35: This is a new dependency. Needs to be added to localbuild/meta.yaml
@ReimarBauer anywhere else?

mslib/plugins/io/gpx.py
```diff
@@ -0,0 +1,58 @@
...
+    # add metadata
+    gpx.name = name
+    gpx.description = "MSS flight track export"
+
+    # create waypoints
+    for i, wp in enumerate(waypoints):
+        # loc = str(wp.location)
+        gpx_segment.points.append(gpxpy.gpx.GPXTrackPoint(wp.lat, wp.lon, name=f"{i}"))
+
+    with _fs.open(_name, "w") as file:
```
joernu76: file is a reserved keyword. use, e.g. "fh" for file handle.

mslib/plugins/io/gpx.py
```diff
@@ -0,0 +1,58 @@
...
+    gpx_track.segments.append(gpx_segment)
+
+    # add metadata
+    gpx.name = name
+    gpx.description = "MSS flight track export"
+
+    # create waypoints
+    for i, wp in enumerate(waypoints):
+        # loc = str(wp.location)
+        gpx_segment.points.append(gpxpy.gpx.GPXTrackPoint(wp.lat, wp.lon, name=f"{i}"))
```
joernu76: not `name=wp.location` ?

joernu76 (changes requested):

mslib/plugins/io/gpx.py
ReimarBauer: it is on conda-forge

gpxpy                          1.4.2            py_0  conda-forge 


please add it  at current state only to localbuild/meta.yaml

At the time this becomes released we add it to the feedstock

The tests install it automagically ;)

mslib/plugins/io/gpx.py
```diff
@@ -0,0 +1,58 @@
...
+"""
+
+    mslib.plugins.io.csv
+    ~~~~~~~~~~~~~~~~~~~~
+
+    plugin for csv format flight track export
+
+    This file is part of mss.
+
+    :copyright: Copyright 2008-2014 Deutsches Zentrum fuer Luft- und Raumfahrt e.V.
```
ReimarBauer, line 11: in this line please add yourself as author.
there may be another line added :copyright: Copyright 2021 by the mss team, see AUTHORS.

ReimarBauer:
a few lines as description are welcome
https://github.com/Open-MSS/MSS/blob/develop/docs/usage.rst

mslib/plugins/io/gpx.py
jroettenbacher: `wp.location` can be an empty string. That is why I opted to assign the corresponding index in the gpx file. However, I added a compromise where I take the location if possible and otherwise use the index if location is an empty string.

mslib/plugins/io/gpx.py
jroettenbacher: I hope I added it at the right place in the right way

jroettenbacher:
> 
> 
> a few lines as description are welcome
> https://github.com/Open-MSS/MSS/blob/develop/docs/usage.rst

Let me know if you were thinking of something more elaborate, such as what gpx can be used for.

mslib/plugins/io/gpx.py
jroettenbacher: like this?

joernu76 (approved):

PR #935, opened 2021-05-07
Fix scaling hitboxes

### LINKED ISSUE #896: star / unstar feature on windows 10 (OS) does not always work
Some of the windows user can set some star but removing seems not to work also further staring seems not be possible

Marilyth:
fixes #896 
When changing the scale of the display, the pixels scale with it.
Instead of hardcoding the position of the star icon, approximately calculate it through the used scale and height of the table rows.
For most scaling options I tried it works fine. 
In some edge cases with too high scaling generally Qt also seems to give up and stops scaling their elements which results in slightly shifted hitboxes I can't really predict.

### DIFF
--- a/mslib/msui/multilayers.py
+++ b/mslib/msui/multilayers.py
@@ -48,6 +48,8 @@ class Multilayers(QtWidgets.QDialog, ui.Ui_MultilayersDialog):
         self.layers_priority = []
         self.current_layer: Layer = None
         self.threads = 0
+        self.height = None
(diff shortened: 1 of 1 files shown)


joernu76 (approved):

joernu76:
Is this a specific to multilayering or a general issue?

Marilyth:
Other than a janky look, everything else should work.
It has been a breaking issue with Multilayering due to my detecting clicks on an icon.

PR #946, opened 2021-05-11
Omit records containing the token

Marilyth:
The token used for authenticating oneself to the mscolab server is currently constantly printed in the engineio.client logs.
This PR filters all messages out which contain it.


joernu76:
Do we have an issue for this? For merging into stable, we need an issue. On a second note, do we need this in stable?

Marilyth:
Discussing and creating issues for security issues in public seems to be frowned upon.

joernu76:
I see. Wouldn't classify it as such, but I am not familiar with the matter.

joernu76 (approved):

ReimarBauer (approved):
thx

PR #954, opened 2021-05-13
added the flight track name to the window title of each view: 916

### LINKED ISSUE #916: show name of the flightpath/project on the views
If one presents a flightpath we currently have no information which it is.

We should display the flightpath name / msco project name on the other views (tableview, sideview, topview)

Perhaps  we could add it as label into the space between Level and Autoupdate

space

jroettenbacher:
This is a very simple addition. In that way one always knows which flight track is currently displayed in any view.
I can't quite figure out where the " - MSS" in the window title comes from though.


ReimarBauer:
To answer the question, we have set it as ApplicationDisplayName :)
https://github.com/Open-MSS/MSS/blob/develop/mslib/msui/mss_pyui.py#L764

ReimarBauer (changes requested):
This is half of the possibilities. ;)
The mscolab topview is using active_project_name for the selected project/flightpath.

Also using the window title is suboptimal for screen sharing. 

This is an example for a window share on BigBlueButton. There the frame is not shown.

topview

During the flightplanning meeting some people requested the name of the flighpaths, this was the driver to get it visible on screen.

A good location may be the lower right after the projection pull down in top view, this has to be checked with the other views.
An other option, which I likly prefer if I'm thinking about screenshots/reports is to add it on the drawing after "Top view" after "vertical flight profile" and don't add to the table view.

Also on the bottom where whe have the EPSG line this information can be added.  e.g. https://github.com/Open-MSS/MSS/blob/develop/mslib/msui/mpl_map.py#L132 

If this is not always wanted to be shown, it probably can get an options flag. 

It would be nice to have in side and top view it on a similar location.

jroettenbacher:
I'll have another go at it then. Thanks for the detailed response!

joernu76:
Mmmh. That the title is not visible always is not a good argument against putting the name there.
I agree that putting it into the figures itself might be a better (additional?) way of visualization.

@ReimarBauer Why does mscolab use a different name?

ReimarBauer:
> Mmmh. That the title is not visible always is not a good argument against putting the name there.
> I agree that putting it into the figures itself might be a better (additional?) way of visualization.
> 
> @ReimarBauer Why does mscolab use a different name?
In the current "new" interface we don't have this list. So there is no entry with the name there.

ReimarBauer:
We could also do two steps  @jroettenbacher do you want add the name first to the mscolab topview window too?

jroettenbacher:
I try to add it to the mscolab topview window first then.

jroettenbacher:
I'm not sure if this is the right place to modify the window title, but it changes it in all three views. I'll see if I can manage to add it to the plot in the TopView as well.

jroettenbacher:
I wonder if that extra if statement is necessary. Better to check twice but in theory there should always be a project name available so it would never be none.

To avoid overplotting onto the actual map I combined CRS and project name in one string with a line break. A saved plot would look like this:
topview_image

ReimarBauer:
on the window title it is but not above the eps code
trying_mscolab

ReimarBauer:
seems we have to set in mscolab UI self.waypoints_model.name

ReimarBauer:
diff --git a/mslib/msui/mscolab.py b/mslib/msui/mscolab.py
```
index 527f6442..08268e88 100644
--- a/mslib/msui/mscolab.py
+++ b/mslib/msui/mscolab.py
@@ -855,8 +855,9 @@ class MSSMscolabWindow(QtWidgets.QMainWindow, ui.Ui_MSSMscolabWindow):
             if active_window.view_type == _type:
                 active_window.raise_()
                 active_window.activateWindow()
-                return
 
+        # ToDo move to the place where the model is created
+        self.waypoints_model.name = self.active_project_name
         if _type == "topview":
             view_window = topview.MSSTopViewWindow(model=self.waypoints_model,
                                                    parent=self.listProjects,
```

ReimarBauer:
@jroettenbacher see comments. if the place is not easy to find, let it there with the Todo.

ReimarBauer (approved):
Thx

PR #958, opened 2021-05-14
fixed mscolab imports in msui #933

### LINKED ISSUE #933: msui start independent from mscolab_settings
Just seen on the windows 10 system. msui is completly independent from the mscolab server. No need to try to load that config py script. We have to remove this dependency if possible.

```
python mslib\msui\mss_pyui.py
INFO: MSS Version: 3.0.2.
```

aravindm711:
Also fixed raising of mscolab windows on closing chat, version history, and admin windows.

### DIFF
--- a/mslib/mscolab/models.py
+++ b/mslib/mscolab/models.py
@@ -33,7 +33,6 @@ from flask_sqlalchemy import SQLAlchemy
 from itsdangerous import (BadSignature, SignatureExpired, TimedJSONWebSignatureSerializer as Serializer)
 from passlib.apps import custom_app_context as pwd_context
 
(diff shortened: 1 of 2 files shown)


ReimarBauer (changes requested):
Now it is not immeadiatly shown to the user. With clicking on chat it gets visible.

I guess we have to move  "from mslib.mscolab.conf import mscolab_settings" into the User class of models.  Add there also a comment, so that it did not get moved by someone back to top.

great that you found that raise_() missing.

aravindm711:
Since the user model contains a staticmethod, I had add the import statement to it as well.

ReimarBauer (approved):

PR #959, opened 2021-05-15
Improve tests

ReimarBauer:


mslib/mscolab/utils.py
```diff
@@ -55,8 +59,20 @@ def get_message_dict(message):
     }
 
 
+def os_fs_create_dir(dir):
```
joernu76, line 62: Mmh. I think we need to go to an fs-only approach in combination with validating/converting any path communicated from an external source (QT-dialogues, JSON files, etc.).

Good start in doing that.

mslib/msui/constants.py
```diff
@@ -27,13 +27,24 @@
...
+import fs
 import os
+import logging
+
 
 HOME = os.path.expanduser(f"~{os.path.sep}")
 MSS_CONFIG_PATH = os.getenv("MSS_CONFIG_PATH", os.path.join(HOME, ".config", "mss"))
-if not os.path.exists(MSS_CONFIG_PATH):
-    os.makedirs(MSS_CONFIG_PATH)
+if '://' in MSS_CONFIG_PATH:
```
joernu76, line 37: needs to be factored into a conversion/fs-validation routine.
Later.

mslib/msui/remotesensing_dockwidget.py
```diff
@@ -55,7 +56,18 @@ def __init__(self, parent=None, view=None):
         self.setupUi(self)
 
         self.view = view
-        self.load = Loader(MSS_CONFIG_PATH, verbose=False)
+        if '://' in MSS_CONFIG_PATH:
```
joernu76, line 59: This needs to be moved. We need a "cache" directory here.
Later.

joernu76 (approved):

PR #964, opened 2021-05-17
improved the workflow, only on merge

### LINKED ISSUE #965: Coveralls action not working
Currently the [coveralls.yml is failing](https://github.com/Open-MSS/MSS/actions/runs/852597895) with the message
```
.github/workflows/coveralls.yml#L1
`pull` is not a valid event name
```
I think the correct term for triggering on pull requests is `pull_request`, not `pull`

ReimarBauer:


Marilyth:
It is hard to verify if this works as expected or not.
Have you done any test runs e.g. on a fork?

ReimarBauer:
I have forked the example
https://github.com/ReimarBauer/github-actions-pr-close-showcase/actions


example_on_merge

the configuration example for coveralls is from
https://github.com/conda-forge/conda-smithy/blob/master/.github/workflows/tests.yml

Marilyth:
I will play around with this a bit. Sadly testing out actions, especially ones as complex as on merged PRs is taking time.

ReimarBauer:
> I will play around with this a bit. Sadly testing out actions, especially ones as complex as on merged PRs is taking time.

Thx. If that works we probably can extend this too. I guess there are other functions on merge interesting. I can manually create updates until we have this productive.

.github/workflows/coveralls.yml
```diff
@@ -16,8 +21,16 @@ jobs:
...
     - name: coveralls
       env: 
-        COVERALLS_REPO_TOKEN: ${{ secrets.COVERALLS_REPO_TOKEN }}
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       run: |
+        echo PR #${{ github.event.number }} has been merged
         python -m pip install --upgrade pip
         pip install python-coveralls
-        coveralls mslib
+        coveralls --service=github mslib
```
Marilyth, line 29: If I am not mistaken, coveralls by itself does nothing but to upload a precreated coverage report, e.g. by pytest-cov.
As it stands here, coveralls will just upload an empty report with 0% coverage.

Also `mslib` must be removed from the `coveralls --service=github mslib` command as that is invalid syntax

.github/workflows/coveralls.yml
```diff
@@ -1,11 +1,16 @@
...
-# For more information see: https://help.github.com/actions/language-and-framework-guides/using-python-with-github-actions
+# This workflow is for coveralls on merges PRs
+# idea from https://brennerm.github.io/posts/trigger-github-actions-on-pr-close.html
 
 name: coveralls
 
-on: [ pull ]
+on:
+  pull_request:
+    types: [ closed ]
```
Marilyth, line 8: This seems very specific.
Why not create a coverage on every push, as opposed to every merged pull request?
It would make the rest of the action simpler.

Marilyth (changes requested):
Generally I think it would be easier to accommodate this into our default test action, as a test need to be performed anyway.
Here is a working example.
https://github.com/Marilyth/MSS/commit/d7b5dddce9ae0a708d1f7ee3d60115302558f900

.github/workflows/coveralls.yml
ReimarBauer: I agree :)
We know then also when a lot new code without tests gets added.

ReimarBauer:
will be differently implemented

PR #972, opened 2021-05-19
Add shortcuts dialog

### LINKED ISSUE #784: cheatsheet in help
We have a lot keys added to the controls of the views.

It may be good to give an overview in the documentation or in the app itselfs.

@joernu76 what do you think about this?

Marilyth:
Adds an application wide Alt+S shortcut and mainwindow menu item under Help, which will open a Dialog listing all shortcuts and their descriptions of all currently open windows.
fixes #784


ReimarBauer (commented):
would it be possible to do this for sideview and topview as well
dockwidgets

joernu76 (dismissed):
Perfect! Minor thing: The Alt+S shortcut does not appear, e.g., for TopView even though it works for that window as well.

Marilyth:
Alt+S now appears for every window

joernu76 (approved):

PR #974, opened 2021-05-19
enables to configure a EXPIRATION time in seconds, fixes #948

### LINKED ISSUE #948: lifetime of mscolab token configurable
we should set a config option for the lifetime of the token in the mscolab server settings script so that it can be individual set.

ReimarBauer:
we catch now various problems for an expired session,
besides on a timelimit a session also expires on server restart.


mslib/msui/mpl_pathinteractor.py
```diff
@@ -211,25 +213,26 @@ def update_from_WaypointsTableModel(self, wps_model):
...
-                break
-
-        self.intermediate_indexes = intermediate_indexes
-        self.ilats = lats
-        self.ilons = lons
-        self.itimes = times
-
-        # Call super method.
-        WaypointsPath.update_from_WaypointsTableModel(self, wps_model)
+        if lats is not None:
```
joernu76, line 216: So far the flightrack is required to consist of at least two points. otherwise a lot of stuff might break...

mslib/msui/mscolab.py
```diff
@@ -641,9 +641,10 @@ def add_projects(self):
             "token": self.token
         }
         r = requests.get(self.mscolab_server_url + '/projects', data=data)
-        _json = json.loads(r.text)
-        self.projects = _json["projects"]
-        self.add_projects_to_ui(self.projects)
+        if r.text != "False":
+            _json = json.loads(r.text)
+            self.projects = _json["projects"]
+            self.add_projects_to_ui(self.projects)
```
joernu76, line 691: No else branch with error dialogue?

mslib/msui/mscolab.py
```diff
@@ -653,12 +654,13 @@ def get_recent_pid(self):
             "token": self.token
         }
         r = requests.get(self.mscolab_server_url + '/projects', data=data)
-        _json = json.loads(r.text)
-        projects = _json["projects"]
-        p_id = None
-        if projects:
-            p_id = projects[-1]["p_id"]
-        return p_id
+        if r.text != "False":
```
joernu76, line 703: see above.

mslib/msui/mscolab.py
```diff
@@ -668,12 +670,13 @@ def get_recent_project(self):
             "token": self.token
         }
         r = requests.get(self.mscolab_server_url + '/projects', data=data)
-        _json = json.loads(r.text)
-        projects = _json["projects"]
-        recent_project = None
-        if projects:
-            recent_project = projects[-1]
-        return recent_project
+        if r.text != "False":
```
joernu76, line 721: see above

mslib/msui/mscolab_version_history.py
```diff
@@ -108,9 +108,12 @@ def load_current_waypoints(self):
...
         res = requests.get(url, data=data)
-        xml_content = json.loads(res.text)["content"]
-        waypoint_model = WaypointsTableModel(name="Current Waypoints", xml_content=xml_content)
-        self.currentWaypointsTable.setModel(waypoint_model)
+        if res.text != "False":
+            xml_content = json.loads(res.text)["content"]
+            waypoint_model = WaypointsTableModel(name="Current Waypoints", xml_content=xml_content)
+            self.currentWaypointsTable.setModel(waypoint_model)
+        else:
+            show_popup(self, "Error", "Some error occurred while fetching data!")
```
joernu76, line 116: see... here an action is taken for the False case.

joernu76 (dismissed):
I am not sure if not some specific action should be taken in case of the r.text == "False"  branch. Eg. disconnecting and showing a dialogue or similar...?

mslib/msui/mscolab.py
ReimarBauer: this has to be looked up later. There are more states. The easiest is network connection lost.

But if the session expires there is still a connection, but the user has no access to data. So there these requests don't have json data. 
session expired can be easily done by a server restart and also happen by reaching the expire time on the server. 

A follow up of these changes is #975

mslib/msui/mscolab_version_history.py
ReimarBauer: this needs to become reviewed with #975

mslib/msui/mscolab_admin_window.py
```diff
@@ -238,12 +242,13 @@ def delete_selected_users(self):
         }
         url = url_join(self.mscolab_server_url, "delete_bulk_permissions")
         res = requests.post(url, data=data)
-        res = res.json()
-        if res["success"]:
-            self.load_users_without_permission()
-            self.load_users_with_permission()
-        else:
-            self.show_error_popup(res["message"])
+        if res.text != "False":
```
Tanish0019, line 253: @ReimarBauer for this request we return JSON response. I believe this false check is only when user's token is expired? If so in else should we not call the logout state? Moreover instead of checking for False maybe we can check with status code instead?

mslib/msui/mpl_pathinteractor.py
ReimarBauer: > So far the flightrack is required to consist of at least two points. otherwise a lot of stuff might break...

when the flightrack can't be loaded then this is set to None, see mslib/utils.py

mslib/msui/mscolab_admin_window.py
ReimarBauer: > @ReimarBauer for this request we return JSON response. I believe this false check is only when user's token is expired? If so in else should we not call the logout state? Moreover instead of checking for False maybe we can check with status code instead?

@Tanish0019 
Yes, on expired.

I think the logout simplifies everything, all other widgets gets closed and can't create various popup messages, only one is needed to inform the user. 

I change to status code.

ReimarBauer: Wait, the problem is that we don't return json, only "False" on text, so there is no way to create json from that and look on the status code.

ReimarBauer: Currently I use only a popup message. further improvements should be done with  https://github.com/Open-MSS/MSS/issues/975

mslib/mscolab/sockets_manager.py
```diff
@@ -130,36 +130,39 @@ def handle_message(self, _json):
...
 
     def handle_message_delete(self, socket_message):
         message_id = socket_message["message_id"]
         p_id = socket_message["p_id"]
         user = User.verify_auth_token(socket_message['token'])
-        perm = self.permission_check_emit(user.id, int(p_id))
-        if perm:
-            self.cm.delete_message(message_id)
-            socketio.emit('delete-message-client', json.dumps({"message_id": message_id}), room=str(p_id))
+        if user is not None:
```
joernu76, line 161: I would probably simplify this to `if (use is not None) and self.permission_check_emit(user.id, int(p_id)):` due to the evaluation order of such expressions.

mslib/msui/mpl_qtwidget.py
```diff
@@ -926,11 +927,14 @@ def set_waypoints_model(self, model):
...
-                linecolor=appearance["colour_ft_vertices"],
-                markerfacecolor=appearance["colour_ft_waypoints"])
-            self.waypoints_interactor.set_vertices_visible(appearance["draw_flighttrack"])
+            try:
+                self.waypoints_interactor = mpl_pi.HPathInteractor(
+                    self.map, self.waypoints_model,
+                    linecolor=appearance["colour_ft_vertices"],
+                    markerfacecolor=appearance["colour_ft_waypoints"])
+                self.waypoints_interactor.set_vertices_visible(appearance["draw_flighttrack"])
+            except IOError as err:
```
joernu76, line 937: Why IOError of all exceptions...?

mslib/msui/mscolab.py
```diff
@@ -653,12 +656,15 @@ def get_recent_pid(self):
...
-        return p_id
+        if r.text != "False":
+            _json = json.loads(r.text)
+            projects = _json["projects"]
+            p_id = None
+            if projects:
+                p_id = projects[-1]["p_id"]
+            return p_id
+        else:
+            show_popup(self, "Error", "Session expired, new login required")
```
joernu76, line 711: Does this do the disconnect also?

mslib/utils.py
```diff
@@ -504,6 +504,9 @@ def path_points(points, numpoints=100, connection='linear'):
     """
     if connection not in ['linear', 'greatcircle']:
         return None, None
+    if points is None or len(points) == 0:
```
joernu76, line 506: two None or three None??

joernu76 (approved):

mslib/msui/mscolab.py
ReimarBauer: we can discuss how we can logout, the connection is still active and data can be saved when a new login is done. This will be continued in #975

mslib/utils.py
ReimarBauer: this what creates this state may need a refactoring. I found points = None and sometimes points = []

mslib/msui/mpl_qtwidget.py
ReimarBauer: Input/Output for now.

another wish?

joernu76: I am just wondering what kind of condition would raise this kind of error within that class. I guess the Exception is there for a reason. No action necessary.

mslib/utils.py
joernu76: I mean, line 506 is `return None, None`. Line 508 is `return None, None, None`. I have doubts that both are correct. Maybe neither...?

mslib/mscolab/sockets_manager.py
joernu76: No action necessary now.

mslib/msui/mpl_qtwidget.py
ReimarBauer: see at the change in mslib/msui/mpl_pathinteractor.py

```
if len(wp_vertices) == 0:
   raise IOError("mscolab session expired")
```

ReimarBauer: this is also something we maybe have to discuss/refactor. 
we have some places where we have an empty list or None .

mslib/utils.py
ReimarBauer: I should have added 
`# ToDo refactor soon there.`
I was surprised about the two kind of states.

Issue #940, opened 2021-05-10
RS testing fails sporadically


joernu76:
To test the remote sensing widget, astronomical data needs to be downloaded. This fails sporadically,
One might add the required data to the docker image to make testing more reliable...?

joernu76:
@Marilyth 
I would say it would be sufficient to testing downloading the data only when creating the docker image. Thus alleviating the IO load on the tests/speeding them up *and* being more reliable.
Also, using the docker image for running MSS would have the correct data already available.
Could you add the necessary data to the docker image?

ReimarBauer:
I assume this also slows down the start of topview and requesting version info

https://github.com/Open-MSS/MSS/issues/922
https://github.com/Open-MSS/MSS/issues/950

It seems also not often updated. 

![de421](https://user-images.githubusercontent.com/278728/118703281-3604af00-b816-11eb-86f8-95ef51b4fc29.png)

ReimarBauer:
Someone did already what I'm thinking about, and it is also on conda-forge
https://pypi.org/project/skyfield-data/
https://github.com/brunobord/skyfield-data

# Name                       Version           Build  Channel             
skyfield-data                  0.1.0            py_0  conda-forge         
skyfield-data                  1.0.0    pyh9f0ad1d_0  conda-forge         
skyfield-data                  1.1.0    pyh9f0ad1d_0  conda-forge         
skyfield-data                  2.0.0    pyhd3deb0d_0  conda-forge         
skyfield-data                  3.0.0    pyhd3deb0d_0  conda-forge 

example
```
from skyfield_data import get_skyfield_data_path
from skyfield.api import Loader
load = Loader(get_skyfield_data_path())
planets = load('de421.bsp') 
```

joernu76:
The de421.bsp is not updated often. Probably needs an update in 2200 or so.
Adding the packet and adopting the code correspondingly seems like the very best solution!

joernu76:
Checked it. Requires update in 2052. There is a function in the data-package called check_expirations, which should be added in a MSS test in the remote sensing widget part.

PR #976, opened 2021-05-20
Use skyfield_data package for big bsp file.

### LINKED ISSUE #693: testing - mock Loader.load of de421.bsp
For the  mslib/msui/_tests/test_topview.py::Test_MSSTopViewWindow::test_open_rs we should not load always de421.bsp 

just use a part or similar data

joernu76:
Fix issue #940


ReimarBauer (approved):
thx

PR #979, opened 2021-05-20
Add additional check for changed files on disk.

### LINKED ISSUE #886: WMS server provides incorrect figure in the case of  dataset update
In the calculations for scientific flight planning, it may happen, that a datafile is updated, having the identical filename, but more timesteps. It seems that the server does not check for the update and returns false figures by not correctly interpolating within the data. After re-starting the user interface and clearing the cache, a correct figure is shown.
It is not clear, whether this is a problem of the user interface or of the server.
wrong figure:
topview_contour_250 0_bug
correct figure:
topview_contour_250 0_correct

joernu76:
Fix #886

### DIFF
--- a/mslib/mswms/_tests/test_demodata.py
+++ b/mslib/mswms/_tests/test_demodata.py
@@ -28,9 +28,8 @@
 
 from past.builtins import basestring
 
-import imp
 import numpy as np
-from mslib._tests.constants import SERVER_CONFIG_FS, DATA_FS, ROOT_FS, SERVER_CONFIG_FILE, SERVER_CONFIG_FILE_PATH
+from mslib._tests.constants import SERVER_CONFIG_FS, DATA_FS, ROOT_FS, SERVER_CONFIG_FILE
 import mslib.mswms.demodata as demodata
 
 
@@ -40,9 +39,6 @@ class TestDemodata(object):
         assert DATA_FS.exists(u'.')
         assert SERVER_CONFIG_FS.exists(SERVER_CONFIG_FILE)
         assert len(DATA_FS.listdir(u'.')) == 23
-
-    def test_server_config_file(self):
-        imp.load_source('mss_wms_settings', SERVER_CONFIG_FILE_PATH)
 
     def test_get_profile(self):
         mean, std = demodata.get_profile("air_pressure", [1000, 10000, 50000], "air_temperature")
--- a/mslib/mswms/dataaccess.py
+++ b/mslib/mswms/dataaccess.py
@@ -102,12 +102,20 @@ class NWPDataAccess(metaclass=ABCMeta):
         fullpath -- if True, the complete path to the file will be returned.
                     Default is False, only the filename will be returned.
         """
-        filename = self._determine_filename(variable, vartype,
-                                            init_time, valid_time)
+        filename = self._determine_filename(
+            variable, vartype, init_time, valid_time)
         if fullpath:
             return os.path.join(self._root_path, filename)
         else:
             return filename
+
+    @abstractmethod
+    def is_reload_required(self, filenames):
+        """
+        Must be overwritten in subclass. Checks if a file was modified since last
+        read-in.
+        """
+        pass
 
     @abstractmethod
     def _determine_filename(self, variable, vartype, init_time, valid_time):
@@ -204,6 +212,7 @@ class DefaultDataAccess(NWPDataAccess):
         self._available_files = None
         self._filetree = None
         self._mfDatasetArgsDict = {"skip_dim_check": skip_dim_check}
+        self._file_cache = {}
 
     def _determine_filename(self, variable, vartype, init_time, valid_time, reload=True):
         """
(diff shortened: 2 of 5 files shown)


joernu76:
@ReimarBauer I had to deactivate the test in test_demodata for MSS tests to keep working. I have no idea how it ever worked before... (otherwise test_get_capabilities fails when executed *together* with test_demodata).
When the WMS class is instantiated, the layers are registered using the data_access classes from mss_wms_settings. The deactivated test re-import another mss_wms_settings and thus replaces the data_access classes.
In test_wms, the new classes are filled with data, but the old ones are still associated to the layers.
I do not know how anything works now, or before. I guess, the Server class should *not* be instantiated as global variable in the module....?

ReimarBauer:
> @ReimarBauer I had to deactivate the test in test_demodata for MSS tests to keep working. I have no idea how it ever worked before... (otherwise test_get_capabilities fails when executed _together_ with test_demodata).
> When the WMS class is instantiated, the layers are registered using the data_access classes from mss_wms_settings. The deactivated test re-import another mss_wms_settings and thus replaces the data_access classes.
> In test_wms, the new classes are filled with data, but the old ones are still associated to the layers.
> I do not know how anything works now, or before. I guess, the Server class should _not_ be instantiated as global variable in the module....?

ok, for deactivating it. I wonder currently too what I wanted with these lines, but I also wonder why it does make trouble. There should be only one mss_wms_settings.py in the test tmp directory in the search path for the tests. I guess in the past python 2 it had not imported because the namespace was already taken by the import in conftest. If it now imports than this may become a difference? At all it is not needed there and we don't want tests dependent of others.

ReimarBauer:
if we add nco to the development requirement, could you think about a test which manipulates a demodata nc file? and use this for an automated test.
All other tests have to cope with that situation too.

joernu76:
I am unsure about the nco comment.?

joernu76:
@Marilyth Why does that "Pytest MSS / Update-Images (push) " thingie fail?

ReimarBauer:
> I am unsure about the nco comment.?

We can do `mamba install nco` for developers
and by nco change a file after the server is started

joernu76:
@Marilyth can you do something about that multi-layer error? That (randomly?) crops up pretty often...

mslib/mswms/_tests/test_demodata.py
```diff
@@ -28,9 +28,10 @@
 
 from past.builtins import basestring
 
-import imp
+# import imp
```
ReimarBauer: comments can be now or in a "comment refactoring" review cleared or now. I do not see any added value to keep them.

mslib/mswms/dataaccess.py
```diff
@@ -213,7 +222,12 @@ def _determine_filename(self, variable, vartype, init_time, valid_time, reload=T
...
         assert self._filetree is not None, "filetree is None. Forgot to call setup()?"
         try:
-            return self._filetree[vartype][init_time][variable][valid_time]
+            filename = self._filetree[vartype][init_time][variable][valid_time]
+            mtime = os.path.getmtime(os.path.join(self._root_path, filename))
+            if filename in self._file_cache and mtime == self._file_cache[filename][0]:
+                return filename
+            else:
+                raise KeyError
         except KeyError:
```
ReimarBauer: os.path.getmtime can fail because of an OSError when path not exists or inaccessible is. The KeyError can occure only when that is passed.

When the KeyError happened the filename can be removed from self._file_cache

mslib/mswms/dataaccess.py
```diff
@@ -224,6 +238,25 @@ def _determine_filename(self, variable, vartype, init_time, valid_time, reload=T
...
+        for filename in filenames:
+            basename = os.path.basename(filename)
+            if basename not in self._file_cache:
+                reload = True
+                break
+            fullname = os.path.join(self._root_path, basename)
+            if not os.path.exists(fullname):
+                reload = True
+                break
+            mtime = os.path.getmtime(fullname)
```
ReimarBauer: Because we change files and while we can do this incorrectly we could try to catch a problem here and log it.

mslib/mswms/dataaccess.py
```diff
@@ -322,23 +355,46 @@ def _add_to_filetree(self, filename, content):
     def setup(self):
         # Get a list of the available data files.
         self._available_files = [
-            _filename for _filename in sorted(os.listdir(self._root_path)) if self._domain_id in _filename]
```
ReimarBauer: I don't care if it's sorted or not,

mslib/mswms/dataaccess.py
```diff
@@ -322,23 +355,46 @@ def _add_to_filetree(self, filename, content):
...
         for filename in self._available_files:
-            logging.info("Opening candidate '%s'", filename)
-            try:
-                content = self._parse_file(filename)
-            except IOError as ex:
-                logging.error("Skipping file '%s' (%s: %s)", filename, type(ex), ex)
-                continue
-            if content["vert_type"] not in self._elevations:
-                self._elevations[content["vert_type"]] = content["elevations"]
+            mtime = os.path.getmtime(os.path.join(self._root_path, filename))
```
ReimarBauer, line 349: I think we should try: .. getmtime except IOError. logging ...

ReimarBauer (changes requested):
We have some data on our servers mounted. NFS mounts behave different atomic as local files also there are caches involved. 

We should log when we can't get a file into the cache.

mslib/mswms/dataaccess.py
joernu76: Maybe. In case that someone removed the file in between listing the directory and parsing them here, it will not work anyway...

ReimarBauer (approved):
thx

PR #981, opened 2021-05-20
Import/Export Extensions, fixes #918

### LINKED ISSUE #497: import/load of files should be separate from project creation
**[Original report](https://bitbucket.org/wxmetvis/mss/issue/497) by Joern Ungermann (Bitbucket: [joernu76](https://bitbucket.org/joernu76), GitHub: [joernu76](https://github.com/joernu76)).**

One might need to update a project after exporting it and working on it in a separate tool by importing it again.

So we need a load/import feature for existing projects.

For non-empty projects a warning message box should be displayed.

The init with existing file feature might then be removed.

### LINKED ISSUE #918: Support more export options in mscolab
Currently when exporting and importing in mscolab, only .ftml is available. This should be extended to more formats.

ReimarBauer:


mslib/msui/mscolab.py
```diff
@@ -259,22 +279,38 @@ def text_changed(self):
...
-        try:
-            model = ft.WaypointsTableModel(xml_content=xml_content)
-        except SyntaxError:
-            show_popup(self, "Import Failed", f"The file - {file_name}, does not contain valid XML")
-            return
-        self.waypoints_model = model
-        if self.workLocallyCheckBox.isChecked():
-            self.waypoints_model.save_to_ftml(self.local_ftml_file)
-            self.waypoints_model.dataChanged.connect(self.handle_waypoints_changed)
+        file_name = fs.path.basename(file_path)
```
joernu76, line 290: does this work on windows? fs.path seems to work only on unix/fs-style paths

joernu76 (approved):

joernu76:
some additional comments:
* We add the csv plugins by default to the import/export plugins. This is currently missing. Might be added but is not that important. 
* We currently have at least two txt-type import plugins, which breaks the idea behind this change... Workaround is obvious.

mslib/msui/mscolab.py
```diff
@@ -259,22 +279,38 @@ def text_changed(self):
         self.loginButton.setEnabled(self.emailid.text() != "" and self.password.text() != "")
 
     def handle_import(self):
-        file_path = get_open_filename(self, "Select a file", "", "Flight track (*.ftml)")
+        if self.workLocallyCheckBox.isChecked():
```
joernu76, line 282: Why not here...?

ReimarBauer:
> some additional comments:
> 
> * We add the csv plugins by default to the import/export plugins. This is currently missing. Might be added but is not that important.
> * We currently have at least two txt-type import plugins, which breaks the idea behind this change... Workaround is obvious.

one can add any plugin to the json config. We can touch this code during the ui project again.

mslib/msui/mscolab.py
ReimarBauer: if there is  request we can add this. Also there is a ui project.

mslib/msui/mscolab.py
ReimarBauer: I assume it does, but will look up windows before the next  release ;)

PR #983, opened 2021-05-21
Changed vertical section interpolation to treat masked values as NaN

### LINKED ISSUE #811: missing values in netCDF data are ignored
values such as -1e30 used to flag missing data using the missing_value attribute are ignored by horizontal plots.

joernu76:
Array is properly masked, but interpolation routine ignored mask.
Filling the array with NaN should solve this issue.
Resulting array is properly masked again.

Fix #811


joernu76:
Has been hot-patched to the ML Cirrus server.

jgrooss (approved):
works now

PR #987, opened 2021-05-22
Verify valid token, fixes #975

### LINKED ISSUE #975: signalling expired mscolab session
A user will recognize when a session is expired on selecting the next flightpath that this has no flightpath, 0 points, no messages.

When the session expires while he/she is using topview or sideview than further drawing is not disabled but nothing is stored on the server. After a new login the user has a new session.

We need a way to inform the user that a new login is required.

ReimarBauer:


joernu76 (commented):
I am not sure I can properly review this before Wednesday.

ReimarBauer:
> I am not sure I can properly review this before Wednesday.

that's ok. I missed the holiday day. 

it is easy to try out by setting EXPIRATION=10 (seconds) into the conf.py temporary.
I dislike always the same if .. else that cries somehow for refactoring, later. For now I have the focus on make it work first.

Tanish0019:
As we're using self.verify_user_token in almost all the functions maybe we can extract it to a decorator in future refactor. That will save us 1 indent and the else case we handle in each function

joernu76 (dismissed):

joernu76 (approved):

Issue #511, opened 2019-11-12
Simplification of performance setup


ReimarBauer:
**[Original report](https://bitbucket.org/wxmetvis/mss/issue/511) by Jens-Uwe Grooß (Bitbucket: [grooss](https://bitbucket.org/grooss), GitHub: [grooss](https://github.com/grooss)).**

----------------------------------------

Currently the performance is calculated from two given numbers provided in the "performance menu" on the table view tab.  
\(1\) The total start mass including fuel  
\(2\) the mass of fuel at start.

One can check, if a flight may be too long, if "remaining fuel" runs into negative numbers". As long as flights have always the same length one does not need to change these numbers, however, for a shorter flight one needs to alter both numbers consistenly.

I propose to change this into the parameters 

mass of empty aircraft \(including fuel safety reserve\)  
and  
maximum possible fuel

Then the "remaining fuel" could be displayed such that it always ends with zero and with variable initial fuel to be calculated. If initial fuel mass would exceed maximum possible fuel mass, this could be indicated e.g. by changing the color in the  "remaining fuel" column to zero.

With this, it is not necessary to alter these numbers given a fixed aircraft with the same setup \(payload\).

\(Version 1.9.0\)

joernu76:
Seems reasonable and probably simple...?

joernu76:
Partially adressed so far.

ReimarBauer:
needs currently no further improvements

PR #1003, opened 2021-05-27
Change performance JSON file format in preparation for #511

joernu76:
Changed documentation and implementation. Old configurations and JSON
files are still supported (i.e. full backwards-compatability).

The second part of the issue of computing fuel backwards has not yet
been implemented, but this change is a prerequisite for that.

See #511


joernu76:
I'd like to have this in 4.0 because it can be interpreted as an API change due to the slightly different file format.
The remaining part of #511 may be implemented for 4.1

ReimarBauer (changes requested):
sounds  good, but I want it currently on hold, because that needs changes on the campaign user setup.

If we want  this in 4.0 @jgrooss and @jroettenbacher should tell if that is now doable. ~40 participants have to use an updated file.

ReimarBauer:
> sounds good, but I want it currently on hold, because that needs changes on the campaign user setup.
> 
> If we want this in 4.0 @jgrooss and @jroettenbacher should tell if that is now doable. ~40 participants have to use an updated file.

discussed in a meeting, changes are backwards compatible

ReimarBauer (approved):

PR #1008, opened 2021-05-28
updated documentation

### LINKED ISSUE #1007: improve docs
update instructions
new features

open-mss-build:


ReimarBauer:
on windows I had to enter

```

mamba update mss=4.0.0

```
  No specs added or removed.

  Package  Version  Build           Channel                 Size
------------------------------------------------------------------
  Install:
------------------------------------------------------------------

  gpxpy      1.4.2  py_0            conda-forge/noarch     38 KB
  mss        4.0.0  py39hcbf5309_0  conda-forge/win-64      3 MB

  Summary:

  Install: 2 packages

  Total download: 3 MB

------------------------------------------------------------------

Confirm changes: [Y/n]
```
 
skyfield-data was on my system already installed

File: docs/installation.rst [hunk 1]
```diff
@@ -70,7 +70,9 @@ variables. ::
 update
 ++++++
 For updating an existing MSS installation to the current version, it is best to install
-it into a new environment.
+it into a new environment. If your current version is not far behind the new version
```
joernu76, line 73: strong disagree. This might fail and leave you without a usable installation. It's IMO fine to do it, but we should not suggest it to arbitrary users.

docs/usage.rst
```diff
@@ -92,14 +92,15 @@ importing changed files back in addition to the main FTML format. These filters
...
+as an example for the definition of own plugins. Take care that added plugins use different file extensions.
+They are listed below. The CSV plugin is enabled by default.
+Enabling the experimental FliteStar text import plugin would require those lines in
 the UI settings file:
 
 .. code:: text
 
     "import_plugins": {
-        "FliteStar": ["txt", "mslib.plugins.io.flitestar", "load_from_flitestar"]
+        "FliteStar": ["fls", "mslib.plugins.io.flitestar", "load_from_flitestar"]
```
joernu76, line 103: I am unsure about this change because the FliteStar tool will use the "txt" extension. This requires a manual rename of the file by the user, which is impossibly difficult for certain users (i.e the OS might scare them out of it).

docs/installation.rst
ReimarBauer: I think we should do. I updated the description, may be that is better. I use it since months.

joernu76 (approved):

docs/usage.rst
ReimarBauer: I added it to be consistant with one extension for one type. But anyway it is always only a suggestion.

PR #1012, opened 2021-05-31
Simplify plotting classes by preparing necessary information in base classes

### LINKED ISSUE #1011: Simplify creation of plotting classes by better preparing necessary data
Currently, the 2-D air_pressure field needs to be manually computed, which should be automated.

Also lon/lat 2-D fields have to always be computed in the same fashion using meshgrid and basemap projection.

Vertical plots require a similar thing to get the 2-D field for the horizontal axis.

All these things should be prepared before calling the _plot_style function.

joernu76:
* Added proper 2-D lon/lat coordinates for horizontal plots
* Added propler 2-D lat coordinate for vertical plots
* Provide air_pressure field for vertical plots for pressure axis data

Fix #1011


ReimarBauer (commented):
@joernu76 I'm not sure if this sample is linked from one of the reST files. Only by a link it is found on our docs over mss.rtfd.io

joernu76:
> @joernu76 I'm not sure if this sample is linked from one of the reST files. Only by a link it is found on our docs over mss.rtfd.io

This is the plotting class from a past campaign, which I port forward as example. This should be restructured when we do the plotting class gallery.

ReimarBauer (approved):
reviewing of refactoring is easier when we have some more tests.

thx for the improvement.

Issue #673, opened 2021-02-10
error message needs to be improved for CERTIFICATE_VERIFY_FAILED on mscolab


ReimarBauer:
the mscolab GUI did not request a connection to a ssl server if  CERTIFICATE_VERIFY_FAILED

and tells only:

MSColab server isn't active

ReimarBauer:
We need to be a bit more verbose for such problems  in mscolab.connect_handler

```
 Max retries exceeded with url: /status (Caused by SSLError(SSLCertVerificationError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:1123)')))
```

nakul-shahdadpuri:
@ReimarBauer can I take this up?

ReimarBauer:
Chain failures just return  "MSColab server isn't active"

ReimarBauer:
seems we have to move requests.exceptions.ConnectionError to the end

```
diff --git a/mslib/msui/mscolab.py b/mslib/msui/mscolab.py
index 6f29272b..f8b0853b 100644
--- a/mslib/msui/mscolab.py
+++ b/mslib/msui/mscolab.py
@@ -215,9 +215,6 @@ class MSSMscolabWindow(QtWidgets.QMainWindow, ui.Ui_MSSMscolabWindow):
                 self.password.setText(config_loader(dataset="MSCOLAB_password"))
             else:
                 show_popup(self, "Error", "Some unexpected error occurred. Please try again.")
-        except requests.exceptions.ConnectionError:
-            logging.debug("MSColab server isn't active")
-            show_popup(self, "Error", "MSColab server isn't active")
         except requests.exceptions.InvalidSchema:
             logging.debug("invalid schema of url")
             show_popup(self, "Error", "Invalid Url Scheme!")
@@ -227,6 +224,9 @@ class MSSMscolabWindow(QtWidgets.QMainWindow, ui.Ui_MSSMscolabWindow):
         except requests.exceptions.SSLError:
             logging.debug("Certificate Verification Failed")
             show_popup(self, "Error", "Certificate Verification Failed")
+        except requests.exceptions.ConnectionError:
+            logging.debug("MSColab server isn't active")
+            show_popup(self, "Error", "MSColab server isn't active")
         except Exception as e:
             logging.debug("Error %s", str(e))
             show_popup(self, "Error", "Some unexpected error occurred. Please try again.")
```


Are there better solutions?




[...]

PR #1020, opened 2021-06-09
Increase coverage

### LINKED ISSUE #1013: More test coverage
Currently around 74% of MSS is covered by tests.
It would be beneficial to bring that number up closer to 80%, to catch more issues and side effects before they are merged.

Marilyth:
This PR adds some tests to increase the test coverage of MSS to 80%
fixes #1013 

Currently there is an odd issue with pytest sometimes crashing midway. Until this is resolved this stays a draft.
Edit: Hasn't happened for the last 2 days, I assume this is fixed.


mslib/msui/_tests/test_mscolab.py
```diff
@@ -211,16 +247,44 @@ def test_set_exported_file(self, mockopen, mockmessage):
         QtWidgets.QApplication.processEvents()
         assert self.window.listProjects.model().rowCount() == 1
 
-    def test_add_project(self):
+    @mock.patch("PyQt5.QtWidgets.QErrorMessage")
+    def test_add_project(self, mockbox):
         # ToDo test needs to be independent from test_user_delete
```
ReimarBauer: If this is now independent, so that each can run without the other the comment can be removed

mslib/msui/_tests/test_mscolab.py
```diff
@@ -211,16 +247,44 @@ def test_set_exported_file(self, mockopen, mockmessage):
...
+        with mock.patch("requests.Session.post", new=ExceptionMock(requests.exceptions.ConnectionError).raise_exc):
+            self._login()
+        with mock.patch("requests.Session.post", return_value=response(201, "False")):
+            self._login()
+        with mock.patch("requests.Session.post", return_value=response(401, "False")):
+            self._login()
+
+        # No return after self.error_dialog.showMessage('Oh no, server authentication were incorrect.')
+        # causes 4 instead of 3 messages, I am not sure if this is on purpose.
+        assert mockbox.return_value.showMessage.call_count == 4
```
ReimarBauer, line 285: from the naming I would have expected mockauth here, or?

joernu76 (dismissed):

mslib/msui/mscolab.py
```diff
@@ -260,6 +260,9 @@ def connect_handler(self):
                     self.loginButton.setEnabled(True)
             else:
                 show_popup(self, "Error", "Some unexpected error occurred. Please try again.")
+        except requests.exceptions.SSLError:
```
ReimarBauer, line 263: I noted some time ago tat I got the wrong ConnectionError and noted this hear
https://github.com/Open-MSS/MSS/issues/673#issuecomment-830837312

Moving the SSL on top is a good idea and helps to find faster when you have e.g. an incomplete chain certificate.

But what is with the others? If they also get catched by the ConnectionError, we better use the chance now to move that on the end of the list.

mslib/msui/mscolab.py
```diff
@@ -260,6 +260,9 @@ def connect_handler(self):
...
+            logging.debug("Certificate Verification Failed")
+            show_popup(self, "Error", "Certificate Verification Failed")
```
ReimarBauer, line 265: good to move this before ConnectionError, https://github.com/Open-MSS/MSS/issues/673

Perhaps moving this ConnectionError to the bottom and a comment will have no impact for others.

ReimarBauer (changes requested):
prinicipial approved ;)
please see comments.

mslib/msui/mscolab.py
Marilyth: As far as I can see ConnectionError is only the superclass of SSLError. The others get caught correctly.
While not specifically tested, I did have a look at the mock output and made sure this was the case.

Should I move it to the bottom regardless, for the future?

mslib/msui/_tests/test_mscolab.py
Marilyth: QErrorMessage is the one on which showMessage is called, as such that needs to be checked on the mockbox.

mslib/msui/mscolab.py
Marilyth: I moved it to the bottom now. In case we need more errors in the future this is safer.

mslib/msui/mscolab.py
Marilyth: ConnectionError is now the last requests Exception ot be caught.

mslib/msui/_tests/test_mscolab.py
ReimarBauer: somehow I thought before that the order of the input arguments is given by the order of the patch commands. Or that was my private rule, not found something in the documentation on a quick search.
The second one is @mock.patch("PyQt5.QtWidgets.QErrorMessage") I automatically would have looked on mockauth.

ReimarBauer: found something, it is by order.
https://stackoverflow.com/questions/33353855/how-does-mock-patch-know-which-parameter-to-use-for-each-mock-object

ReimarBauer: it is correct from, the variable, but for reading I would have epected mockauth connected to the auth dialog mock patch.

mslib/msui/mscolab.py
ReimarBauer: Thx

mslib/msui/_tests/test_mscolab.py
Marilyth: I wasn't sure about the order either. Basically I found the top decorator is the last function argument.
QErrorMessage is the mockbox and the Auth dialogue is the mockauth.

ReimarBauer (approved):
\o/

PR #1021, opened 2021-06-09
Linear view mscolab bug

### LINKED ISSUE #1022: Table view not opening once Linear view is created in MSColab
Trying to create tableview after creating linearview in an MSColab Project will just raise the already existing linearview window.

aravindm711:
1. Tableview does not open after opening linearview
2. On closing a view main window appears instead of mscolab window

### DIFF
--- a/mslib/msui/mscolab.py
+++ b/mslib/msui/mscolab.py
@@ -1271,6 +1271,7 @@ class MSSMscolabWindow(QtWidgets.QMainWindow, ui.Ui_MSSMscolabWindow):
         for index, window in enumerate(self.active_windows):
             if window._id == value:
                 del self.active_windows[index]
+                self.raise_()
 
     def setIdentifier(self, identifier):
         self.identifier = identifier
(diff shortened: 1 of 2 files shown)


mslib/msui/mscolab.py
```diff
@@ -1271,6 +1271,7 @@ def handle_view_close(self, value):
         for index, window in enumerate(self.active_windows):
             if window._id == value:
                 del self.active_windows[index]
+                self.raise_()
```
Marilyth, line 1274: I am not sure why, but this line does nothing on my end. Same for self.show(), self.activateWindow() and self.showNormal().
Maybe that is a Linux issue.
On Windows I can not even change the order of the windows. MSColab is always in front of Main and the views are always in front of MSColab.
What OS are you using for the main window to pop up after closing a window?

Marilyth (commented):
Thanks. The linearview fix is also part of #1020 but it can't hurt to fix it sooner.

mslib/msui/mscolab.py
aravindm711: I'm using a mac for development. It's usually the thing with Macs, the first created window would raise to the front.

Marilyth: Unfortunately I can't test this on mac, but I will take your word for it working the way it should.

Marilyth (approved):

PR #1028, opened 2021-06-11
Allow specification of file for linear defaultstyle

### LINKED ISSUE #1027: Make linear styles work on non .ml files
So far the linear styles require .ml files, which not every server provides.
It would be beneficial to make it work for other file type like .pl as well.

Marilyth:
Makes the filetype for the linear defaultstyle choosable, so long as it contains an air_pressure field or is a pressure level file.
fixes #1027

### DIFF
--- a/mslib/mswms/mss_plot_driver.py
+++ b/mslib/mswms/mss_plot_driver.py
@@ -797,10 +797,10 @@ class LinearSectionDriver(VerticalSectionDriver):
             cross_section = utils.interpolate_vertsec(var_data, self.lat_data, lon_data, self.lats, self.lons)
             # Create vertical interpolation factors and indices for subsequent variables
             # TODO: Improve performance for this interpolation in general
-            if name == "air_pressure" or "air_pressure" not in self.data_vars and len(factors) == 0:
+            if len(factors) == 0:
                 for index_lonlat, alt in enumerate(self.alts):
-                    pressures = cross_section[:, index_lonlat] if "air_pressure" in self.data_vars \
-                        else self.vert_data[::-self.vert_order] * 100 if self.vert_units.lower() == "hpa" else 1
+                    pressures = cross_section[:, index_lonlat] if name == "air_pressure" \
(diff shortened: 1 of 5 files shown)


mslib/mswms/mss_plot_driver.py
```diff
@@ -796,9 +797,10 @@ def _load_interpolate_timestep(self):
             cross_section = utils.interpolate_vertsec(var_data, self.lat_data, lon_data, self.lats, self.lons)
             # Create vertical interpolation factors and indices for subsequent variables
             # TODO: Improve performance for this interpolation in general
-            if name == "air_pressure":
+            if name == "air_pressure" or "air_pressure" not in self.data_vars and len(factors) == 0:
```
joernu76: I recently "automagically" added air_pressure to vertical sections in mslib/mswms/mpl_vsec.py lines 166ff. Maybe yoou could that similarly beforehand?

joernu76 (dismissed):

mslib/mswms/mss_plot_driver.py
Marilyth: I actually did use your vsec solution, but it turned out to be not quite what I needed. And instead of modifying your solution I decided to just use self.vert_data

ReimarBauer: I'm not sure if I get this if condition right, 

factors creates the limitation that only once this is called. if we had "air_pressure" in the self.data_vars it would be on index 0. And if we don't have it while factors is empty it runs, but it can't not be executed more than once, so could that be simplified to len(factors) == 0 ?

ReimarBauer (changes requested):
see my question

mslib/mswms/mss_plot_driver.py
Marilyth: You are right of course. I changed it now.

ReimarBauer (dismissed):
Thx

ReimarBauer (approved):
Thx

PR #1031, opened 2021-06-14
Add file change test

Marilyth:
Adds a test which modifies the .pl demodata file and asserts whether the image the server returns has changed.
It may cause undefined race conditions if run in parallel with other tests, so it is skipped in that case.

### DIFF
--- a/mslib/mswms/dataaccess.py
+++ b/mslib/mswms/dataaccess.py
@@ -391,6 +391,8 @@ class DefaultDataAccess(NWPDataAccess):
                     logging.error("Skipping file '%s' (%s: %s)", filename, type(ex), ex)
                     continue
                 self._file_cache[filename] = (mtime, content)
+                if content["vert_type"] not in self._elevations:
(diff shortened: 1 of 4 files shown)


mslib/mswms/dataaccess.py
```diff
@@ -391,6 +391,8 @@ def setup(self):
                     logging.error("Skipping file '%s' (%s: %s)", filename, type(ex), ex)
                     continue
                 self._file_cache[filename] = (mtime, content)
+                if content["vert_type"] not in self._elevations:
+                    self._elevations[content["vert_type"]] = content["elevations"]
```
joernu76, line 395: nice catch. did you run into an error or simply see that this is wrong?

joernu76 (approved):

mslib/mswms/dataaccess.py
Marilyth: I ran into an error trying to mock another data directory. It didn't work out in the end but at least the error is caught.

PR #1033, opened 2021-06-16
Fix pyproj 3.1.0 error

### LINKED ISSUE #1032: Docker containers fail, pyproj new release
All containers failed at the same time, at the same test: https://github.com/Open-MSS/dockerhub/runs/2836364461?check_suite_focus=true

Changed libraries from yesterday are:
`pcre                      8.44                 he6710b0_0` ->`pcre                      8.45                 h9c3ff4c_0    conda-forge`
`pyproj                    3.0.1            py39h50a64a9_0    conda-forge` -> `pyproj                    3.1.0            py39h50a64a9_3    conda-forge`
`six                       1.16.0             pyh6c4a22f_0    conda-forge` -> `six                       1.16.0             pyhd3eb1b0_0`

I assume this has something to do with the pyproj update.

Marilyth:
A [recent update in pyproj](https://github.com/pyproj4/pyproj/commit/e0b9453d29e7c9475467268d6838dac8503ab145#diff-884caf2af6e10fc8e0b305d38b427edc7b43d9ed3ec739aed1be2de94049ab7dR279) now throws an error when trying to determine 0 points between A and B, which is the case when trying to draw a path between the same waypoints, where previously it returned an empty list.
This fix calls gc.npts only if npoints is bigger than 0 to avoid this error.
Fixes #1032

### DIFF
--- a/mslib/msui/mpl_map.py
+++ b/mslib/msui/mpl_map.py
@@ -561,7 +561,9 @@ class MapCanvas(basemap.Basemap):
             # projection, gc.npts() returns lons that connect lon1 and lat2, not lon1 and
             # lon2 ... I cannot figure out why, maybe this is an issue in certain versions
             # of pyproj?? (mr, 16Oct2012)
-            lonlats = gc.npts(lons[i], lats[i], lons[i + 1], lats[i + 1], npoints)
+            lonlats = []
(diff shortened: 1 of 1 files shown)


ReimarBauer (approved):
ok

PR #1039, opened 2021-06-17
displaying project names with a dash correctly, fixes #1036

### LINKED ISSUE #1036: "-" in mscolab project name causes project name to be not properly displayed on plots and prompts
Problem in Top View plot:
grafik
Problem when trying to delete project:
grafik

ReimarBauer:


mslib/msui/mscolab.py
```diff
@@ -845,7 +845,7 @@ def set_active_pid(self, item):
             # set active_pid here
             self.active_pid = item.p_id
             self.access_level = item.access_level
-            self.active_project_name = item.text().split("-")[0].strip()
+            self.active_project_name = item.text().rsplit("-", 1)[0].strip()
```
jroettenbacher, line 848: This will split on the most right dash, correct? That way it avoids to already split at the project name, right?
If so, then this should be fine.
However, I wonder why we need to retrieve the project name this way and not have it as a variable after a project is created? But I'm not too deep into mscolab...

jroettenbacher (approved):

mslib/msui/mscolab.py
ReimarBauer: We can improve this, the improvement of the API will go into a feature release. 

We have a client server communication, or better said a server to many client communication.  In mscolab we have an additional layer to the local file based storage of a flightpath. The additional layer manages the access by different users to the flightpath object. At the moment you get added to edit a flightpath and you edit this is all communicated by socket/IO and happens by the server part. You see this in the client and it behaves similiar to the "single user" file based mode.

joernu76: I would prefer to set the project_name to the QWidget similar to how the access_level and p_id are set and handled.
This is 99.9% correct now but certainly more brittle than a solution that does *not* try to extract something from a string, which we could have simply directly stored.

Also some other "split" usages have been missed.
What about this:

```
 (e) base  ~  src  mss  git diff                                                                                                                                                                                                                                                                               stable 
diff --git a/mslib/msui/mscolab.py b/mslib/msui/mscolab.py
index d50799e0..5b02ec95 100644
--- a/mslib/msui/mscolab.py
+++ b/mslib/msui/mscolab.py
@@ -816,6 +816,7 @@ class MSSMscolabWindow(QtWidgets.QMainWindow, ui.Ui_MSSMscolabWindow):
             widgetItem = QtWidgets.QListWidgetItem(project_desc, parent=self.listProjects)
             widgetItem.p_id = project["p_id"]
             widgetItem.access_level = project["access_level"]
+            widgetItem.project_path = project["path"]
             if widgetItem.p_id == self.active_pid:
                 selectedProject = widgetItem
             self.listProjects.addItem(widgetItem)
@@ -845,7 +846,7 @@ class MSSMscolabWindow(QtWidgets.QMainWindow, ui.Ui_MSSMscolabWindow):
             # set active_pid here
             self.active_pid = item.p_id
             self.access_level = item.access_level
-            self.active_project_name = item.text().split("-")[0].strip()
+            self.active_project_name = item.project_path
             self.waypoints_model = None
             # set active flightpath here
             self.load_wps_from_server()
@@ -1165,12 +1166,9 @@ class MSSMscolabWindow(QtWidgets.QMainWindow, ui.Ui_MSSMscolabWindow):
             for i in range(self.listProjects.count()):
                 item = self.listProjects.item(i)
                 if item.p_id == p_id:
-                    desc = item.text().split(' - ')
-                    project_name = desc[0]
-                    desc[-1] = access_level
-                    desc = ' - '.join(desc)
-                    item.setText(desc)
+                    project_name = item.project_path
                     item.access_level = access_level
+                    item.setText(f'{project_name} - {item.access_level}')
                     break
             if project_name is not None:
                 show_popup(self, "Permission Updated",
@@ -1212,10 +1210,11 @@ class MSSMscolabWindow(QtWidgets.QMainWindow, ui.Ui_MSSMscolabWindow):
             item = self.listProjects.item(i)
             if item.p_id == p_id:
                 remove_item = item
+                break
         if remove_item is not None:
-            logging.debug("remove_item: %s" % remove_item)
+            logging.debug("remove_item: %s", remove_item)
             self.listProjects.takeItem(self.listProjects.row(remove_item))
-            return remove_item.text().split(' - ')[0]
+            return remove_item.project_name
 
     @QtCore.Slot(int, int)
     def handle_revoke_permission(self, p_id, u_id):
```

joernu76 (changes requested):

mslib/msui/mscolab.py
ReimarBauer: @joernu76  I'd added a test too, may be we close this PR and you can propose your solution and the test?
If it is a bigger API change that need to go into a feature release.

joernu76: I can prepare that as a MR as well, if you like.

ReimarBauer:
solved in a different PR

PR #1042, opened 2021-06-18
Plot Gallery

Marilyth:
- [x] Server side plot generation of all available layers
- [x] WMS hosted plot website
- [x] Command line options for plot/code generation 
- [x] Integration of plots into Sphinx Documentation
- [x] Make Sphinx documentation automatic via readthedocs
- [x] Add an in depth example of plot creation via a template
- [x] Generation of code for each plot
- [x] Generation of mss_wms_settings.py for specific plot
- [x] Extension of existing demodata to show off every plot available
- [x] Tests

This is the current WIP for the plot gallery, an extension of the wms.
An example (might be outdated) can be found here https://mss-test.readthedocs.io/en/latest/gallery/index.html


mslib/index.py
```diff
@@ -104,6 +106,7 @@ def get_topmenu():
              ((url_for('about'), 'About'),
               (url_for('install'), 'Install'),
               (url_for('help'), 'Help'),
+              (url_for('plots'), 'Plots'),
```
ReimarBauer: 'Help' should be the last option in the list

ReimarBauer:
We are based on  the framework twitter bootstrap. Have you looked on gallery implementations for this? e.g. https://freefrontend.com/bootstrap-galleries/

ReimarBauer:
We have to take care that a user knows that he has to call a command after he configured his e.g. apache or nginx server to get the gallery online. Probably also to repeat this when mss_wms_setting.py gets changes.

When we could setup the gallery with the installation of mss, we could add that command to  https://github.com/conda-forge/mss-feedstock/blob/master/recipe/post-link.sh  and also the "undo" to pre-unlink.sh

After removing mss we have to take care that all installation dirs got emptied.

ReimarBauer:
Our servers are configured based on this pattern
https://mss.readthedocs.io/en/stable/deployment.html#one-instance

We decoupled the installation code from the users instance configuration. I also would prefer when such a config is used to create the files in such a directory below the INSTANCE or wsgi directory or the user get asked when there is a mss_wms_settings.py file used,

mslib/mswms/gallery_builder.py
```diff
@@ -0,0 +1,279 @@
...
+
+static_location = os.path.join(os.path.dirname(os.path.abspath(__file__)), "..", "static")
+
+begin = """
+<!DOCTYPE html>
+<html>
+<head>
+<meta name="viewport" content="width=device-width, initial-scale=1">
+<style>
+body {font-family: Arial;}
```
ReimarBauer: Arial is a none free font
https://www.fonts.com/font/monotype/arial/licenses

choose an alternative

ReimarBauer (changes requested):

mslib/index.py
Marilyth: Done

mslib/mswms/gallery_builder.py
Marilyth: Done

mslib/mswms/gallery_builder.py
```diff
@@ -0,0 +1,363 @@
...
+from PIL import Image
+import io
+from matplotlib import pyplot as plt
+import defusedxml.ElementTree as etree
+import inspect
+from mslib.mswms.mpl_vsec import AbstractVerticalSectionStyle
+from mslib.mswms.mpl_lsec import AbstractLinearSectionStyle
+
+
+static_location = os.path.join(os.path.dirname(os.path.abspath(__file__)), "..", "static")
```
ReimarBauer: in the past those where written in uppercase to identify as global constants for the module.

mslib/mswms/gallery_builder.py
```diff
@@ -0,0 +1,363 @@
...
+
+{files}
+""")
+
+
+def write_js(sphinx=False):
+    """
+    Writes the plots.js file containing the gallery
+    """
+    location = docs_location if sphinx else os.path.join(static_location, "docs")
```
ReimarBauer: I would prefer to be able to use also directory which is not in the installed environment.

mslib/mswms/demodata.py
```diff
@@ -80,7 +80,13 @@
...
 vertically_integrated_probability_of_wcb_occurrence
 dimensionless
- 0.25 0.2"""
+ 0.25 0.2
+tropopause_altitude
+km
+ 10 3
+secondary_tropopause_altitude
+km
+ 30 3"""
```
joernu76: 30 -> 12

mslib/mswms/demodata.py
```diff
@@ -1196,7 +1202,10 @@ def create_data(self):
                 ("U", "eastward_wind"),
                 ("V", "northward_wind"),
                 ("W", "lagrangian_tendency_of_air_pressure"),
-                ("Q", "specific_humidity")):
+                ("Q", "specific_humidity"),
+                ("GPH", "geopotential_height"),
+                ("PT", "air_potential_temperature"),
```
joernu76: PT -> THETA

mslib/mswms/gallery_builder.py
```diff
@@ -0,0 +1,363 @@
...
+
+    with open(os.path.join(location, "code", f"{l_type}_{plot_object.name}.md"), "w+") as md:
+        md.write(f"**How to use this plot**  \n1. Make sure you have the required datafields "
+                 f"({', '.join(f'`{field[1]}`' for field in plot_object.required_datafields)})  \n"
+                 f"2. [Download this file](/mss/code/{l_type}_{plot_object.name}.md?download=True)  \n"
+                 f"3. Put this file into your mss_wms_settings.py directory, e.g. `~/mss`  \n"
+                 f"4. Append this code into your `mss_wms_settings.py`:  \n")
+        md.write(f"---\n```python\nfrom {l_type}_{plot_object.name} import {plot_object.__class__.__name__}\n"
+                 f"register_{layer}_layers = [] if not register_{layer}_layers else register_{layer}_layers\n"
+                 f"register_{layer}_layers.append(({plot_object.__class__.__name__}, [next(iter(data))]))\n```\n---\n")
```
joernu76: "next(iter(data)) " does this work for any server?

mslib/mswms/mss_plot_driver.py
```diff
@@ -274,7 +274,7 @@ def update_plot_parameters(self, plot_object=None, figsize=None, style=None,
         figsize = figsize if figsize is not None else self.figsize
         noframe = noframe if noframe is not None else self.noframe
         init_time = init_time if init_time is not None else self.init_time
-        valid_time = valid_time if valid_time is not None else self.valid_time
```
joernu76, line 277: I propose to consistently use in the server "valid_time" instead of fc_time. Maybe in a separate issue.

mslib/mswms/demodata.py
```diff
@@ -1196,7 +1202,10 @@ def create_data(self):
...
+                ("D", "divergence_of_wind")):
```
joernu76: "D" -> "DIV"

mslib/mswms/mss_plot_driver.py
Marilyth: I agree. I am just reacting to errors here, since self.valid_time actually doesn't exist.

mslib/mswms/demodata.py
Marilyth: Done

mslib/mswms/demodata.py
Marilyth: Done

mslib/mswms/demodata.py
Marilyth: Done

mslib/mswms/gallery_builder.py
Marilyth: It takes the first key of the data defined in the mss_wms_settings.py, so for most cases it works fine.
If there are more than one datasets and the plot doesn't derive from the first one, the user should change it.
Alternatively one could just list all datasets there `list(data.keys())` but that feels dirty.

ReimarBauer:
for counting purposes
xdist test rerun because of
mslib/mswms/_tests/test_mss_plot_driver.py::Test_VSec::test_VS_HorizontalVelocityStyle_01

ReimarBauer:
@Marilyth this happens on a fresh fork
```
(mssdev) reimar@infinte:~/PycharmProjects/2021/Marilyth/MSS/docs$ make html
sphinx-build -b html -d _build/doctrees   . _build/html
Sphinx v4.0.2 in Verwendung

/!\ existing server config: "mss_wms_settings.py" for demodata not overwritten!
            

/!\ existing server auth config: "mss_wms_auth.py" for demodata not overwritten!
                
2021-06-28 17:06:01               setup || Files identified for domain 'EUR_LL015': ['20121017_12_ecmwf_forecast.ALTITUDE_LEVELS.EUR_LL015.036.al.nc', '20121017_12_ecmwf_forecast.CC.EUR_LL015.036.ml.nc', '20121017_12_ecmwf_forecast.CIWC.EUR_LL015.036.ml.nc', '20121017_12_ecmwf_forecast.CLWC.EUR_LL015.036.ml.nc', '20121017_12_ecmwf_forecast.DIV.EUR_LL015.036.ml.nc', '20121017_12_ecmwf_forecast.EMAC.EUR_LL015.036.ml.nc', '20121017_12_ecmwf_forecast.GPH.EUR_LL015.036.ml.nc', 

Configuration error:
There is a programmable error in your configuration file:

  File "/home/reimar/miniconda3/envs/mssdev/lib/python3.9/site-packages/sphinx/config.py", line 323, in eval_config_file
    exec(code, namespace)
  File "/home/reimar/PycharmProjects/2021/Marilyth/MSS/docs/conf.py", line 53, in <module>
    from gallery.plot_examples import HS_template, VS_template
ModuleNotFoundError: No module named 'gallery'

Makefile:58: recipe for target 'html' failed
make: *** [html] Error 2
(mssdev) reimar@infinte:~/PycharmProjects/2021/Marilyth/MSS/docs$ git branch
* i521-plots
```

ReimarBauer:
@Marilyth this happens on a fresh clone. It works when the path of the gallery dir was added to the search path. I do prefer a solution where we just can do a standard sphinx `make html`. 

```
(mssdev) reimar@infinte:~/PycharmProjects/2021/Marilyth/MSS/docs$ make html
sphinx-build -b html -d _build/doctrees   . _build/html
Sphinx v4.0.2 in Verwendung

/!\ existing server config: "mss_wms_settings.py" for demodata not overwritten!
            

/!\ existing server auth config: "mss_wms_auth.py" for demodata not overwritten!
                
2021-06-28 17:06:01               setup || Files identified for domain 'EUR_LL015': ['20121017_12_ecmwf_forecast.ALTITUDE_LEVELS.EUR_LL015.036.al.nc', '20121017_12_ecmwf_forecast.CC.EUR_LL015.036.ml.nc', '20121017_12_ecmwf_forecast.CIWC.EUR_LL015.036.ml.nc', '20121017_12_ecmwf_forecast.CLWC.EUR_LL015.036.ml.nc', '20121017_12_ecmwf_forecast.DIV.EUR_LL015.036.ml.nc', '20121017_12_ecmwf_forecast.EMAC.EUR_LL015.036.ml.nc', '20121017_12_ecmwf_forecast.GPH.EUR_LL015.036.ml.nc', 

Configuration error:
There is a programmable error in your configuration file:

  File "/home/reimar/miniconda3/envs/mssdev/lib/python3.9/site-packages/sphinx/config.py", line 323, in eval_config_file
    exec(code, namespace)
  File "/home/reimar/PycharmProjects/2021/Marilyth/MSS/docs/conf.py", line 53, in <module>
    from gallery.plot_examples import HS_template, VS_template
ModuleNotFoundError: No module named 'gallery'

Makefile:58: recipe for target 'html' failed
make: *** [html] Error 2
(mssdev) reimar@infinte:~/PycharmProjects/2021/Marilyth/MSS/docs$ git branch
* i521-plots
```

Marilyth:
Oddly this issue never happened on the readthedocs runs.
I changed the line to docs.gallery, since the MSS directory is already included before this should be found as well @ReimarBauer

joernu76:
Looks good. Did you test the gallery on a server with hundreds of horizontal  and hundreds of vertical layers? (Thinking about timeouts etc.)

ReimarBauer:
there should be one comment line on the downloaded examples in first line.  
```
# Example of MSS (https://github.com/Open-MSS/MSS), License Apache 2.0 (https://opensource.org/licenses/Apache-2.0)
```

Anyone using this for something should know which license it has and what freedoms result from this

We should have this only on downloaded examples. The snippets are on a page where one could read this information.

ReimarBauer (commented):
> there should be one comment line on the downloaded examples in first line.
> 
> ```
> # Example of MSS (https://github.com/Open-MSS/MSS), License Apache 2.0 (https://opensource.org/licenses/Apache-2.0)
> ```
> 
> Anyone using this for something should know which license it has and what freedoms result from this
> 
> We should have this only on downloaded examples. The snippets are on a page where one could read this information.

Just seen that most have the complete header. My first one seems not to have it. Full header is much better. Looking which one it was

@Marilyth it is on the *_template.py missing

docs/gallery/plot_examples/HS_template.py
```diff
@@ -0,0 +1,33 @@
+import numpy as np
```
ReimarBauer, line 20: This should also have our header

mslib/index.py
```diff
@@ -103,15 +107,17 @@ def get_topmenu():
             (url_for('index'), 'Mission Support System',
              ((url_for('about'), 'About'),
               (url_for('install'), 'Install'),
+              (url_for("plots"), 'Gallery'),
```
ReimarBauer: we share index.py for mscolab too. This now has an empty page stating 

"Gallery was not generated for this server."

We should extend this with a link to mss.rtfd.io/gallery/index.html

An example can be seen on http://mss.rtfd.io/gallery/index.html  To build this overview see the examples on ... 

Or on mscolab server we don't show that menu entry, this would be cleaner.

mslib/mswms/gallery_builder.py
Marilyth: You must be looking at an old version. The newest version defaults to the mss_wms_settings directory, with an option to change it inside the mss_wms_settings itself.

mslib/mswms/mswms.py
```diff
@@ -46,6 +46,14 @@ def main():
     parser.add_argument("--logfile", help="If set to a name log output goes to that file", dest="logfile",
                         default=None)
     parser.add_argument("--update", help="Updates MSS to the newest version", action="store_true", default=False)
+
+    subparsers = parser.add_subparsers(help='Available actions', dest='action')
+    gallery = subparsers.add_parser("gallery", help="Generates plots of all layers not already present")
+    gallery.add_argument("--force-regenerate", action="store_true", default=False,
```
ReimarBauer: this naming let assume that there could be also a  --init or --seed to start with.

a shorter word for force-regenerate is refresh. And a common naming approach is 
- create
- clean
- refresh would do first a clean and a create afterwards

We do need clean too, to get to an origin state back.

ReimarBauer (changes requested):
This a very big step forward to bring transparency into how the server create those plots.

I added a few thoughts currently about little changes based on a functional review. I do  continue tommorrow.

ReimarBauer:
I do see on fresh demodata a problem

```
ERROR: Could not identify filename. mole_fraction_of_ozone_in_air ml 2012-10-17 12:00:00 2012-10-19 00:00:00 <class 'KeyError'> 'mole_fraction_of_ozone_in_air'
  File "/home/reimar/PycharmProjects/2021/Marilyth/MSS/mslib/mswms/dataaccess.py", line 216, in _determine_filename
    return self._filetree[vartype][init_time][variable][valid_time]
KeyError: 'mole_fraction_of_ozone_in_air'
```

mslib/mswms/mswms.py
ReimarBauer: Someone who wants to serve these static files by a nginx or apache2 wants to have an option to give a dir name where the /static/plots are stored

for explaination
https://code.google.com/archive/p/modwsgi/wikis/ConfigurationGuidelines.wiki#Hosting_Of_Static_Files

ReimarBauer:
The plot.js has more html or css than javascript code. Is there a need to name it as js?

mslib/mswms/gallery_builder.py
Marilyth: The locations are now uppercase

mslib/index.py
Marilyth: Done

docs/gallery/plot_examples/HS_template.py
Marilyth: Done

mslib/mswms/mswms.py
Marilyth: Parameters were renamed.
The gallery is now stored inside the _gallerypath defined in mss_wms_settings, which defaults to the mss_wms_settings location.

ReimarBauer (changes requested):
some features of mswms gallery are too expensive.

e.g. mswms gallery -h should only show
```2021-06-29 11:50:57,004 (wms.__init__:205): INFO: Since 4.0.0 MSS has support for linear layers in the mss_wms_settings.py.
Look at the documentation for an example https://mss.readthedocs.io/en/stable/deployment.html#configuration-file-of-the-wms-server
usage: mswms gallery [-h] [--create] [--clear] [--refresh] [--show-code]

optional arguments:
  -h, --help   show this help message and exit
  --create     Generates plots of all layers not already present
  --clear      Deletes all plots and corresponding code
  --refresh    Deletes all plots and regenerates them, a mix of --clear and --create
  --show-code  Generates plots of all layers not already present, and generates code snippets for each plot when clicking on the image
```

currently it seems to lookup all files it can proceeed

example

same with --clear, there I won't expect that it did a lookup what can be created

Marilyth:
I agree, but this is not an issue with the gallery but with mswms in general. It happens even when just requesting the version.
The server loads on wms import, which happens on the top.
The issue would be resolved by importing it only once it is needed, not before, or loading the server only once requested.

ReimarBauer:
We have to prefix by the SCRIPT_NAME and use it plots.html

mslib/mswms/gallery_builder.py
```diff
@@ -0,0 +1,450 @@
...
+        else:
+            with Image.open(io.BytesIO(plot)) as image:
+                image.save(os.path.join(location, "plots", l_type + "_" + plot_object.name + ".png"),
+                           format="PNG")
+
+    if generate_code:
+        write_plot_details(plot_object, l_type, sphinx)
+
+    img_path = f"../_images/{l_type}_{plot_object.name}.png" if sphinx \
+        else f"/static/plots/{l_type}_{plot_object.name}.png"
```
ReimarBauer: here wird should take for the SCRIPT_NAME

joernu76:
The order of importing shouldn't be changed lightly. We currently need to "patch" certain things in MSS from the settings.
We need a cleaner design for this, but changing this would force our hands to actually do that...

ReimarBauer (dismissed):
We played with that BIG improvement today on our staging system. I'm happy with the current state. Further work if needed can be done in smaller changes. 

Great also the addon to the rtfd document :)

mslib/mswms/wms.py
```diff
@@ -274,6 +274,7 @@ def generate_gallery(self, create=False, clear=False, generate_code=False, sphin
                                 init_time = plot_driver.get_init_times()[-1]
                                 valid_time = plot_driver.get_valid_times(plot_object.required_datafields[0][1],
                                                                          file_type, init_time)[-1]
+                                style = plot_object.styles[0][0] if plot_object.styles else None
```
joernu76, line 279: Isnt't this one [0] to much? styles is a list of strings...?

mslib/mswms/gallery_builder.py
```diff
@@ -0,0 +1,523 @@
...
+    from_text = f"{l_type}_{dataset}{plot_object.name}" if not native_import else native_import
+    instruction = f"from {from_text} import {plot_object.__class__.__name__}\n" \
+                  f"register_{layer}_layers = [] if not register_{layer}_layers else register_{layer}_layers\n"
+    # Generic LS class
+    if isinstance(plot_object, LS_DefaultStyle):
+        instruction += f"register_{layer}_layers.append(({plot_object.__class__.__name__}, " \
+                       f"\"{plot_object.required_datafields[-1][1]}\", \"{plot_object.required_datafields[-1][0]}\", " \
+                       f"[next(iter(data))]))"
+    # Generic HS or VS class without custom _prepare_datafields
+    elif (isinstance(plot_object, VS_GenericStyle) or isinstance(plot_object, HS_GenericStyle)) and "pass" in \
```
joernu76, line 300: Mmmh. Other classes can also be programmatically implemented. Not sure that we must handle this.

mslib/mswms/gallery_builder.py
```diff
@@ -0,0 +1,523 @@
...
+            inspect.getsource(plot_object._prepare_datafields):
+        style = "hsec" if isinstance(plot_object, HS_GenericStyle) else "vsec"
+        if plot_object.contours:
+            for i in range(len(plot_object.contours)):
```
joernu76, line 306: `
for i, temp in enumerate(plot_object.contours):
`
?

joernu76: Why is this even necessary? Should a repr() below in the string deal with this?
We could also modify the class to provide better instructions. It is our "go-to" class for generic plotting and should be improved to handle more use cases in the future.

mslib/mswms/mpl_hsec.py
```diff
@@ -63,6 +65,19 @@ def plot_hsection(self):
         """
         pass
 
+    def add_colorbar(self, contour, label=None, tick_levels=None, width="3%", height="30%", format=None):
```
joernu76: I think there were a couple of plots with deviating parameters.
Please add optional parameters for the ""magic numbers" and "magic strings". (i.e. fraction, pad, shrink, loc, extend, ticks_position)
E.g. to allow color bar placement on the left, etc.
Defaults should be the current values.

mslib/mswms/wms.py
Marilyth: styles is a list of 2-tuples containing strings. e.g.
```
styles = [
        ("auto", "auto colour scale"),
        ("autolog", "auto log colour scale")]
```

style, in this case, would be "auto"

mslib/mswms/mpl_vsec.py
```diff
@@ -59,6 +61,18 @@ def __init__(self, driver=None):
         """
         super(AbstractVerticalSectionStyle, self).__init__(driver=driver)
 
+    def add_colorbar(self, contour, label=None, tick_levels=None, width="3%", height="30%", format=None):
```
joernu76: Isn't this the same as for the horizontal hsecs?

mslib/mswms/mswms.py
```diff
@@ -46,6 +46,23 @@ def main():
     parser.add_argument("--logfile", help="If set to a name log output goes to that file", dest="logfile",
                         default=None)
     parser.add_argument("--update", help="Updates MSS to the newest version", action="store_true", default=False)
+
+    subparsers = parser.add_subparsers(help='Available actions', dest='action')
```
joernu76, line 50: I am not sure I like subparsers. At least not with many "commands". Examples are "git".

mslib/mswms/wms.py
```diff
@@ -211,6 +214,114 @@ def __init__(self):
...
+                        plot_object = registry[dataset][plot]
+                        l_type = "Linear" if driver == self.lsec_drivers else \
+                            "Side" if driver == self.vsec_drivers else "Top"
+
+                        try:
+                            if not os.path.exists(os.path.join(location, "plots",
+                                                               f"{l_type}_{dataset if multiple_datasets else ''}"
+                                                               f"{plot_object.name}.png")):
+                                # Plot doesn't already exist, generate it
+                                file_type = next((field[0] for field in plot_object.required_datafields
```
joernu76: I dislike the use of next here. Many people are not that familiar with iterators. I would spend the extra line for better maintainability.

docs/conf.py
```diff
@@ -12,7 +12,55 @@
...
+import mslib.mswms.gallery_builder
+
+# Generate template plots
+from docs.gallery.plot_examples import HS_template, VS_template
+dataset = [next(iter(mslib.mswms.wms.mss_wms_settings.data))]
+mslib.mswms.wms.mss_wms_settings.register_horizontal_layers = [(HS_template.HS_Template, dataset)]
+mslib.mswms.wms.mss_wms_settings.register_vertical_layers = [(VS_template.VS_Template, dataset)]
+mslib.mswms.wms.mss_wms_settings.register_linear_layers = []
+mslib.mswms.wms.server.__init__()
+mslib.mswms.wms.server.generate_gallery(sphinx=True, create=True, clear=True)
```
joernu76, line 59: For the documentation I would prefer to only have a few Generic Plot examples, e.g. only Ozone. They look all the same...

mslib/mswms/_tests/test_wms.py
```diff
@@ -367,11 +371,46 @@ def test_multiple_xml(self):
...
         with mock.patch.dict("sys.modules", {"mss_wms_settings": None, "mss_wms_auth": None}):
             reload(mslib.mswms.wms)
             assert mslib.mswms.wms.mss_wms_settings.__file__ is None
             assert mslib.mswms.wms.mss_wms_auth.__file__ is None
         reload(mslib.mswms.wms)
         assert mslib.mswms.wms.mss_wms_settings.__file__ is not None
         assert mslib.mswms.wms.mss_wms_auth.__file__ is not None
+
+    def test_gallery(self):
+        tempdir = tempfile.mkdtemp()
```
joernu76: Are this directories removed after the test? Perhaps use tmpdir fixture and two sub-directories. pytest then takes care of removing the tmpdir.

joernu76 (changes requested):

mslib/mswms/gallery_builder.py
Marilyth: Having any numpy array and representing it as a string will yield an unparseable syntax like "array([1, 2, 3])" or "[1 2 3]".
Converting it to a normal list beforehand makes sure the class can be used as is.

joernu76: Mmh. I see the problem. But the intent of the code is not obvious. In this case a comment w.r.t. to the intent (replacement of numpy arrays) would be helpful. Maybe move the conversion towards the make_generic functions (with a helper). The gallery code shows the need for better self-documentation/introspection of that classes.

mslib/mswms/gallery_builder.py
Marilyth: I am not sure what you are suggesting here. This branch is responsible for all make_generic_class instances without add_prepare

mslib/mswms/mpl_vsec.py
Marilyth: The code between adding colourbars between hsecs and vsecs was slightly different. I added the most common one to their respective super class which reduces the need for providing parameters.

mslib/mswms/gallery_builder.py
joernu76: Me neither. Let's discuss this after the meeting.
There are a couple of things.
a) I am nervous about the prepare_datafield
b) I am nervous that programmatically created classes require special treatment

joernu76 (dismissed):

mslib/mswms/wms.py
Marilyth: I will abide.
I always found the use of next quite nice as an equivalent to what other languages offer out of the box for enumerables, particularly C#'s Enumerable.First(predicate), which is more performant for less code.
I don't often think in Python terms, so oddities like that arise occasionally.

docs/conf.py
Marilyth: There are 0 generic plot examples in the docs gallery. They just happen to look alike anyway.

mslib/mswms/gallery_builder.py
Marilyth: This is a question for another time, once the problems arise, I think.

mslib/mswms/mpl_hsec.py
Marilyth: Done

mslib/mswms/_tests/test_wms.py
Marilyth: Done

mslib/mswms/mswms.py
Marilyth: I disagree. I find subparsers for commands which inherently belong together reasonable.

mslib/mswms/wms.py
joernu76: I concur that a "first" function would be very readable (or "car" for me, as I had a lisp/scheme phase). 
"next" is confusing IMO.
Even something like `([x for x in ...] + ["sfc"])[0]` would be better.

Writing code for other people/your older self is something one takes more into account when getting older ;-)

mslib/mswms/mswms.py
joernu76: Fine for me.

joernu76: And my comment was misleading, rereading it. For `git` I find it totally fine and the only way to handle it, but for a *single* action, it might be a bit confusing.

ReimarBauer (approved):
extensliy tested :)

PR #1043, opened 2021-06-18
Fix behaviour for projects with dash "-" in name.

### LINKED ISSUE #1036: "-" in mscolab project name causes project name to be not properly displayed on plots and prompts

joernu76:
Fix #1036


ReimarBauer (approved):
Thx, works like a charm ;)

PR #1054, opened 2021-06-28
Minor bug fixes.

joernu76:
* Return proper unit for eq. potential temperature (degC->K)
* Do not assume units of pressure levels for several old plotting
  styles.
* Fix values of geopotential altitude in demodata


joernu76:
@may I found out why the equivalent potential temperature plot looked blank. This may conflict with your changes and you may also include this into your MR and close this, if it is simpler.

Marilyth (approved):
I can't exactly confirm that this is correct-er than the previous values, but I will trust you.
The plots look almost the same on my end.
If this is a fix of something wrong, should this maybe go to stable?

PR #1055, opened 2021-06-28
Simplification of flightlevel 2 pressure and back computations

joernu76:
More use of metpy, better standard conformance, less copy paste.

### DIFF
--- a/mslib/_tests/test_thermolib.py
+++ b/mslib/_tests/test_thermolib.py
@@ -31,13 +31,27 @@ import pytest
 import mslib.thermolib as tl
 
 
+def test_flightlevel2pressure2flightlevel():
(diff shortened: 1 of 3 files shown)


joernu76:
Hi, I simplified the code at some cost of efficiency (the array method is much more costly than the single value method).
The code now uses the US standard atmosphere values. With the metpy Rd value, the discontinuities at 32 km and 51 km are now small enough such I do not think we need to do a correction. Everything relates now back to the constant table of P/T/gamma values.
I think that can serve as a good basis for metpy and if not, our code got a lot more consistent and maintainable.

If you can conjure up a more efficient way of doing the array computation I am very happy to see it.

Marilyth (approved):
Looks very similar to what I have done myself so far.
Merging it this way will decrease my PR size once it comes.

PR #1062, opened 2021-06-30
Added missing filename key.

joernu76:
See #1066


Marilyth (approved):
Seems to run fine on the server now.

PR #1063, opened 2021-06-30
Update development.txt

### LINKED ISSUE #1056: add conda-verify to the development.txt
conda-verify can be used to test the localbuild meta.yaml

We should have it in our developer stack

himanshu007-creator:
resolves #1056

### DIFF
--- a/localbuild/meta.yaml
+++ b/localbuild/meta.yaml
@@ -81,7 +81,6 @@ requirements:
     - sqlite <3.35.1
     - gpxpy >=1.4.2
     - metpy
-    - conda-verify
 
 test:
   imports:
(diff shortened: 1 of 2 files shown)


ReimarBauer (changes requested):
Hi @himanshu007-creator we only need this for development. A common released installation does not have benefit from this. It would only increase the size for users.

For all development related packages, e.g. pytest etc. we have a different file.

we want this added here
https://github.com/Open-MSS/MSS/blob/develop/requirements.d/development.txt

ReimarBauer:
This is the released section about development, this section will be improved for the next release
https://mss.readthedocs.io/en/stable/development.html#setup-a-development-environment

ReimarBauer (approved):
thx

PR #1070, opened 2021-07-01
Rerun failed tests up to 5 times before complete fail

Marilyth:
Band aid fix for all the random failing tests on GitHub.
If tests fail, give them 5 more chances to pass before declaring the action failed.
Does not get rid of the occasional xdist hang up.

### DIFF
--- a/.github/workflows/xdist_testing.yml
+++ b/.github/workflows/xdist_testing.yml
@@ -41,6 +41,6 @@ jobs:
         && cd $GITHUB_WORKSPACE \
         && pytest -n 6 --dist loadscope --max-worker-restart 0 mslib \
         || (for i in {1..5} \
-          ; do pytest mslib --last-failed --lfnf=none \
+          ; do pytest -n 6 --dist loadscope --max-worker-restart 0 mslib --last-failed --lfnf=none \
           && break \
         ; done)
--- a/.github/workflows/xdist_testing.yml
+++ b/.github/workflows/xdist_testing.yml
@@ -39,4 +39,8 @@ jobs:
       run: |
         source /opt/conda/bin/activate mssenv \
         && cd $GITHUB_WORKSPACE \
-        && pytest -n 6 --dist loadscope --max-worker-restart 0 mslib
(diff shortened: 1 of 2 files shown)


joernu76:
I run sometimes into timeouts already now (at least with my modified time check. NFS??). How does this relate to our maximal runtime?

joernu76:
What about
https://docs.pytest.org/en/6.2.x/flaky.html

via pytest.mark.xfail?

Especially this deeper link

https://github.com/pytest-dev/pytest-rerunfailures
Seems to do something similar, but allows more fine grained control via

`@pytest.mark.flaky(reruns=5)`

Marilyth:
Some testing with pytest-rerunfailures caused me a lot of headache and segmentation faults after x reruns.
Considering pytest comes with failure reruns already, I don't see the need for it.

The runtime increases with each failed test. Running them again 5 times will take maybe 10 seconds of overhead + the time needed for the tests * 5. If it passes after 1 retry then the test finishes right there.
It isn't much of an increase.

.github/workflows/xdist_testing.yml
```diff
@@ -39,4 +39,8 @@ jobs:
       run: |
         source /opt/conda/bin/activate mssenv \
         && cd $GITHUB_WORKSPACE \
-        && pytest -n 6 --dist loadscope --max-worker-restart 0 mslib
+        && pytest -n 6 --dist loadscope --max-worker-restart 0 mslib \
+        || (for i in {1..5} \
+          ; do pytest mslib --last-failed --lfnf=none \
```
ReimarBauer: this means the fallback is to run serial tests? because there is not -n option given.

i think we should keep the parallel option as the other is already taken by the other test.

parallel for sure will fail if we have a test dependency included which comes from another test and is in a process missing. A failing by this would be correct because we don't want that.

ReimarBauer (changes requested):
see comment

.github/workflows/xdist_testing.yml
Marilyth: xdist now runs the reruns in parallel also.

ReimarBauer (approved):
looks good :)

PR #1071, opened 2021-07-01
Dynamically insert mss version and search result

### LINKED ISSUE #1052: organize / reorganize Topics of the Development part
By  the PR fixing 
https://github.com/Open-MSS/MSS/issues/719
we now have a lot more Level 1 Topics in our documentation.

The documentation is both for new users and also developers. 

The list of main topics (first level) should be short.
All development related documentations have to be in the section Development. May be they can be grouped or sorted into that section.

The current state is:

menu_development

That needs to become improved

### LINKED ISSUE #1057: Update docs mss version etc automatically
Currently when there is a new version being released, the docs need to reflect that version, which currently warrants manually modifying each string containing the version and updating it.
[PR 1042](https://github.com/Open-MSS/MSS/pull/1042) has shown that dynamic docs generation is possible, and this should also be used to capture and present the newest version automatically whenever the docs are created.

Marilyth:
Replace the version and mss search in the .rst with template variables, and replace them on doc generation.
fixes #1057 and fixes #1052
Generated example here https://mss-test.readthedocs.io/en/latest/installation.html (4.0.1 since develop is still showing that version)


docs/conf.py
```diff
@@ -117,6 +119,20 @@
 # The full version, including alpha/beta/rc tags.
 release = __version__
 
+# Replace $variables in the .rst files if on a readthedocs worker
+if "/home/docs/checkouts" in " ".join(sys.argv):
+    mss_search = subprocess.run(["conda", "search", "-c", "conda-forge", "mss"], stdout=subprocess.PIPE,
+                                stderr=subprocess.STDOUT, encoding="utf8").stdout
+    mss_search = "   ".join([line for line in mss_search.splitlines(True) if line.startswith("mss ")][-2:])
+
+    for file in os.listdir():
```
ReimarBauer, line 129: just a comment someone else falls over this
file was in Py 2 a builtin and this was removed in Py 3 by the io framework

docs/conf.py
```diff
@@ -117,6 +119,20 @@
 # The full version, including alpha/beta/rc tags.
 release = __version__
 
+# Replace $variables in the .rst files if on a readthedocs worker
+if "/home/docs/checkouts" in " ".join(sys.argv):
+    mss_search = subprocess.run(["conda", "search", "-c", "conda-forge", "mss"], stdout=subprocess.PIPE,
```
ReimarBauer, line 124: mamba search is here much faster

docs/conf.py
```diff
@@ -117,6 +119,20 @@
...
+                                stderr=subprocess.STDOUT, encoding="utf8").stdout
+    mss_search = "   ".join([line for line in mss_search.splitlines(True) if line.startswith("mss ")][-2:])
```
ReimarBauer: comment: another option to create blanks without other  hidden chars.

`" " * 4`
chr(32) * 4

docs/development.rst
```diff
@@ -22,7 +22,7 @@ Setting Up local Environement
 ============================
 Forking the Repo
 ============================
```
ReimarBauer: because you are in the file, please replace those chapter headings by a normal heading, so we remove the exposure as main level in the index.

ReimarBauer (changes requested):
see comments

docs/conf.py
Marilyth: Is it also faster to install mamba only for the sake of searching for mss afterwards?
I am not sure if readthedocs comes with mamba preinstalled, but I doubt it.

ReimarBauer: not thought on that. keep it.

docs/development.rst
Marilyth: I also took the liberty of rearranging them.

docs/conf.py
Marilyth: Done

ReimarBauer (approved):
thx

PR #1074, opened 2021-07-05
Waypoints Tutorial : Working on Linux system now.

risehr:
I have made the code more efficient in terms of automation.
Now, it has been made to work and tried both on windows and linux systems (also will work on mac)

Made the clicking points relative to the buttons which were absolute points earlier. This is helpful because different systems has the opened topview window at different positions on the screen.

On Linux system type on the terminal:
`sudo apt-get install scrot`
This will make sure the locateOnScreen functions work correctly on linux.


ReimarBauer (approved):
thx

PR #1087, opened 2021-07-20
Less intrusive layer error handling

### LINKED ISSUE #1085: multilayer shows often cannot determine init time format
A user reported often seeing

init_time

Marilyth:
Instead of popping up a message box for every failed layer, log it and disable it.


ReimarBauer (approved):
Thx :)

PR #1089, opened 2021-07-20
Include airports and airspaces in topview

### LINKED ISSUE #1069: integrate recent Airbase data, e.g. from openaip
https://www.openaip.net/

Marilyth:
Allows the visualisation of airports and airspaces through the topview.
Data will automatically be downloaded.
fixes #1069


mslib/msui/mpl_map.py
```diff
@@ -325,6 +336,154 @@ def set_graticule_visible(self, visible=True):
...
+            map_polygon = Polygon([(self.llcrnrx, self.llcrnry), (self.urcrnrx, self.llcrnry),
+                                  (self.urcrnrx, self.urcrnry), (self.llcrnrx, self.urcrnry)])
+            airspaces = [airspace for airspace in airspaces if
+                         (not range_km or range_km[0] <= airspace["bottom"] <= range_km[1]) and
+                         Polygon(airspace["polygon"]).intersects(map_polygon)]
+            if not airspaces:
+                return
+
+            if "openaip.net" not in self.crs_text.get_text():
+                self.crs_text.set_text("Airspaces provided by openaip.net\n" + self.crs_text.get_text())
```
ReimarBauer: theire suggestion is:

Airspaces Data used comes from openAIP. Visit openAIP.net and contribute to better aviation data, free for everyone to use and share.

ReimarBauer (changes requested):
not finished with review.

mslib/msui/mpl_map.py
Marilyth: Updated

ReimarBauer (commented):
in the mss ui About we should add a section for displaying NOTICE (rendered) or by some new menu entry.

in the index.py we could add a menu which shows NOTICE (rendered) too

joernu76:
Nothing serious, but I got this RuntimeError message when pressing the "this directory" button:
```
INFO: Launching user interface...
Initializing caja-open-terminal extension
Initializing caja-image-converter extension
RuntimeError: object at 0x7fede458b540 of type RenameMenu is not initialized
```
The correct directory *does* open, though.

joernu76:
I also guess, the airports should be plotted *above* the airspaces. Currently, they are hidden.

joernu76:
Last, I suggest to move the functionality to a dockwidget. I feel that we need to add some more filters/functions to this feature, which would be a bit much for this option dialogue...?

E.g. airspaces might also be filtered according to category. They might also be drawn unfilled to play better with WMS maps. etc.

joernu76:
Pretty nifty, overall!

Marilyth:
@joernu76 
Airspaces now don't need a manual download anymore, so the RuntimeError should be gone?
Airports are now rendered above airspaces.
Functionality has been moved to a dockwidget.

ReimarBauer (commented):
I would prefer a progress bar for the Update/Download feature. I do see currently info entries on the terminal.

ReimarBauer:
the selection dropdown may get a select all / unselect all feature.

Or probably a selection by continents? e.g. europe or the map selection.

ReimarBauer (approved):
Thx, that is a big step further.

Further improvements anyway will be done after this gets used.

PR #1091, opened 2021-07-22
Return default settings on pytest

### LINKED ISSUE #1090: Test should not use the mss-core config
I ran into issues today with a failing test, which was caused by my topview map appearance settings.
I would expect tests to not save or load settings persistently, as that is a hard to trace reason of error.
The error was gone after I forced tests to run on the default settings and being unable to save anything, and no other test was failing.
Is there any reason to keep the saving/loading for tests, as every test needs to be independent anyway?

Marilyth:
If pytest was imported, saving and loading qsettings only returns the default.
Fixes #1090


ReimarBauer (approved):
Thx

PR #1092, opened 2021-07-23
Gsoc config editor

aravindm711:


NOTICE
```diff
@@ -108,7 +108,13 @@ Further Information: http://tango.freedesktop.org
 
 
 
+Qt-json-view
+------------
 
+Author: Paul Schweizer (https://github.com/PaulSchweizer)
+
+Package for working with JSON files in PyQt5.
+Obtained from Github (https://github.com/PaulSchweizer/qt-json-view), on 23/7/2021.
```
ReimarBauer, line 118: add the License of this code too

mslib/msui/editor.py
```diff
@@ -29,11 +29,234 @@
...
+
+
+class ConfigurationEditorWindow(QtWidgets.QMainWindow, ui_conf.Ui_ConfigurationEditorWindow):
+    def __init__(self, parent=None):
+        super(ConfigurationEditorWindow, self).__init__(parent)
+        self.setupUi(self)
+
+        self.frame.setStyleSheet("QFrame {border: 0;}")
+
+        settings_options = [
```
ReimarBauer, line 50: we should use the "options" by the config class. We cannot have different names for those. Only names in the config class can be used. This may need a refactoring of them to have a grouping/categorizing. Also there is a need for a decription explaination which is then shown in the widget.

ReimarBauer (changes requested):
see comments

mslib/msui/editor.py
aravindm711: I have just added this code for checking if the importing of the library works. I'm going to completely rewrite this from scratch after finalizing the design mockup, which I haven't done yet.

NOTICE
aravindm711: I have added a README.txt file in the support dir (mslib/support/README.txt), which mentions the changes I've made to the code and also mentions the license. Can you please confirm if that is enough or should I update it here too?

ReimarBauer:
We need a tab size for Key which is len longest key + 5. 
All keys need a description. That can be shown on mouse over or another column. Perhaps if we have defined values to be used, we may also show these values.

The pulldown selector may be have an alphabetical sorted list.

The word Key may be renamed to Option in the table.

keyval

On saving only the differences of default should go into the json file

NOTICE
ReimarBauer: also here add the License type, because we cannot convert the MIT to the apache license.

mslib/msui/editor.py
ReimarBauer: because of my network situation, I just added ideas ;)

NOTICE
ReimarBauer: The README is good :)

aravindm711:
Closing this PR since this branch shares commits with PR that was merged parallely. Opened new PR for a clean update on config editor https://github.com/Open-MSS/MSS/pull/1103.

Issue #1093, opened 2021-07-23
Tests aborting silently


Marilyth:
I observed on multiple occasions the tests just aborting silently, no error output or anything, which looks something like that 
```                                                                                                                                                                                                              
mslib/_tests/test_convert.py .                                                                                                                                                                                                   [  0%]
mslib/_tests/test_index.py ..                                                                                                                                                                                                    [  0%]
mslib/_tests/test_netCDF4tools.py ........s..                                                                                                                                                                                    [  3%]
mslib/_tests/test_thermolib.py .....s                                                                                                                                                                                            [  4%]
mslib/_tests/test_utils.py ..........................                                                                                                                                                                            [ 10%]
mslib/mscolab/_tests/test_chat.py .(mssenv) mayb@icg1705:~/Desktop/MSS/MSS$
```

It seems like @aravindm711 managed to make this a consistent error on his repo which can be seen [here](https://github.com/aravindm711/MSS/runs/3142033187?check_suite_focus=true)

After some digging and debugging it seems to happen due to X11 connections breaking. Particularly with the message
`The X11 connection broke: No error (code 0)`
A log of this debug can be seen [here](https://github.com/Marilyth/MSS/runs/3141549177?check_suite_focus=true)

I haven't gotten further than that yet.

Things I tried:
- Catching the error, which is impossible
- Updating X11, which had no effect
- Throwing out test_mscolab.py which made the test pass, but obviously isn't the root of the issue

Marilyth:
Investigating the random test aborts on github, I discovered a complete run of pytest on the github actions left open 200 unresponsive windows at the end of the test, surely playing into the issue. #1080 might be the cause for this issue as well
```
Root window id: 0x388 (the root window) (has no name)
  Parent window id: 0x0 (none)
     200 children:
     0x2005a1 "WMS Dock Widget": ("__main__.py" "__main__.py")  756x142+0+0  +0+0
     0x20059b "WMS Dock Widget": ("__main__.py" "__main__.py")  756x142+0+0  +0+0
     0x200595 "WMS Dock Widget": ("__main__.py" "__main__.py")  756x142+0+0  +0+0
     0x20058f "WMS Dock Widget": ("__main__.py" "__main__.py")  756x142+0+0  +0+0
     0x200589 "WMS Dock Widget": ("__main__.py" "__main__.py")  756x142+0+0  +0+0
[...]

PR #1097, opened 2021-07-29
Close windows and timeout after 60 seconds

### LINKED ISSUE #1080: pytest windows stays open
currently using pytest almost all windows stay open until the end

We should close them on teardown.

Marilyth:
Fixes #1080 and hopefully a fix for #1093
After each test loop through all open windows and widgets and destroy them.
Fail a test that takes more than a minute to hopefully circumvent freezing tests.
Mock all messageboxes to avoid unexpected modal error popups never returning to pytest.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -29,6 +29,7 @@ import importlib
 import importlib.machinery
 import os
 import sys
+import mock
 from PyQt5 import QtWidgets
 # Disable pyc files
 sys.dont_write_bytecode = True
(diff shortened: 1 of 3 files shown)


conftest.py
```diff
@@ -152,6 +154,25 @@ class mscolab_settings(object):
...
+            mock.patch("PyQt5.QtWidgets.QMessageBox.critical"), mock.patch("PyQt5.QtWidgets.QMessageBox.warning"):
+        yield
+
+    # Try to close all remaining widgets after each test
+    for qobject in set(QtWidgets.QApplication.topLevelWindows() + QtWidgets.QApplication.topLevelWidgets()):
+        try:
+            qobject.destroy()
+        # Some objects deny permission, pass in that case
+        except RuntimeError:
+            pass
```
joernu76, line 181: Should we act on an *unexpected* message window?

Marilyth: I am not sure what you mean here.
Should we do more than pass on RuntimeError?
Windows/Widgets which throw a RuntimeError on destroyed seem to be not persistent or too important to destroy. So ignoring them will usually solve the issue and leave no unwanted window open anyway.

Should we act on unexpected message windows? Absolutely. Message windows block pytest, any test which unexpectedly cause a message window to pop up without handling it is an infinite freeze. This is solved by the mocks before yield.

joernu76: question: what happens when now an unexpected message window appears?
The test should fail, shouldn't it?

joernu76: Previously it froze, so now it's obviously much better already.

joernu76 (approved):

Issue #1098, opened 2021-07-30
make satellite predictor link clickable.


joernu76:

joernu76:
Fixed by #1100

PR #1100, opened 2021-07-30
Made NASA LARC URL clickable.

joernu76:
Fix issue #1098


Marilyth (approved):

joernu76:
@Marilyth What happened here with the update/images?

Marilyth:
@joernu76 Workflows disable themselves after 60 days of repo inactivity.
Open-MSS/dockerhub unfortunately doesn't get a lot of pushes so that can happen. I enabled it again.

PR #1103, opened 2021-07-30
GSOC Config editor update

aravindm711:
- [x]  The non-default options should be visually marked, e.g., by bolding the value. 
- [x] After removing my json config file, I got an error on opening the editor:
```
  File "mslib/msui/mss_pyui.py", line 743, in open_config_editor
    self.config_editor = editor.ConfigurationEditorWindow(parent=self)
  File "/home/icg173/src/mss/mslib/msui/editor.py", line 152, in __init__
    if json_file_data:
UnboundLocalError: local variable 'json_file_data' referenced before assignment
```
- [x] For the "grouping" of configuration options, we need to come up with "categories" for the drop-down box. Not necessarily your task, though.
- [x] Because changing the configuration requires a restart I would suggest to make the Editor a blocking dialogue.
- [x] Pushing the "+ Add" "Restore Defaults" and "Import" buttons does not seem to do anything???
- [x] Add more logging.debug messages to roughly allow tracing the code execution from the log on the terminal (after calling with --debug)
- [x] The whole "buttonbox" does not seem to be working? Where is the code handling the buttons? I would suggest implementing the buttons "manually" With QPushButtons. (edited) 
- [x] It would be nice, not essentiaol  if you would follow the naming style of the old UI. See e.g.  mslib/msui/ui/ui_topview_window.ui I.e. ComboBoxes are named "cb"Whatever and pushbuttons "pb"SomethinElse.
- [x] The documentaion should be displayed with some stand-in text (like document me)
- [x] I propose to decouple the Adding of Options from the Filter. If one has selected a list-like option (i.e. it is highlighted with a bar), the add option should become clickable. I.e. Add/Remove should be enabled/disabled depending on if they can actually act. Rename the filter to be obviously "only" a filter.
- [x] "Project->Project Properties->Edit Project" does not seem to do anything.


joernu76:
Moved task list to top.

aravindm711:
I have covered almost all of the points you've mentioned, except implementing import, cancel, save and save as. Could you please take a look and verify? @joernu76

joernu76:
I will. Can you tick the boxes of the adressed points in my comment?

joernu76:
@aravindm711 

From a first glance it works better much than before.

I am undecided how to bold/not bold the lists. E.g. the list of "locations" has default values, which are very different from my current own list. Maybe here, also all entries not from the default list should be marked in bold? Similar to other lists/dictionaries...?

The restore to defaults only restores the selected option to defaults. Which is very useful, but I did not expect it due to its placement. Maybe move this to the add/remove buttons to make a trio add/remove/restore. The enabling/disabling of all three buttons should work similar, anyway.

I would rename "Save As" to Export, as this might imply otherwise MSS would use the newly "save ased" file.

I am also unsure if the split of buttons to top and button makes sense. You should logically group the buttons according to functionality (individual option related / file handling related), obviously, but doing that all, e.g., below the editor might be cleaner? Just a thought.

aravindm711:
Even I was confused about bolding list/dict items, maybe I'll bold the entire list/dict (values only) if it exactly matches the default ones. What do you think about that?

add/remove/restore sounds like a good idea. I'll change it.

Since we've agreed on moving the restore button to the top, I think save/cancel should be at the right bottom because its the most logical place for it to be. Although, I'm not sure where to place import/export. I'll think of something and let you know.

Is it possible for me to tick points in your comments, UI-wise its not clickable? Or should I copy/paste it and tick them in markdown.

joernu76:
I moved my list to the top of the PR. Only there, ticking of boxes seems possible.

Bolding the whole list is certainly better than not bolding anything.

ReimarBauer (changes requested):
currently I have not a mss_settings.json
```
  File "/home/reimar/PycharmProjects/2021/aravindm711/MSS/mslib/msui/mss_pyui.py", line 743, in open_config_editor
    self.config_editor = editor.ConfigurationEditorWindow(parent=self)
  File "/home/reimar/PycharmProjects/2021/aravindm711/MSS/mslib/msui/editor.py", line 135, in __init__
    if json_file_data:
UnboundLocalError: local variable 'json_file_data' referenced before assignment
```

ReimarBauer (commented):
If the old editor is not needed do we need pyperclip? May be that can be removed from the meta.yaml?

ReimarBauer:
> currently I have not a mss_settings.json
> 
> ```
> Traceback (most recent call last):
>   File "/home/reimar/PycharmProjects/2021/aravindm711/MSS/mslib/msui/mss_pyui.py", line 743, in open_config_editor
>     self.config_editor = editor.ConfigurationEditorWindow(parent=self)
>   File "/home/reimar/PycharmProjects/2021/aravindm711/MSS/mslib/msui/editor.py", line 135, in __init__
>     if json_file_data:
> UnboundLocalError: local variable 'json_file_data' referenced before assignment
> ```

outdated clone

ReimarBauer (approved):
Thx

PR #1096, opened 2021-07-28
Automated tutorial for how to create a project in MSS using all the views(Completed)

risehr:
Automated tutorial on how to create a project using all the view. For now this is just for Windows. (May or may not correctly in linux)... I will be adding a commit to this PR soon for running smoothly on Linux systems.

PR #1104, opened 2021-07-31
[Completed] Automated tutorial for how to create a project in MSS using all the views.

risehr:
This involves the demonstration of how to create a project with all the views, that are, top view, side view, table view and linear view.
I have tested in both windows and linux and it works smoothly.

(The previous PR opened had the random tests failing. Hence, created this new PR and closed the previous #1096 PR)


Marilyth:
Your tests will probably pass when next merged with develop

risehr:
> Your tests will probably pass when next merged with develop

Okay. Thanks.

tutorials/tutorial_views.py
```diff
@@ -0,0 +1,944 @@
...
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+"""
+
+import pyautogui as pag
+import multiprocessing
+import sys
+from sys import platform
```
ReimarBauer, line 30: When you import "sys" by thee line before, it would make sense to me to use sys.platform

tutorials/tutorial_views.py
```diff
@@ -0,0 +1,944 @@
...
+    Calls the main MSS GUI window since operations are to be performed on it only.
+    """
+    mss_pyui.main()
+
+
+def automate_views():
+    """
+    This is the main automating script of the MSS views tutorial which will cover all the views(topview, sideview,
+    tableview, linear view) in demonstrating how to create a project. This will be recorded and savedto a file having
+     dateframe nomenclature with a .mp4 extension(codec).
```
ReimarBauer: left blank too much?

tutorials/tutorial_views.py
```diff
@@ -0,0 +1,944 @@
...
+        elif platform == 'win32':
+            pag.hotkey('win', 'up')
+    except Exception:
+        print("\nException : Enable Shortcuts for your system or try again!")
+    pag.sleep(2)
+    pag.hotkey('ctrl', 'h')
+    pag.sleep(2)
+
+    # Shfting topview window to upper right corner
+    try:
```
ReimarBauer, line 103: comment: usually we do capsulate in try exception blocks only the part of the lines which could make trouble

tutorials/tutorial_views.py
```diff
@@ -0,0 +1,944 @@
...
+        pag.move(-50, 150, duration=1)
+        pag.click(interval=2)
+        pag.sleep(1)
+        pag.move(65, 65, duration=1)
+        pag.click(interval=2)
+        pag.sleep(1)
+
+        pag.move(-150, 30, duration=1)
+        x1, y1 = pag.position()
+        # print(x1, y1)
```
ReimarBauer: remove not needed comments

tutorials/tutorial_views.py
```diff
@@ -0,0 +1,944 @@
...
+        pag.click(interval=1)
+        if platform == 'win32' or platform == 'linux' or platform == 'linux2':
+            pag.hotkey('ctrl', 'v')
+        elif platform == 'darwin':
+            pag.hotkey('command', 'v')
+        pag.sleep(4)
+        # Shifting Sideview window to upper left corner.
+        try:
+            x1, y1 = pag.locateCenterOnScreen(f'{dir_path}add_waypoint.png')
+            if platform == 'win32' or platform == 'darwin':
```
ReimarBauer, line 122: the pag code lines can be moved to a utils functions and has the platform as input value

tutorials/tutorial_views.py
```diff
@@ -0,0 +1,944 @@
...
+    """
+    This is the main automating script of the MSS views tutorial which will cover all the views(topview, sideview,
+    tableview, linear view) in demonstrating how to create a project. This will be recorded and savedto a file having
+     dateframe nomenclature with a .mp4 extension(codec).
+    """
+    # Giving time for loading of the MSS GUI.
+    pag.sleep(5)
+
+    if platform == 'linux' or platform == 'linux2' or platform == 'darwin':
+        dir_path = 'pictures/tutorial_views/win/'
```
ReimarBauer, line 79: there could be in utils a pictures functions, similiar to the icons functions which sets the path and also returns the needed image

ReimarBauer (changes requested):
I would like to see a bit modularisation in some utils functions for the different platform parts.

This could also be done in a refactoring PR. What is your opinion about this?

tutorials/tutorial_views.py
risehr: The `import sys` has a different use such as sys.exit() but the `from sys import platform` had been done because platform could be independently used as the name suggests for the platforms such as windows or linux or mac.

Should I remove `from sys import platform` and use only `sys.platform`??

tutorials/tutorial_views.py
risehr: Again by mistake, Will remove that.

tutorials/tutorial_views.py
risehr: I have used try except in every part where it is locating an image on the screen to click upon. If the try block would be removed and if it is not able to find the image then the user will not know where the error in program is or it would simply stop the execution.
Wherever there is `pag.locateCentreOnScreen()` function I have kept it inside the try block and all the following code that is dependent on that image found by the same function.

tutorials/tutorial_views.py
risehr: Again by mistake. I will remove those. Sorry!

tutorials/tutorial_views.py
risehr: Okay. I will try to find a way in which it can be done. Since, there are many small things which are platform dependent and even if when added to utils function, somethings still need to be there in the source code regarding platform.

tutorials/tutorial_views.py
risehr: I will try to do that.

risehr:
> I would like to see a bit modularisation in some utils functions for the different platform parts.
> 
> This could also be done in a refactoring PR. What is your opinion about this?

Yeah, it would be helpful if I do all those utils' part in another refactoring PR. Moreover, I am thinking to complete the rest tutorials first and then do this refactoring around utils part? Because, I guess I should complete the tutorials first. What say?

tutorials/tutorial_views.py
ReimarBauer: I think I would unify the usage. you could do 

from sys import platform, exit

or use sys.platform 

I would use sys.platform because it is already imported and I am used to read sys.exit()

tutorials/tutorial_views.py
ReimarBauer: you can do this later, when you have a module ready. I guess it is better readable afterwards.

ReimarBauer:
> > I would like to see a bit modularisation in some utils functions for the different platform parts.
> > This could also be done in a refactoring PR. What is your opinion about this?
> 
> Yeah, it would be helpful if I do all those utils' part in another refactoring PR. Moreover, I am thinking to complete the rest tutorials first and then do this refactoring around utils part? Because, I guess I should complete the tutorials first. What say?

Agreed

risehr:
> > > I would like to see a bit modularisation in some utils functions for the different platform parts.
> > > This could also be done in a refactoring PR. What is your opinion about this?
> > 
> > 
> > Yeah, it would be helpful if I do all those utils' part in another refactoring PR. Moreover, I am thinking to complete the rest tutorials first and then do this refactoring around utils part? Because, I guess I should complete the tutorials first. What say?
> 
> Agreed

Okay then. I am starting my work on mscolab tutorial as of now.

ReimarBauer (approved):
Ok Thx

PR #1106, opened 2021-08-01
Use mamba for all commands if possible

Marilyth:
Previously some updater commands used conda instead of mamba, slowing down the process.
This fixes it.


ReimarBauer (approved):
Thx

PR #1121, opened 2021-08-03
responsive images and respecting SCRIPT_NAME, fixes #1120

### LINKED ISSUE #1120: problems with campaign logos on the servers Frontpage
- size of logos on the wms or msco server Frontpage are not responsive resizing to browser size
- SCRIPT_NAME not used

ReimarBauer:


joernu76 (approved):
What do the classes do? Is the width compatible with different devices? Where do the images actually appear? I was thinking they would be part of the MSS rtfd.io documentation, e.g. on the About part?

ReimarBauer:
> What do the classes do? Is the width compatible with different devices? Where do the images actually appear? I was thinking they would be part of the MSS rtfd.io documentation, e.g. on the About part?

currently they are on the front page of a mswms/mscolab server e.g. http://open-mss.org/  

It is a start :)

RTFD needs more details on campaigns, and a dedicated campaign server will need only one logo, other informations.

the width attribute of the images together with the added class makes it responsive on different screen sizes.

removing the leading slash uses the base url which gives automatically the right URL e.g. for a /demo/ base /demo/mss_theme/img/emerge_2017.jpeg

PR #1136, opened 2021-08-05
Remote Sensing and Satellite Track Tutorial

risehr:
Updated the Screen Recorder class with options to record a particular size of screen. Made the recording frame screen changeble/flexible.
Created two tutorials: one of remote sensing in topview and another is satellite track prediction in top view.


ReimarBauer:
both work very well. 

When I use a terminal on linux to start the tutorials they move the terminal in foreground on finishing not the screen recorder.
The terminal shows `qq`  which should close the screen recorder.

Besides that everything fine. Look on this please after merging. I guess we have this on all existing tutorials.

ReimarBauer (approved):
Thx

PR #1148, opened 2021-08-13
hides subprocess windows on os.system = "nt"

### LINKED ISSUE #1145: Starting MSS under Windows opens confusing mamba and/or conda windows
Now: Starting MSS opens several command windows for mama (and or conda) upon starting on Windows.

These Windows hide the main window and stay open rather long.

The main window is also hidden behind other application

This is very confusing.

Desired state: No windows open for those processes.

ReimarBauer:


ReimarBauer:
I verified these changes on my guacamole windows MSS 5.0 installation by copying the changed utils.py to the installed location. 
If you have an option to try it out before the next release I'm interested in Feedback.

mslib/utils.py
```diff
@@ -101,6 +101,19 @@
 UR.define("pptv = 1e-12 fraction")
 
 
+def windows_subprocess_startupinfo():
```
joernu76: you could simplify the code quite a bit by returning "None" for non-"nt" OS and always supplying it as an argument.
Alternatively, we could shim the subprocess.run command (https://en.wikipedia.org/wiki/Shim_(computing))

Functionally this is fine.

ReimarBauer: I first thought I just replace it by EasyProcess, https://github.com/ponty/EasyProcess
But that is an API change and it won't go into 5.0.1. I packaged this for conda-forge for pyvirtualdisplay as a dependency. 

That this happens now only on windows desktop start is the reason I choosed that filter. But anyway if it breaks after next installation something else we get faster a 5.0.2.

joernu76 (approved):

PR #1154, opened 2021-08-15
GSoC config editor update 2 - validations

aravindm711:


mslib/msui/editor.py
```diff
@@ -537,11 +686,8 @@ def remove_option_handler(self):
                 else:
                     non_removable.append(index)
 
-        # dialog_box = QtWidgets.QMessageBox.question(self, 'Remove option',
-        #                 'Are you sure about removing the selected options?',
-        #                 QtWidgets.QMessageBox.Yes, QtWidgets.QMessageBox.No)
-        # if dialog_box == QtWidgets.QMessageBox.No:
-        #     return
+        # self.error_dialog = QtWidgets.QErrorMessage()
```
ReimarBauer: if the functionality is not needed, better remove it completly ;)

aravindm711: I'm still thinking about how to include an confirmation message. I'll remove it.

aravindm711:
I've added validations for checking duplicate, empty, invalid and dummy keys/values in the config options. If there are duplicate/empty/invalid keys, the related rows will be colored in red, and the config will not be saved. Whereas, If there are dummy values found, the row will be colored in gray and will ask for confirmation before saving. Did you get time to give it a try @ReimarBauer?

ReimarBauer:
@aravindm711 good idea with the color codeing for mistakes. :)
If there is no change done than nothing must be saved. We will have users which do everything which is in the dialog. So maybe the message should be changed to "Do you want to close the editor" for this case
The message on "nothing is changed" and close window could be handled different 

dialog_onrevert

Sometimes it sends me into a loop, when I have corrected, I can't leave.

aravindm711:
Thanks for the feedback. What you demonstrated with the gif was actually supposed to work like you wanted. If there are no changes from the last save, nothing will be saved. I'll look into it.

ReimarBauer:
This is may be a better video, because it is nothing to save. Also me clicked on yes ;) I get into this with cycling 1 change. Without doing that it closes without dialog.

dialog_onrevert

mslib/msui/editor.py
```diff
@@ -114,41 +118,78 @@
...
-                if isinstance(model, QtCore.QAbstractProxyModel):
-                    model = model.sourceModel()
-                source_data = model.serialize()
-                if source_data[root_option] != default_options[root_option]:
+                if model_data != default_data:
                     option.font.setWeight(QtGui.QFont.Bold)
 
+            return super(JsonDelegate, self).paint(painter, option, index)
+
+#         # Bold the value if not default
```
ReimarBauer: not sure if this only temporary removed, but such large segments if that not needed can also be removed

mslib/msui/editor.py
```diff
@@ -284,15 +336,20 @@ def __init__(self, parent=None):
...
-            parent_list.append(index)
-        parent_list.reverse()
-        if parents:
-            return index, parent_list
-        return index
+        # Add invalidity roles and update status of keys
+        self.update_view()
+
+    def dict_raise_on_duplicates_empty(self, ordered_pairs):
+        """Reject duplicate and empty keys."""
```
ReimarBauer, line 313: :)

mslib/msui/editor.py
ReimarBauer: You could leave it but also add a `# ToDo` explaining what should be achieved. 

Later someone could have an idea how to solve it.  Without a ToDo likly someone on the next refactoring will remove it.

ReimarBauer (changes requested):
a few things, see also the earlier comments. Very useful! Good ideas added!

mslib/msui/editor.py
aravindm711: Forgot to remove those unwanted lines. I'll remove them.

aravindm711:
About the bug you demonstrated with the gif, I found out that only "filepicker_default" behaves this way and I've fixed it. And, thanks for suggesting to add 'Todos', I'll do that.

mslib/msui/editor.py
joernu76: Indeed, commenting something out is typically not a good idea - long term wise. As such it is always good to remove that before mergeing it.

mslib/msui/editor.py
```diff
@@ -284,15 +336,20 @@ def __init__(self, parent=None):
...
+        d = {}
```
joernu76: d is not a good name. I would use "result" to indicate from the beginning that "d" will be returned.
k, v is borderline okay, but key, value would be more apparent and not that much more to type either.

mslib/msui/editor.py
```diff
@@ -329,6 +402,79 @@ def tree_selection_changed(self, selected, deselected):
...
+                duplicate = False
+                if len(child_keys) != rows or empty:
+                    duplicate = True
+                    color = QtCore.Qt.red
+                root_item.setData([empty, duplicate, False], InvalidityRole)
+                root_item.setData(dummy, DummyRole)
+            elif key in list_option_structure:
+                empty, duplicate, dummy = [False] * 3
+                values_list = data[key]
+                if any([values_list[i] == list_option_structure[key][0] for i in range(len(values_list))]):
```
joernu76: `if any([v == list_option_structure[key][0] for v in values_list]):` ?

mslib/msui/editor.py
```diff
@@ -678,55 +824,81 @@ def import_config(self):
...
         with fs.open_fs(dir_name) as _fs:
             _fs.writetext(file_name, json.dumps(save_data, indent=4))
 
+    def validate_data(self):
+        invalid, dummy = False, False
+        parent = QtCore.QModelIndex()
+        for r in range(self.json_model.rowCount(parent)):
+            index = self.json_model.index(r, 0, parent)
+            item = self.json_model.itemFromIndex(index)
+            if any(item.data(InvalidityRole)):
```
joernu76: simply `invalid |= any(item.data(InvalidityRole))` ?
But probably better readable as it is now.

mslib/msui/mss_pyui.py
```diff
@@ -235,6 +235,12 @@ def __init__(self, mscolab_data_dir=None, *args):
         self.active_flight_track = None
         self.last_save_directory = config_loader(dataset="data_dir")
 
+        # bind keyboard shortcuts to view actions
+        self.actionTopView.setShortcut(QtGui.QKeySequence("Ctrl+h"))
+        self.actionSideView.setShortcut(QtGui.QKeySequence("Ctrl+v"))
+        self.actionTableView.setShortcut(QtGui.QKeySequence("Ctrl+t"))
+        self.actionLinearView.setShortcut(QtGui.QKeySequence("Ctrl+l"))
```
joernu76: can't this be implemented using the QtCreator and the UI file?

mslib/msui/editor.py
aravindm711: I'll update them.

mslib/msui/editor.py
aravindm711: Better. Thanks.

mslib/msui/mss_pyui.py
aravindm711: When set as Ctrl+... in a mac, the shortcuts translate to "meta" which maps to no keys in linux systems. And also, Ctrl+h translates to command+h in mac, which is a system wide shortcut, so setting it here resolves these problems.

aravindm711: Or we could find a different shortcut for opening topview and set it through QtCreator itself in a linux/windows system.

ReimarBauer: We should keep your solution, please add a comment string above. e.g. ctrl+h is a special command on Mac, we can't use it in the QtCreator.

aravindm711: okay!

aravindm711:
@ReimarBauer I have changed actionView shortcuts to meta (ctrl in macOS), but will remain ctrl in linux/windows systems.

ReimarBauer (approved):
Thx :)

ReimarBauer (approved):
Thx

PR #1159, opened 2021-08-19
GSoC config editor update 3 - refactor of config_loader

aravindm711:
I have attempted to refactor the order in which configuration is loaded. With the changes I've made, the settings is now read only once while starting the app, and is updated whenever restarted.


mslib/msui/ui/ui_mainwindow.ui
```diff
@@ -379,9 +379,6 @@ Save a flight track to name it.</string>
    <property name="text">
     <string>&amp;Configuration</string>
    </property>
-   <property name="shortcut">
```
ReimarBauer, line 382: removed lines intended?

mslib/utils/_tests/tests.py
```diff
@@ -64,6 +65,7 @@ def test_load_settings(self):
         assert settings["foo"] == "bar"
 
 
+@pytest.mark.skip("To be done for new refactor")
 class TestConfigLoader(object):
     """
```
ReimarBauer: tests needs also be splitted in multiple modules
preferable we should have "module"  and "test_module"
the current tests.py is a placeholder from the long file.

mslib/utils/config.py
```diff
@@ -0,0 +1,299 @@
...
+                raise FatalUserError(error_message)
+            except ValueError as e:
+                logging.error(f"Error while loading json file {e}")
+                error_message = f"Invalid keys detected in config\n{e}"
+                raise FatalUserError(error_message)
+        else:
+            error_message = f"MSS config File '{path}' not found"
+            raise FatalUserError(error_message)
+
+    global user_options
```
ReimarBauer, line 373: with global I'm very carefully and I had already a lot of the problems here
https://stackoverflow.com/questions/19158339/why-are-global-variables-evil

ReimarBauer (changes requested):
A solution whithout using global or a good reason commented prefered
multithreading may get limited by that.

a field test follows

mslib/msui/ui/ui_mainwindow.ui
aravindm711: I removed this because this shortcut applies only for mac. Should I include it anyways?

mslib/utils/_tests/tests.py
aravindm711: I was planning to do that once we decided if the refactor was okay.

mslib/utils/config.py
aravindm711: I'll look into this

joernu76: evil is a strong word. I think the use is justified here. Please not that you do not need it here, actually, as user_options is a dictionary, which you might modify in-place. 
E.g.
`
copied_config = copy.deepcopy(user_options)
user_options.clear()
user_config.update(copied_config)
`
or similar.

mslib/utils/config.py
```diff
@@ -0,0 +1,299 @@
...
+    Function for loading json config data
+
+    Args:
+        config_file: json file, parameters for initializing mss,
+        dataset: section to pull from json file
+
+    Returns: a the dataset value or the config as dictionary
+
+    """
+    global user_options
```
joernu76: global not needed here, I think.

mslib/utils/config.py
```diff
@@ -0,0 +1,299 @@
...
+import logging
+import fs
+
+from mslib.utils import FatalUserError
+from mslib.msui import constants, MissionSupportSystemDefaultConfig
+from mslib.support.qt_json_view.datatypes import match_type, UrlType, StrType
+
+
+# system default options as dictionary
+default_options = dict(MissionSupportSystemDefaultConfig.__dict__)
```
joernu76: wouldn't it make sense to move the config class here? And maybe instrument it with the structure info stored in the dicts below?

mslib/msui/mss_pyui.py
```diff
@@ -767,13 +769,13 @@ def show_shortcuts(self):
...
         dlg.exec_()
 
-    def status(self):
-        if constants.CACHED_CONFIG_FILE is None:
-            return ("Status : System Configuration")
-        else:
-            filename = constants.CACHED_CONFIG_FILE
-            head_filename, tail_filename = os.path.split(filename)
-            return("Status : User Configuration '" + tail_filename + "' loaded")
+    # def status(self):
```
joernu76: commented out sections should not appear in non-draft/WIP merge requests.

mslib/utils/config.py
joernu76: Actually, many arguments against global variables can be made against classes and in Python the distinction between a class and a module is marginal. So object oriented programming is...?
"State" in any form is "evil", IMHO, but functional programming languages (e.g. Haskell) lacking state have somehow problems of not being adopted due to being very difficult to work with. But man, the code is easily verifiable and testable.

mslib/msui/mss_pyui.py
aravindm711: I'll remove them.

mslib/utils/config.py
aravindm711: yes, just wanted to be explicit.

mslib/utils/config.py
aravindm711: I'll look into this.

aravindm711:
@joernu76  Do you think I can go ahead with using the global variables?

ReimarBauer:
On client side global is not that dangerous as used on server side. 

https://stackoverflow.com/questions/32815451/are-global-variables-thread-safe-in-flask-how-do-i-share-data-between-requests

joernu76:
> @joernu76 Do you think I can go ahead with using the global variables?

yes.

ReimarBauer:
locally always this test hangs `mslib/msui/_tests/test_wms_control.py::Test_HSecWMSControlWidget::test_preload`

pytest -v -n 8 --dist loadscope  mslib

The test uses a static method of MSSMainWindow. Can the change of config create a problem?
https://github.com/Open-MSS/MSS/blob/40ba47598c63a504f2ba6837f56c9f7a58857a63/mslib/msui/mss_pyui.py#L318

ReimarBauer (changes requested):
I have a bit too much values stored and when I update, e.g. restore topview they are coming back, value 20. But they are not in the saved file.

Also when I say yes to the restart because of changes to the config editor, it did only close the views, it does not disconnect me or log out me from mscolab.

too_much


When I Quit MSS and restart it, then defaults of topview font size are used.

if I change other values I get again that problem

can_t_save_changes

Now the values are different to the stored ones and different to the default, but I can't store.

aravindm711:
@ReimarBauer Thanks for catching this. It turns out I was trying to merge previously loaded config with the new config, which ended up as a reason for this pretty serious bug. I think I've fixed it, please give it a try and let me know.

Now I've made restarting mandatory while trying to save the config, because if not, we are allowing a state where the loaded config is different from the saved config. Now it works perfectly.

About the failing wms_preload test, I'm not able to run the test on my machine. It gives me this output,
```
mslib/msui/_tests/test_wms_control.py E                              [100%]

================================== ERRORS ==================================
_________ ERROR at setup of Test_HSecWMSControlWidget.test_preload _________

self = <test_wms_control.Test_HSecWMSControlWidget object at 0x1470a16a0>

    def setup(self):
>       self._setup("hsec")

mslib/msui/_tests/test_wms_control.py:114:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
mslib/msui/_tests/test_wms_control.py:74: in _setup
    self.thread.start()
../../../../opt/anaconda3/envs/mssenv/lib/python3.9/multiprocessing/process.py:121: in start
    self._popen = self._Popen(self)
../../../../opt/anaconda3/envs/mssenv/lib/python3.9/multiprocessing/context.py:224: in _Popen
    return _default_context.get_context().Process._Popen(process_obj)
../../../../opt/anaconda3/envs/mssenv/lib/python3.9/multiprocessing/context.py:284: in _Popen
    return Popen(process_obj)
../../../../opt/anaconda3/envs/mssenv/lib/python3.9/multiprocessing/popen_spawn_posix.py:32: in __init__
    super().__init__(process_obj)
../../../../opt/anaconda3/envs/mssenv/lib/python3.9/multiprocessing/popen_fork.py:19: in __init__
    self._launch(process_obj)
../../../../opt/anaconda3/envs/mssenv/lib/python3.9/multiprocessing/popen_spawn_posix.py:47: in _launch
    reduction.dump(process_obj, fp)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

obj = <Process name='Process-1' parent=14521 initial>
file = <_io.BytesIO object at 0x145ad5040>, protocol = None

    def dump(obj, file, protocol=None):
        '''Replacement for pickle.dump() using ForkingPickler.'''
>       ForkingPickler(file, protocol).dump(obj)
E       AttributeError: Can't pickle local object 'app_loader.<locals>.files'

../../../../opt/anaconda3/envs/mssenv/lib/python3.9/multiprocessing/reduction.py:60: AttributeError
============================= warnings summary =============================
../../../../opt/anaconda3/envs/mssenv/lib/python3.9/site-packages/future/standard_library/__init__.py:65
  /Users/aravindmurali/opt/anaconda3/envs/mssenv/lib/python3.9/site-packages/future/standard_library/__init__.py:65: DeprecationWarning: the imp module is deprecated in favour of importlib; see the module's documentation for alternative uses
    import imp

-- Docs: https://docs.pytest.org/en/stable/warnings.html
========================= short test summary info ==========================
ERROR mslib/msui/_tests/test_wms_control.py::Test_HSecWMSControlWidget::test_preload
======================= 1 warning, 1 error in 6.43s ========================

```
But, I think config_loader wouldn't be a problem as it returns default values when the config file is not read.

ReimarBauer (approved):
Thx

PR #1160, opened 2021-08-19
upgrading pallet projects modules e.g. werkzeug 2.0

### LINKED ISSUE #970: refactor for updated pallets projects
https://palletsprojects.com/blog/flask-2-0-released/

ReimarBauer:


mslib/mscolab/_tests/test_chat.py
```diff
@@ -79,24 +79,24 @@ def handle_incoming_message(msg):
             msg = json.loads(msg)
             messages.append(msg)
 
-        sio.on('chat-message-client', handler=handle_incoming_message)
-        sio.connect(self.url)
-        sio.emit('start', response)
+        sio.on('chat-message-client', handler=handle_incoming_message, namespace="/")
```
joernu76: This namespace for *each call* seems pretty dangerous and verbose. Is this really necessary or can this somehow be configured more smartly?

According to "https://python-socketio.readthedocs.io/en/latest/server.html#namespaces" "/" *is* the default namespace, so this seems superfluous?

Alternatively, one can use class based namespaces.

joernu76 (dismissed):

mslib/mscolab/_tests/test_chat.py
ReimarBauer: see 
https://flask-socketio.readthedocs.io/en/latest/upgrading.html

We do have an event based communication, which is secured by a token. 
On any http request the Url is verbose.

ReimarBauer: example for using more than one namespace. 
https://socket.io/docs/v3/namespaces/index.html

separation, improving, organisation can be done in a different issue.

joernu76 (dismissed):

joernu76 (approved):

PR #1161, opened 2021-08-20
GSoC update - Updated MSColab help dialog

aravindm711:


ReimarBauer (changes requested):
Recently collaborator role was changed  but it was missed to change it in "Adding Users to project". 

3. Collaborator - Collaborator can edit the flight track path and use the version history but has no ability to manage users.

ReimarBauer (approved):
Thx

Issue #1153, opened 2021-08-14
shortcuts commands ideas


ReimarBauer:
I'm thinking about the "Current Shortcuts" by Alt+S

Currently I can read them and then use on my keyboard that shown Keystroke.
I also could select a line. e.g. "Table View Ctrl +T". 

The shortest way  to select now "Table View Ctrl +T" could be double click and open it from there.

When we could show there also all Buttons, Menu, Input entries, also those without a key code and add a search field this will give a kind of overview and a different navigation option.

For using pyautogui this also means we can enter a search string to that helper and it finds the button, without the need to have an image  to locate it on screen. Currently we need different images dependent on users screen depth and fonts used.

We could also think about having a Ctrl+F on any view which then searches this helper data and highlights the element we are looking for.

This behaviour is than quite similiar to browsers.

For pyautogui that would mean we highlight something and search for e.g. a yellow box
set the mouse pointer there and click.

Marilyth:
So you are suggesting
- Being able to execute shortcuts via double click in the Shortcut dialog
- Include within the Shortcut dialog all interactive QObjects (sounds like it would be very confusing for a normal user)
- Include a filter/search for the Shortcut dialog
- Include a Ctrl+F shortcut on every window to highlight QObjects by their name (for ease of use for pyautogui)

Yes?

PR #1172, opened 2021-08-23
Add more shortcut functionality

Marilyth:
Partially implements #1153 

Expands the shortcuts by:
- Execute shortcut through double clicking
- Include non-shortcut object if so desired
- Highlight currently selected shortcut with a yellow background
- Include a filter
- Add Ctrl+F Shortcut to directly search for plain text in the UI via the shortcuts dialog


ReimarBauer:
great to enhance by "Advanced Settings" the information density

The list shows a ":" I've no idea what is connected to that, the redraw button starts with `&R...`
There also some to select which are in menus 

example

Would it be posible to show a tooltip, when it is selected? 

Or a checkbox to extend to tooltips and to have it disabled on default?

ReimarBauer (changes requested):
see coment

Marilyth:
Hovering over an item now reveals all available information about it.

ReimarBauer (changes requested):
That feature showing ObjectName gets a very good development feature too. Great!

Found a small problem,
two_selections

that happens when I select with "Show items without shortcut" enabled in Text Type. Then  toggle "Show items without shortcut" to disabled and switch Display type to ObjectName. Afterwards select something.

Marilyth:
Should be fixed now

ReimarBauer (approved):

PR #1179, opened 2021-08-25
Gallery extension

### LINKED ISSUE #1129: Campaign-Defined Plot Gallery
A very useful tool would be to have a campaign-defined plot gallery available on the MSS server URL.
That means a set of pre-defined (top view) plots that are important for the specific measurement campaign.
It could be built upon the "plot gallery" structure.
Pre-definition of the plots should include map projection section, list of parameters (also allow multilayer plots),
a list of  levels. The exact plot settings would be defined in a config file, like in retriever.py that uses "automated_plotting" in mss_settings.json

An exact website layout needs to be defined/discussed: 
There could be a page displaying all plots in a gallery and one for single plots.
Maybe also different galleries for different scientific themes (e.g. chemistry, clouds, gravitiy waves)
There could be a choice box to set the initialisation time and valid time. 
There could be a clickable arrow to move in valid time and initialisation time.

Marilyth:
Partially fixes #1129 
Example here https://mss-test.readthedocs.io/en/latest/gallery/index.html
Includes the ability to specify one or more levels, init times and valid times. 
In case of multiple the user can go through them via a combobox on the website.
Does not yet include multilayering.


joernu76:
This is a very good start.
1) I have some troubles with the UI. I can switch between the 18th and the 19th, but not select the bolded 17th. (Firefox on Linux).
2) The unified levels drop down box will not be usable on a "real" server. I think one needs to either have a more complicated interface with different leveltypes and stuff, or have a reduction to a (much) smaller set of layers as in the example.
3) Switch pressure levels seems to work, selecting, e.g., 2PVU does not seem to do anything. This is confusing. I assume that the PV plots change, but that is not visible. The disconnect between action and appearance need be adressed. E.g. either only display plots of a type or have multiple boxes for different level types...?
4) We could also restrict a specific gallery to one type of plot, i.e. offer multiple galleries on one server for different products to circumvent the problem.
I see that this is contradictory to the purpose of documentation, where we want all plots on one page.
5) It would also be nice to have a similar interface on the individual plot page, because the plots on the main page are very small and I do not see how I could enlarge them (and if they were enlarged the UI would be far away).

Marilyth:
The 17th is the init-time, the 18th and 19th are the valid times for this init-time. It looks odd mostly because there aren't multiple init times. One click in the time selection directly chooses init and valid times.
I am also very undecided on how to do the level selection, as layer have different units for it. Right now, changing a level only changes the layers whose unit is in the selected level, the others stay as is.

joernu76:
Integrating the init time as such is very efficient, but with the given times it was a bit confusing. With "real" data, this will likely be obvious.

Marilyth:
Level selection now hides all layers which don't have the specified unit. There is now also a time and level select inside the individual plot pages.

joernu76:
Pretty cool!
This serves as a nice getting-an-overview feature and might be developed further in additional PRs.
@ReimarBauer I am for merging it now.

joernu76 (approved):

ReimarBauer (approved):
Thx

PR #1180, opened 2021-08-26
Fix wrong GET data retrieval, and missing CORS header

### LINKED ISSUE #1095: mscolab not JavaScript friendly
While developing the windy plugin I encountered a few oddities of mscolab which should be solved in order to be used with JavaScript.

1. API and Websocket don't return a CORS header, which is required for communication through JavaScript.
2. Some GET endpoints require request.form data. GET requests don't have a body, as such either query string request.args or the combined request.values should be used.

Marilyth:
Fixes #1095 as far as I can see.
Includes a CORS header for mscolab requests, and replaces false request.form instructions with request.values


ReimarBauer (changes requested):
some documentation of this 
https://auth0.com/blog/cors-tutorial-a-guide-to-cross-origin-resource-sharing/

Want to have the cors_allowed_origins="*" configurable by the msolab server configuration. A server owner can restrict the access to an url or open it for all or limit it to own adress. Currently I would prefer it at least configurable. The default can be any origin. The url for windy should be documented.

Marilyth:
It is now configurable. Why should the url for windy be documented? Just as an example other than *?

ReimarBauer:
> It is now configurable. Why should the url for windy be documented? Just as an example other than *?

I'm currently not aware of the state of a discussed plugin. At the moment that is available it can be added "for example limit only for ..."

ReimarBauer (approved):
Thx

PR #1185, opened 2021-08-27
Don't get confused by python-mss

### LINKED ISSUE #1184: Updater doesn't work with python-mss installed
The Updater mistakes python-mss for the current mss version.

Marilyth:
Within conda list only search for mss as a full name, as such ignore all other packages including mss.
Fixes #1184


ReimarBauer (approved):
Thx

PR #1189, opened 2021-09-02
Don't trigger multilayer click if favourite

### LINKED ISSUE #1188: Favourite Selection in Multilayer Window
Current state:
Sselecting a layer as favourite also selects the layer, triggering (typically) a map request.

Desired state:
Selecting the "star" of a layer do make/unmake it a favourite should not select the layer.

Marilyth:
Fixes #1188 
Trigger favourite check before multilayer click, if the user did change the favourite status skip the multilayer click event once.


joernu76 (approved):
nice!

PR #1191, opened 2021-09-05
changes on UI and models for using categories

### LINKED ISSUE #1053: further ideas for organisation of the projects on the mscolab server
Longterm we will have almost the same users on a server which do different campaigns.
A campaign is the organisational form of many scientists to rent several flight hours of an aircraft and to equipp it with instruments. 

We need an option to group the projects by campaign name. 
This enables a search by this and to list only those for a user.
A list var in the mss_settings should be used to preselect.

### LINKED ISSUE #1183: mscolab projects need categories
currently all projects are shown in a large growing lists.

We need a possibility to group and filter them by categories. 
The database needs an entry for adding a category. This category can be a campaign name. This corresponds to a namespace

On the client site we need a filter for a list of category entries the user has access too (given from the projects)

On the mss_settings we need a default for selecting those to be seen in the UI list of projects

On the new project dialog we need a field to selet from those mss_settings list or add a new one.

ReimarBauer:


ReimarBauer:
I look next on filing a default on project creation from the mss_settings.json

Also missing is a validation on input of the category name.

mslib/msui/mscolab.py
```diff
@@ -687,8 +694,9 @@ def browse():
             self.add_proj_dialog.buttonBox.accepted.connect(self.add_project)
             self.add_proj_dialog.buttonBox.button(QtWidgets.QDialogButtonBox.Ok).setEnabled(False)
             self.add_proj_dialog.path.textChanged.connect(check_and_enable_project_accept)
-            self.add_proj_dialog.description.textChanged.connect(check_and_enable_project_accept)
```
joernu76, line 690: This should not have been removed.?

joernu76 (changes requested):

mslib/msui/mscolab.py
ReimarBauer: good catch :)

mslib/msui/mscolab.py
```diff
@@ -687,8 +693,10 @@ def browse():
...
+            self.add_proj_dialog.description.textChanged.connect(check_and_enable_project)
```
joernu76: "_accept" is missing

joernu76 (approved):

PR #1194, opened 2021-09-05
default is not having a mss_settings.json

### LINKED ISSUE #1176: none existing mss_settings.json creates an error
We need a fix and  a test for this case too :)

```
 python mslib/msui/mss_pyui.py 
INFO: MSS Version: 5.0.1.
INFO: Python Version: 3.9.6 | packaged by conda-forge | (default, Jul 11 2021, 03:39:48) 
[GCC 9.3.0]
INFO: Platform: Linux-5.4.0-81-generic-x86_64-with-glibc2.27 (('64bit', 'ELF'))
  File "/home/reimar/MASTER/MSS/mslib/msui/mss_pyui.py", line 913, in <module>
    main()
  File "/home/reimar/MASTER/MSS/mslib/msui/mss_pyui.py", line 898, in main
    read_config_file()
  File "/home/reimar/MASTER/MSS/mslib/utils/config.py", line 371, in read_config_file
    raise FatalUserError(error_message)
mslib.utils.FatalUserError: MSS config File '/home/reimar/.config/mss/mss_settings.json' not found
CRITICAL: MSS Version: 5.0.1.
CRITICAL: Python Version: 3.9.6 | packaged by conda-forge | (default, Jul 11 2021, 03:39:48) 
[GCC 9.3.0]
CRITICAL: Platform: Linux-5.4.0-81-generic-x86_64-with-glibc2.27 (('64bit', 'ELF'))
  File "/home/reimar/MASTER/MSS/mslib/msui/mss_pyui.py", line 913, in <module>
    main()
  File "/home/reimar/MASTER/MSS/mslib/msui/mss_pyui.py", line 898, in main
    read_config_file()
  File "/home/reimar/MASTER/MSS/mslib/utils/config.py", line 371, in read_config_file
    raise FatalUserError(error_message)
mslib.utils.FatalUserError: MSS config File '/home/reimar/.config/mss/mss_settings.json' not found

QWidget: Must construct a QApplication before a QWidget
Abgebrochen
```

ReimarBauer:


joernu76:
Mmmh. We are pretty agressive nowadays with creating the non-existing config directory. Maybe we should simply place a default configuration there in case it does not exist? 

I certainly do not like the pop-up window.

ReimarBauer:
Currently, a new user can't start the msui after installation when she did not know where to place a file. And that is bad. Also automatically creating one I dislike because I want to use the default configuration.  The MSUI has to start based on defaults not based on different user settings.

I prefer the old situation without a popup and only the logging.info message.  I want to have MSS always use the default configuration when there is no user settings.

The config editor stores only differences to the defaults. So the minimum file has to have 
```
{
}
```

joernu76:
Maybe simply place a "{}" file in the directory upon creation?

I also liked the old behaviour, but I am unsure that the exception is caught at the right level.
Previously, an error was thrown in case a *configured config filename* was not found. Now, the code does not distinguish between the default file location and a configured one, which causes the problem.

ReimarBauer:
This needs further tests, if this still can be used. We use this to setup the tests too,
os.environ["MSS_CONFIG_PATH"] = MSS_CONFIG_PATH

ReimarBauer:
for tests that works, have to fix other problems too.

mslib/msui/mss_pyui.py
```diff
@@ -894,8 +894,11 @@ def main():
     logging.info("Python Version: %s", sys.version)
     logging.info("Platform: %s (%s)", platform.platform(), platform.architecture())
 
-    read_config_file()
-    logging.info("Launching user interface...")
+    try:
+        read_config_file()
+    except FileNotFoundError as ex:
```
Marilyth: Now that the settings file is being created, does this error ever occur with the default parameter for read_config_file?

mslib/msui/constants.py
```diff
@@ -48,6 +48,14 @@
 
 MSS_SETTINGS = os.getenv('MSS_SETTINGS', os.path.join(MSS_CONFIG_PATH, "mss_settings.json"))
 
+if '://' in MSS_SETTINGS:
+    if not _fs.exists('mss_settings.json'):
+        _fs.writetext("mss_settings.json", "{}")
```
Marilyth: Shouldn't people with an fs path also be able to set another filename than mss_settings.json?

Marilyth (commented):
A few questions.

mslib/msui/mss_pyui.py
joernu76: Agree, it's "double fixed" now. And the fs/non-fs code path is different.

ReimarBauer: interesting, we have  two tests which don't fail. So yes that can be reached.

mslib/msui/constants.py
ReimarBauer: yes ;)

mslib/msui/mss_pyui.py
ReimarBauer: It still can happen, but I guess rather seldom.  One case is you set a path where you don't have create rights. 

decided to exit for that case and throw an error message.

mslib/msui/mss_pyui.py
```diff
@@ -894,8 +894,11 @@ def main():
...
+        message = f"Setup your configuration.\nUse the comfortable editor in the File menu.\n{ex}"
+        logging.info(message)
```
joernu76: see other discussion. Also, most people will never see this.

joernu76 (approved):

PR #1201, opened 2021-09-09
Fix no init/valid time support

### LINKED ISSUE #1200: MSWMS server does not work for data without init/valid times
Some layers may have no "init" or "valid" time. This is supported by the server, but this support contains a bug.

joernu76:
Fix #1200

### DIFF
--- a/mslib/mswms/mss_2D_sections.py
+++ b/mslib/mswms/mss_2D_sections.py
@@ -135,21 +135,15 @@ class Abstract2DSectionStyle(metaclass=ABCMeta):
 
     def uses_inittime_dimension(self):
         """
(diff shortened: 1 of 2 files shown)


mslib/mswms/mss_2D_sections.py
```diff
@@ -140,7 +140,7 @@ def uses_inittime_dimension(self):
 
         Currently redirected to check for valid_time.
```
ReimarBauer: Can you rephrase this too. It sounds wrong, because there is not another call here.
"Currently redirected to check for valid_time."

ReimarBauer (changes requested):
the docstrings of the changed functions may need changes too.

ReimarBauer (approved):

Issue #536, opened 2020-02-27
refactor  logging-format-interpolation findings by pylint


ReimarBauer:
**[Original report](https://bitbucket.org/wxmetvis/mss/issue/536) by Reimar Bauer (Bitbucket: [ReimarBauer](https://bitbucket.org/ReimarBauer), GitHub: [ReimarBauer](https://github.com/ReimarBauer)).**

----------------------------------------

pylint: logging-format-interpolation / Use % formatting in logging functions and pass the % parameters as arguments 

* mslib/mscolab/add\_bulk\_permissions.py
* mslib/utils.py
* mslib/mscolab/conf.py
* mslib/msui/mpl\_map.py
* mslib/msui/mscolab.py
* mslib/thermolib.py

‌

ReimarBauer:
**Original comment by Reimar Bauer (Bitbucket: [ReimarBauer](https://bitbucket.org/ReimarBauer), GitHub: [ReimarBauer](https://github.com/ReimarBauer)).**

----------------------------------------

resoved by 14470a1

PR #1202, opened 2021-09-12
option added for user verification

### LINKED ISSUE #632: verify user creation for mscolab
Currently we can create users in the UI of mscolab without verification by a valid email

we need to send an email to the user and verify by a token that his email address is correct.
only verified users should be added to the database by the gui.

ReimarBauer:


ReimarBauer:
This was a good start for this implementation
- https://realpython.com/handling-email-confirmation-in-flask/
- https://pythonhosted.org/Flask-Mail/

mslib/mscolab/server.py
```diff
@@ -28,11 +28,15 @@
 import json
 import logging
 import time
+import datetime
 import secrets
 import fs
 import socketio
+from itsdangerous import URLSafeTimedSerializer
```
joernu76, line 35: I do not see this in meta.yaml

mslib/mscolab/server.py
```diff
@@ -89,6 +102,37 @@ def verify_pw(username, password):
...
+    msg = Message(
+        subject,
+        recipients=[to],
+        html=template,
+        sender=APP.config['MAIL_DEFAULT_SENDER']
+    )
+    try:
+        mail.send(msg)
+    except IOError:
+        logging.debug(f"Can't send email to {to}")
```
joernu76: Do not use f-strings. Use % for performance reasons etc.

mslib/msui/mscolab.py
```diff
@@ -1159,22 +1169,23 @@ def handle_project_deleted(self, p_id):
...
-                projects = _json["projects"]
-                self.ui.filterCategoryCb.clear()
-                categories = set(["ANY"])
-                for project in projects:
-                    categories.add(project["category"])
-                categories.remove("ANY")
-                categories = list(categories)
-                categories.insert(0, "ANY")
-                self.ui.filterCategoryCb.addItems(categories)
+        if self.mscolab_server_url is not None:
```
joernu76: use "and" to avoid indentation

mslib/msui/mscolab.py
```diff
@@ -1159,22 +1169,23 @@ def handle_project_deleted(self, p_id):
...
+                if r.text != "False":
+                    _json = json.loads(r.text)
+                    projects = _json["projects"]
+                    self.ui.filterCategoryCb.clear()
+                    categories = set(["ANY"])
+                    for project in projects:
+                        categories.add(project["category"])
+                    categories.remove("ANY")
+                    categories = list(categories)
+                    categories.insert(0, "ANY")
```
joernu76: evtk. `categories = ["ANY"] + sorted(categroies)`

joernu76 (changes requested):

mslib/mscolab/server.py
Marilyth: My personal timeit performance check turned out, at best, inconclusive and at worst show c-strings being worse than f-strings at high numbers, but the difference is negligible.
I would recommend using what is most readable, which is f-strings.

mslib/mscolab/server.py
ReimarBauer: it is a dependency of flask, see
https://github.com/conda-forge/flask-feedstock/blob/master/recipe/meta.yaml

mslib/mscolab/server.py
ReimarBauer: we have already other f-strings for logging.

joernu76: f-formatting is almost always faster than c-style formatting due to some particularly interesting implementation shenanigans. That is not the point here, however.

The idea here is that the formatting will *not take place at all* in case the logging level is above debug. Using an f-string always does the formatting even if nothing will be output. That can have an detrimental effect on performance *with logging disabled*.
Thus it is good style to *always* employ conditional formatting, which by necessity of backwards-compatability, is c-style formatting. For info/warn/error, which are basically always dumped or where the formatted string is needed anyway (e.g. for a popup message), one may deviate at discretion. 

There is a "style" parameter, which can be triggered on a *by call* basis to use other formatters (e.g. the {} one). See
https://docs.python.org/3/howto/logging-cookbook.html#formatting-styles
This seems very cumbersome for little gain.
f-strings cannot be used as the decision whether to format cannot be controlled by the logging class.

mslib/msui/mscolab.py
ReimarBauer: this line can be removed, this was left over from some trials

mslib/mscolab/server.py
joernu76: > we have already other f-strings for logging.

Then these need to be replaced as well.

ReimarBauer: this is than another refactoring, fixed that one

joernu76: Fine for me if it is fixed here.
We had that discussion - and fixes - in the past.
E.g.
https://github.com/Open-MSS/MSS/issues/536

joernu76 (approved):

mslib/mscolab/server.py
ReimarBauer: I remember ;)

PR #1204, opened 2021-09-12
msc project renamed to operation

### LINKED ISSUE #1195: rename UI wordings
I do want to change the wordings of the UI, mainly because of strongen that we do have Missions on focus. I always think Projects are Missions, so lets discuss about this idea first.

any Project word becomes Mission

- s/Open Flight Tracks/Flight Tracks/
- s/Open Projects/Missions/
- s/Connect to MSColab/Connect MSC/
- s/No Project Selected/No Mission selected/
- s/MSColab Url/MSC Url/
- s/Clone Project Permissions/Clone Mission Permissions/
- s/Project Properties/Mission Properties/

It is also easier to describe what we are doing when we keep our wordings.
Instead of using the word Mission I also thought about Operation. For me each of them fits better than Project.

ReimarBauer:


joernu76 (dismissed):
looks dangerous ...

mslib/msui/qt5/ui_add_operation_dialog.py
```diff
@@ -0,0 +1,69 @@
+# -*- coding: utf-8 -*-
+
+# Form implementation generated from reading ui file 'ui_add_operation.ui'
```
Marilyth: ui_add_operation creating ui_add_operation_dialog? Is there a naming error?
Either way, ui_add_operation.ui and ui_add_operation_dialog.py differ and retranslating it throws errors on creating operations.

Marilyth (changes requested):
Please look out for changes you made to the UI .py files which are not reflected in the .ui files.
Here are a few things I found

https://github.com/ReimarBauer/MSS/blob/a077a63cc5953173f53d468d33d555b6941bcdbc/mslib/msui/ui/ui_mainwindow.ui#L122 this group still uses project instead of operation

https://github.com/ReimarBauer/MSS/blob/renaming/mslib/msui/ui/ui_mscolab_version_history.ui
https://github.com/ReimarBauer/MSS/blob/renaming/mslib/msui/ui/ui_mscolab_operation_window.ui
https://github.com/ReimarBauer/MSS/blob/renaming/mslib/msui/ui/ui_mscolab_admin_window.ui
Many elements here still default to Project, even if it isn't visible in the end

https://github.com/ReimarBauer/MSS/blob/renaming/mslib/msui/ui/ui_mscolab_help_dialog.ui The help window still uses Project instead of Operation

mslib/msui/qt5/ui_add_operation_dialog.py
ReimarBauer: > Please look out for changes you made to the UI .py files which are not reflected in the .ui files.
> Here are a few things I found
> 
> https://github.com/ReimarBauer/MSS/blob/a077a63cc5953173f53d468d33d555b6941bcdbc/mslib/msui/ui/ui_mainwindow.ui#L122 this group still uses project instead of operation
> 
> https://github.com/ReimarBauer/MSS/blob/renaming/mslib/msui/ui/ui_mscolab_version_history.ui
> https://github.com/ReimarBauer/MSS/blob/renaming/mslib/msui/ui/ui_mscolab_operation_window.ui
> https://github.com/ReimarBauer/MSS/blob/renaming/mslib/msui/ui/ui_mscolab_admin_window.ui
> Many elements here still default to Project, even if it isn't visible in the end
> 
> https://github.com/ReimarBauer/MSS/blob/renaming/mslib/msui/ui/ui_mscolab_help_dialog.ui The help window still uses Project instead of Operation

thx.

I've done those too. Please have a look again.

Marilyth: https://github.com/ReimarBauer/MSS/blob/79723e123db87f9dea98299a51516626e813e265/mslib/msui/mpl_map.py#L87
Perhaps project_name should be renamed, now that it is called operation. @ReimarBauer 
I don't think I see anything else.

Marilyth (dismissed):

ReimarBauer:
This will be the last in the row. The incoming changed need also resolving merge problems.

joernu76 (approved):

PR #1206, opened 2021-09-14
Merge stable to develop

ReimarBauer:
this needs a linear merge


mslib/mscolab/server.py
```diff
@@ -182,7 +184,7 @@ def verify_user(func):
     @functools.wraps(func)
     def wrapper(*args, **kwargs):
         try:
-            user = User.verify_auth_token(request.form.get('token', False))
+            user = User.verify_auth_token(request.values.get('token', False))
```
ReimarBauer: this needs to become reverted because of newer pallets projects

ReimarBauer:
That will be redone, too much to learn

On that merge is also a refactoring needed

PR #1208, opened 2021-09-15
change views on given permissions especially for operations with viewer

### LINKED ISSUE #1203: adding/revoking permissions have to add/remove flightpathes
In develop to 6.0 revoking permissions don't remove flightpathes from the project list and also don't remove flightpathes from opened views. In the released version the windows were closed for the case that no further access was given.

Giving permissions did not update flightpathes on the project list. A toggle between categories does a manual update. This needs to get automated again.

ReimarBauer:


mslib/mscolab/utils.py
```diff
@@ -30,6 +30,54 @@
 from mslib.mscolab.conf import mscolab_settings
 
 
+def enable_navbar_action_buttons(_type, view_window):
```
joernu76: Not the best object-oriented programmer here, but wouldn't this be better a method of the views itself instead of a function?

Marilyth: I agree. The _type can also be determined by the view_window class name through `isinstance()` in the MSSViewWindow, or just reimplementing it for each subclass. So effectively both parameters can be removed.

mslib/msui/mscolab.py
```diff
@@ -1262,6 +1276,8 @@ def set_active_pid(self, item):
             # set new waypoints model to open views
             for window in self.ui.get_active_views():
                 window.setFlightTrackModel(self.waypoints_model)
+                if self.access_level == "viewer":
```
Marilyth, line 1280: Isn't there an else missing here?
Suppose you have a TopView open as a viewer, and change to another operation where you are a collaborator, the navbar does not enable itself again.

Marilyth (dismissed):

mslib/mscolab/utils.py
ReimarBauer: agreed

mslib/msui/mscolab.py
ReimarBauer: good catch :)

joernu76 (dismissed):

joernu76 (approved):

PR #1210, opened 2021-09-15
Fix flighttrack name not updating in views

### LINKED ISSUE #1192: change of mscolab project does not update title on opened views
The Topview and other views keep the path of the initial used flightpath in the title.

e.g. starting with example switching to measurement keeps example in the title

also the text in TopView above the EPSG code is not changed.

### LINKED ISSUE #1209: Linear view vertical lines not resetting on flighttrack change
When changing a flight track, the linear view does not erase old vertical lines, as such it becomes quite convoluted and confusing.

Marilyth:
Fixes #1192 and fixes #1209
Changing the project/operation in the main UI now also changes the views window title and other references to it in the plot.

### DIFF
--- a/mslib/msui/mpl_qtwidget.py
+++ b/mslib/msui/mpl_qtwidget.py
@@ -967,7 +967,7 @@ class MplLinearViewCanvas(MplCanvas):
         If no model had been set before, create a new interactor object on the model
         """
         self.waypoints_model = model
-        pass
+
         if self.waypoints_interactor:
             self.waypoints_interactor.set_waypoints_model(model)
         else:
@@ -1029,6 +1029,12 @@ class MplLinearViewCanvas(MplCanvas):
                                      for d in zip(lats[::tick_index_step],
                                                   lons[::tick_index_step])],
                                     rotation=25, horizontalalignment="right")
+
+            # Remove all vertical lines
+            vertical_lines = [line for line in self.ax.lines if
+                              all(x == line.get_path().vertices[0, 0] for x in line.get_path().vertices[:, 0])]
+            for line in vertical_lines:
+                self.ax.lines.remove(line)
 
             ipoint = 0
             highlight = [[wp.lat, wp.lon] for wp in self.waypoints_model.waypoints]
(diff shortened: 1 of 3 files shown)


mslib/msui/mpl_qtwidget.py
```diff
@@ -1030,6 +1030,12 @@ def redraw_xaxis(self):
                                                   lons[::tick_index_step])],
                                     rotation=25, horizontalalignment="right")
 
+            # Remove all vertical lines
```
joernu76, line 1038: Who is drawing these lines? We should store the lines "properly" in a member to remove it cleanly...?

mslib/msui/viewwindows.py
```diff
@@ -173,10 +178,20 @@ def setFlightTrackModel(self, model):
         """
         Set the QAbstractItemModel instance that the view displays.
         """
-        self.waypoints_model = model
+        super().setFlightTrackModel(model)
```
joernu76, line 180: oh, python 3 super is much easier to use. Nice.

mslib/msui/viewwindows.py
```diff
@@ -173,10 +178,20 @@ def setFlightTrackModel(self, model):
...
+
         if self.mpl is not None:
             self.mpl.canvas.set_waypoints_model(model)
 
+            # Update Top View flighttrack name
```
joernu76, line 185: Same as above. I would prefer using members/methods for this instead of "parsing" the text. Not very maintainable.

joernu76 (changes requested):

mslib/msui/viewwindows.py
Marilyth: Done

mslib/msui/mpl_qtwidget.py
Marilyth: Done, also for Side View.

ReimarBauer:
the operation name gets used and replaced by clicking on the map to the "internal name" Flight track. This is also added to the window title

namechangeonclick

ReimarBauer (changes requested):
By an operation the name changes  on e.g. inserting a waypoint to Flight track, see peek  comment.

joernu76 (dismissed):

Marilyth:
Should be fixed now

ReimarBauer (approved):
retried, works now. Thx.

PR #1215, opened 2021-09-17
Allow GET requests to use query parameters

### LINKED ISSUE #1095: mscolab not JavaScript friendly

Marilyth:
fixes #1095 
GET requests normally don't have a body, as such it should have used `args`.
For backwards compatibility both should be allowed, and `args` should be preferred over `form`.
Werkzeug 2.0 removed the usability of `values` for GET requests.

### DIFF
--- a/mslib/mscolab/server.py
+++ b/mslib/mscolab/server.py
@@ -184,7 +184,7 @@ def verify_user(func):
     @functools.wraps(func)
     def wrapper(*args, **kwargs):
         try:
-            user = User.verify_auth_token(request.form.get('token', False))
+            user = User.verify_auth_token(request.values.get('token', False))
         except TypeError:
             logging.debug("no token in request form")
             abort(404)
@@ -242,7 +242,7 @@ def get_auth_token():
 
 @APP.route('/test_authorized')
 def authorized():
-    token = request.form['token']
+    token = request.values['token']
     user = User.verify_auth_token(token)
     if user is not None:
         if mscolab_settings.USER_VERIFICATION:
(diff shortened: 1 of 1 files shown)


ReimarBauer (changes requested):
please undo thes also to the new syntax
https://github.com/Open-MSS/MSS/commit/9a873bb0c5dadb2ac231c90c182c0c401056670f

create_project etc.

Marilyth:
@ReimarBauer you want `request.form` of the POST requests to be converted to `request.values`?

ReimarBauer:
> @ReimarBauer you want `request.form` of the POST requests to be converted to `request.values`?

In the past we had:
```
   path = request.values['path']
   description = request.values['description']
   content = request.values.get('content', None)
```

I changed this to
```
    path = request.form['path']
    content = request.form.get('content', None)
    description = request.form.get('description', None)
```
I'm currently undecided if I changed the behaviour or it was in the past wrong, what is your opinion?

Marilyth:
Both behaviours are fine. It depends on if you even want query parameters to work for POST requests.
Sometimes they may contain sensitive information, sometimes the data would be too long for query parameters, etc.

Personally I would keep it the way it currently is. Forcing all information to be in the body for POST requests isn't bad, that's what POST requests are for.

ReimarBauer (approved):
Thx

PR #1223, opened 2021-09-22
Made code more robust for different projections.

### LINKED ISSUE #246: Different relations between map pixels and km not treated correctly
**[Original report](https://bitbucket.org/wxmetvis/mss/issue/246) by Joern Ungermann (Bitbucket: [joernu76](https://bitbucket.org/joernu76), GitHub: [joernu76](https://github.com/joernu76)).**

Depending on the map, the map coordinates and thus the relation between pixels and km vary. 
Cylindrical projection uses longitude/latitude as unit, whereas stereographic projections use meters.

This distinctions is important and several places in the code check for "lcc" and "stere" to treat it differently. This is not consistently used and the chance of< introducing errors increases drastically.

We should review the code for places where the projection is relevant for the behaviour and maybe encapsulate this.

We should add checks to the code where the distinction is important that the map type is known and covered.

See issue #245 for an example of the code not being careful about map projections.

joernu76:
Fix #246


mslib/msui/mpl_pathinteractor.py
```diff
@@ -975,11 +975,9 @@ def appropriate_epsilon_km(self, px=5):
         # (bounds = left, bottom, width, height)
         ax_bounds = self.ax.bbox.bounds
         diagonal = math.hypot(round(ax_bounds[2]), round(ax_bounds[3]))
-        if self.map.projection in ['stere', 'lcc']:
-            map_delta = np.hypot(self.map.llcrnry - self.map.urcrnry, self.map.llcrnrx - self.map.urcrnrx) / 1000.
-        else:
-            map_delta = get_distance((self.map.llcrnry, self.map.llcrnrx), (self.map.urcrnry, self.map.urcrnrx))
+        map_delta = get_distance((self.map.llcrnrlat, self.map.llcrnrlon), (self.map.urcrnrlat, self.map.urcrnrlon))
```
ReimarBauer, line 978: having for all projection the same method is great,

ReimarBauer (approved):
Thx

PR #1224, opened 2021-09-22
Restrict skyfield data access to preloaded data.

### LINKED ISSUE #1133: mock data from external ressource in our tests
We have to mock all data fetched from external ressources in our tests.

Currently we can easily have blocked tests when we run in a surge protection by some data provider. This can result in a blocked IP.
Also we should take care for not too much requests on these providers just for tests.

joernu76:
See #1133


ReimarBauer (approved):
Thx

PR #1231, opened 2021-09-23
Warnings and bug fixes.

### LINKED ISSUE #1226: Lots of unhandled warnings during test execution
Several of those need to be adressed. The most noise generates this
```
 mslib/msui/_tests/test_mscolab_merge_waypoints.py::Test_Save_Keep_Server_Points::test_save_keep_server_points
  /__w/MSS/MSS/conftest.py:191: UserWarning: An unhandled message box popped up during your test!
  PyQt5.QtWidgets.QMessageBox.critical(): [call(<mslib.msui.mss_pyui.MSSMainWindow object at 0x7f2fcc0e8430>, 'file io plugin error import plugins', "ERROR: Configuration\n\n{'csv': <function load_from_csv at 0x7f300de6f280>}\n\nthrows <class 'ValueError'> error:\n'actionImportFlightTrackcsv' has already been set!"), call(<mslib.msui.mss_pyui.MSSMainWindow object at 0x7f2fcc0e8430>, 'file io plugin error import plugins', "ERROR: Configuration\n\n{'csv': <function save_to_csv at 0x7f300de6f1f0>}\n\nthrows <class 'ValueError'> error:\n'actionExportFlightTrackcsv' has already been set!")]
    warnings.warn(f"An unhandled message box popped up during your test!\n{summary}")
```
which happens in many test cases.

joernu76:
Fix #1226

### DIFF
--- a/mslib/msui/_tests/test_mscolab.py
+++ b/mslib/msui/_tests/test_mscolab.py
@@ -378,8 +378,8 @@ class Test_Mscolab(object):
         self._activate_operation_at_index(1)
         assert self.window.mscolab.active_operation_name == "reproduce-test"
 
-    @mock.patch("mslib.msui.mscolab.QtWidgets.QMessageBox")
-    @mock.patch("mslib.msui.mscolab.QtWidgets.QInputDialog.getText", return_value=("flight7", True))
+    @mock.patch("PyQt5.QtWidgets.QMessageBox.information")
+    @mock.patch("PyQt5.QtWidgets.QInputDialog.getText", return_value=("flight7", True))
     def test_handle_delete_operation(self, mocktext, mockbox):
         # pytest.skip('needs a review for the delete button pressed. Seems to delete a None operation')
         self._connect_to_mscolab()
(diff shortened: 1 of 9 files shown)


mslib/mswms/mpl_hsec_styles.py
```diff
@@ -180,6 +180,11 @@ def _plot_style(self):
         mslp = data['air_pressure_at_sea_level']
 
         # Colors in python2.6/site-packages/matplotlib/colors.py
```
ReimarBauer: the comment is likly old

and the prints will make trouble on the server

ReimarBauer (changes requested):
see comments

mslib/mswms/mpl_hsec_styles.py
joernu76: prints were an oversight. removed

mslib/msui/_tests/test_mscolab.py
```diff
@@ -398,6 +398,9 @@ def test_handle_delete_operation(self, mocktext, mockbox):
         QtWidgets.QApplication.processEvents()
         op_id = self.window.mscolab.get_recent_op_id()
         assert op_id is None
+        QtWidgets.QApplication.processEvents()
+        QtTest.QTest.qWait(0)
```
ReimarBauer, line 402: this is 0 ms. Does it makes a difference, without?

ReimarBauer (approved):
not sure if the qWait with 0ms does something, we can also review all these later

joernu76:
> not sure if the qWait with 0ms does something, we can also review all these later

It shouldn't do anything, but it does. There is an interaction between May catch-all-popup thingie and QT. With the qWait, the window gets destroyed/closed as it should. Without it stays around (maybe invisible). Even a qWait(0) has a quite different effect compared to a processEvents. Argh.

PR #1239, opened 2021-09-30
I1234, multiple popups, your connection is expired

### LINKED ISSUE #1234: multiple popups, your connection is expired
sometimes there are more than three popups are pushed to me when I expire the server

### LINKED ISSUE #1240: own user delete did not trigger logout
when a user deletes his account he did not get disconnected, or get disconnected when deletion not works.

this needs to be verified deletion and changed to get disconnected on success.

ReimarBauer:


joernu76 (approved):
Quite a lot of changes. How close are we to release?

ReimarBauer:
> Quite a lot of changes. How close are we to release?

likly next week. There is one UI bug left, I take next. After that tutorials can be worked on. I setup later today a development msc server too. So we can try out what all those changes do.

PR #1268, opened 2021-10-05
added config option to appropriate type

aravindm711:
* added mscolab_skip_verify_user_token to key_value_options type
* added test to catch missing out on adding the option to the appropriate type


ReimarBauer (approved):
thx

PR #1270, opened 2021-10-05
changes in response to always %s syntax (old string formatting) not .…

kawalpreettkaur:
…format() that has the same problem as f'strings for logging

Replacement done with logging.error('Make sure that the FS url "%s" exists' % dir)


ReimarBauer (changes requested):
please don't open various patch pull requests. 

commit this to the first you did, this will update automatically the pull request, and we have one place to review.

PR #1284, opened 2021-10-18
Show Operation Description Below Connected URL

### LINKED ISSUE #1281: show operation decription below connected url
A text field with the operation description have to be shown below the "Connect to" line for a selected operation.

### LINKED ISSUE #1286: extend tooltip on "operation list" by "doubleclick line"
The doubleclick tooltip is only shown on the left side. The line should be added to the Operations list too

doubleclick_tooltip

Jatin2020-24:
Fixes: #1281 
Fixes: #1286

### DIFF
--- a/mslib/msui/mscolab.py
+++ b/mslib/msui/mscolab.py
@@ -1262,8 +1262,10 @@ class MSSMscolab(QtCore.QObject):
             self.waypoints_model = None
 
             # Show operation description
-            self.error_dialog = QtWidgets.QErrorMessage()
-            self.ui.actionDescription.triggered.connect(lambda: self.error_dialog.showMessage(f"{self.active_operation_desc}"))
+            # ToDo review closing of widget
+            self.ui.actionDescription.triggered.connect(lambda: QtWidgets.QMessageBox.information(None,
+                                                                                                  "Operation Description",
+                                                                                                  f"{self.active_operation_desc}"))
 
             # Set active operation description
             desc_count = len(str(self.active_operation_desc))
--- a/mslib/msui/mscolab.py
+++ b/mslib/msui/mscolab.py
@@ -1263,8 +1263,7 @@ class MSSMscolab(QtCore.QObject):
 
             # Show operation description
             self.error_dialog = QtWidgets.QErrorMessage()
-            self.ui.actionDescription.triggered.connect \
-                (lambda: self.error_dialog.showMessage(f"{self.active_operation_desc}"))
+            self.ui.actionDescription.triggered.connect(lambda: self.error_dialog.showMessage(f"{self.active_operation_desc}"))
 
             # Set active operation description
             desc_count = len(str(self.active_operation_desc))
--- a/mslib/msui/mscolab.py
+++ b/mslib/msui/mscolab.py
@@ -374,9 +374,11 @@ class MSSMscolab(QtCore.QObject):
         self.hide_operation_options()
         self.ui.activeOperationDesc.setHidden(True)
 
-        # hide operation description for flight tracks and open views
-        self.ui.listFlightTracks.itemDoubleClicked.connect(lambda: self.ui.activeOperationDesc.setHidden(True))
-        self.ui.listViews.itemDoubleClicked.connect(lambda: self.ui.activeOperationDesc.setHidden(True))
+        # reset operation description label for flight tracks and open views
+        self.ui.listFlightTracks.itemDoubleClicked.connect(
+            lambda: self.ui.activeOperationDesc.setText("Select Operation to View Description."))
+        self.ui.listViews.itemDoubleClicked.connect(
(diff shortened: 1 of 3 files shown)


mslib/msui/qt5/ui_mainwindow.py
```diff
@@ -287,3 +292,13 @@ def retranslateUi(self, MSSMainWindow):
         self.actionSearch.setText(_translate("MSSMainWindow", "Search"))
         self.actionSearch.setToolTip(_translate("MSSMainWindow", "Search for interactive text in the UI"))
         self.actionSearch.setShortcut(_translate("MSSMainWindow", "Ctrl+F"))
+
```
ReimarBauer, line 300: I guess this is created by the -x flag of pyuic5?
We don't need this extra code to test and display the class.

Jatin2020-24: Removed extra code.

mslib/msui/qt5/ui_mainwindow.py
```diff
@@ -241,10 +249,7 @@ def retranslateUi(self, MSSMainWindow):
         self.filterCategoryCb.setWhatsThis(_translate("MSSMainWindow", "filter by operation category"))
         self.filterCategoryCb.setCurrentText(_translate("MSSMainWindow", "ANY"))
         self.filterCategoryCb.setItemText(0, _translate("MSSMainWindow", "ANY"))
-        self.openFlightTracksLabel.setText(_translate("MSSMainWindow", "Flight Tracks:"))
-        self.listFlightTracks.setToolTip(_translate("MSSMainWindow", "List of open flight tracks.\n"
-"Double-click a flight track to activate it.\n"
-"Save a flight track to name it."))
+        self.activeOperationDesc.setText(_translate("MSSMainWindow", "Select Operation to View Description."))
```
ReimarBauer, line 256: @Jatin2020-24 I tried to find where this text is shown. When and where is it?

ReimarBauer (approved):
see a question in the comment.

I have looked on the last used descriptions of the CIRRUS campaign. All used fits to the line.
We can change this when someone has a usecase for much more words.

mslib/msui/qt5/ui_mainwindow.py
Jatin2020-24: if we add command `self.ui.activeOperationDesc.setHidden(False)` in `after_login` function then this text will be visible.
It will be good idea to show this text immediately after login, so user can easily understand how to see operation description.
InkedScreenshot (318)_LI

ReimarBauer: that's a good idea, please do so.

ReimarBauer (changes requested):
a small change as suggested :)

ReimarBauer:
(I tried there the feature)

found another problem.

with logout the description must be cleared. Currently after the next login without selection the old text is shown again

oldinfo

Jatin2020-24:
now long descriptions (more than 95 chars) can be viewed from operations menu and for short desc. (less than 95 chars) can be viewed either from MSColab main window or in operation menu.
Tooltip has also been extended by "double-click" line.
new

ReimarBauer (commented):
We have to take care on users which are stressed and so I think we should write 

Double click a operation to activate and view its description.

That is usually what a user won't know, how to get it activated.

Then it is "viewing description" an addon to the tooltip which is given from "Flight Tracks"

ReimarBauer:
Somehow I have the same problem as you showed in the gif too, it does not react on my first "ok" click. Also a click on the corner "x" does not destroy it on first attempt. 

interestingly sometimes it works perfect.

mslib/msui/mscolab.py
```diff
@@ -1243,8 +1258,21 @@ def set_active_op_id(self, item):
             self.active_op_id = item.op_id
             self.access_level = item.access_level
             self.active_operation_name = item.operation_path
+            self.active_operation_desc = item.active_operation_desc
             self.waypoints_model = None
 
+            # Show operation description
+            self.error_dialog = QtWidgets.QErrorMessage()
```
ReimarBauer: in this case it is no "error" likly information, or description.

please have a look if we can use too:
QtWidgets.QMessageBox.information

ReimarBauer (changes requested):
see comments.

if the closing of the window did not work as expected (many times of ok clicking) and there is no solution yet please add a # ToDo review closing of widget to that definition.

mslib/msui/mscolab.py
Jatin2020-24: changed error diaog to QMessage.information.

But, it is also not closing on single click.

ReimarBauer: I created another issue to review this. It can be also closing and immeaditly reopen again.
But that is independent from this feature.

ReimarBauer (approved):
Thx

Issue #1287, opened 2021-10-21
upload of avatar image for MSC 


ReimarBauer:
currently we can use a remote url with an image from gravatar after enabled to be used instead of the letter symbol

We want to have an option to replace the gravatar url by an uploaded image.
The upload feature has to become added to the user profile 

The image has to be stored as 200x200 px and served as static file. It should be used on the main interface and the chat. (currently using in the chat is not implemented)

sratslla:
I am working on this issue!

ReimarBauer:
thx :)

PR #1292, opened 2021-11-10
Profile window update

aravindm711:
* Simplified gravatar implementation to directly retrieve png format of the gravatar image
* Added comments

### DIFF
--- a/mslib/msui/mscolab.py
+++ b/mslib/msui/mscolab.py
@@ -537,31 +537,34 @@ class MSSMscolab(QtCore.QObject):
     def fetch_gravatar(self, refresh=False):
         email_hash = hashlib.md5(bytes(self.email.encode('utf-8')).lower()).hexdigest()
         email_in_config = self.email in config_loader(dataset="gravatar_ids")
-        gravatar_path = os.path.join(constants.MSS_CONFIG_PATH, 'gravatars')
-        gravatar = os.path.join(gravatar_path, f"{email_hash}.png")
+        gravatar_img_path = fs.path.join(constants.GRAVATAR_DIR_PATH, f"{email_hash}.png")
+        config_fs = fs.open_fs(constants.MSS_CONFIG_PATH)
 
         # refresh is used to fetch new gravatar associated with the email
         if refresh or email_in_config:
             # create directory to store cached gravatar images
-            if not os.path.exists(gravatar_path):
+            if not config_fs.exists("gravatars"):
                 try:
-                    os.makedirs(gravatar_path)
-                except Exception as e:
-                    logging.debug("Error %s", str(e))
-                    show_popup(self.prof_diag, "Error", "Could not create gravatar folder in config folder")
+                    config_fs.makedirs("gravatars")
+                except fs.errors.CreateFailed:
+                    logging.error('Creation of gravatar directory failed')
+                    return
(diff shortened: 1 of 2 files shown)


mslib/msui/mscolab.py
```diff
@@ -536,11 +536,13 @@ def after_login(self, emailid, url, r):
...
         gravatar_path = os.path.join(constants.MSS_CONFIG_PATH, 'gravatars')
         gravatar = os.path.join(gravatar_path, f"{email_hash}.png")
 
-        if refresh or email_in_settings:
+        # refresh is used to fetch new gravatar associated with the email
+        if refresh or email_in_config:
+            # create directory to store cached gravatar images
             if not os.path.exists(gravatar_path):
                 try:
                     os.makedirs(gravatar_path)
```
ReimarBauer: we should use here **fs** to create and test the directory. On running tests this currently creates a directory outside of tempfs. 
On my linux system it creates various dirs similiar to  osfs:/tmp/tmp9y4tzpcemssf8b90094fc in the MSS folder because os.makedirs does not understand fs syntax.

mslib/msui/mscolab.py
```diff
@@ -549,17 +551,17 @@ def fetch_gravatar(self, refresh=False):
...
-            if not refresh and email_in_settings and os.path.exists(gravatar):
+            # use cached image if refresh not requested
+            if not refresh and email_in_config and os.path.exists(gravatar):
                 self.set_gravatar(gravatar)
+                return
 
-            gravatar = os.path.join(gravatar_path, f"{email_hash}.jpg")
-            gravatar_url = f"https://www.gravatar.com/avatar/{email_hash}?s=80&d=404"
+            # fetch gravatar image
+            gravatar_url = f"https://www.gravatar.com/avatar/{email_hash}.png?s=80&d=404"
```
ReimarBauer, line 563: it would be good to make the url configurabel and also to add an option to the UI to upload an image. 
ideas can be discussed on
https://github.com/Open-MSS/MSS/issues/1287

ReimarBauer (changes requested):
please review for further os modules used and replace them by the fs module

mslib/msui/mscolab.py
aravindm711: I tried using fs.open_fs for creating the 'gravatars' folder, but it doesn't seem to work. How else can I create a dir using fs?

mslib/msui/mscolab.py
aravindm711: For this PR, I just wanted to focus on simplifying the retrieval of the gravatar images. I'll discuss ideas on the issue thread if I get any.

mslib/msui/mscolab.py
ReimarBauer: e.g.
example uses fs but not solves the problem ;) just using home also for tests
```
home_fs = fs.open_fs("~/")
if not home_fs.exists("gravatars"):
    home_fs.makedirs("gravatars")
```
look at the msui.constants. maybe the path can be created there and then used dependent on the MSS_SETTINGS env var

mslib/msui/mscolab.py
ReimarBauer: sure :) that's why it is a new issue.

ReimarBauer (approved):
Thx

PR #1296, opened 2021-11-22
Config Editor update

aravindm711:
* Displaying invalid values in config through messagebox, on startup.


swsrkty:
Hi @aravindm711, are you still working on it? I would be happy to look into it. : )

ReimarBauer:
We can reopen this when we continue

PR #1297, opened 2021-11-24
defintions for singularity added

ReimarBauer:


singularity/Singularity.def
```diff
@@ -0,0 +1,51 @@
...
+
+# get keyboard working for mss gui
+apt-get --yes update && DEBIAN_FRONTEND=noninteractive \
+apt-get --yes install xserver-xorg-video-dummy \
+&& apt-get --yes upgrade \
+&& apt-get clean \
+&& rm -rf /var/lib/apt/lists/*
+
+# Set up conda-forge channel
+conda config --add channels conda-forge && conda config --add channels defaults &&\
```
joernu76, line 34: doesnt that put defaults on top of forge, and didn't we stop doing that?

singularity/Singularity.def
```diff
@@ -0,0 +1,51 @@
...
+
+
+# create some desktop user directories
+# if there is no data attached e.g. demodata /srv/mss is the preferred dir
+mkdir -p /root/.local/share/applications/ \
+&& mkdir -p /root/.local/share/icons/hicolor/48x48/apps/ \
+&& mkdir /srv/mss
+
+# Install Mission Support System Software
+mamba create -n mssenv mss xorg-libxcb libxcb xorg-libxrender -y
```
joernu76, line 45: mmh. Would install the x11 stuff via apt-get. Dependencies should be much more reliable.

singularity/Singularity.def
ReimarBauer: ah, good catch.

singularity/Singularity.def
ReimarBauer: I do verify if I need all three. But I assume that we need at least the xcb stuff in the same path as the QT libs.

singularity/Singularity.def
ReimarBauer: no it doesn't it appends, only when it is in the list already it would go on top.
Because of mamba-forge, we have already conda-forge set and that is not needed.

But we need defaults for the next command. No better solution yet.

ReimarBauer:
moved to
https://github.com/Open-MSS/singularity/pull/1

Issue #613, opened 2021-01-04
Mamba problem on installing mss


Marilyth:
Hey there!

I tried following the instructions for installing MSS in the README, namely:
```
$ conda config --add channels conda-forge
$ conda config --add channels defaults
$ conda create -n mssenv python=3
$ conda activate mssenv
$ conda install mamba
$ mamba install mss
```
Everything works fine until the last step, `mamba install mss`.
I tried doing it on a relatively new Linux system, as well as on Windows 10.
Both receive the following output, and error:
```
Looking for: ['mss']

pkgs/msys2/win-64        [====================] (00m:00s) Done
pkgs/main/noarch         [====================] (00m:00s) Done
pkgs/r/noarch            [====================] (00m:00s) Done
pkgs/msys2/noarch        [====================] (00m:00s) Done
pkgs/r/win-64            [====================] (00m:00s) Done
pkgs/main/win-64         [====================] (00m:01s) Done
conda-forge/noarch       [====================] (00m:01s) Done
conda-forge/win-64       [====================] (00m:04s) Done
Encountered problems while solving.
Problem: nothing provides basemap 1.0.7 needed by mss-1.2.1-py27_0
```
I guess there is a reason the old instruction was changed, but using `conda install mss` instead seemed to install and work fine.

ReimarBauer:
that's interesting. it seems to try to install a py27 version mss-1.2.1-py27 
The reason may an old catalogue. Which version of anaconda was used?

Marilyth:
I installed anaconda fresh on both systems today, I checked a few versions which might be relevant:

mamba 0.7.6
conda 4.9.2
anaconda Command line client (version 1.7.2)

Nothing was done with anaconda other than the few lines provided in the installation instructions.
Something, somewhere must be outdated then, I think.

ReimarBauer:
reproduced

we have to use for mamba
`conda create -n mssenv python=3.8`

mamba seems not to be able to decrease the python version. And currently we have not a mss build for python=3.9

Marilyth:
I can confirm specifying python=3.8 has solved the issue for me.

wolfv:
Hi @Marilyth thanks for trying mamba. 

Indeed, the error message is slightly confusing.

The problem is that conda & mamba handle Python in a special way: the minor version is pinned, so that mamba will always keep python 3.9 unless explicitly asked to change the python version.

You can work around this by either specifying `mamba install mss python=3.8`.

[...]

PR #1298, opened 2021-11-27
Fix mamba installation

swsrkty:
Fixes mamba failing to install MSS. 

While installing MSS, the following error appears:
```
Encountered problems while solving.
Problem: nothing provides basemap 1.0.7 needed by mss-1.2.1-py27_0
```
This occurs because Mamba keeps python 3.10 unless explicitly mentioned, and currently we don't have a python 3.10 release.
Related: #613

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,8 +33,9 @@ Installing `MSS` from the `conda-forge` channel can be achieved by adding `conda
 
 Once the `conda-forge` channel has been enabled, `mss` can be installed with:
 
-    $ conda create -n mssenv mamba
+    $ conda create -n mssenv python=3.9
     $ conda activate mssenv
+    $ conda install mamba
     $ mamba install mss
 
 It is possible to list all versions of `mss` available on your platform with:


README.md
```diff
@@ -33,8 +33,9 @@ Installing `MSS` from the `conda-forge` channel can be achieved by adding `conda
 
 Once the `conda-forge` channel has been enabled, `mss` can be installed with:
 
-    $ conda create -n mssenv mamba
+    $ conda create -n mssenv python=3.9
```
ReimarBauer, line 36: please remove =3.9

seems there are more occurencies of this problem interpreting a version string as float number ;)

adding the python modul will release automatically the fixation and we will not need to adjust when we can use 3.10 again

README.md
```diff
@@ -33,8 +33,9 @@ Installing `MSS` from the `conda-forge` channel can be achieved by adding `conda
...
     $ conda activate mssenv
+    $ conda install mamba
```
ReimarBauer, line 38: this depends but will not cause trouble if you are using mambaforge

ReimarBauer:
> Fixes mamba failing to install MSS.
> 
> While installing MSS, the following error appears:
> 
> ```
> Encountered problems while solving.
> Problem: nothing provides basemap 1.0.7 needed by mss-1.2.1-py27_0
> ```
> 
> This occurs because Mamba keeps python 3.10 unless explicitly mentioned, and currently we don't have a python 3.10 release. Related: #613

not exactly but close ;)

someone implemented to use float comparison instead of version string comparison. It searches for python 3.1 not "3.10"

ReimarBauer (changes requested):
please send a PR to stable.  but bugreport first, we like to document before.

I will later merge stable to development.

README.md
ReimarBauer: and if "conda create -n mssenv mamba python" works we should do that

swsrkty:
Hi @ReimarBauer , 
I tried removing =3.9 and I also tried `conda create -n mssenv mamba python`, and neither of them worked.
When I ran the following, I didn't get any errors related to Python 3.1:
```
conda create -n mssenv python
conda activate mssenv
conda install mamba
```

Can you tell me which conda version are you using? Mine is 4.11.0. I think the error depends on the conda version.
I  found out a bug on conda: https://github.com/conda/conda/issues/10969, which could be responsibe for the error, but I didn't get that error.

I also wanted to ask: Isn't it good to specify the Python version while creating the conda environment? Related: https://github.com/mamba-org/mamba/issues/1245.

I'll open the bug report issue and open the new PR once we come to a conclusion on the issue here. Is that alright?
Thank you! : )

ReimarBauer:
We should always start with an issue before fixing something. Because all discussions are now hidden in the PR after it is merged/closed and it is quite difficult to reread or search for this.

When an issue gets added I could have labled it by 6.0.1 so it is clear that we want the fix in stable.

We link issues in the releases so everyone can find the PR.

A first issue could have been just the description of the problem. The PR afterwards brings the solution ;)

ReimarBauer:
I do also have the recent version and I tried the instructions from
https://mss.readthedocs.io/en/stable/installation.html#install

The version string is added on build time of the documentation. May be this can be done by an variable similiar in the README?
```
$ conda create -n mssenv mamba
$ conda activate mssenv
(mssenv) $ mamba install mss=6.0.0 python
(mssenv) $ mss
```

ReimarBauer:
We do build MSS for many python versions https://anaconda.org/conda-forge/mss/files

The fixation of the (most recent) python version is currently not mandatory for using MSS. Updating packages to a newer python version sometimes eats time. There can be an OS user which can't use a specified version of python because of some dependencies are not solved. 

We came to the solution  to force the version of MSS and let the system solve based on the fixations in the meta.yaml what the user can get.

swsrkty:
>We should always start with an issue before fixing something. Because all discussions are now hidden in the PR after it is merged/closed and it is quite difficult to reread or search for this.
When an issue gets added I could have labled it by 6.0.1 so it is clear that we want the fix in stable.
We link issues in the releases so everyone can find the PR.
A first issue could have been just the description of the problem. The PR afterwards brings the solution ;)

Ah, got it. Thank you for explaining.
I've opened an issue and a PR #1301 for the same. Pls review it.

PR #1294, opened 2021-11-19
module added for websocket and polling for windows

ReimarBauer:
seems login on windows needs the package installed. 

tested on linux, seems not to beak tests

PR #1301, opened 2021-12-06
Fix mamba failing to install MSS

### LINKED ISSUE #1299: mamba failing to install MSS through instructions in README
Mamba failing to install MSS.
While following the instructions for installing MSS in README, namely 
```
conda create -n mssenv mamba
$ conda activate mssenv
(mssenv) $ mamba install mss
```

the following error appears:
```
Encountered problems while solving.
Problem: nothing provides basemap 1.0.7 needed by mss-1.2.1-py27_0
```
Related: #613

swsrkty:
Fixes #1299
Discussion: #1298

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,7 +35,7 @@ Once the `conda-forge` channel has been enabled, `mss` can be installed with:
 
     $ conda create -n mssenv mamba
     $ conda activate mssenv
-    (mssenv) $ mamba install mss=6.0.0 python
+    (mssenv) $ mamba install mss python
     (mssenv) $ mss
 
 It is possible to list all versions of `mss` available on your platform with:
--- a/README.md
+++ b/README.md
@@ -35,7 +35,8 @@ Once the `conda-forge` channel has been enabled, `mss` can be installed with:
 
     $ conda create -n mssenv mamba
     $ conda activate mssenv
-    $ mamba install mss
+    (mssenv) $ mamba install mss=6.0.0 python
+    (mssenv) $ mss
 
 It is possible to list all versions of `mss` available on your platform with:
 


README.md
```diff
@@ -35,7 +35,8 @@ Once the `conda-forge` channel has been enabled, `mss` can be installed with:
 
     $ conda create -n mssenv mamba
     $ conda activate mssenv
-    $ mamba install mss
+    (mssenv) $ mamba install mss=6.0.0 python
+    (mssenv) $ mss
```
ReimarBauer, line 39: instead of the hardcoded 6.0.0 I would like to use the recent tag of the branch, But I've no idea if this can be filled in. If there is a var which we can use.

on command line this can be done by e.g.
https://newbedev.com/how-to-get-the-latest-tag-name-in-current-branch-in-git

If we commit a hardcoded version string this file needs on any further release a change too.

Marilyth: Doesn't `mamba install mss python` suffice?
Once the python fixation is gone, it should look for the appropriate version by itself.

That is also how our installer works
https://github.com/Open-MSS/mss-install/blob/main/LinuxMac.sh#L65

ReimarBauer: for the test environment I needed to add always python only for any new single package added
https://github.com/Open-MSS/MSS/pull/1294/files#diff-7dc87d4394e1756c519dbfd0b80d3b31377f643f0bc25d3ed807ce8a3794023d

ReimarBauer: on windows the install script fetches now the latest version, so no need to change this.

ReimarBauer: `mamba install mss python` works, We should not force the version number in the README here.

ReimarBauer (changes requested):
see discussions, comments

README.md
swsrkty: > 

Ah interesting, so it works because mss finds the appropriate version by itself and we don't need to specify it.

swsrkty:
> mamba install mss python works, We should not force the version number in the README here.

I've checked and removed the version from the instructions.

Marilyth (approved):
Should work this way.

ReimarBauer (approved):

PR #1302, opened 2021-12-07
Switch to OpenAIP google cloud storage

Marilyth:
OpenAIP now has a storage bucket through which airspace files can be downloaded.
The google cloud storage library and the OpenAIP API require authentication, so the google cloud storage API is used instead.

### DIFF
--- a/mslib/utils/airdata.py
+++ b/mslib/utils/airdata.py
@@ -42,24 +42,21 @@ _airspaces = []
 _airports = []
 _airports_mtime = 0
 _airspaces_mtime = {}
-_airspace_url = "https://www.openaip.net/customer_export_akfshb9237tgwiuvb4tgiwbf"
-# Updated Jul 27 2021
+_airspace_url = "https://storage.googleapis.com/29f98e10-a489-4c82-ae5e-489dbcd4912f"
+_airspace_download_url = "https://storage.googleapis.com/storage/v1/b/29f98e10-a489-4c82-ae5e-489dbcd4912f/o/" \
+                         "{}_asp.xml?alt=media"
+# Updated Dec 07 2021
 _airspace_cache = \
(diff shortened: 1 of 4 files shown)


ReimarBauer (changes requested):
I have a problem to use the downloaded data, see

download_and_use_problem

also clicking the Update/Download button downloads files and then by apply I get the error message

```
ERROR: Tried to draw airspaces without asp files.
INFO: Downloading to /home/reimar/.config/mss/airports.csv. This might take a while.
INFO: 3152KB Downloaded
INFO: 6542KB Downloaded
INFO: 9823KB Downloaded
ERROR: Tried to draw airspaces without asp files
```

Marilyth:
You selected airports and airspaces to be drawn, but selected none to be drawn in the combobox. Which caused this error.
I suppose it would make sense to disable the checkbox in those cases.

Marilyth:
The button/checkbox now gets enabled/disabled by itself, depending on the current situation with the combobox.
This should prevent the user from applying empty airports or airspaces to begin with.

mslib/utils/airdata.py
```diff
@@ -145,7 +141,7 @@ def update_airspace(force_download=False, countries=["de"]):
         if (force_download or is_outdated or not file_exists) \
                 and QtWidgets.QMessageBox.question(
                     None, "Allow download",
-                    f"The selected {country} airspace needs to be downloaded ({data[-1]})"
+                    f"The selected {country} airspace needs to be downloaded ({round(int(data[-1]) / (1024**2), 3)}MB)"
```
ReimarBauer: have a look on this ;) 
```
import humanfriendly
humanfriendly.format_size(NUMBER)
```
more examples https://pypi.org/project/humanfriendly/#description

we have this in our stack.

ReimarBauer (changes requested):
a small change for using humanfriendly.

mslib/utils/airdata.py
Marilyth: Done

ReimarBauer (approved):
Thx

PR #1303, opened 2021-12-07
Fix operation description not closing

### LINKED ISSUE #1288: review usage of QtWidgets.QErrorMessage() and behaviour on "ok" click
while working on 
https://github.com/Open-MSS/MSS/pull/1284
see ToDo in mscolab.py

we recognized that the widget to show the description does not close on first click of the OK button.

we have to review if that happens more often, find a solution or report a bug to QT

we have also to review the events on that place.

Marilyth:
Fixes #1288 
Every time the active operation changes, actionDescription connects to a new slot, but is never disconnected.
This causes the description window to open multiple times in a row if the active operation ever changes.

A simple fix is to just move this part of the code to the constructor.


ReimarBauer (approved):
Thx :)

PR #1306, opened 2021-12-17
Some fixes for plotting on the dateline

### LINKED ISSUE #1304: Flight paths crossing the date line (longitude=180)
When moving the top view map towards a shifted map that reaches over the date line (longitude=180), it is possible that the longitude registred of a way point gets below -180. A flight path whith this coordinate is correclty displayed, however the resulting side view panel is not.
The following plots show the temperature curtain for the 2 waypoints (lat:-40 lon -150) and (lat: -40, lon:130) and the same when the second lon is replaced by -230. Although the values below -180 are considered as not in the valid range in the table view, these way points can be created by clicking on the top view. For the such created invalid logitudes the sideview shows no data.
sideview_Temperature1
sideview_Temperature2

This is important for missions near the date line e.g. in the Pacific

joernu76:
* Vertical and linear plots remap the longitude such that
  interpolation works correctly
* For horizontal maps, a column with NaN is added to prevent
  interpolation.

### DIFF
--- a/mslib/mswms/mpl_hsec.py
+++ b/mslib/mswms/mpl_hsec.py
@@ -448,6 +448,16 @@ class MPLBasemapHorizontalSectionStyle(AbstractHorizontalSectionStyle):
         left_longitude = min(self.bm.llcrnrlon, ulcrnrlon)
         logging.debug("shifting data grid to leftmost longitude in map %2.f..", left_longitude)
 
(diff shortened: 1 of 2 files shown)


ReimarBauer (commented):
looks good to me currently verifying on our dataset

ReimarBauer (approved):
Thx

PR #1307, opened 2021-12-20
added extensions key as namespace for functions

### LINKED ISSUE #1305: Can only use one 'export_plugins' and 'import_plugins' entry.
Greetings from Boulder!,

In version 6.0.0 of the software when I put more than one plugin into "export_plugins": {} in the mss_settings.json file, it will always load the first plugin perfectly fine regardless of which one of my plugins I choose first, but every other one in the list after that does not install when starting up the MSS planner.  I am getting the following message box:

ERROR: Configuration
{'csv': <function load_from_csv at 0x7fc5d8571b80>, 'txt': <function import_foreflight at 0x7fc5e1b31310>}
throws <class 'ValueError'> error:
'actionImportFlightTracktxt' has already been set!

I get this message for each of my plugins (minus the first one listed in "export_plugins":{}), then the program comes all the way up, but only with the first plugin loaded.  The same happens for my "import_plugins":{} entries.

Thanks!

-Shawn Honomichl

ReimarBauer:


mslib/msui/_tests/test_mss_pyui.py
```diff
@@ -118,6 +118,15 @@ class Test_MSSSideViewWindow(object):
...
         # "GPX": ["gpx", "mslib.plugins.io.gpx", "save_to_gpx"]
     }
+    actions = {
+        "actionImportFlightTrackFTML": "ftml",
+        "actionImportFlightTrackCSV": "csv",
+        "actionImportFlightTrackText": "txt",
+        "actionImportFlightTrackFliteStar": "fls",
+        "actionExportFlightTrackFTML": "ftml",
+        "actionExportFlightTrackCSV": "csv",
+        "actionExportFlightTrackText": "Text",
```
joernu76: Import&Export Text has inconsistent ending.

mslib/msui/_tests/test_mss_pyui.py
```diff
@@ -107,7 +107,7 @@ class Test_MSSSideViewWindow(object):
     save_csv = os.path.join(ROOT_DIR, "example.csv")
     save_ftml = os.path.join(ROOT_DIR, "example.ftml")
     save_ftml = save_ftml.replace('\\', '/')
-    save_txt = os.path.join(ROOT_DIR, "example.txt")
+    save_txt = os.path.join(ROOT_DIR, "example.Text")
```
joernu76: why?

ReimarBauer: this is currently to satisfy tests. Import/Export/SaveAs were unified based on the extension which makes trouble when we have multiple times same exension, e.g. save_as_csv. It did in the past the mapping by the extension and not by the key e.g. DLR had than loaded the other csv method. 

I use now the key as identifier to destiguish plugins. The Key for the builtin .txt extension is Text in our setup.

By this changes we can have multiple times plugins which create csv or others with same extension. Also  on reading a file we will first select the Type and then choose the file. So we get  always the right function used

I recognized than that the mscolab Operation Import Feature needs a UI and API change, because there is the mapping to the type missing. So also all tests for this feature needs a refactoring. select first type than export/import by the prefered extension. 

This was currently a mimum change not to have using the UI for this. Test needs more changes.

joernu76 (approved):

PR #1313, opened 2021-12-21
Rename mscolab operation

### LINKED ISSUE #910: mscolab UI misses a rename Project
We have in mscolab UI not implemented to rename a project. 

update_project can be used, from the file_manager

Jatin2020-24:
Fixes: #910

### DIFF
--- a/mslib/mscolab/server.py
+++ b/mslib/mscolab/server.py
@@ -485,7 +485,6 @@ def update_operation():
     return r
 
 
-
 @APP.route('/operation_details', methods=["GET"])
 @verify_user
 def get_operation_details():
--- a/mslib/mscolab/server.py
+++ b/mslib/mscolab/server.py
@@ -477,12 +477,7 @@ def update_operation():
     attribute = request.form['attribute']
     value = request.form['value']
     user = g.user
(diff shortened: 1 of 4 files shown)


ReimarBauer (changes requested):
I had a first view on it. 
A few things more needs a trigger.

The renamed OP is active, the description should be shown too.  
The OP name is also on the *View Windows
The title of the window needs also to update the name.

For ideas about the latter two see https://github.com/Open-MSS/MSS/pull/1210

newExample

ReimarBauer (changes requested):
For a user per operation that is now correct. Have a look on an other user in the OP having that open while it gets renamed. multiuser has the old name in the list.
We have to send an event over socketio so that the other user interface also gets updated. For an example:


https://github.com/Open-MSS/MSS/pull/1216/files#diff-e56d16fe754ea534464abbcdab7d56ef6752a86f7a83c5f1beeb500db7dc91e5R138

renaming

ReimarBauer (approved):

ReimarBauer:
We have to review on all existing actions the communication state over socketio.

PR #1321, opened 2021-12-30
Further properties for Operations

### LINKED ISSUE #1280: further properties for Operations
A creator role needs the option to rename the operation.

An admin role needs the possibility to change the description and the category of an operation

Jatin2020-24:
Fixes: #1280 

Issues Resolved: 

- Admin can change description and category of operation

### DIFF
--- a/mslib/msui/mscolab.py
+++ b/mslib/msui/mscolab.py
@@ -386,6 +386,7 @@ class MSSMscolab(QtCore.QObject):
         self.ui.actionVersionHistory.triggered.connect(self.operation_options_handler)
         self.ui.actionManageUsers.triggered.connect(self.operation_options_handler)
         self.ui.actionDeleteOperation.triggered.connect(self.operation_options_handler)
+        self.ui.actionUpdateOperationDesc.triggered.connect(self.update_description_handler)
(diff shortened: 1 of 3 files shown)


mslib/msui/mscolab.py
```diff
@@ -930,6 +937,49 @@ def handle_delete_operation(self):
...
+        if desc_count < 95:
+            self.ui.activeOperationDesc.setText(
+                self.ui.tr(f"{self.active_operation_name}: {self.active_operation_desc}"))
+        else:
+            self.ui.activeOperationDesc.setText(
+                "Description is too long to show here, for long descriptions go "
+                "to operations menu.")
+
+    def update_description_handler(self):
+        # only after login
```
ReimarBauer, line 952: comment hint because of the comment.

this must be protected on the server site (if not done)

The verify_user_token as client site function name is a bit misleading. It does lookup if not surpressed that the token of the user on the socket layer is still valiid. This token exists after login, but we can also disable this check. The token gets invalid when it expires on the server. e.g. the server gets restartet. A liftetime is exceeded. ....

Because anyone can create other clients, we have to protect such functions on the server.

joernu76 (approved):

PR #1329, opened 2022-01-13
Rename operation name

### LINKED ISSUE #1328: review communication state over socketio.
We have to review on all existing actions the communication state over socketio.

Jatin2020-24:


Jatin2020-24:
rename_operation

This will show current changes.

ReimarBauer (changes requested):
checked a few more things now. merged too early. sorry. 

we should warn when a name is taken. Not by lookup because this can be used to guess OPs. Just if it not possible to tell "this name is taken" by  QtWidgets.QMessageBox.Warning / Error

exists

ReimarBauer:
The list is updated
start

now we need also to update the image on the particpants opened views (window title and text)
renamed


When a client is triggered to reload_operations, we should check as client if the name changes of the selected operation and update afterwards the open views. If the client ist looking on a different operation or none we don't have to update.

ReimarBauer:
if an OP is selected where the user is not creator the "rename Operation" dialog should become inactive. Currently I can use the action but it doesn't rename (that's correct behaviour)

ReimarBauer (approved):
thx for working on this.

PR #1334, opened 2022-01-18
Make conversion to float instead of int for addition accuracy.

joernu76:


joernu76:
superceded

ReimarBauer (changes requested):
should perhaps go to stable

PR #1336, opened 2022-01-18
Made code robust against no data/nan in data.

### LINKED ISSUE #1335: Server generates error message when displaying section with no data
The error is that NaN cannot be converted to int.
This stems from clabel calls with int formatting and no data.

joernu76:
Fix #1335


ReimarBauer (approved):
thx

PR #1339, opened 2022-01-26
MSCOLAB: making "add user" more user-friendly

### LINKED ISSUE #1244: MSCOLAB: make "add user" user-friendly
In case a user has no log in to the MSCOLAB server, he has to add a user first.

Now:
1) The user creates a user with mail + password
2) The user has to type the login details he just used into the login fields and click the login button

Desired:
1) The user creates a user just as now
2) The user is automatically logged in
Bonus: 
The login details are automatically stored in the config file for future use.

swsrkty:
Fixes #1244


ReimarBauer (changes requested):
tested with open-mss.org

please have a look on the additional part to store the user credentials in the mss_settings.json, so that on a restart of MSS the data is provided. 

it needs to set "MSCOLAB_mailid" and "MSCOLAB_password" there is an API in the config editor which can be used to write only changes to that file

joernu76 (dismissed):

swsrkty:
Moin @ReimarBauer, I have a doubt. Consider the following scenario:

User A registers on MSColab server and the MSCOLAB_mailid and MSCOLAB_password are set in mss_settings.json. Then. when user B registers on the same computer on MSColab server, should the MSCOLAB_mailid and MSCOLAB_password be changed in mss_settings.json?

ReimarBauer:
> Moin @ReimarBauer, I have a doubt. Consider the following scenario:
> 
> User A registers on MSColab server and the MSCOLAB_mailid and MSCOLAB_password are set in mss_settings.json. Then. when user B registers on the same computer on MSColab server, should the MSCOLAB_mailid and MSCOLAB_password be changed in mss_settings.json?

if it is the same value then there is no update needed. If it is a new value we should update it. You could also ask the user. Hi you are using new credentials, should your settings files become updated [Y/N]? Then the user can decide.

We have in a further step in the future to reorganize the settings file. We need to improve grouping and also that we can have more servers.

joernu76:
Should the "INFO" file be in the repo?

swsrkty:
> Should the "INFO" file be in the repo?

There is no need for that file to be in the repo. I created one to check the logs and I accidentally pushed it.

ReimarBauer (changes requested):
currently this did not update the existing mss_settings.json parameters, it writes a new one.

We want to keep the old settings, similiar as when you use the Configuration Editor below the File menu.
e.g. if you have this in your file, it should be after storing the credential also be there.
```
{
    "num_labels": 20,
    "MSCOLAB_category": "EXMPLE"
}
```

and it might be also a problem in the login procedure if the login protection is in place but the account information is not set in the mss_settings.json. This currently toggles the two login forms.

We can use a simple protection to avoid public access, see
https://mss.readthedocs.io/en/stable/mscolab.html#protecting-login

Add a file into the search path, see doc
mss_mscolab_auth.py

The behaviour should be when the auth is not set in the configuration that only this dialog should be shown and its data can also be stored in the config.

some images how it looks now.

new_user1
new_user2
new_user_login3
user_login4

ReimarBauer:
@aravindm711 please help a bit on the updating of config settings.

swsrkty:
> currently this did not update the existing mss_settings.json parameters, it writes a new one.

Thanks for pointing it out! I've fixed it now.

> and it might be also a problem in the login procedure if the login protection is in place but the account information is not set in the mss_settings.json. This currently toggles the two login forms.

Ah, I'll look into it.

aravindm711:
Hey @sr1j , modify_config_file looks good but it only updates the mss_settings.json file, which is not what serves the config options in runtime. It's rather the global variable "user_options", which is updated at the end of read_config_file method. 

I found this by adding a new user, which automatically logs in and updates the settings file, but the configuration editor does not reflect this change.

I guess you could just call the read_config_file method at the end of modify_config_file method to fix this.

swsrkty:
> Hey @sr1j , modify_config_file looks good but it only updates the mss_settings.json file, which is not what serves the config options in runtime. It's rather the global variable "user_options", which is updated at the end of read_config_file method.

Thank you @aravindm711 for pointing it out!

> I guess you could just call the read_config_file method at the end of modify_config_file method to fix this.

Yep, that worked. : )

mslib/msui/mscolab.py
```diff
@@ -276,21 +275,68 @@ def login_handler(self):
...
+                "Failed to establish a new connection" f' to "{self.mscolab_server_url}". Try in a moment again.',
+            )
+            self.disconnect_handler()
+            return
+        data_to_save_in_config_file = {
+            "MSCOLAB_mailid": emailid,
+            "MSCOLAB_password": password
+        }
+        if config_loader(dataset="MSCOLAB_mailid") == "" and config_loader(dataset="MSCOLAB_password") == "":
+            modify_config_file(data_to_save_in_config_file)
```
ReimarBauer: the test runs in an error now

```
2022-02-16 19:25:13 CRITICAL Fatal error: Traceback (most recent call last):
  File "/home/reimar/PycharmProjects/2022/sr1j/MSS/mslib/msui/mscolab.py", line 376, in new_user_handler
    self.new_user_login_handler(emailid, password)
  File "/home/reimar/PycharmProjects/2022/sr1j/MSS/mslib/msui/mscolab.py", line 307, in new_user_login_handler
    modify_config_file(data_to_save_in_config_file)
  File "/home/reimar/PycharmProjects/2022/sr1j/MSS/mslib/utils/config.py", line 409, in modify_config_file
    modified_data = merge_data(json_file_data_copy, data)
  File "/home/reimar/PycharmProjects/2022/sr1j/MSS/mslib/utils/config.py", line 553, in merge_data
    data, match = compare_data(options[key], json_file_data[key])
KeyError: 'MSCOLAB_mailid'
```

ReimarBauer (changes requested):
please have a look on the tests. You can ran these specific tests by

pytest -k test_mscolab.py

The tests covers a lot situations. 
(I care for the skyfield data test)  Currently this should be the only one which fails in a PR.

mslib/msui/mscolab.py
ReimarBauer: the functional tests trying your changes gives for me the wanted result :)

But have also a look on the tests because, we try to protect by tests given functionality.

swsrkty: Moin @ReimarBauer, yep, I was looking into the tests.

ReimarBauer:
We should have a test which reads the changed mss_settings file and checks if the changes are stored.

mslib/utils/config.py
```diff
@@ -464,19 +504,21 @@ def load_settings_qsettings(tag, default_settings=None, ignore_test=False):
     return default_settings
 
 
-def merge_data(options, json_file_data):
+def merge_data(dict_to_merge_new_dict_into, new_dict):
```
swsrkty: I have renamed the parameters of the merge data function as it was a bit confusing. Please let me know what you think.

mslib/msui/_tests/test_mscolab.py
```diff
@@ -113,11 +116,40 @@ def test_login(self):
...
 
+    @mock.patch("PyQt5.QtWidgets.QMessageBox.question", return_value=QtWidgets.QMessageBox.Yes)
+    def test_add_users_with_updating_credentials_in_config_file(self, mockmessage):
+        self._connect_to_mscolab()
+        self._create_user("something", "something@something.org", "something")
+        assert config_loader(dataset="MSCOLAB_mailid") == "something@something.org"
+        assert config_loader(dataset="MSCOLAB_password") == "something"
+        assert self.main_window.usernameLabel.text() == "something"
+        self.main_window.mscolab.logout_action.trigger()
+        QtWidgets.QApplication.processEvents()
```
swsrkty: I was trying to logout but it's not working. Can you please help with this? : )

mslib/utils/config.py
ReimarBauer: the function could also be merge_dict? shorter parameter could be perhaps then given_dict, new_dict or just given, new.

mslib/msui/mscolab.py
ReimarBauer: I'll take a look

docs/samples/config/mscolab/mss_mscolab_auth.py.sample
```diff
@@ -1,3 +1,3 @@
 class mss_mscolab_auth(object):
      password = "please use the methods to save only the encrypted value"
-     allowed_users = [("user", hashlib.md5(password.encode('utf-8')).hexdigest())),]
```
ReimarBauer, line 3: good catch

mslib/utils/config.py
swsrkty: Yes, and how about existing_dict and new_dict as parameters?

ReimarBauer: :)

mslib/msui/_tests/test_mscolab.py
ReimarBauer: I look on this.

swsrkty:
Hi @ReimarBauer, I just found out that the email and password are not being stored in the config file when the email verification is set. I'm looking into it. : )

ReimarBauer (approved):
great! Thx!

PR #1340, opened 2022-01-31
do not run tests twice

joernu76:


joernu76:
@ReimarBauer Do you have an idea, why this does not work? On data-retrieval, this (seems to?) works as expected...

.github/workflows/python-flake8.yml
```diff
@@ -3,7 +3,14 @@
 
 name: flake8
 
-on: [ push, pull_request ]
+on:
+  push:
+    branches:
+    - develop, stable
```
ReimarBauer: from the yaml tutorial this should as list one item per line 
or you need brackets, e.g.  branches: [ develop, stable ]

see 
- https://www.cloudbees.com/blog/yaml-tutorial-everything-you-need-get-started
- https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions

(I guess it searches now the whole string of the line)

.github/workflows/testing.yml
```diff
@@ -1,6 +1,13 @@
 name: Pytest MSS
 
-on: [ push, pull_request ]
+on:
+  push:
+    branches:
+    - develop, stable
```
ReimarBauer: same "develop, stable" may not work

ReimarBauer (changes requested):
use one list item per line.

ReimarBauer (approved):
thx

PR #1342, opened 2022-01-31
Issue1341

joernu76:


gisi90 (approved):
looks good

joernu76:
@ReimarBauer is sonja not allowed to approve or what else is the reason for this still being blocked?

ReimarBauer:
> @ReimarBauer is sonja not allowed to approve or what else is the reason for this still being blocked?

see chat, I assume that there is a hidden feature/bug? I have not found something enabled in the settings which forces this. 
It behaves like a "review by an owner is required". This shows also the hint "At least 1 approving review is required by reviewers with write access." But this rule is not enabled.

ReimarBauer (approved):
thx

PR #1349, opened 2022-02-04
Fixed level selection.

### LINKED ISSUE #1348: gallery: horizontal layers without level cannot be selected and visualized

joernu76:
Works now with units containing spaces and "-" signs. Has an additional
"None" selection for level that reveals layers without levels.

Fix #1348


ReimarBauer (approved):
thx

PR #1350, opened 2022-02-04
"unwrap" longitude before determining valid area

### LINKED ISSUE #1347: gallery: automatic deduction of plotting area fails for 360 degree jump in longitude in data
See current TI3GER server.

joernu76:
fix #1347


ReimarBauer (approved):
On a further bigger refactoring that code needs to become testable.

PR #1352, opened 2022-02-04
Make the gallery usable while it is being updated.

### LINKED ISSUE #1351: gallery: updating the gallery should be atomic
I.e. the old gallery should work while the new one is being created. Then an atomic switch (e.g. OS mv) should be performed.

### LINKED ISSUE #1353: gallery: horizontal layers with spaces (etc) in level units are duplicated

### LINKED ISSUE #1354: temperature hsec style
valid time is not formatted (f missing)

joernu76:


ReimarBauer (dismissed):
looks so  far well done :)

I would have named path and old_path because reading a wsgi web server program, I would not think that location is meant as path.

mslib/mswms/gallery_builder.py
```diff
@@ -712,7 +712,7 @@ def add_image(path, plot, plot_object, generate_code=False, sphinx=False, url_pr
             markdown = add_times(itime, [vtime], markdown)
         plot_htmls[f"{l_type}_{dataset}{plot_object.name}"] = markdown
 
-    id = img_path.split(f"-{level}")[0]
+    id = img_path.split("-" + f"{level}".replace(" ", "_").replace(":", "_").replace("-", "_"))[0]
```
ReimarBauer, line 715: should we use here secure_filename?
https://flask.palletsprojects.com/en/1.1.x/patterns/fileuploads/

ReimarBauer (approved):
thx, good enough for now

mslib/mswms/gallery_builder.py
joernu76: That solves a different problem.
But personally, the mechanism of trying to "guess" poorly thought out filenames in the JavaScript is the root problem.
It would be much, much safer to build an associative array while creating the plots and store it in a static structure.

*Or* add a div element for all plots and simply toggle the visibility.  I am not enough of a web programmer to see, if this'd be a good idea.
The current approach seems efficient, but dangerous.

PR #1383, opened 2022-03-03
Doc improvements

### LINKED ISSUE #1382: Documentation improvements
- In the [Protecting Login](https://github.com/Open-MSS/MSS/blob/5c2376b6264e8a78b27ea714b46663d880f483d6/docs/mscolab.rst#protecting-login) section, it would be better to specify the path of mss_mscolab_auth.py.
- In the [Installing dependencies](https://github.com/Open-MSS/MSS/blob/0755dabd39f9fa7f0f7b578f5b44724624b95b31/docs/development.rst#installing-dependencies) section, we are using the old installation command `mamba install mss=$mss_version --only-deps` which doesn't work now.
- In the [Setup mscolab server](https://github.com/Open-MSS/MSS/blob/0755dabd39f9fa7f0f7b578f5b44724624b95b31/docs/development.rst#setup-mscolab-server) section, the command `python mslib/mscolab/mscolab db --init` has a typo (it should be mscolab.py).

swsrkty:
Fixes #1382


docs/development.rst
```diff
@@ -142,7 +142,7 @@ Create an environment and install the whole mss package dependencies then remove
 
   $ conda create -n mssdev mamba
   $ conda activate mssdev
-  $ mamba install mss=$mss_version --only-deps
```
ReimarBauer, line 145: this will be replaced on build on RTFD to the current number, we should keep this, because that worked alwasys in the past.  adding python is correct ;) but keep the env var

ReimarBauer: triggerd the build, that's still manually, the package must be first in the channel to get it installed.

docs/mscolab.rst
```diff
@@ -23,10 +23,10 @@ Protecting Login
 ~~~~~~~~~~~~~~~~
 The login to the MSColab server can be protected by an additional auth method.
 
-**mss_mscolab_auth.py**
-
  .. literalinclude:: samples/config/mscolab/mss_mscolab_auth.py.sample
 
+Copy the above file to the mslib/mscolab directory, rename it to mss_mscolab_auth.py and make the necessary changes in the file. 
+
```
ReimarBauer, line 29: on an installed MSS that's not that easy. The prefered solution is to have it in the PYTHONPATH. e.g. you can have a CONFIG dir with all these pyscripts and add this to the PYTHONPATH env var. 
Similiar to the section above https://mss.readthedocs.io/en/stable/mscolab.html?highlight=PYTHONPATH#configuring-your-mscolab-server

it must be only in a path where python finds it. Usualy on server setup you add this path to the PYTHONPATH environment.

ReimarBauer (changes requested):
see comments, small fixes needed

docs/development.rst
swsrkty: Thank you. I've fixed it. : )

docs/mscolab.rst
swsrkty: Ah, got it. Thank you.

ReimarBauer (approved):
Thx

PR #1401, opened 2022-03-23
Remove leftover __future__ and other Py2 stuff

### LINKED ISSUE #1400: remove __future__ imports
From the past migration we have some leftover "from __future__ import" lines 
e.g.

```
from __future__ import absolute_import
from __future__ import division
from __future__ import with_statement
```

since python 3 this is the default, so we can remove these lines. You do find them at least in these modules

```
mslib/mscolab/migrations/env.py:from __future__ import with_statement
mslib/msui/_tests/test_topview.py:from __future__ import division
mslib/plugins/io/text.py:from __future__ import division
mslib/plugins/io/gpx.py:from __future__ import absolute_import
mslib/plugins/io/gpx.py:from __future__ import division
mslib/plugins/io/flitestar.py:from __future__ import division
mslib/plugins/io/csv.py:from __future__ import absolute_import
mslib/plugins/io/csv.py:from __future__ import division

```

joernu76:
Close #1400

### DIFF
--- a/mslib/mscolab/migrations/env.py
+++ b/mslib/mscolab/migrations/env.py
@@ -23,8 +23,6 @@ SOFTWARE.
 """
 # source https://github.com/sunscrapers/flask-boilerplate/blob/master/migrations/env.py
 
-from __future__ import with_statement
-
 import logging
 from logging.config import fileConfig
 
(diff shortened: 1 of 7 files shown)


ReimarBauer (approved):
thx

PR #1402, opened 2022-03-24
Issue 1363, navaid

### LINKED ISSUE #1363: Include the use of NAVAID Waypoint naming convention
Pilots use the naming convention for waypoints in the format WPT012034
where WPT is the NAVAID ID, 012 is the heading and 034 is the distance
in nautical miles. This waypoint format is only used over land.  It
would be helpful to have the aibility to handle this format.

I was provided a python code developed by Jenna Samra and Jonathan
Franklin (SAO and Harvard University) and described by Pavel Romashkin
(NCAR RAF). I need to check, whether we are allowed to use this code
officially.

There is a database of these NAVAID waypoints e.g. in OPENAIP
https://www.openaip.net/navaids

Envisaged would be
(1) to create a flight path output plugin that is able to
automatically create the waypoint names in this format, either in a
ASCII text table or csv.

(2) an input plugin in which waypoints of this convention are
translated into real latitudes and longitudes

(3) potentially an automatic detection in the table view panel, if the
waypoints in this convention are typed into the name field.  It could
also include the NAVAID waypoints themselves.

jgrooss:


ReimarBauer (changes requested):
I change a bit later

jgrooss (commented):
I would approve these changes (I cannot officially do that as I am the PR author )
in navaid.rst, line 20  it should be "config dir" instead of "condif dir"

ReimarBauer (approved):
Thx

PR #1411, opened 2022-04-01
Server and client may crash if no coastline is visible in selected re…

### LINKED ISSUE #1410: MSS and WMS crashes sometimes if no coastline is present in the selected area
Similar to the problem described here

https://stackoverflow.com/questions/59855017/determining-if-a-coastline-is-present-when-using-basemap

I think this is an issue with basemap and/or matplotlib, but we can handle it in the proposed manner with a simple try/except ourselves.

joernu76:
…ctangle

This was partially captured already in the client, but could be provoked
by enabling/disabling visibility of coastlines in inopportune moments.
The server always crashed. The problem is now safely avoided/treated.

Fix #1410

### DIFF
--- a/mslib/msui/mpl_map.py
+++ b/mslib/msui/mpl_map.py
@@ -122,6 +122,7 @@ class MapCanvas(basemap.Basemap):
 
         # Set up the map appearance.
         if self.appearance["draw_coastlines"]:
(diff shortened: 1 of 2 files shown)


ReimarBauer (approved):
thx

PR #1412, opened 2022-04-01
Moved Target-singleton to use unit-aware ranges and thresholds

### LINKED ISSUE #1408: mslib.mswms.utils still contains fixed units
The THRESHOLDS variable of the Targets class contains values without units. This is "implicetly" defined by the UNITS tag, but this doesn't work correctly, when a different unit is specified in the generic plot.

Solution: The units tag must remove the conversion factor; instead, a unit must be given to the Thresholds, which is then converted to the actual plotting units.

joernu76:
Annotating the ranges allows setting them with file-specific units and
converting them ad-hoc the the configured unit.

Fix #1408

### DIFF
--- a/mslib/mswms/utils.py
+++ b/mslib/mswms/utils.py
@@ -27,9 +27,9 @@
 
 import numpy as np
 import matplotlib
-import pint
 
-UR = pint.UnitRegistry()
+from mslib.utils.units import convert_to
+
 N_LEVELS = 16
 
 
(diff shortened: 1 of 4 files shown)


mslib/mswms/utils.py
```diff
@@ -256,8 +258,11 @@ def get_thresholds(standard_name, level=None, type=None):
         Returns:
             Tuple of threshold values to be supplied to a BoundaryNorm.
         """
+        # put here so this file may be imported without triggering an import of whole mslib
+        from mslib.utils.units import convert_to
```
ReimarBauer: units uses only


```
import logging

from metpy.units import units, check_units  # noqa
import pint
```
so this import should be moved on top

ReimarBauer (changes requested):
see comment

ReimarBauer (approved):
Thx

PR #1413, opened 2022-04-01
don't add servers without layers

### LINKED ISSUE #1376: remove a layer-less WMS server using the "trashcan" icon causes a segmentation fault
Happens on stable 6.0.3

Steps to reproduce:
1) load current SouthTrac server (serving no layers)
2) remove the server using the trashcan icon in the layer selection
3) program crashes with a segmentation fault, probably within QT

ReimarBauer:


joernu76 (commented):
Mmmh. I feel that this "hack" does not belong to the general Worker class, but more into the wms_control.py, where one could do the check....?

joernu76:
Also, having a look at the usage of the Worker, there seem to be some issues.
I also do not understand the use of the Workers anymore. There are "named" Workers and "unnamed" ones...
I need to check something.

joernu76:
OK, the dialogue for entering login information is broken as its result is not connected anymore.

This explains for problem of today, where the get_capabilities timed out!

joernu76:
shall I create an issue for that? I think this needs some serious refactoring to fix that bug  and provide the workaround of yours in a clean fashion.

joernu76 (approved):

PR #1416, opened 2022-04-12
Make linear default layer names unique

### LINKED ISSUE #1415: Default Linear Plotting classes for different quantities share same name property
The name property of default linear plotting classes is defined from some abbreviation of the plotted variable, which causes overwriting of registered layers in registration.

joernu76:
Currently, similar variables result in layers of same name, which
overwrirte each other in registration.

Fix #1415

### DIFF
--- a/mslib/mswms/mpl_lsec_styles.py
+++ b/mslib/mswms/mpl_lsec_styles.py
@@ -43,8 +43,7 @@ class LS_DefaultStyle(AbstractLinearSectionStyle):
         self.required_datafields = [(filetype, self.variable, None)]
         if filetype != "pl":
             self.required_datafields.insert(0, (filetype, "air_pressure", "Pa"))
-        abbreviation = "".join([text[0] for text in self.variable.split("_")])
-        self.name = f"LS_{str.upper(abbreviation)}"
+        self.name = f"LS_Default_{variable}"
         self.title = f"{self.variable} Linear Plot"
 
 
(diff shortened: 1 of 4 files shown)


mslib/mswms/mpl_lsec_styles.py
```diff
@@ -43,8 +43,7 @@ def __init__(self, driver, variable="air_temperature", filetype="ml"):
         self.required_datafields = [(filetype, self.variable, None)]
         if filetype != "pl":
             self.required_datafields.insert(0, (filetype, "air_pressure", "Pa"))
-        abbreviation = "".join([text[0] for text in self.variable.split("_")])
-        self.name = f"LS_{str.upper(abbreviation)}"
+        self.name = f"LS_Default_{variable}"
```
ReimarBauer, line 46: noting, this will change length of filename too, old one:
Linear_LS_AT-Noneit2022_02_04_00_00_00vt2022_02_05_12_00_00.png

it makes sense to use the origin variable name.

ReimarBauer (dismissed):
thx, good catch.

mslib/mswms/mpl_lsec_styles.py
ReimarBauer: test_wms tests have to become adapted too

joernu76: The error messages in relation to libgsl are... related?

ReimarBauer (approved):
thx

PR #1419, opened 2022-04-14
skip test because of timeout on gh

ReimarBauer:


ReimarBauer:
We can much easier disable a test on the image creation process if this disturbs to build the same.

PR #1422, opened 2022-04-14
I1421, taking care on github fix

### LINKED ISSUE #1421: Error: fatal: unsafe repository on github actions
Because of https://github.blog/2022-04-12-git-security-vulnerability-announced/ we have to change our testing configuration

```
 Initialized empty Git repository in /__w/MSS/MSS/.git/
33
  /opt/conda/envs/mssenv/bin/git remote add origin https://github.com/Open-MSS/MSS
34
  Error: fatal: unsafe repository ('/__w/MSS/MSS' is owned by someone else)
35
  To add an exception for this directory, call:
36
  
37
  	git config --global --add safe.directory /__w/MSS/MSS
38
  Error: The process '/opt/conda/envs/mssenv/bin/git' failed with exit code 128

```

ReimarBauer:


joernu76 (approved):
Can you elaborate a bit what this is doing?

ReimarBauer:
added my source "actions" to the issue. This security fix was also the cause for the release of git 2.35.2. in the actions issue https://github.com/actions/checkout/issues/760 are also workarounds before that git version.
cited: 
Same issue with the Windows runners, and it is indeed because of the security fix for CVE-2022-24765.

The quick fix is to add a step before the checkout action to run git config --global --add safe.directory <path>. The better fix will be for the runners / containers / &c. to set this automatically; on the GitHub runners, at least, because they're not shared, I don't believe there's any risk of actually hitting this vulnerability.

Issue #1205, opened 2021-09-13
For logging, do not use f-strings. Use % for performance reasons etc.


ReimarBauer:
refactor all logging strings

engineerrbae:
Hey Reimar,  to show you a sample I have added % to the below code. Is this how you wish to have changes in all the .py files

https://github.com/engineerrbae/MSS/blob/develop/conftest.py#L189


summary = "\n".join(["PyQt5.QtWidgets.QMessageBox.%s: %s" %(box()._extract_mock_name(),box.mock_calls[:-1])
for box in [q, i, c, w] if box.call_count > 0])
 warnings.warn("An unhandled message box popped up during your test!\n%s" %(summary))
 
 OR
 
 summary = "\n".join(["PyQt5.QtWidgets.QMessageBox.{ }: { }".format(box()._extract_mock_name(),box.mock_calls[:-1])
for box in [q, i, c, w] if box.call_count > 0])
 warnings.warn("An unhandled message box popped up during your test!\n{ }".format(summary))

Once you share I will move ahead with refactoring. 

Thanks.

ReimarBauer:
We do want to change only the use of f-string in logging....() to `%` syntax. On all other places f-strings are better.

Joern wrote an explanaition copied from https://github.com/Open-MSS/MSS/pull/1202#discussion_r707259404

f-formatting is almost always faster than c-style formatting due to some particularly interesting implementation shenanigans. That is not the point here, however.

The idea here is that the formatting will not take place at all in case the logging level is above debug. Using an f-string always does the formatting even if nothing will be output. That can have an detrimental effect on performance with logging disabled.
Thus it is good style to always employ conditional formatting, which by necessity of backwards-compatability, is c-style formatting. For info/warn/error, which are basically always dumped or where the formatted string is needed anyway (e.g. for a popup message), one may deviate at discretion.

There is a "style" parameter, which can be triggered on a by call basis to use other formatters (e.g. the {} one). See
https://docs.python.org/3/howto/logging-cookbook.html#formatting-styles
This seems very cumbersome for little gain.
f-strings cannot be used as the decision whether to format cannot be controlled by the logging class.

engineerrbae:
Got it. f-string is an expression that is evaluated at runtime and it lacks logging optimizations. Also, logging by default follows % standards & adoption of its default standards going to make the logging fast & accurate & better statistically.

[...]

PR #1425, opened 2022-04-17
converted all f-strings to % for logging an mentioned in issue #1205

DipanshuShukla:
converted all f-strings to % for logging an mentioned in issue #1205

### DIFF
--- a/mslib/mscolab/file_manager.py
+++ b/mslib/mscolab/file_manager.py
@@ -151,7 +151,7 @@ class FileManager(object):
         operation = Operation.query.filter_by(id=op_id).first()
         if attribute == "path":
             if value.find("/") != -1 or value.find("\\") != -1 or (" " in value):
-                logging.debug("malicious request: %s", user)
+                logging.debug("malicious request: %s" % user)
                 return False
             data = fs.open_fs(self.data_dir)
             if data.exists(value):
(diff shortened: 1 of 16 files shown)


DipanshuShukla:
why did these tests fail?

ReimarBauer:
Have a look on the "Details" for flake8
https://github.com/Open-MSS/MSS/runs/6059325533?check_suite_focus=true
The other two I Iook at.

mslib/mscolab/file_manager.py
```diff
@@ -151,7 +151,7 @@ def update_operation(self, op_id, attribute, value, user):
         operation = Operation.query.filter_by(id=op_id).first()
         if attribute == "path":
             if value.find("/") != -1 or value.find("\\") != -1 or (" " in value):
-                logging.debug("malicious request: %s", user)
+                logging.debug("malicious request: %s" % user)
```
ReimarBauer: the "," should be correct here

ReimarBauer (changes requested):
Have a look on the oldstyle syntax and check also erros of tests

ReimarBauer:
> why did these tests fail?

the dockerimage was not correct, fixed that temporary (PR is open). But tests have errors, please 
check.
You can also run tests locally, see
https://mss.readthedocs.io/en/stable/development.html?highlight=pytest#run-tests

ReimarBauer:
@DipanshuShukla please have a look on https://docs.python.org/3/howto/logging.html#logging-variable-data

ReimarBauer:
we can reopen it when continuing

PR #1426, opened 2022-04-19
replaced f' strings with % for logging

Ritwikrajsingh:


ReimarBauer:
You could run the tests locally, see https://mss.readthedocs.io/en/stable/development.html?highlight=pytest#run-tests

mslib/mswms/mswms.py
```diff
@@ -106,8 +106,8 @@ def main():
         logging.info("Gallery generation done.")
         sys.exit()
 
-    updater.on_update_available.connect(lambda old, new: logging.info(f"MSS can be updated from {old} to {new}.\nRun"
-                                                                      " the --update argument to update the server."))
+    updater.on_update_available.connect(lambda old, new: logging.info("MSS can be updated from %s to %s.\nRun"
```
joernu76: There are several issues with PEP8 compliance here (see failed pipeline).
Here, you need to either align the second string with the line before or join the two strings in a new line (more options are likely possible).
Use pycodestyle or flake8 to test locally. We use some specific configurations to reduce the nagginess of the tools, depending on how you call them, you might get slightly different results than the pipeline. See setup.cfg.

mslib/mswms/wms.py
```diff
@@ -173,137 +173,139 @@ def squash_multiple_xml(xml_strings):
...
-                                                             i_vtimes if vtimes == "all" else [i_vtimes[-1]])] or [None]
-
-                                for vtime in sorted(valid_times):
-                                    if vtime and i_vtimes and vtime not in i_vtimes:
-                                        logging.warning(f"Requested vtime {vtime} at {itime} not present for "
-                                                        f"{dataset} {plot_object.name}! vtimes present: {i_vtimes}")
-                                        continue
-                                    elif not i_vtimes:
-                                        vtime = None
+	def __init__(self):
```
ReimarBauer, line 176: This looks like an indenting problem. 

In python a tab is 8 blanks. Indenting for one level is 4 blanks.

Which IDE do you use?

ReimarBauer (changes requested):
Hi, there is also a flake 8 failure in 

mslib/mscolab/mscolab.py:144:128: E501 line too long (134 > 127 characters)
1     E501 line too long (134 > 127 characters)
1     E999 IndentationError: unindent does not match any outer indentation level

mslib/mswms/wms.py
ReimarBauer: This is a good tutorial for using flake8
https://siderlabs.com/blog/automatically-review-code-for-python-projects-using-flake8-6fcb056a001a/

Ritwikrajsingh: I use vscode

ReimarBauer: You may want to configure it to use python indenting style, see https://peps.python.org/pep-0008/#indentation

I assume there is a config option to use python indenting.  @joernu76 have you a picture where that needs to become configured.

The tests fail because of an indenting problem.
https://github.com/Open-MSS/MSS/runs/6159304098?check_suite_focus=true#step:6:1123

ReimarBauer:
we can reopen it when continuing

PR #1429, opened 2022-04-26
show an indicator in the title for multilayer plots

### LINKED ISSUE #1381: multilayer does not show used layer names on plot
When I select multiple layers the plot title keeps the first one. 
The UI shows in a line what was selected, this can also be optimized, because we don't have much space there to get many names concatenated in a row.

On the UI the selection in the ML view is where we have this information on the correct place, 
The information in the row of the server/layer button could show "multilayer"

On the plot we maybe could enhance the title or put another documentation above the lines where we write the flight track name. The plot is the region one looks on after multilayer were selected. So we have to add that information there.

ReimarBauer:
This is a first step for showing on a plot that there is more data by multilayer seen

### DIFF
--- a/mslib/msui/wms_control.py
+++ b/mslib/msui/wms_control.py
@@ -1547,10 +1547,7 @@ class VSecWMSControlWidget(WMSControlWidget):
         self.view.draw_image(self.squash_multiple_images(imgs))
         self.view.draw_legend(self.append_multiple_images(legend_imgs))
         style_title = self.multilayers.get_current_layer().get_style()
-        title = self.multilayers.get_current_layer().layerobj.title
-        if len(self.multilayers.get_active_layers()) > 1 and self.multilayers.get_current_layer().checkState(0):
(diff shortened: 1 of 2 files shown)


mslib/msui/wms_control.py
```diff
@@ -1547,7 +1547,10 @@ def display_retrieved_image(self, imgs, legend_imgs, layer, style, init_time, va
         self.view.draw_image(self.squash_multiple_images(imgs))
         self.view.draw_legend(self.append_multiple_images(legend_imgs))
         style_title = self.multilayers.get_current_layer().get_style()
-        self.view.draw_metadata(title=self.multilayers.get_current_layer().layerobj.title,
+        title = self.multilayers.get_current_layer().layerobj.title
```
joernu76: with this complexity and repition, I argue that this should be refactored to a get_title() method of self.multilayers

joernu76 (approved):

PR #1436, opened 2022-04-29
Switch testing structure to new docker images

### LINKED ISSUE #1433: MSS testing infrastructure
Vision/Requirements:
* have separate testing docker images for stable & develop
* non-dependency changing PR shall not modify installed packages
* have periodic scheduled tests against "bleeding edge" conda-forge packages (similar to "mamba --update --all; mamba remove mss --force; mamba install mss");
  - this shall replicate building mss on conda-forge
  - this shall replicate install mss on a user system
* no manual interaction
* noticeable trigger of failed periodic runs to core team
* as simple as possible

First steps:
* New repo for docker generation. Single branch with multiple directories. Share as much as possible between develop and stable images.
* New tags for images. testing-develop-latest testing-stable-latest (testing-develop-candidate,...)
* Make periodic tests from "conda local build" from fresh mamba-forge container.

joernu76:


.github/workflows/testing.yml
```diff
@@ -21,15 +19,19 @@ jobs:
...
-      image: openmss/mss-test:latest
+      image: openmss/develop:latest
 
     steps:
     - name: Trust My Directory
       run: git config --global --add safe.directory /__w/MSS/MSS
     - uses: actions/checkout@v2
         
-    - name: Reinstall dependencies if changed
+    - name: Reinstall dependencies if changeda
```
ReimarBauer: typo

ReimarBauer (changes requested):
When we remove xdist testing, then we should have another testcall which does not a reverse order of tests.  xdist testing does normal order based on dist loadfile. We need both to find not good written tests which are based on other tests. 

We can move xdist testing to the developer description and also can cleanup the  development.txt but we should not change tests. I want to have the option to run them parallel.

.github/workflows/testing.yml
```diff
@@ -55,9 +68,20 @@ jobs:
           && break \
         ; done)
 
-        
-    - name: coveralls
-      if: ${{ always() && github.event_name == 'push' && github.ref == 'refs/heads/develop' }}
+    - name: Run tests in parallel
+      if: ${{ always() && inputs.xdist == 'yes' }}
+      timeout-minutes: 25
```
ReimarBauer, line 73: ah, if condition :)

we should try again if -n auto takes more than 2. So that we can take github management.
Last time I fixated it to 6, because auto seems not to work on that platform

.github/workflows/testing.yml
```diff
@@ -21,17 +28,22 @@ jobs:
...
+      image: openmss/testing-${{ inputs.branch_name }}
 
     steps:
     - name: Trust My Directory
       run: git config --global --add safe.directory /__w/MSS/MSS
+
     - uses: actions/checkout@v2
-        
     - name: Reinstall dependencies if changed
+      id: step1
```
ReimarBauer, line 43: There is not an enumeration of step1, step2 ..
Maybe rename this into check ?

Then this is like a question:

output1: ${{ steps.check.outputs.triggerdockerbuild }}

.github/workflows/testing.yml
```diff
@@ -21,17 +28,22 @@ jobs:
       run:
         shell: bash
 
+    outputs:
+      output1: ${{ steps.step1.outputs.triggerdockerbuild }}
```
ReimarBauer, line 32: output1: ${{ steps.check.outputs.triggerdockerbuild }}

.github/workflows/testing-stable.yml
```diff
@@ -0,0 +1,30 @@
...
+on:
+  push:
+    branches:
+    - stable
+  pull_request:
+    branches:
+    - stable
+
+jobs:     
+  test-develop:
```
ReimarBauer: should that be named test-stable ?

.github/workflows/testing-stable.yml
```diff
@@ -0,0 +1,30 @@
...
+    uses: 
+      ./.github/workflows/testing.yml
+    with:
+      xdist: no
+      branch_name: stable
+      event_name: ${{ github.event_name }}
+    secrets: inherit
+
+  test-develop-xdist:
+    uses:
```
ReimarBauer, line 22: also test-stable-xdist

or we can make this more generic between both definition files, by using

test-seriell
test-parallel

.github/workflows/testing.yml
```diff
@@ -21,17 +28,22 @@ jobs:
...
     - name: Trust My Directory
       run: git config --global --add safe.directory /__w/MSS/MSS
+
     - uses: actions/checkout@v2
-        
     - name: Reinstall dependencies if changed
+      id: step1
       run: |
         cmp -s /meta.yaml localbuild/meta.yaml && cmp -s /development.txt requirements.d/development.txt || (echo Dependencies differ \
+        && echo "::set-output name=triggerdockerbuild::yes" \
```
ReimarBauer, line 46: good :)

ReimarBauer (changes requested):
Please have a look on the "nameings" suggestions.

.github/workflows/testing.yml
ReimarBauer: Could we add this python version to a git var? 
current should be 3.9.13 

&& mamba install pyvirtualdisplay python=3.9.12 \

.github/workflows/testing.yml
joernu76: this is mostly copied from the github documentation.
I have to admit, that I understand that yml-Stuff only 80%.
Sticking close to the documentation might be helpful?

.github/workflows/testing-stable.yml
joernu76: Once this works, I'll investigate whether these copy-pasted scripts can be unified as well.

.github/workflows/testing.yml
joernu76: upgraded to 3.9.13. I hope to remove this once basemap is fixed.

.github/workflows/testing.yml
ReimarBauer: the "steps:" is a keyword, we can't change, but the id: "lable" should be something we could give a meaningful name. In https://docs.github.com/en/actions/learn-github-actions/contexts a lable of "github_context_step" is used. 
A rename is not so important for me now,

.github/workflows/testing.yml
ReimarBauer: the "steps:" is a keyword, we can't change, but the id: "label" should be something we could give a meaningful name. In https://docs.github.com/en/actions/learn-github-actions/contexts a label of "github_context_step" is used. 
A rename is not so important for me now,

.github/workflows/testing.yml
ReimarBauer: see discussion on https://github.com/conda-forge/basemap-feedstock/issues/97

ReimarBauer (approved):
Please peek on the comments on a further version.

Thx.

PR #1438, opened 2022-05-01
Adding leave operation option in properties menu

### LINKED ISSUE #1322: leave Operation in properties menu
Any user role besides creator needs to have the option to leave an operation.

After doing that the user needs to be disconnected from that operation, e.g. close open views

swsrkty:
Fixes #1322


mslib/mscolab/file_manager.py
```diff
@@ -396,16 +396,18 @@ def modify_bulk_permission(self, op_id, user, u_ids, new_access_level):
             return False
 
     def delete_bulk_permission(self, op_id, user, u_ids):
-        if not self.is_admin(user.id, op_id):
+        if self.is_admin(user.id, op_id) and user.id in u_ids:
```
ReimarBauer: our current tests only cover the else condition. You could create a local coverage by
``` pytest --cov-report html --cov mslib ```
this stores in MSS/htmlcov the report.

add a decriptive comment to the conditions for which case they are needed.  A test which covers the lines helps us not to remove/refactor this, because than the test will break,

We should make sure that a user can only remove himself as none admin from the OP and not use "/delete_bulk_permissions" to remove users from any other OP. 
Besides is_admin we also have is_collaborator. Perhaps a "is_member" can be created which covers all accesslevels or also a "is_viewer" . A test which shows that it is safe would be pretty good.

in the else part we should verify that the user is in that OP ;)

file_mananger


not self.is_admin(user.id, op_id)

**Review by ReimarBauer — CHANGES_REQUESTED** [review]
see comments

File: mslib/mscolab/file_manager.py [hunk 2]
```diff
@@ -104,6 +104,17 @@ def list_operations(self, user):
             })
         return operations
 
+    def is_member(self, u_id, op_id):
+        """
+        op_id: operation id
+        u_id: user-id
+        """
+        # return true only if the user is a member
+        permi = Permission.query.filter_by(u_id=u_id, op_id=op_id).first()
```
ReimarBauer: please keep the word "perm" someone else has used in the other methods.

mslib/mscolab/file_manager.py
```diff
@@ -104,6 +104,17 @@ def list_operations(self, user):
...
+        if not permi:
```
ReimarBauer: better compare please to None

if perm is not None:

 we abuse boolean here and some other none cpython interpreter can do it wrong

mslib/mscolab/file_manager.py
```diff
@@ -117,6 +128,19 @@ def is_admin(self, u_id, op_id):
...
         return True
 
+    def is_non_admin_member(self, u_id, op_id):
+        """
+        op_id: operation id
+        u_id: user-id
+        """
+        # return true only if the user is a member but not admin
+        perm = Permission.query.filter_by(u_id=u_id, op_id=op_id).first()
+        if not perm:
```
ReimarBauer: if perm is not None:

mslib/mscolab/file_manager.py
```diff
@@ -130,6 +154,19 @@ def is_collaborator(self, u_id, op_id):
             return False
         return True
 
+    def is_viewer(self, u_id, op_id):
+        """
+        op_id: operation id
+        u_id: user-id
+        """
+        # return true only if the user is viewer
+        permi = Permission.query.filter_by(u_id=u_id, op_id=op_id).first()
```
ReimarBauer: please use  perm = 

so that someone reading this on some place will know Permission is meant

ReimarBauer (changes requested):
a few little things to improve, mostly done.

see also on our chat.

ReimarBauer (changes requested):
I double checked the differnces between creator and admin role.

creator is only the user who created the operation. This one can delete the operation. 
The user has automatically the admin role which allows to add other users.
There can be more users with admin role. A user with the admin role should be able to remove himself too.

please have a look on this too. The test was missing before you started already.

```
--- a/mslib/mscolab/_tests/test_file_manager.py
+++ b/mslib/mscolab/_tests/test_file_manager.py
@@ -68,6 +68,8 @@ class Test_FileManager(TestCase):
         self.vieweruser = get_user('UV30@uv30')
         assert add_user('UV40@uv40', 'UV40', 'uv40')
         self.collaboratoruser = get_user('UV40@uv40')
+        assert add_user('UV80@uv80', 'UV80', 'uv80')
+        self.adminuser = get_user('UV80@uv80')
         assert add_user('UV50@uv50', 'UV50', 'uv50')
         self.op2user = get_user('UV50@uv50')
         assert add_user('UV60@uv60', 'UV60', 'uv60')
@@ -79,6 +81,60 @@ class Test_FileManager(TestCase):
     def tearDown(self):
         pass
```
```
    def test_admins_when_not_creator_should_be_able_to_remove_themselves(self):
        # creator is the one who creates the operation
        # a creator can't be removed from an operation and can't change this. 
        # Only creator role gives the option to delete the operation
        # admin is a role to add users.
        # the creator has also this role. But there could be more admins
        # A user with the admin role should be able to remove himself
        with self.app.test_client():
            flight_path, operation = self._create_operation(flight_path="saturn")
            assert operation.path == flight_path
            self.fm.add_bulk_permission(operation.id, self.user, [self.adminuser.id], "admin")
            assert self.fm.is_admin(self.adminuser.id, operation.id) is True
            assert self.fm.delete_bulk_permission(operation.id, self.user, [self.user.id]) is False
            assert self.fm.delete_bulk_permission(operation.id, self.user, [self.adminuser.id]) is True
            assert self.fm.delete_bulk_permission(operation.id, self.adminuser, [self.adminuser.id]) is True
            assert self.fm.delete_bulk_permission(operation.id, self.adminuser, [self.user.id]) is False
            ```

**Comment by swsrkty** [comment]
Moin @ReimarBauer, thank you for explaining the difference between creator and admin. I've updated the tests, please take a look : ).

File: mslib/mscolab/_tests/test_file_manager.py [hunk 6]
```diff
@@ -198,13 +236,89 @@ def test_undo(self):
...
+            assert self.fm.is_creator(self.user.id, operation1.id)
+            assert self.user.id is not None
+            self.fm.add_bulk_permission(operation1.id, self.user, [self.anotheruser.id], "collaborator")
+            assert self.fm.is_collaborator(self.anotheruser.id, operation1.id)
+            self.fm.add_bulk_permission(operation1.id, self.user, [self.collaboratoruser.id], "collaborator")
+            assert self.fm.is_collaborator(self.collaboratoruser.id, operation1.id)
+            flight_path1, operation2 = self._create_operation(flight_path="testflight1", user=self.op2user)
+            assert self.fm.is_creator(self.op2user.id, operation2.id)
+            assert self.op2user.id is not None
+            # user (admin of operation1) has no right leave operation1
```
ReimarBauer: in this case self.user was the creator and therefore has no right to leave operation1 It is only the comment which is not exact. This one at same time has also admin role ;)

You can leave it like this because then it is consistent with the next lines. Or if you like change it.

ReimarBauer (dismissed):
great :)

sr1j:
Hi @ReimarBauer , I just realised that i've some more things to fix.

mslib/mscolab/_tests/test_file_manager.py
swsrkty: I've fixed it : )

swsrkty:
Moin @ReimarBauer, I've added one more change. Please check.

mslib/mscolab/_tests/test_file_manager.py
```diff
@@ -198,13 +235,89 @@ def test_undo(self):
...
+            self.fm.add_bulk_permission(operation1.id, self.user, [self.vieweruser.id], "viewer")
+            assert self.fm.is_viewer(self.vieweruser.id, operation1.id)
+            flight_path1, operation2 = self._create_operation(flight_path="testflight1", user=self.op2user)
+            assert self.fm.is_creator(self.op2user.id, operation2.id)
+            assert self.op2user.id is not None
+            self.fm.add_bulk_permission(operation2.id, self.op2user, [self.op2vieweruser.id], "collaborator")
+            # collaboratoruser of operation1 has no right to remove vieweruser from the operation1
+            assert self.fm.delete_bulk_permission(operation1.id, self.collaboratoruser, [self.vieweruser.id]) is False
+            # The below assertion fails in stable 6.1
+            # The change is so that vieweruser (any user other than admin) can leave the operation
```
ReimarBauer: just seen that, 

any user other than **creator**)

mslib/mscolab/_tests/test_file_manager.py
```diff
@@ -217,7 +330,7 @@ def test_import_permission(self):
             result = (False, None, 'Permissions are already given')
             assert self.fm.import_permissions(operation10.id, operation11.id, self.user.id) == result
             # no admin rights
-            result = (False, None, 'Not an admin of this operation')
```
ReimarBauer, line 220: This should be possible for the admin role too. This squence shows what the meaning is. This is an important feature because when we have several users on a server we can copy all roles and users from one operation to an other. 
The feature is similar to manual add members for the admin role. creator and admin role should be able to use it.

admin_should_be_able_to_import

ReimarBauer (changes requested):
Found by using the ui a difference to the past. Please have a look on the gifsycle. Please have a look if this easy solvalble.

mslib/mscolab/_tests/test_file_manager.py
swsrkty: Ah, got it. I've fixed it : ).

ReimarBauer (approved):
Thx

PR #1442, opened 2022-05-07
User Profile Options Updated

### LINKED ISSUE #1282: user profile options
currently we show the account informations which are used to login.

We maybe should provide a fullname, nickname feature to be added from the profile 
This can be used on chat and also on a created image by one of the views.

Jatin2020-24:
Fixes: #1282


mslib/mscolab/server.py
```diff
@@ -322,6 +322,28 @@ def delete_user():
...
+def edit_full_name():
+    user = g.user
+    fullname = request.form.get("fullname")
+    name = User.query.filter_by(id=int(user.id)).first()
+    name.fullname = str(fullname)
+    db.session.commit()
+    return jsonify({"success": True}), 200
+
+
+# @APP.route("/edit_nick_name", methods=["POST"])
```
ReimarBauer, line 335: maybe we should have only one /edit_profile for the edits of userdata. 

It must be blockable that users can edit once set email Adress and nickname.
That may be can done by some parameter in the mscolab configuration, e.g.
something user_form_disable = []
The server has then to drop changes and the UI may disable the fields.

ReimarBauer:
we can reopen it, when you want to continue

PR #1445, opened 2022-05-11
Update roles for "delete operation" and "update description"

### LINKED ISSUE #1444: role option "delete operation" only for "creator" role
The property option to delete an operation can only be done by a creator role. We have to update the menu based on the selected operation.

e.g. a viewer role can't do that, there is no need to have it activated in the menu

viewer_option_review

Also update description is not possible for a viewer role. That also should be fixed in same PR.

Jatin2020-24:
Fixes: #1444


ReimarBauer (changes requested):
please send the PR to stable branch. This is a bug fix for current stable release.

Jatin2020-24:
@ReimarBauer This PR is sended to stable. Check #1447 .

PR #1447, opened 2022-05-12
Update roles for "delete operation" and "update description"

### LINKED ISSUE #1444: role option "delete operation" only for "creator" role

Jatin2020-24:
Fixes: #1444

### DIFF
--- a/mslib/msui/mscolab.py
+++ b/mslib/msui/mscolab.py
@@ -1480,11 +1480,14 @@ class MSSMscolab(QtCore.QObject):
         self.ui.actionManageUsers.setEnabled(False)
         self.ui.menuProperties.setEnabled(True)
         self.ui.actionRenameOperation.setEnabled(False)
+        self.ui.actionDeleteOperation.setEnabled(False)
+        self.ui.actionUpdateOperationDesc.setEnabled(False)
         if self.access_level == "viewer":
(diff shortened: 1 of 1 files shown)


mslib/msui/mscolab.py
```diff
@@ -1480,11 +1480,14 @@ def show_operation_options(self):
...
         self.ui.menuProperties.setEnabled(True)
         self.ui.actionRenameOperation.setEnabled(False)
+        self.ui.actionDeleteOperation.setEnabled(False)
+        self.ui.actionUpdateOperationDesc.setEnabled(False)
         if self.access_level == "viewer":
             self.ui.menuImportFlightTrack.setEnabled(False)
             return
 
         if self.access_level in ["creator", "admin", "collaborator"]:
+            self.ui.actionUpdateOperationDesc.setEnabled(True)
```
ReimarBauer: only createor and admin role can change the decription. The line may be moved to 1506?

ReimarBauer (changes requested):

ReimarBauer (approved):
Thx

PR #1456, opened 2022-05-31
add flight level as vertical coordinate for files

### LINKED ISSUE #1455: Allow data with flight level as vertical coordinate
For some applications (e.g. contrail forecasts), data are stored on flight levels and should be recognized by msswms and available as such also in the horizontal view (rather than strange intermediate pressure levels/values).

gisi90:
see issue #1455


ReimarBauer:
Our testdata does currently not know this coordinate, we maybe should add data for it?

https://github.com/Open-MSS/MSS/blob/develop/mslib/mswms/demodata.py#L841

joernu76:
I think fixing/adding to the testcases is something we can do in a second step.

I can handle that then.

joernu76 (commented):
The "fl" type needs to be added to 
mslib/mswms/wms.py lines ~776+780

Thinking about it, it would be best to replace the list [ ... ] with
mslib.utils.netCDF4tools.VERTICAL_AXIS

In this fashion, it *should* be feasible to import the netCDF4tools and add stuff to the dictionary from the config file....

The new type needs to be added to "docs/mswms.rst" as well.

joernu76 (approved):

PR #1458, opened 2022-06-01
improved mss build for python 3.10 and recent basemap

ReimarBauer:


ReimarBauer:
This one has a problem with the new setup of our test infrastructure. We need to investigage this further.

PR #1468, opened 2022-06-06
builtin CSV plugin have to become removed too

### LINKED ISSUE #1467: fatal error on saving a configuration file
When using the configuration editor and saving a change the 
dialog appears
"
Do you want to restart the application?
(This is necessary to apply changes)

Please note that clicking 'No' will not save the current configuration
"
and as follow up on "Yes" the error happens.

```
  File "/home/reimar/MSS/mslib/msui/msui.py", line 881, in restart_application
    self.add_plugins()
  File "/home/reimar/MSS/mslib/msui/msui.py", line 487, in add_plugins
    self.add_plugin_submenu("CSV", "csv", load_from_csv, picker_default, plugin_type="Import")
  File "/home/reimar//MSS/mslib/msui/msui.py", line 505, in add_plugin_submenu
    raise ValueError(f"'{action_name}' has already been set!")
ValueError: 'actionImportFlightTrackCSV' has already been set!
```

The configuation change is saved.

ReimarBauer:


joernu76:
Mmmh. This is a mess. Why are the CSV plugins added at a different place than the FTML "plugins"? Why do we initialize self.import_plugins in a method without checking it is empty?

I suggest moving the CSV adding to the constructor, where the FTML stuff is added. This seems to be a more sensible change (make things consistent, not complicated. In addition we might want to make an issue for making this whole thing more robust (do not initialize values in a subfunction, check on proper class-state before adding plugins, etc.) as an easy issue?

ReimarBauer:
when you save a configuration, the msui gets restarted. The old one kept the initial state of CSV in the menu because we missed to remove it too.
The cleanest solution is not to add io plugins silently and have all only by configuration of the users msui_settings.conf.

save/load _ ftml is defined in flightrack.py. This is the only builtin we have.

If we would remove CSV as silent default and a user has to add it in configuration this is also solved. If it gets builtin we should refactor it to the module flightrack.py. Each is more than a bug fix,

We should fix it now in a simple way and refactor after 7.0 is released. -> new issue

PR #1469, opened 2022-06-07
added workflow for scheduled testing

joernu76:


.github/workflows/testing-scheduled.yml
```diff
@@ -0,0 +1,30 @@
+name: new dependency test (scheduled)
+
+on:
+  schedule:
+    - cron: '30 5 * * 1' 
+
```
ReimarBauer, line 6: I don't use cron much. I am also not sure by reading 
https://crontab.guru/

Is it once per month or weekly? I would prefer weekly.

ReimarBauer: https://crontab.guru/every-week
weekly
```0 0 * * 0```

ReimarBauer (approved):
lets see what github makes

PR #1470, opened 2022-06-07
Simple fix for restarting UI and CSV support.

### LINKED ISSUE #1467: fatal error on saving a configuration file

joernu76:
To be remodelled at a point in tim eafter 7.0.0 release.

Fix #1467

### DIFF
--- a/localbuild/meta.yaml
+++ b/localbuild/meta.yaml
@@ -38,7 +38,7 @@ requirements:
     - hdf4
     - pillow
     - pytz
-    - pyqt >=5, <6
+    - pyqt >=5, <5.13
     - pygeoif
     - qt >=5.10, <6
     - requests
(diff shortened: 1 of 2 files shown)


ReimarBauer:
we should add this to stable. And also with the QT fixation we should create a new build on conda-forge. That works better for every user than an increased build number.

joernu76:
Updated.

ReimarBauer (approved):
thx, better solution :)

PR #1474, opened 2022-06-07
I440b

joernu76:


requirements.d/addons.txt
```diff
@@ -1,6 +1,7 @@
 # This file may be used to create an environment using:
 # $ conda create --name <env> --file <this file>
 # platform: linux-64
+gsl==2.7.0
```
ReimarBauer: gsl is needed by some tests for the nco library. Has that changed?

ReimarBauer (approved):

PR #1487, opened 2022-06-14
I1486

### LINKED ISSUE #1486: wrong units for wind in mswms utils
Is 1/ms. Should be m/s

joernu76:


mslib/mswms/utils.py
```diff
@@ -102,13 +102,13 @@ class Targets(object):
...
-        "eastward_wind": "1/ms",
+        "eastward_wind": "m/s",
         "equivalent_latitude": "degree N",
         "ertel_potential_vorticity": "PVU",
         "gravity_wave_temperature_perturbation": "K",
         "mean_age_of_air": "month",
         "median_of_age_of_air_spectrum": "month",
-        "northward_wind": "1/ms",
+        "northward_wind": "m/s",
         "square_of_brunt_vaisala_frequency_in_air": "1/s²",
```
ReimarBauer, line 112: Is this float always culprit, then we should review the whole file. There is another 1/

joernu76: The problem is that 1/ms means "per millisecond", i.e. kHz, whereas we want a speed, i.e. meters per second. The other units should be fine.

ReimarBauer (approved):
Thx

PR #1491, opened 2022-06-23
I1489

### LINKED ISSUE #1489: Linear view does not support layers without init_time
Log:

```
[Sun Jun 19 20:48:39.716374 2022] [wsgi:error] [pid 269769:tid 140012983400192] [remote 134.94.152.46:36348]     image = self.plot_object.plot_lsection(data, self.lats, self.lons,
[Sun Jun 19 20:48:39.716376 2022] [wsgi:error] [pid 269769:tid 140012983400192] [remote 134.94.152.46:36348]   File "/home/mss-campaigns2019/miniconda3/envs/stable/lib/python3.9/site-packages/mslib/mswms/mpl_lsec.py", line 107, in plot_lsection
[Sun Jun 19 20:48:39.716377 2022] [wsgi:error] [pid 269769:tid 140012983400192] [remote 134.94.152.46:36348]     node.appendChild(xmldoc.createTextNode(self.init_time.strftime("%Y-%m-%dT%H:%M:%SZ")))
[Sun Jun 19 20:48:39.716379 2022] [wsgi:error] [pid 269769:tid 140012983400192] [remote 134.94.152.46:36348] AttributeError: 'NoneType' object has no attribute 'strftime'

```

joernu76:


mslib/mswms/mpl_vsec.py
```diff
@@ -346,3 +348,5 @@ def plot_vsection(self, data, lats, lons, valid_time, init_time,
 
             # Return the XML document as formatted string.
             return xmldoc.toprettyxml(indent="  ")
+        else:
+            raise RuntimeError
```
ReimarBauer, line 352: should we log this error?

mslib/mswms/mss_plot_driver.py
```diff
@@ -533,6 +534,9 @@ def plot(self):
         else:
             resolution = (-1, -1)
 
+        if self.return_format not in ("image/png", "text/xml"):
```
ReimarBauer, line 537: I'm not sure if we have to name it by a prefix of "return_"  Also we could use the name mimetype which links it to e.g. https://wiki.selfhtml.org/wiki/MIME-Type/%C3%9Cbersicht

mslib/mswms/wms.py
```diff
@@ -329,7 +329,8 @@ def generate_gallery(self, create=False, clear=False, generate_code=False, sphin
                                         plot_driver.set_plot_parameters(**kwargs,
                                                                         lsec_path=[[0, 0, 20000], [1, 1, 20000]],
                                                                         lsec_numpoints=201,
-                                                                        lsec_path_connection="linear")
+                                                                        lsec_path_connection="linear",
+                                                                        return_format="text/xml")
```
ReimarBauer, line 333: mimetype we used already on few other places (e.g. wms_control doc string, index.py), I think that is a better name,

ReimarBauer (changes requested):
see comments

mslib/mswms/mpl_vsec.py
joernu76: The code should not be able to get here at all as wrong formats should be caught before going here.
The RuntimeError will be caught in wms.py and logged there. This will not be helpful to anyone but a developer as such a log entry is not needed.

mslib/mswms/mss_plot_driver.py
joernu76: maybe. I did not invent the name. As this goes to stable, I try to keep changes minimal. These are already more changes than I intended to do.

mslib/mswms/wms.py
joernu76: I would do such a rename in develop, not in stable.

ReimarBauer (approved):
Thx

PR #1496, opened 2022-06-26
Refactor tutorials

ReimarBauer:


ReimarBauer:
in principle we now can use `xvfb-run --server-args="-screen 0 1920x1080x24" ./start_tutorial.sh python ./tutorial_NAME`
by enabling the fluxbox window manager in the start_tutorial.sh

I continue with that after I learned how to configure QT backgrounds of the window manager to the same grey as I have stored in the images for my common window manager.

joernu76 (approved):

Issue #1511, opened 2022-07-21
WMS server does not discard dataless layer under certain circumstances


joernu76:
Normal behaviour: WMS server does not serve data-less layers in its capability document
Current behavour: under certain circumstances, they are listed with wrong time information, causing a later failure when trying to retrieve data.

Cause: The logic for collecting "valid times" from multiple data variables is flawed, if *only* the first checked variable has *no* data.

PR #1512, opened 2022-07-21
Now properly aggregate available data datetimes

joernu76:
Fix issue #1511


jgrooss (approved):
looks ok to me and does work for me

PR #1513, opened 2022-08-04
Improve updater

### LINKED ISSUE #1409: Automatic updater does not work correctly
currently the newest mss version is 6.0.5
If on 6.0.4 UI the updater in the Help menu of the UI is executed, it does not reckognise the new version.
Similar behaviour was seen before

ReimarBauer:


mslib/msui/updater.py
```diff
@@ -40,19 +42,26 @@ def __init__(self, parent=None):
...
-        self.updater.on_update_finished.connect(lambda: self.btRestart.setEnabled(True))
-        self.btUpdate.clicked.connect(lambda: (self.updater.update_mss(), self.btUpdate.setEnabled(False)))
-        self.btRestart.clicked.connect(self.updater._restart_msui)
-        self.updater.run()
+        self.updater = None
+        try:
+            self.updater = Updater()
+        except ModuleNotFoundError as ex:
+            # https://github.com/Open-MSS/MSS/issues/1409#issuecomment-1200946622
+            logging.error("unexpected error in updater: %s %s in version: %s", type(ex), ex, __version__)
```
ReimarBauer, line 50: The updater is already on startup called. 
As soon as we know other problems, we can issue a tutorial here. The problem that led to these lines was already very complex.

mslib/msui/updater.py
```diff
@@ -40,19 +42,26 @@ def __init__(self, parent=None):
...
+
+        if self.updater is not None:
```
joernu76, line 52: else: in the sense of try: except: else: ?

mslib/utils/qt.py
```diff
@@ -446,13 +446,15 @@ def __init__(self, parent=None):
         self.is_git_env = False
         self.new_version = None
         self.old_version = None
-        self.command = "conda"
-
+        # we are using the installer version of the env
+        self.conda_prefix = os.getenv("CONDA_PREFIX")
+        self.command = os.path.join(self.conda_prefix, 'bin', "conda")
```
joernu76, line 451: There is a naked "conda" below in line ~490

mslib/utils/qt.py
```diff
@@ -446,13 +446,15 @@ def __init__(self, parent=None):
...
+        self.mamba_cmd = os.path.join(self.conda_prefix, 'bin', 'mamba')
```
joernu76: Why self.mamba_cmd? No "self.", please

joernu76 (dismissed):

mslib/utils/qt.py
ReimarBauer: ok, looking

mslib/utils/qt.py
ReimarBauer: ouch, that has to become self.command

joernu76 (approved):

PR #1514, opened 2022-08-08
Draft for making styles configurable.

### LINKED ISSUE #1509: Make "styles" for generic plots configurable.
Styles for generic plots are hard-coded but should be sensibly extendable.

joernu76:
Fix #1509


joernu76:
@crolf Moin, Can you have a look at the ice cloud styles? I do not understand the use of dataname. In case a "different" dataname is used the program will crash because clev was not set. Is this intended?

crolf:
Moin moin @joernu76, yes that's true. And of course it is not intended to crash the program ;-) The plotting classes where only intended to use for the ice_cloud products. I wanted to avoid to create too many styles. But maybe it would be better to create one plot style for each of the three parameter:
- ice_cloud_number
- ice_cloud_iwc
- ice_cloud_radius
and not to use the dataname if statements.
What do you think ?

joernu76:
> Moin moin @joernu76, yes that's true. And of course it is not intended to crash the program ;-) The plotting classes where only intended to use for the ice_cloud products. I wanted to avoid to create too many styles. But maybe it would be better to create one plot style for each of the three parameter:
> 
>     * ice_cloud_number
> 
>     * ice_cloud_iwc
> 
>     * ice_cloud_radius
>       and not to use the dataname if statements.
>       What do you think ?

Agree. The intent of this change is to move the definition of new styles *from* MSS core code *to* mswms_settings.py files such that the list does not grow too large.

I'll prepare a draft.

crolf (dismissed):
Looks good to me! It is a much cleaner solution than before.

mslib/mswms/utils.py
```diff
@@ -295,6 +299,235 @@ def get_log_levels(cmin, cmax, levels=N_LEVELS):
     return clev
 
 
+def _style_default(_dataname, _style, cmin, cmax, cmap, _data):
```
ReimarBauer, line 302: _data and _style is not used in the functions?

joernu76:
@ReimarBauer Why isn't Christian's review sufficient?

mslib/mswms/utils.py
```diff
@@ -295,6 +299,235 @@ def get_log_levels(cmin, cmax, levels=N_LEVELS):
...
+    return cmin, cmax, clev, cmap, norm, None
+
+
+def _style_auto(_dataname, _style, cmin, cmax, cmap, data):
+    cmin_p = data.min()
+    cmax_p = data.max()
+    if not any([isinstance(_x, np.ma.core.MaskedConstant) for _x in (cmin_p, cmax_p)]):
+        cmin, cmax = cmin_p, cmax_p
+    if cmin == cmax:
+        cmin, cmax = 0, 1
```
ReimarBauer: cmin, cmax are here integers

mslib/mswms/utils.py
```diff
@@ -295,6 +299,235 @@ def get_log_levels(cmin, cmax, levels=N_LEVELS):
...
+    if 0 < cmin < 0.05 * cmax:
+        cmin = 0.
```
ReimarBauer, line 316: here it is a float

mslib/mswms/utils.py
```diff
@@ -295,6 +299,235 @@ def get_log_levels(cmin, cmax, levels=N_LEVELS):
...
+
+
+def _style_tropopause_altitude(_dataname, _style, cmin, cmax, cmap, _data):
+    cmap = matplotlib.pyplot.cm.terrain
+    clev = np.arange(5, 18.1, 0.25)
+    return cmin, cmax, clev, cmap, None, None
+
+
+def _style_ertel_potential_vorticity(_dataname, style, cmin, cmax, cmap, _data, hemisphere):
+    colors = [
```
ReimarBauer, line 353: hmm. some don't have the hemisphere as _hemisphere placeholder

mslib/mswms/utils.py
```diff
@@ -309,153 +542,12 @@ def get_style_parameters(dataname, style, cmin, cmax, data):
...
-             (0.212, 0.114, 0.075, 1.0),
-             ], name="ice_map")
-        if dataname == "mean_mass_radius_of_cloud_ice_crystals":
-            clev = np.array([0, 1, 10, 20, 30, 40, 50, 60, 70, 80, 90, 100])
-        norm = matplotlib.colors.BoundaryNorm(clev, cmap.N)
-    else:
-        raise RuntimeError(f"Illegal plotting style?! ({style})")
+    try:
+        cmin, cmax, clev, cmap, norm, ticks = \
+            _styles[style](dataname, style, cmin, cmax, cmap, data)
```
ReimarBauer: ok, here are always 6 arguments passed and there are few by a name convention which have 7, which get translated

mslib/mswms/utils.py
```diff
@@ -309,153 +542,12 @@ def get_style_parameters(dataname, style, cmin, cmax, data):
...
+    except BaseException:
+        logging.error("Illegal plotting style '%s' for dataname '%s'", style, dataname)
+        raise
     if clev[0] == clev[-1]:
```
ReimarBauer, line 578: min len of clev is 2

ReimarBauer (dismissed):
We should document in one of the functions or in the docs the meaning of the parameters and theire type.

ReimarBauer:
> Why isn't Christian's review sufficient?

github destinguishs between contributor and first time contributor.

crolf has not done yet a MR. 

On the first MR of a new user also one of us have to enable the tests 

Christian has done code commits on our old home on bitbucket with a different name, https://www.openhub.net/p/mss/contributors?query=christian&time_span=&sort=commits

From the documentation another point:
https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/about-pull-request-reviews
> If your repository requires approving reviews from people with write or admin permissions, then any approvals from people with these permissions are denoted with a green check mark, and approvals from people without these permissions have a gray check mark. Approvals with a gray check mark do not affect whether the pull request can be merged.

Last time we run in this I enabled owner for Sonja. But that don't scale. We should verify if a simple commit does the same.

mslib/mswms/utils.py
```diff
@@ -295,167 +299,282 @@ def get_log_levels(cmin, cmax, levels=N_LEVELS):
...
+    except (ValueError, TypeError):
+        logging.error("Illegal number of arguments/return values for plotting style '%s' "
+                      "and dataname '%s'", style, dataname)
+        raise
+    except BaseException:
+        logging.error("Unknown error in style call: plotting style '%s' "
+                      "and dataname '%s'", style, dataname)
+        raise
     if clev[0] == clev[-1]:
         cmin, cmax = 0, 1
```
ReimarBauer: Here is another INT definition

ReimarBauer (changes requested):
missing int/float in utils

ReimarBauer (approved):
Thx

PR #1543, opened 2022-08-29
updater needs to take care not only started from an env

### LINKED ISSUE #1541: The updater should not crash in a docker container or on start without an environment (linux msui icon on program manager)
`docker run   --net=host -ti --rm -e DISPLAY=$DISPLAY -v /tmp/.X11-unix/:/tmp/.X11-unix openmss/mss MSS`

```
  File "/opt/conda/envs/mssenv/bin/mswms", line 11, in <module>
    sys.exit(main())
  File "/opt/conda/envs/mssenv/lib/python3.9/site-packages/mslib/mswms/mswms.py", line 89, in main
    updater = Updater()
  File "/opt/conda/envs/mssenv/lib/python3.9/site-packages/mslib/utils/qt.py", line 451, in __init__
    self.command = os.path.join(self.conda_prefix, 'bin', "conda")
  File "/opt/conda/envs/mssenv/lib/python3.9/posixpath.py", line 76, in join
    a = os.fspath(a)
TypeError: expected str, bytes or os.PathLike object, not NoneType
2022-08-29 18:21:03          excepthook || MSS Version: 7.0.3.
2022-08-29 18:21:03          excepthook || Python Version: 3.9.13 | packaged by conda-forge | (main, May 27 2022, 16:56:21) 
[GCC 10.3.0]
2022-08-29 18:21:03          excepthook || Platform: Linux-5.4.0-124-generic-x86_64-with-glibc2.31 (('64bit', 'ELF'))
2022-08-29 18:21:03          excepthook || Fatal error: Traceback (most recent call last):
  File "/opt/conda/envs/mssenv/bin/mswms", line 11, in <module>
    sys.exit(main())
  File "/opt/conda/envs/mssenv/lib/python3.9/site-packages/mslib/mswms/mswms.py", line 89, in main
    updater = Updater()
  File "/opt/conda/envs/mssenv/lib/python3.9/site-packages/mslib/utils/qt.py", line 451, in __init__
    self.command = os.path.join(self.conda_prefix, 'bin', "conda")
  File "/opt/conda/envs/mssenv/lib/python3.9/posixpath.py", line 76, in join
    a = os.fspath(a)
TypeError: expected str, bytes or os.PathLike object, not NoneType
```

The problem was introduced by https://github.com/Open-MSS/MSS/pull/1513/files#diff-5c39c5f9af4c2325fd21b9d539f813296d580fb6b67b960340a4bb8a9b1e6fbeR451

os.getenv("CONDA_PREFIX") seems to be None in the docker container and also when we start msui by the icon.

in case of None we should fall back to the conda command.

ReimarBauer:


joernu76 (commented):
I am not sure that this fix properly addresses the issue as it addresses the symptom not the cause.
MSS should not be started in a non-activated environment as this will cause all kinds of issues.
We thus need to 
a) make sure that the environment is activated within the docker. I have not looked at the "execute MSS" container in a while. We should update it along the lines/based on of our "testing container".
b) Switch over the start via icon from calling into the environment towards starting a script that activates the correct environment first.

ReimarBauer:
@joernu76 agreed to a) we should combine the old dockerhub with the new dockertesting

solving b) is in stable not that easy for us.  It looks to me that the issue currently blocks linux and mac entry points. 
For windows it is different implemented. Best solution is that this gets upstream fixed.

I checked that 7.0.3 on windows can be started by the icon. A gifcycle is in the issue.

Working on an improved dockersetup is independent and will be done after we merged this or similar. The different behaviour between windows and linux is adressed by that minor release.

joernu76 (approved):

PR #1544, opened 2022-09-02
updated_multiple_flightpath_dockwidget

Jatin2020-24:
listviews


ReimarBauer:
fine, the one which is activated (highlighted bold) should have a check. So that activation means check it too. And the state of this should only be changeable by activating another one on the msui main window.

ReimarBauer:
you need to update names, when one saves

name_update

a new topview has updated names

ReimarBauer:
When I have many in the view and I toggle the one which I can edit, the other previous seen should toggle to a line (visible). Currently I would have to reactivate them again.

previous_should_be_shown_too

mslib/msui/multiple_flightpath_dockwidget.py
```diff
@@ -92,177 +92,132 @@ def __init__(self, parent=None, view=None, listView=None, activeFlightTrack=None
...
+    def update(self):
+        for entry in self.dict_files.values():
+            entry["patch"].update()
+
+    def remove(self):
+        for entry in self.dict_files.values():
+            entry["patch"].remove()
+
+    def wait(self, parent, start, end):
+        t1 = threading.Timer(1.0, self.flighttrackAdded, [parent, start, end])
```
ReimarBauer: maybe make one method with parameters instead of enumbering?

mslib/msui/multiple_flightpath_dockwidget.py
```diff
@@ -92,177 +92,132 @@ def __init__(self, parent=None, view=None, listView=None, activeFlightTrack=None
...
+        t1.start()
+
+    def wait2(self):
+        t2 = threading.Timer(2.0, self.flagop)
+        t2.start()
+
+    def flagop(self):
+        if self.flag == 1:       # New flighttrack added
```
ReimarBauer: use better names here, e.g. flighttrack_added


If 1 is meant as True use boolean too.

mslib/msui/multiple_flightpath_dockwidget.py
```diff
@@ -92,177 +92,132 @@ def __init__(self, parent=None, view=None, listView=None, activeFlightTrack=None
...
+            self.flag = 0
+        elif self.flag1 == 1:    # Flighttrack activated
```
ReimarBauer: naming

mslib/msui/multiple_flightpath_dockwidget.py
```diff
@@ -92,177 +92,132 @@ def __init__(self, parent=None, view=None, listView=None, activeFlightTrack=None
...
+        wp_model = self.listView.item(start).flighttrack_model
+        listItem = self.create_list_item(wp_model)
+        self.save_waypoint_model_data(wp_model)
 
-    def select_file(self, filenames):
+        self.drawInactiveFlighttracks()
+        self.activate_flighttrack(listItem)
+
+    def save_waypoint_model_data(self, wp_model):
+        lst = []
```
ReimarBauer: better name please,  or may be a list comprehension?

mslib/msui/multiple_flightpath_dockwidget.py
```diff
@@ -92,177 +92,132 @@ def __init__(self, parent=None, view=None, listView=None, activeFlightTrack=None
...
-        Save Flighttrack settings after closing of view.
+        Activate flighttrack
         """
-        for entry in self.dict_files.values():
-            entry["track"] = None
-        settings = {
-            "saved_files": self.dict_files
-        }
-        save_settings_qsettings(self.settings_tag, settings)
+        if self.flag != 1:
```
ReimarBauer: when it could be a boolean this would be better readable

ReimarBauer (changes requested):
Hi @Jatin2020-24 this already looks good
see comments.

ReimarBauer:
When we use the color changer from options it somehow toggles color with the other 
color

mslib/msui/multiple_flightpath_dockwidget.py
Jatin2020-24: I'll do that

mslib/msui/multiple_flightpath_dockwidget.py
Jatin2020-24: I'll change it

ReimarBauer:
we can switch between docking widgets.  The selection what is shown should only been controlledf by the multiple Flightpath widget. so e.g. adding data by the layer chooser needs to take this in account too. (all other docking widgets too)

using_layerchooser

mslib/msui/multiple_flightpath_dockwidget.py
```diff
@@ -24,33 +24,29 @@
...
-from PyQt5 import QtWidgets, QtCore
+from PyQt5 import QtWidgets, QtGui, QtCore
 from mslib.msui.qt5 import ui_multiple_flightpath_dockwidget as ui
-from mslib.utils.qt import get_open_filenames
-from mslib.utils.config import load_settings_qsettings, save_settings_qsettings
-from fs import open_fs
-import mslib.msui.flighttrack as ft
 from mslib.msui import msui
-import xml
+import threading
```
ReimarBauer, line 34: have a look on https://www.pythonguis.com/tutorials/multithreading-pyqt-applications-qthreadpool/
 QThreadPool()

for  QtCore.QThread() we have a worker in https://github.com/Open-MSS/MSS/blob/develop/mslib/utils/qt.py#L370
https://www.youtube.com/watch?v=1FZO4VKANbM

ReimarBauer:
The previous visible selected flightpath for editing has no mark in [a]. This should switch to marked, So that it is on, when one selects [b] for editing. A user has to unselect [a] to get it removed from multiple flightpath docking widget.
The image shows now [a] and [b] unmarked.

If I would select [c] for editing and then [d] all are unmarked. 

toggle

ReimarBauer:
When I remove a not saved flight path it gives a traceback, it is then removed in the flightpath list, not yet in the docking widget.
remove_problem


```
Fatal error in MSS 7.0.1. on Linux-5.4.0-128-generic-x86_64-with-glibc2.27
Python 3.10.4 | packaged by conda-forge | (main, Mar 24 2022, 17:38:57) [GCC 10.3.0]

  File "/home/reimar/PycharmProjects/2021/Jatin2020-24/MSS/mslib/msui/multiple_flightpath_dockwidget.py", line 231, in flighttrackRemoved
    self.dict_files[self.list_flighttrack.item(start).flighttrack_model]["patch"].remove()
AttributeError: 'NoneType' object has no attribute 'remove'
```

ReimarBauer:
When I don't save local flightpathes, they are not correct shown in the docking widget

local_unsaved_flightpath

mslib/msui/multiple_flightpath_dockwidget.py
```diff
@@ -84,185 +90,200 @@ class MultipleFlightpathControlWidget(QtWidgets.QWidget, ui.Ui_MultipleViewWidge
...
 
-    def syncListViews(self):
-        self.list_flighttrack.clear()
+        # Load flighttracks
         for index in range(self.listView.count()):
             wp_model = self.listView.item(index).flighttrack_model
-            msui.QFlightTrackListWidgetItem(wp_model, self.list_flighttrack)
+            listItem = self.create_list_item(wp_model, self.list_flighttrack)
+
+        # for index in range(self.listOperationsMSC.count()):
```
ReimarBauer: when these lines are not needed may be remove them

mslib/msui/multiple_flightpath_dockwidget.py
```diff
@@ -84,185 +90,200 @@ class MultipleFlightpathControlWidget(QtWidgets.QWidget, ui.Ui_MultipleViewWidge
...
+    def __init__(self, parent=None, view=None, listView=None,
+                 listOperationsMSC=None, activeFlightTrack=None):
         super(MultipleFlightpathControlWidget, self).__init__(parent)
+        self.listOperationsMSC = listOperationsMSC
         self.listView = listView
         self.ui = parent
         self.setupUi(self)
         self.view = view  # canvas
         self.flight_path = None  # flightpath object
+        self.colour = None
```
ReimarBauer: self.color?

Just seen that we used it different written all over MSS, on other places you use self.color.

mslib/msui/multiple_flightpath_dockwidget.py
```diff
@@ -84,185 +90,200 @@ class MultipleFlightpathControlWidget(QtWidgets.QWidget, ui.Ui_MultipleViewWidge
...
+            entry["patch"].update()
 
-    def get_file(self):
+    def remove(self):
+        for entry in self.dict_files.values():
+            entry["patch"].remove()
+
+    def wait(self, parent, start, end):
+        self.flighttrack_added = True
+        t1 = threading.Timer(0.5, self.flighttrackAdded, [parent, start, end])
```
ReimarBauer, line 228: I'm not sure if this is needed when signals are used, is it?

mslib/msui/multiple_flightpath_dockwidget.py
Jatin2020-24: Removed

mslib/msui/multiple_flightpath_dockwidget.py
Jatin2020-24: these lines are needed to add mscolab operations in list

ReimarBauer:
good improvement :)

I found another problem, or inconsity ;)

We currently define the flight track color in the options (you get this by the options button) This has precedence. We don't have there line thickness.

For the color we maybe can solve this by:
Use color defined in options for initial color of active flight path. afterwards deactivate the color change button in options and it needs also the check mark for enabled, but can't be changed (disabled). At the moment the dockingwidget is closed the button and checkmark has to become activated again.


options

When the precidence is given any activation of a flight track will toggle the defined color in the widget.


color

Other ideas for solving this are welcome too.

ReimarBauer:
When the docking widget is closed we should only see the activated flight path

still_four

ReimarBauer:
When we activate an operation we need to be able to uncheck all multiple local flightpath. Currently the last activated is unchangeable.

mscolab

mslib/msui/multiple_flightpath_dockwidget.py
```diff
@@ -106,14 +104,18 @@ def __init__(self, parent=None, view=None, listView=None,
         # Set flags
         self.flighttrack_added = False
         self.flighttrack_activated = False
-        self.color_change = None
+        self.color_change = False
+        self.change_linewidth = False
+        self.dsbx_linewidth.setValue(2.0)
```
ReimarBauer, line 142: Add a `# ToDo` to use an invented constant from https://github.com/Open-MSS/MSS/blob/develop/mslib/utils/config.py#L227 for initializing this.

mslib/msui/multiple_flightpath_dockwidget.py
```diff
@@ -172,7 +174,8 @@ def create_list_item(self, wp_model, listWidget):
         # Create new key in dict
         self.dict_flighttrack[wp_model] = {}
         self.dict_flighttrack[wp_model]["patch"] = None
-        self.dict_flighttrack[wp_model]["color"] = 'blue'
+        self.dict_flighttrack[wp_model]["color"] = None
+        self.dict_flighttrack[wp_model]["linewidth"] = 2
```
ReimarBauer, line 298: This should be a float

ReimarBauer:
changing linewidth works, when toggling activation it falls to defaul. Linewidth is currently not so important.

I got a traceback on the activated flight track playing wiht line width and color
```
  File "/home/reimar/PycharmProjects/2021/Jatin2020-24/MSS/mslib/msui/multiple_flightpath_dockwidget.py", line 207, in select_color
    self.dict_flighttrack[wp_model]["patch"].update(color=
AttributeError: 'NoneType' object has no attribute 'update'
CRITICAL: MSS Version: 7.0.1.
CRITICAL: Python Version: 3.10.6 | packaged by conda-forge | (main, Aug 22 2022, 20:35:26) [GCC 10.4.0]
CRITICAL: Platform: Linux-5.4.0-131-generic-x86_64-with-glibc2.27 (('64bit', 'ELF'))
  File "/home/reimar/PycharmProjects/2021/Jatin2020-24/MSS/mslib/msui/multiple_flightpath_dockwidget.py", line 207, in select_color
    self.dict_flighttrack[wp_model]["patch"].update(color=
AttributeError: 'NoneType' object has no attribute 'update'
```

Jatin2020-24:
I've written most of the functions to plot operations in dockwidget. Only connections of signals and slots is remaining.

mslib/msui/msui.py
```diff
@@ -491,6 +494,10 @@ def add_plugins(self):
         self.add_import_plugins(picker_default)
         self.add_export_plugins(picker_default)
 
+    @QtCore.Slot(int)
+    def ss(self, active_op_id):
```
ReimarBauer: use a meaningful name here

ReimarBauer:
Only the activated Operation or flight track should be protected against removing the checkmark
In my example (d) as operation is activated. I can't change the state of the "new flight track (1)".

protected

mslib/msui/multiple_flightpath_dockwidget.py
```diff
@@ -84,185 +105,423 @@ class MultipleFlightpathControlWidget(QtWidgets.QWidget, ui.Ui_MultipleViewWidge
...
+        self.view = view
+
+        self.operation_added = False
+        self.operation_removed = False
+        self.operation_activated = False
+
+        # Connect signals and slots
+        self.list_operation_track.itemChanged.connect(self.set_flag)
+        self.listOperationsMSC.model().rowsInserted.connect(self.wait2)
+        self.listOperationsMSC.model().rowsRemoved.connect(self.operationRemoved)
```
Jatin2020-24: On removing operations, `rowsInserted()` signal is also emitted and on inserting operations -- `rowsRemoved()` signal is also emitted. I'm not able to find why it's happening?

mslib/msui/kmloverlay_dockwidget.py
```diff
@@ -478,7 +478,7 @@ def select_file(self, filenames):
                 return
             text = filename
             if text not in self.dict_files:  # prevents same file being added twice
-                # initializing the nested dictionary dict_files
+                # initializing the nested dictionary dict_flighttrack
                 self.dict_files[text] = {}
                 self.dict_files[text]["patch"] = None
                 self.dict_files[text]["color"] = (0, 0, 0, 1)
```
ReimarBauer, line 484: search for "colour_ft_vertices" this may be the initial color

mslib/msui/multiple_flightpath_dockwidget.py
ReimarBauer: have a look on this bug fix. May be this is causing what you've seeing
https://github.com/Open-MSS/MSS/pull/1561/files

ReimarBauer: Because many OP creator can change access to an operation it may be better to use something similar to the reload_operation_list  https://github.com/Jatin2020-24/MSS/blob/update-multiple-flightpath-dockwidget/mslib/msui/mscolab.py#L1269
or get the data after this was called

In principle.
A user who adds me to an operation, triggers an event on the server. This triggers my UI to request the new list.

If I remove an operation, I send a request to the server. The server sends an event to users. This triggers the UI to update.

every time reload_operation_list should have set the data you need in the docking widget

ReimarBauer:
Found a problem, when on an activated mscolab operation the change color button is pressed

(b) is activated

updating_color_for_active_msc

ReimarBauer:
on logout the list is not cleared

on_logout

ReimarBauer:
when I login again and the dockingwidget is opened I see a traceback and no update of the list (empty)

on_login_by_opened_widget

multiple time pressing ok, gets over this.

After closing/reopening the view I get the operations into the list

The error message comes then on any activation click of an OP.


```
Fatal error in MSS 7.0.1. on Linux-5.4.0-131-generic-x86_64-with-glibc2.27
Python 3.9.13 | packaged by conda-forge | (main, May 27 2022, 16:58:50) 
[GCC 10.3.0]

  File "/home/reimar/PycharmProjects/2021/Jatin2020-24/MSS/mslib/msui/multiple_flightpath_dockwidget.py", line 158, in <lambda>
    self.listOperationsMSC.itemClicked.connect(lambda: self.list_flighttrack.setCurrentItem(None))
RuntimeError: wrapped C/C++ object of type QListWidget has been deleted

```

ReimarBauer:
Importing an ftml file gives a traceback
```
Fatal error in MSS 7.0.1. on Linux-5.4.0-131-generic-x86_64-with-glibc2.27
Python 3.9.13 | packaged by conda-forge | (main, May 27 2022, 16:58:50) 
[GCC 10.3.0]

  File "/home/reimar/PycharmProjects/2021/Jatin2020-24/MSS/mslib/msui/msui.py", line 622, in handle_import_local
    self.create_new_flight_track(filename=filename, function=function)
  File "/home/reimar/PycharmProjects/2021/Jatin2020-24/MSS/mslib/msui/msui.py", line 681, in create_new_flight_track
    waypoints_model = ft.WaypointsTableModel(filename=filename)
  File "/home/reimar/PycharmProjects/2021/Jatin2020-24/MSS/mslib/msui/flighttrack.py", line 169, in __init__
    self.load_from_ftml(filename)
  File "/home/reimar/PycharmProjects/2021/Jatin2020-24/MSS/mslib/msui/flighttrack.py", line 612, in load_from_ftml
    _dirname, _name = os.path.split(filename)
  File "/home/reimar/miniconda3/envs/mssdev/lib/python3.9/posixpath.py", line 103, in split
    p = os.fspath(p)
TypeError: expected str, bytes or os.PathLike object, not list
```


crash

ReimarBauer:
good idea with the hint, I'm able to show mscolab data but color can't be changed?

mscolab_and_hint

The limitation is also for flight pathes given when an operation is activated.

ReimarBauer:
changing by options an activated flighpath gives an error. It is independent if highlighted or not.

```
Fatal error in MSS 7.0.1. on Linux-5.4.0-131-generic-x86_64-with-glibc2.27
Python 3.9.13 | packaged by conda-forge | (main, May 27 2022, 16:58:50) 
[GCC 10.3.0]

  File "/home/reimar/PycharmProjects/2021/Jatin2020-24/MSS/mslib/msui/multiple_flightpath_dockwidget.py", line 251, in ft_vertices_color
    self.operations.ft_color_update(color)
AttributeError: 'MultipleFlightpathControlWidget' object has no attribute 'operations'
```

option_color_change_problem

By clicking ok on the error it continues and sets the color

mslib/msui/qt5/ui_multiple_flightpath_dockwidget.py
```diff
@@ -60,6 +113,20 @@ def setupUi(self, MultipleViewWidget):
     def retranslateUi(self, MultipleViewWidget):
         _translate = QtCore.QCoreApplication.translate
         MultipleViewWidget.setWindowTitle(_translate("MultipleViewWidget", "Form"))
-        self.bt_addFile.setText(_translate("MultipleViewWidget", "Add FTML Files"))
-        self.btRemove_file.setText(_translate("MultipleViewWidget", "Remove Flighttrack"))
+        self.ft_color_label.setText(_translate("MultipleViewWidget", "(Color shown before flighttracks/operations are only visible in \"deactivated\" state) 𝐀𝐜𝐭𝐢𝐯𝐚𝐭𝐞𝐝 𝐓𝐫𝐚𝐜𝐤 𝐕𝐞𝐫𝐭𝐢𝐜𝐞𝐬 𝐂𝐨𝐥𝐨𝐫 : "))
```
ReimarBauer: I'm not sure about the meaning of "(Color shown before flighttracks/operations are only visible in \"deactivated\" state)"
I would skip it the most important message is the bold  𝐀𝐜𝐭𝐢𝐯𝐚𝐭𝐞𝐝 𝐓𝐫𝐚𝐜𝐤 𝐕𝐞𝐫𝐭𝐢𝐜𝐞𝐬 𝐂𝐨𝐥𝐨𝐫 :

ReimarBauer:
updating permissions is not yet implemented

updating_permissions_not_yet

When access is removed the old operation persists in the docking widget
Peek 2022-11-06 15-39

mslib/msui/qt5/ui_multiple_flightpath_dockwidget.py
```diff
@@ -60,6 +113,20 @@ def setupUi(self, MultipleViewWidget):
...
+        self.list_flighttrack.setToolTip(_translate("MultipleViewWidget", "List of Open Flighttracks.\n"
+"Check box to activate and display track on topview."))
+        self.list_operation_track.setToolTip(_translate("MultipleViewWidget", "List of Mscolab Operations.\n"
+"Check box to activate and display track on topview."))
+        self.pushButton_color.setText(_translate("MultipleViewWidget", "Change Color"))
         self.labelStatus.setText(_translate("MultipleViewWidget", "Status: "))
+
+
+if __name__ == "__main__":
```
ReimarBauer: remove the main part, this is likly created by an option of the ui convertor

mslib/msui/qt5/ui_multiple_flightpath_dockwidget.py
ReimarBauer: May be it should also state Flighttrack or Operation.

ReimarBauer:
When I start with flighttracks, c is selected and I compare with data from mscolab server and check there other operations.
At the moment I activate an other flighttrack the selecction of mscolab is cleared

clears_checkmarks

ReimarBauer:
on mscolab operations linethick seems not implemented. I can only change this on flight path.

ReimarBauer:
Trying line thick on acivated flight path shows a problem

 
```
Fatal error in MSS 7.0.1. on Linux-5.4.0-131-generic-x86_64-with-glibc2.27
Python 3.9.13 | packaged by conda-forge | (main, May 27 2022, 16:58:50) 
[GCC 10.3.0]

  File "/home/reimar/PycharmProjects/2021/Jatin2020-24/MSS/mslib/msui/multiple_flightpath_dockwidget.py", line 364, in set_linewidth
    self.dict_flighttrack[wp_model]["patch"].remove()
AttributeError: 'NoneType' object has no attribute 'remove'

```
changing_line_thick_on_acktivated_flightpath_crash

ReimarBauer:
please remove the "test develop / test-filter-flight-paths" from the test configuration

ReimarBauer:
mostly good :) found a little UI problem. When a user gets access to an operation but a flight track is activated we get now twice bold names in both navigations. 

access_in_mscolab

When it is mscolab then it changes the activation bold formatting to the new one. E.g- (a) is activated but vvv gets bold formatting.

ReimarBauer:
activation_clear_checkmarks_on_mscolab

only the user should uncheck. Currently we can uncheck one site by selecting a flight track or a mscolab operation

ReimarBauer:
Currently I can't change thickness for mscolab lines.

ReimarBauer:
status

Seems I don't understand the text in the brackets. For me it would be enough to read the part after

ReimarBauer:
I set red in options for an activated track. Starting with (d) I activated (b) this keeps the green color, in the options it is red. 
This needs to be consistent, it could be get red or have to change the options color to green and also change the activated track color to green.
Keeping it green would not give a visual change to the user but then the options color for the active flight track has to get this color.
choosing  the other color now in option will get everything changed when a user toggles throught the tracks. That is why I prefer to keep it in the users color in this case green.



activated_color_update

ReimarBauer:
I slected now (d) in mscolab. I set in options blue as color. The line is also drawn in blue. the Active color is blue. The color in the widget remains green.
color_update_in_mscolab

ReimarBauer:
In mscolab one can create a new operation, this is not getting default activated. Also getting access to an operation does not mean this gets activated. 

This is different in the standalone flight tracks setup, there any new flight track is activated. This makes no sense in mscolab. (We likly change this behaviour also with local flight tracks)

I think this mismatch creates the wrong bold names in the docking widget for mscolab operations.

ReimarBauer:
We need a way to leave the mode. One option could be on closing the docking widget. When doing this we should keep only the activated one. Nothing else.

get_back_only_one

ReimarBauer:
We decided earlier that any activated gets the check mark. This has a site effect by clicking any available flight track means activates all.

This should become documented, e.g. comment or description

ReimarBauer:
There is now a little delay on activation until it is seen, can the speed improved?

ReimarBauer:
When a topview is started when some data is loaded, all is shown in the docking widget with flight track color. Changing the color in the options menu does not update the color for the active flight track.
The first part of the sample video shows single data mode, without the docking widget. The problem is shown on the second top view.

color_changed_needs_to_mark active_one

This is done on a new activation.

ReimarBauer (approved):
Thx

PR #1550, opened 2022-09-04
Refactor plotting of views

swsrkty:


mslib/msui/mpl_qtwidget.py
```diff
@@ -490,57 +922,32 @@ class MplSideViewCanvas(MplCanvas):
...
-            self.settings_dict.update(settings)
+            self.myfig.settings_dict.update(settings)
 
         # Setup the plot.
         self.update_vertical_extent_from_settings(init=True)
 
         self.numlabels = numlabels
-        self.ax2 = self.ax.twinx()
-        self.ax.patch.set_facecolor("None")
+        self.ax2 = self.myfig.ax.twinx()
```
joernu76: shouldn't the ax2 be part of the MySideViewFigure class?

joernu76: Indeed, you have added it to the correct class already. This must be removed.

PR #1552, opened 2022-09-04
tests moved to a seperate dir

### LINKED ISSUE #1545: Index Error pop from empty list when pytest is launched
While launching pytest mslib on WIndows, an index error is observed. Potentially due to the backslash that is considered as an escape key.

issue_mss

### LINKED ISSUE #1547: move tests to its own dir parallel to mslib
We don't need to package tests and can easily exclude them

joernu76:


ReimarBauer:
We use this one. No idea yet why I can't solve myself in my branch that Merge Conflict

ReimarBauer (approved):
approved

PR #1568, opened 2022-09-27
Better restrict longitudes to map range for cylindric projections

### LINKED ISSUE #1565: Remote sensing tool tangent points not correct west of longitude 180°
I was playing around with the "Remote Sensing Tools" in the Top View window, while experiencing the following (see screenshot):
grafik

As you can see, tangent point locations are only displayed correctly east of longitude 180°, when crossing 180°. Adjusting the flight track to points only east of 180°, all tangent points are shown incorrectly. West of 180°, everything seems to be fine.
Tested with version 7.0.4.

joernu76:
Fix #1565

### DIFF
--- a/mslib/msui/remotesensing_dockwidget.py
+++ b/mslib/msui/remotesensing_dockwidget.py
@@ -35,7 +35,7 @@ import skyfield_data
 from PyQt5 import QtGui, QtWidgets
 from mslib.utils.qt import ui_remotesensing_dockwidget as ui
 from mslib.utils.time import jsec_to_datetime, datetime_to_jsec
-from mslib.utils.coordinate import get_distance, rotate_point, fix_angle
+from mslib.utils.coordinate import get_distance, rotate_point, fix_angle, normalize_longitude
 
 
 EARTH_RADIUS = 6371.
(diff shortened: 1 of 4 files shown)


mslib/msui/remotesensing_dockwidget.py
```diff
@@ -323,7 +331,7 @@ def direction_coordinates(self, lon_lin, lat_lin):
         tps = [[(x0 / np.cos(np.deg2rad(y0)), y0), (x1 / np.cos(np.deg2rad(y0)), y1)] for (x0, y0, x1, y1) in tps]
         return tps
 
-    @staticmethod
+    @ staticmethod
```
ReimarBauer: this extra blank sounds wrong. I would have expected a syntax error, which didn't occure, however there is no example with a blank in.

https://book.pythontips.com/en/latest/decorators.html

mslib/msui/remotesensing_dockwidget.py
```diff
@@ -266,7 +274,7 @@ def tangent_point_coordinates(self, lon_lin, lat_lin, flight_alt=14, cut_height=
         Returns: List of tuples of longitude/latitude coordinates
 
         """
-        lins = list(zip(lon_lin[0:-1], lon_lin[1:], lat_lin[0:-1], lat_lin[1:]))
+        lins = list(zip(lon_lin[0: -1], lon_lin[1:], lat_lin[0: -1], lat_lin[1:]))
```
ReimarBauer: Definitly better readable, this unbalanced whitespace around the colon got me to look into 
https://realpython.com/python-pep8/
For extended slices it is different.

ReimarBauer (changes requested):
have a look on my comment about the decorator `@` syntax and also the image comparison.

ReimarBauer:
I'm not familar with this so I have no idea, if this lower left  "red circle" is expected or there should be only a parallel dotted line?

ex2

in stereo it looks like this

stereo

and also in cyl (zoomed) 

cyl

mslib/msui/remotesensing_dockwidget.py
joernu76: That was autopep8 from my Visual Studio Code. I am not certain whether to agree, but I thought that autopep8 probably overrules my taste?
As there were only few changes to the file, I did not revert them.

joernu76:
The style changes were introduced by my autopep8 in VS Code during a syntactically incorrect file save and not reverted after fixing the error. Stupid, useless tool. I thought there were fine, but you convinced me of the opposite.

joernu76:
The first image, how did you produce that? Both cylindric and stereographic should look very smilar. And they do on my end...

ReimarBauer:
I get this always by

example

ReimarBauer:
> The style changes were introduced by my autopep8 in VS Code during a syntactically incorrect file save and not reverted after fixing the error. Stupid, useless tool. I thought there were fine, but you convinced me of the opposite.

They could have used the common autopep8 from pypi, but I am also not using this. So no idea if it is better.

joernu76:
> > The style changes were introduced by my autopep8 in VS Code during a syntactically incorrect file save and not reverted after fixing the error. Stupid, useless tool. I thought there were fine, but you convinced me of the opposite.
> 
> They could have used the common autopep8 from pypi, but I am also not using this. So no idea if it is better.

That's what they use. When there is a syntactically wrong string or class definition, autopep8 sometimes screws over the formatting. When the error is fixed, it only reverts part of the stuff. :-(

joernu76:
> I get this always by
> 
> example [ example ](https://user-images.githubusercontent.com/278728/192712788-1ff094a0-85ca-4a7f-847e-3109cc242955.gif) [ ](https://user-images.githubusercontent.com/278728/192712788-1ff094a0-85ca-4a7f-847e-3109cc242955.gif)

got it!

ReimarBauer (approved):
Thx

PR #1570, opened 2022-09-28
Refactoring and testing of Interactor classes

swsrkty:
Refactoring Interactor module to separate Interactor classes from Plotting classes.


mslib/msui/mpl_pathinteractor.py
```diff
@@ -354,62 +355,31 @@ class PathInteractor(QtCore.QObject):
...
-                                       edgecolor=edgecolor, alpha=0.15)
-        ax.add_patch(pathpatch)
+        path = self.myfig.mplpath
+        pathpatch = mpatches.PathPatch(path, facecolor=self.myfig.facecolor,
+                                       edgecolor=self.myfig.edgecolor, alpha=0.15)
+        self.myfig.ax.add_patch(pathpatch)
 
-        self.ax = ax
-        self.path = path
         self.pathpatch = pathpatch
```
joernu76: I would assume, naively, that the pathpatch need be maintained by the plotting-class. The PathPatch *is* the line, AFAIK.

joernu76: OK, had a closer look. This is a mess. There are *two* related objects, the pathpatch (which is not plotted) and the line (which is plotted). Both should be identical and the pathpatch is used to interact with the "path", whereas the line is plotting it.
This might be very useful for your disentanglement, but ultimately, it seems overcomplicated and superfluous.
To get the pathpatch to be plotted, it should be sufficient to set "edgecolor" to "linecolor" (and remove the self.line-plot). This works reasonably well - except for drag-dropping points. The latter could likely be fixed in some fashion, probably, by exchanging the line.set_data-stuff with something similar for the pathpatch.
The whole thing is a tad complicated also due to the distinction between wp_vertices and vertices etc needed by the great-circle paths...

mslib/msui/mpl_pathinteractor.py
```diff
@@ -407,8 +397,467 @@ def __init__(self, ax=None, waypoints=None, mplpath=None,
...
+        self.show_tangent_points = visible
+
+    def set_solar_angle_visible(self, visible):
+        self.show_solar_angle = visible
+
+    def set_remote_sensing(self, ref):
+        self.remote_sensing = ref
+
+
+class PathV_GCPlotter(PathPlotter):
```
joernu76: The GC stands for "great circle". There are two, separate, path classes (probably for historical reasons; doesn't seem to fill any particular purpose). PathH and PathH_GC, which, presumably deal with horizontal paths in an Euclidian geometry, on the one hand, and on a sphere, on the other.

Please use the GC only for the Horizontal classes.

mslib/msui/mpl_pathinteractor.py
```diff
@@ -407,8 +397,467 @@ def __init__(self, ax=None, waypoints=None, mplpath=None,
...
+            self.canvas.restore_region(self.background)
+        try:
+            self.ax.draw_artist(self.pathpatch)
+        except ValueError as error:
+            logging.debug("ValueError Exception '%s'", error)
+        for t in self.wp_labels:
+            self.ax.draw_artist(t)
+
+
+class PathL_GCPlotter(PathPlotter):
```
joernu76: see above.

mslib/msui/mpl_pathinteractor.py
```diff
@@ -711,36 +1062,9 @@ def button_release_insert_callback(self, event):
...
-            wpm.waypoint_data(best_index - 1).lat, wpm.waypoint_data(best_index - 1).lon,
-            wpm.waypoint_data(best_index).lat, wpm.waypoint_data(best_index).lon,
-            number_of_intermediate_points, connection="greatcircle")
-
-        # best_index1 is the best index among the intermediate coordinates to fit the hovered point
-        # if x axis has increasing coordinates
-        best_index1 = np.argmin(abs(vert_xs - x))
-        # depends if best_index1 or best_index1 - 1 on closeness to left or right neighbourhood
-        return (lats[best_index1], lons[best_index1]), best_index
+        lat_lon, ind = self.plotter.get_lat_lon(event, wpm)
```
joernu76: I would prefer, for consistency, when you do not supply the wpm model, but only the contained data (similar to redraw_path).
Think about testing! For testing this get_lat_lon function, one needs to create/mock a complicated QT structure, where a simple list of coordinates'd work as well.

mslib/msui/mpl_qtwidget.py
```diff
@@ -1194,10 +1193,6 @@ def redraw_xaxis(self, lats, lons, times):
                     ys.extend(ceil)
                 xs.append(vx[-1])
                 ys.append(aircraft.get_ceiling_altitude(wpd[-1].weight))
-
-                self.ceiling_alt = self.ax.plot(
```
joernu76, line 1198: Is it correct that this is removed? Who plots it now?

mslib/msui/mpl_pathinteractor.py
swsrkty: Ah, got it.

mslib/msui/mpl_qtwidget.py
swsrkty: I assumed it was related to the plotting using line instead of pathpatch and so I removed it and I thought the below line is used in plotting the pathpatch. I'll change it.
> self.ax.draw_artist(self.pathpatch)

mslib/plot.py
```diff
@@ -65,8 +69,10 @@ def load_from_xml_data(datasource):
         else:
             comments = ''
 
-        waypoints_list.append((lat, lon, flightlevel, location, comments))
-    return waypoints_list
+        waypoints_list1.append((lat, lon, flightlevel, location, comments))
+        waypoints_list2.append(ft.Waypoint(lat, lon, flightlevel, location, comments))
```
joernu76, line 99: As most of the mslib stuff uses the list of ft.Waypoint's, I would try to remove the need of the waypoints_list1.

mslib/plot.py
```diff
@@ -87,84 +93,55 @@ def __init__(self):
         self.params["basemap"].update(self.config["predefined_map_sections"][section]["map"])
         print(self.params)
         self.bbox_units = self.params["bbox"]
-        self.wps = load_from_ftml(filename)
+        self.wpslist = []
```
joernu76, line 122: not necessary

mslib/plot.py
```diff
@@ -87,84 +93,55 @@ def __init__(self):
...
+        self.wpslist = load_from_ftml(filename)
+        self.wps = self.wpslist[0]
+        self.wp_model_data = self.wpslist[1]
         self.wp_lats, self.wp_lons, self.wp_locs = [[x[i] for x in self.wps] for i in [0, 1, 3]]
-        self.wp_presss = [mslib.utils.thermolib.flightlevel2pressure(wp[2] * units.hft).to("Pa").m for wp in self.wps]
+        self.wp_press = [mslib.utils.thermolib.flightlevel2pressure(wp[2] * units.hft).to("Pa").m for wp in self.wps]
         self.fig.clear()
         self.ax = self.fig.add_subplot(111, zorder=99)
         self.path = [(wp[0], wp[1], datetime.datetime.now()) for wp in self.wps]
+        self.vertices = [list(a) for a in (zip(self.wp_lons, self.wp_lats))]
```
joernu76, line 131: This should be removed. The "vertices" of the msui classes are often in different units then lon/lat. The restructuring of using update_from_waypoints should have made this superfluous anyway.

mslib/plot.py
```diff
@@ -87,84 +93,55 @@ def __init__(self):
...
-    def TopViewPath(self):
         matplotlib.backends.backend_agg.FigureCanvasAgg(self.myfig.fig)
         self.ax = self.myfig.ax
         self.fig = self.myfig.fig
         self.myfig.init_map(**(self.params["basemap"]))
         self.myfig.set_map()
+        self.plotter = mpath.PathH_GCPlotter(self.myfig.map)
 
-        # plot path
+    def TopViewPath(self):
```
joernu76, line 151: The function names do not follow the PEP8 convention. Also adding TopView here seems redundant.

I think it is sensible to have small functions plotting the individual plot components. However calling three functions without any argument in main() seems like a strange design. Why not only one? *In principle* you could put everything in the constructor - or make a simple function doing all that.

What makes sense is reusing the MyFigure class for all plots of one type. But maybe the draw() function should not contain the loop. Instead the loop should IMO go into the main() function and call a draw function with flightpath etc. as  *arguments*.
I also suspect that this does not work properly for multiple flights as, e.g., the label functions must be called after changing the flight.
But maybe I am mistaken, I haven't tested it in practice.

autoplot.py
```diff
@@ -0,0 +1,58 @@
+import click
```
ReimarBauer: see template.py, please add the common header. Add your name as initial author

ReimarBauer: the module should go to mslib.utils

autoplot.py
```diff
@@ -0,0 +1,58 @@
...
+@click.option('--stime', default="", help='Starting time for downloading multiple plots with a fixed interval.')
+@click.option('--etime', default="", help='Ending time for downloading multiple plots with a fixed interval.')
+def autoplot(cpath, ftrack, itime, vtime, intv, stime, etime):
+    conf.read_config_file(path=cpath)
+    config = conf.config_loader()
+    a = plot.TopViewPlotting()
+    for flight, section, vertical, filename, init_time, time in \
+        config["automated_plotting_flights"]:
+        for url, layer, style, elevation in config["automated_plotting_hsecs"]:
+            if(intv == 0):
```
ReimarBauer: the brackets are unusual, not needed.

brackets are used for multiline statements and they have a blank between if and (

ReimarBauer: if is a condition, not a function.

autoplot.py
```diff
@@ -0,0 +1,58 @@
...
+                if(itime != ""):
+                    inittime = datetime.strptime(itime, "%Y-%m-%dT" "%H:%M:%S")
+                else:
+                    inittime = itime
+                starttime = datetime.strptime(stime, "%Y-%m-%dT" "%H:%M:%S")
+                endtime = datetime.strptime(etime, "%Y-%m-%dT" "%H:%M:%S")
+                i = 1
+                time = starttime
+                while(time <= endtime):
+                    print(time)
```
ReimarBauer: may be replace print by logging.info, logging.debug

mslib/msui/constants.py
```diff
@@ -69,6 +69,27 @@
         except IOError:
             logging.error(f'"{MSUI_SETTINGS}" can''t be created')
 
+    MSS_AUTOPLOT = os.getenv('MSS_AUTOPLOT', os.path.join(MSUI_CONFIG_PATH, "mss_autoplot.json"))
```
ReimarBauer, line 72: add a ToDo, # ToDo refactor to a function

autoplot.py
ReimarBauer: https://peps.python.org/pep-0008/ search for if-statement

mslib/utils/autoplot.py
```diff
@@ -0,0 +1,58 @@
...
+    for flight, section, vertical, filename, init_time, time in \
+        config["automated_plotting_flights"]:
+        for url, layer, style, elevation in config["automated_plotting_hsecs"]:
+            if(intv == 0):
+                if(itime != ""):
+                    inittime = datetime.strptime(itime, "%Y-%m-%dT" "%H:%M:%S")
+                else:
+                    inittime = itime
+                time = datetime.strptime(vtime, "%Y-%m-%dT" "%H:%M:%S")
+                if(ftrack != ""):
```
ReimarBauer, line 29: this "(" is only used for multiline statements and then with a blank after the "if"

see if-statement in
https://peps.python.org/pep-0008/

joernu76:
This was part of another pull request already.

PR #1574, opened 2022-09-30
Do not crash for consecutive points with identical coordinates.

### LINKED ISSUE #1571: Remote sensing tool results in crash, if two consecutive way points are in the same horizontal position
As mentioned in #1516 , I was able to provoke a crash if I was either

- cloning a way point in the table view with enabled tangent points of the remote sensing tools 
- enabling the tangent points of the remote sensing tools when having two consecutive way points in the same horizontal position

An easy workaround for me as a user is to ensure that there are no two identical consecutive way points, before enabling the remote sensing tools, but maybe it could be checked by the remote sensing tool itself?

joernu76:
Fix #1571

### DIFF
--- a/mslib/msui/remotesensing_dockwidget.py
+++ b/mslib/msui/remotesensing_dockwidget.py
@@ -316,10 +316,12 @@ class RemoteSensingControlWidget(QtWidgets.QWidget, ui.Ui_RemoteSensingDockWidge
         Returns: List of tuples of longitude/latitude coordinates
 
         """
-        lins = [(_line[0][mid], _line[0][mid + 1],
-                 _line[1][mid], _line[1][mid + 1]) for _line, mid in
-                zip(gc_lines, [len(_line[0]) // 2 for _line in gc_lines])]
-        lens = [np.hypot(_line[0][0] - _line[0][-1], _line[0][0] - _line[0][-1]) * 110. for _line in gc_lines]
+        lins = [(_line[0][mid], _line[0][mid + 1], _line[1][mid], _line[1][mid + 1])
+                for _line, mid in zip(gc_lines, [len(_line[0]) // 2 for _line in gc_lines])
+                if len(_line[0]) > 2]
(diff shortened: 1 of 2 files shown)


mslib/msui/remotesensing_dockwidget.py
```diff
@@ -316,10 +316,12 @@ def direction_coordinates(self, gc_lines):
         Returns: List of tuples of longitude/latitude coordinates
 
         """
-        lins = [(_line[0][mid], _line[0][mid + 1],
-                 _line[1][mid], _line[1][mid + 1]) for _line, mid in
-                zip(gc_lines, [len(_line[0]) // 2 for _line in gc_lines])]
-        lens = [np.hypot(_line[0][0] - _line[0][-1], _line[0][0] - _line[0][-1]) * 110. for _line in gc_lines]
+        lins = [(_line[0][mid], _line[0][mid + 1], _line[1][mid], _line[1][mid + 1])
```
ReimarBauer, line 319: as bug fix okay, but that is not getting better readable. When there are tests written may be some of the methods can be  extracted as functions to a remotesensing utils

joernu76: I agree.
However, I did also not see how to clean this up easily, or how to rewrite this using "numpy" terminology. The whole thing should be very slow and I am surprised it is still real-time capable.

ReimarBauer:
sometimes I see 
 ```
RuntimeWarning: invalid value encountered in double_scalars
  direction = [(_x / np.hypot(_x, _y), _y / np.hypot(_x, _y))
/home/reimar/MASTER/MSS/mslib/msui/mpl_pathinteractor.py:77: RuntimeWarning: invalid value encountered in double_scalars
  r = (np.dot(p - l1, l2 - l1) / np.linalg.norm(l2 - l1) ** 2)
```

ReimarBauer (approved):
solves the issue, Thx

PR #1583, opened 2022-10-27
Unwrap longitudes of flightpath before computing the left_longitude.

### LINKED ISSUE #1566: Wrapping issue in side view (and linear view) when crossing 180° longitude
When having a flight path crossing 180° longitude, there are missing values at these longitudes in the side view (and also in linear view), as shown in the screenshot:
grafik

It seems like, the model data is not wrapped correctly and simply goes from -180 to +180 degrees longitude?

joernu76:
Currently, if the flight path crosses -180, -180 degree will be found as left_longitude, which is potentially smack in the middle. unwrapping the longitude helps shifting everything to the proper range.

Fix #1566


ReimarBauer (approved):
Thx

PR #1584, opened 2022-10-27
this style of unit definitions should work with old and new pint

joernu76:


mslib/utils/units.py
```diff
@@ -34,19 +34,15 @@
...
-                       ("pptv", 1e-12)):
-    units.define(pint.unit.UnitDefinition(_name, '', (),
-                 pint.converters.ScaleConverter(_factor)))
+units.define("level = 1"),
+units.define("sigma = 1"),
+units.define("permille = 1e-3"),
+units.define("ppm = 1e-6"),
+units.define("ppb = 1e-9"),
+units.define("ppt = 1e-12"),
+units.define("pptv = 1e-12")
```
ReimarBauer: that would get better readable, when same order 
ppm
ppb
ppt
ppmv
ppbv
pptv

I readed it wrong, compared pattern of scale factors

ReimarBauer (changes requested):
please make the order of mixing ratio better readable. 
either as before or same order for the volume mixing ratio. 
my pattern reading stumbles now ;)

ReimarBauer:
At all this looks very simple and makes a fixation unnessary. I checked it with the tests, 0.19 and 0.20.1 Do you want to undo the fixation and add it to stable?

I was blocked by a call, not seen that you did the other change as I clicked Update ...

ReimarBauer (approved):
thx

PR #1592, opened 2022-11-04
enable to select multiple files for import

### LINKED ISSUE #1578: select multiple files for local flightpath upload
I want to be able to select many ftml files and load them by one file dialog. Currently for 4 files I need 4 times File Import Open.

ReimarBauer:
closes #1578

### DIFF
--- a/mslib/msui/msui.py
+++ b/mslib/msui/msui.py
@@ -57,7 +57,7 @@ from mslib.msui.updater import UpdaterUI
 from mslib.utils import setup_logging
 from mslib.plugins.io.csv import load_from_csv, save_to_csv
 from mslib.msui.icons import icons, python_powered
-from mslib.utils.qt import get_open_filename, get_save_filename, Worker, Updater
+from mslib.utils.qt import get_open_filenames, get_save_filename, Worker, Updater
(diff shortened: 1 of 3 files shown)


mslib/msui/msui.py
```diff
@@ -591,17 +591,19 @@ def remove_plugins(self):
...
+        filename = get_open_filenames(
             self, "Import Flight Track",
             self.last_save_directory,
             f"Flight Track (*.{extension});;All files (*.*)",
             pickertype=pickertype)
         if self.local_active:
             if filename is not None:
-                self.last_save_directory = fs.path.dirname(filename)
-                self.create_new_flight_track(filename=filename, function=function)
+                for name in filename:
```
joernu76: maybe filenames instead of filename?

joernu76 (dismissed):

mslib/msui/msui.py
ReimarBauer: true, better wording. it can still be only one file.

joernu76: Fine either way. I tend to use plural for lists or tuples independent of length (can be empty as well).

joernu76 (approved):

PR #1596, opened 2022-11-14
show creator in manage user interface

### LINKED ISSUE #1448: info in Manage Users interface about the creator of the operation
currently one user with admin role can't see in the manage users interface who created the operation.
The creator role user of course is not shown in the list of usernames. 

We should add there where we have Operation: flight1  in next line "created by" creator name

who_is_creator
 
There already two users in that operation. The creator and the admin. The logged in user has the admin role. The creator is invisible because that role can't be changed by this UI.

ReimarBauer:
solved also a few test uncleanliness

closes #1448


joernu76 (approved):

ReimarBauer:
needs to be redone

PR #1605, opened 2022-11-22
refactor

joernu76:


mslib/utils/mssautoplot.py
```diff
@@ -24,12 +24,316 @@
...
-import click
 import logging
-from mslib import plot
 from datetime import datetime, timedelta
-from mslib.utils import config as conf
+import hashlib
+import io
+import os
+import xml
+from fs import open_fs
```
ReimarBauer: maybe move this from line to the from section?

mslib/utils/mssautoplot.py
```diff
@@ -24,12 +24,316 @@
...
+
+import PIL.Image
+import matplotlib
+import click
+
+import mslib
```
ReimarBauer, line 42: not sure if this is needed?

mslib/utils/mssautoplot.py
```diff
@@ -24,12 +24,316 @@
...
+import mslib.utils
+import mslib.msui
+import mslib.msui.mpl_map
+import mslib.utils.qt
+import mslib.utils.thermolib
+from mslib.utils.config import config_loader, read_config_file
+from mslib.utils.units import units
+from mslib.msui.wms_control import MSUIWebMapService
+import matplotlib.pyplot as plt
+import defusedxml.ElementTree as etree
```
ReimarBauer: this should be above in the import section.

see suggestions from https://peps.python.org/pep-0008/#imports

mslib/utils/mssautoplot.py
```diff
@@ -24,12 +24,316 @@
...
+        else:
+            comments = ''
+
+        waypoints_list1.append((lat, lon, flightlevel, location, comments))
+        waypoints_list2.append(ft.Waypoint(lat, lon, flightlevel, location, comments))
+        waypoints_list2[-1].utc_time = datetime.now()
+    return waypoints_list1, waypoints_list2
+
+
+class Plotting():
```
ReimarBauer: only class Plotting:

see https://docs.python.org/3/tutorial/classes.html#class-definition-syntax

mslib/utils/mssautoplot.py
```diff
@@ -24,12 +24,316 @@
...
+                          "size": (width, height)
+                         }
+
+                xmls = wms.getmap(**kwargs)
+                urlstr = wms.getmap(return_only_url=True, **kwargs)
+                ending = ".xml"
+
+                if not os.path.exists(self.wms_cache):
+                    os.makedirs(self.wms_cache)
+                md5_filename = os.path.join(self.wms_cache, hashlib.md5(urlstr.encode('utf-8')).hexdigest() + ending)
```
ReimarBauer, line 322: I do later a refactoring for a fs path.

mslib/utils/mssautoplot.py
```diff
@@ -24,12 +24,316 @@
...
+    datasource = _fs.open(_name)
+    wp_list = load_from_xml_data(datasource)
+    return wp_list
+
+
+def load_from_xml_data(datasource):
+    try:
+        doc = xml.dom.minidom.parse(datasource)
+    except xml.parsers.expat.ExpatError as ex:
+        raise SyntaxError(str(ex))
```
ReimarBauer, line 78: maybe a logging entry

joernu76: later. this code needs to be joined with the other file parsing code. This will be multiple steps to not overcomplicate the pull request.

mslib/utils/mssautoplot.py
```diff
@@ -24,12 +24,316 @@
...
+        self.wp_press = [mslib.utils.thermolib.flightlevel2pressure(wp[2] * units.hft).to("Pa").m for wp in self.wps]
+        self.fig.clear()
+        self.ax = self.fig.add_subplot(111, zorder=99)
+        self.path = [(wp[0], wp[1], datetime.now()) for wp in self.wps]
+        self.vertices = [list(a) for a in (zip(self.wp_lons, self.wp_lats))]
+        self.lats, self.lons = mslib.utils.coordinate.path_points([_x[0] for _x in self.path],
+                                                                  [_x[1] for _x in self.path],
+                                                                  numpoints=self.num_interpolation_points + 1,
+                                                                  connection="greatcircle")
+
```
ReimarBauer, line 134: we could also think about moving path, draw, setup method to the Plotting class,

https://www.techversantinfotech.com/python-naming-conventions-points-you-should-know/#:~:text=Method%20name%20in%20Python,it%20needs%20to%20be%20mangled.

mslib/utils/mssautoplot.py
```diff
@@ -24,12 +24,316 @@
...
+        self.line = None
+        matplotlib.backends.backend_agg.FigureCanvasAgg(self.myfig.fig)
+        self.ax = self.myfig.ax
+        self.fig = self.myfig.fig
+        self.myfig.init_map(**(self.params["basemap"]))
+        self.myfig.set_map()
+        self.plotter = mpath.PathH_GCPlotter(self.myfig.map)
+
+    def TopViewPath(self):
+        # plot path and label
```
ReimarBauer, line 150: method names should be lowercase

ReimarBauer (changes requested):
see comments

mslib/utils/mssautoplot.py
joernu76: yes. I focused here mostly on the mslib/msui stuff and left the CLI tool mostly untouched except for moving it to a single file..

mslib/utils/mssautoplot.py
joernu76: again, refactoring the CLI tool is a next step and maybe something that Sree could do, if she wants.

ReimarBauer (approved):
see discussions for further steps

PR #1608, opened 2022-11-23
refactor

joernu76:


mslib/msui/flighttrack.py
```diff
@@ -94,6 +94,34 @@ def seconds_to_string(seconds):
...
+        location = wp_el.getAttribute("location")
+        lat = float(wp_el.getAttribute("lat"))
+        lon = float(wp_el.getAttribute("lon"))
+        flightlevel = float(wp_el.getAttribute("flightlevel"))
+        comments = wp_el.getElementsByTagName("Comments")[0]
+        # If num of comments is 0(null comment), then return ''
+        if len(comments.childNodes):
+            comments = comments.childNodes[0].data.strip()
+        else:
+            comments = str('')
```
ReimarBauer: comments = ''

makes it already to a class str

mslib/msui/flighttrack.py
```diff
@@ -94,6 +94,34 @@ def seconds_to_string(seconds):
...
+
+        waypoints_list.append(Waypoint(lat, lon, flightlevel,
+                                       location=location,
+                                       comments=comments))
+    return waypoints_list
+
+
 class Waypoint(object):
```
ReimarBauer: class Waypoint:

since python version 3 this is a replacement for (object) 
(this can also be done in a further refactoring)

mslib/msui/mpl_pathinteractor.py
```diff
@@ -441,6 +436,7 @@ def redraw_path(self, vertices=None, waypoints_model_data=None):
         output. Otherwise the vertex array obtained from the path patch
         will be used.
         """
+        assert False
```
ReimarBauer: We can not rely on asserts in production because

for performance reasons, assert statements can be disabled

see also
https://towardsdatascience.com/2-caveats-to-avoid-when-using-python-assert-statement-8060e123c26a

mslib/msui/flighttrack.py
```diff
@@ -616,33 +644,9 @@ def load_from_ftml(self, filename):
         self.load_from_xml_data(xml_content, name)
 
     def load_from_xml_data(self, xml_content, name="Flight track"):
```
ReimarBauer, line 646: hmm, we have load_from_xml_data, load_from_ftml,  read_ftml 

is that basically the same?

mslib/msui/mpl_qtwidget.py
```diff
@@ -106,6 +107,13 @@ def get_plot_size_in_px(self):
         height = int(round(ax_bounds[3]))
         return width, height
 
+    def get_settings(self):
+        return self.settings
+
+    def set_settings(self, settings):
+        if settings is not None:
```
ReimarBauer, line 124: a common docstring of the parameters could help, e.g. settings: dict ..

mslib/msui/mpl_qtwidget.py
```diff
@@ -119,40 +127,34 @@ def __init__(self, fig=None, ax=None):
...
+        self.map = mpl_map.MapCanvas(appearance=self.get_settings(),
                                      resolution="l", area_thresh=1000., ax=self.ax,
                                      **kwargs)
 
         # Sets the selected fontsize only if draw_graticule box from topview options is checked in.
-        if self.appearance_settings["draw_graticule"]:
+        if self.settings["draw_graticule"]:
             try:
                 self.map._draw_auto_graticule(self.tov_als)
             except Exception as ex:
```
ReimarBauer, line 164: we should add the specific exceptions here

mslib/msui/mpl_qtwidget.py
```diff
@@ -327,25 +320,24 @@ def __init__(self, fig=None, ax=None, settings=None, numlabels=None, num_interpo
...
+                           "fill_flighttrack": True,
+                           "label_flighttrack": True,
+                           "draw_verticals": True,
+                           "draw_marker": True,
+                           "draw_ceiling": True,
+                           "colour_ft_vertices": (0, 0, 1, 1),
+                           "colour_ft_waypoints": (1, 0, 0, 1),
+                           "colour_ft_fill": (0, 0, 1, 0.15),
+                           "colour_ceiling": (0, 0, 1, 0.15)})
+        self.set_settings(settings)
```
ReimarBauer, line 350: "draw_graticule" is not defined here, but looks like mandatory

we maybe should have one place with all possible keywords

mslib/utils/mssautoplot.py
```diff
@@ -66,38 +63,14 @@ def load_from_ftml(filename):
...
-            comments = comments.childNodes[0].data.strip()
-        else:
-            comments = ''
-
-        waypoints_list1.append((lat, lon, flightlevel, location, comments))
-        waypoints_list2.append(ft.Waypoint(lat, lon, flightlevel, location, comments))
-        waypoints_list2[-1].utc_time = datetime.now()
-    return waypoints_list1, waypoints_list2
+    datasource = _fs.readtext(_name)
+    wp_list = ft.load_from_xml_data(datasource)
```
ReimarBauer, line 67: important to postfix by the type?

ReimarBauer (changes requested):
see comments

mslib/msui/flighttrack.py
joernu76: One could probably further tidy that up, but: no
The current three functions are not redundant.

mslib/msui/mpl_pathinteractor.py
joernu76: I put that there to identify dead code and forgot about it. Nice catch.

Then this whole method is dead code.

mslib/utils/mssautoplot.py
joernu76: ?
The naming of this functions is probaly not ideal.

ReimarBauer: I meant more always adding "_list" to all these names, we don't do that regularly.

mslib/msui/mpl_qtwidget.py
joernu76: can probably be removed. this code is riddled with work-arounds for matplotlib 0.x

mslib/msui/mpl_qtwidget.py
joernu76: only for TopView. These settings are heavily related to the Options in the UI. One can think about a better way of handling this without welding this code again with QT.

mslib/msui/mpl_qtwidget.py
joernu76: Maybe make an issue for this? Investigate matplotlib-workarounds?

ReimarBauer: https://github.com/Open-MSS/MSS/issues/1618

ReimarBauer (approved):
Thx

PR #1615, opened 2022-11-23
renamed SaveAs to Rename, fixed usage of handler

### LINKED ISSUE #1311: Replace "Save As" by Rename
Currently we do use two different "Save As" features. The other one is named export. 
Export can use different functions for same extension while "Save As" can only use one extension per function.
If there are several csv plugins "Save As" can only use one.  "Save As" also renames the  active local flight path.

We should simplify this, Rename should replace "Save As" and could be used to rename any active flight path. 
All other file exchange functionality should go to "Import" / "Export".

The creation of an operation should use the same Import definition as for Import in the menu. We need to replace the simplified file browse functionality by select first the import function then choose a file. By this also the extension is not a master key.

ReimarBauer:
closes #1311


ReimarBauer:
Changed to:

renamed

joernu76:
Mmh. This is not consistent with traditional Programs. Typically you use save as and continue working on the newly saved file with a new name.
E.g. Word does not use "Rename" for this.
"Rename" implies to me that the original is removed as well.

ReimarBauer:
we can leave it as it was, and I fix only the handler / function call in a new PR-
"save as" also is called on a new name on first save. This should not use the handler.

Rename may later be moved to a context menu on the name or so. The dialog looks also strange to have that added to rename.

PR #1623, opened 2022-11-28
Hide mscolab ui when not needed

### LINKED ISSUE #1625: MSColab UI should only be displayed when necessary
The MSColab interface for working with operations consumes a great deal of space of the main window, but is only needed when one connects to an MSCOlab server.

It would be desirable to hide this part until one connects to a server and rehides it upon disconnection.

joernu76:


ReimarBauer (commented):
nice idea :)

have a look if the operation menu entry can be also disabled, removing may dissort the UI

ReimarBauer:
@joernu76 please link with an issue (create that too)

joernu76:
I juggled the other UI elements around a bit. Seems a bit more consistent now.
The MSColab connectivity has a permanent group box above with a more consistent status message.
I move the operation description to the MSCOlab operations box.

joernu76:
I am not convinced that people will find the "disconnect" button. Toggling the connect button to disconnect'd be more intuitive, I think.
Web services try to make disconnecting/signing out purposely difficult; I am not sure we should follow that particular UI design philosophy, but that'd be a change for a different MR.

ReimarBauer:
currently I like how it is, a lot web services have the logout feature in that position.  
Closing is also a logout, I guess when we would have something similar as closing the operation list, it may be an compromise. 
I look on the other changes after the it-forum ;)

joernu76:
> currently I like how it is, a lot web services have the logout feature in that position. Closing is also a logout, I guess when we would have something similar as closing the operation list, it may be an compromise. I look on the other changes after the it-forum ;)

Currently, I did not change this.

ReimarBauer:
Functionality so far ok, but I found something new in the UI which we hadn't before. It seems we can move things arround, somehow interesting but I think we don't have something to get into these elements yet?

unknown_element

When those are removed we would also save space.

ReimarBauer (changes requested):
there seems now a moveable toolbar in the UI see peek video. This can also be duplicated by the user.

I guess that got somehow in, and we don't have something to add there yet?

joernu76:
Yes, there were *two* toolbars. I have no idea how they got added by qtcreator. I removed them. Should be fine now...

ReimarBauer:
fun fact was that I somehow clicked on that and wondered what that is. Then I wanted to decribe this and needed long to figure that I have to click into the edge

ReimarBauer (approved):
Thx

Issue #1332, opened 2022-01-14
download folder in .config/msui for openaip


ReimarBauer:
currently we download directly to the .config/MSS folder.

We should have an aip folder and use this.

ReimarBauer:
from a users point of view I think we should have a downloads folder and maybe below this other subdirs. This helps to destinguish which files have been installed by using MSS to the system. Everything which is in downloads should be removable without making problems.

.config/msui
├── downloads
├── gravatars
├── plugins

And below downloads we could create e.g. an folder aip

joernu76:
seems very reasonable!

rootxrishabh:
To my understanding, we need to add some folders for openaip data to organize the data efficiently, so I will make a PR.

PR #1627, opened 2022-11-30
improve airdata

### LINKED ISSUE #1626: get_airspaces can crash
because not finding airspaces in defined files

ReimarBauer:
fixes #1626

### DIFF
--- a/mslib/utils/_tests/test_airdata.py
+++ b/mslib/utils/_tests/test_airdata.py
@@ -1,0 +1,166 @@
+# -*- coding: utf-8 -*-
+"""
+
+    tests._test_utils.test_airdata
+    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+    This module provides pytest functions to test mslib.utils.airdata
+
+    This file is part of MSS.
+
+    :copyright: Copyright 2022-2022 Reimar Bauer
+    :copyright: Copyright 2022-2022 by the MSS team, see AUTHORS.
+    :license: APACHE-2.0, see LICENSE for details.
+
+    Licensed under the Apache License, Version 2.0 (the "License");
+    you may not use this file except in compliance with the License.
+    You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+"""
+import os
+import mock
+from PyQt5 import QtWidgets
+from mslib.utils.airdata import download_progress, get_airports,\
+    get_available_airspaces, update_airspace, get_airspaces
+from tests.constants import ROOT_DIR
+
+
+def _download_progress_airports(path, url):
+    """ mock expensive download from external site"""
+    assert path is not None
+    assert url is not None
+    text = '''"id","ident","type","name","latitude_deg","longitude_deg","elevation_ft","continent",\
+"iso_country","iso_region","municipality","scheduled_service","gps_code","iata_code","local_code",\
+"home_link","wikipedia_link","keywords"
+6523,"00A","heliport","Total Rf Heliport",40.07080078125,-74.93360137939453,11,"NA","US",\
+"US-PA","Bensalem","no","00A",,"00A",,,\
+323361,"00AA","small_airport","Aero B Ranch Airport",38.704022,-101.473911,3435,"NA",\
+"US","US-KS","Leoti","no","00AA",,"00AA",,,'''
+    file_path = os.path.join(ROOT_DIR, "airports.csv")
+    with open(file_path, "w") as f:
+        f.write(text)
+
+
+def _download_progress_airspace(path, url):
+    """ mock expensive download from external site"""
+    assert path is not None
+    assert url is not None
+    text = '''<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+ <!-- For Testing ONLY -->
+<OPENAIP VERSION="1668386405436" DATAFORMAT="1.1" xmlns="https://www.openaip.net"\
+ xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"\
+  xsi:schemaLocation="https://www.openaip.net https://storage.googleapis.com/d644c13a-ed49-48ad-8493-d3e55d3281a5\
+  /assets/common/openaip-schema-v1.xsd">
+<AIRSPACES><ASP CATEGORY="C"><VERSION>1669682436912</VERSION><ID>631330d834a543d6cb4a83fe\
+</ID><COUNTRY>BG</COUNTRY><NAME>TMA 1 SOFIA 123.700</NAME><ALTLIMIT_TOP REFERENCE="STD">\
+<ALT UNIT="FL">245</ALT></ALTLIMIT_TOP><ALTLIMIT_BOTTOM REFERENCE="MSL"><ALT UNIT="F">8500</ALT>\
+</ALTLIMIT_BOTTOM><GEOMETRY><POLYGON>22.4899253237031 43.56620830068364,22.79416666666667 \
+43.478611111111114,23.09111111111111 43.39222222222222,23.31277777777778 43.32833333333333,\
+23.80777777777778 43.151111111111106,24.220555555555553 43.09527777777778,24.16333333333333 \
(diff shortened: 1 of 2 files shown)


mslib/utils/airdata.py
```diff
@@ -79,19 +80,24 @@ def download_progress(file_path, url, progress_callback=lambda f: logging.info(f
...
     Gets or downloads the airports.csv in ~/.config/msui and returns all airports within
     """
     global _airports, _airports_mtime
-    file_exists = os.path.exists(os.path.join(MSUI_CONFIG_PATH, "airports.csv"))
+    if url is None:
+        url = "https://ourairports.com/data/airports.csv"
+    data = fs.open_fs(MSUI_CONFIG_PATH)
+    osdir = data.root_path
+
+    file_exists = os.path.exists(os.path.join(osdir, "airports.csv"))
```
joernu76: is there a good reason not to store the joined path in a variable? This repetion is (as demonstrated here) tedious.

mslib/utils/airdata.py
```diff
@@ -173,41 +192,55 @@ def get_airspaces(countries=[]):
...
-
-                airspace_data["polygon"] = [(float(data.split(" ")[0]), float(data.split(" ")[-1]))
-                                            for data in airspace_data["polygon"].split(",")]
-                _airspaces.append(airspace_data)
-                _airspaces_mtime[file] = os.path.getmtime(os.path.join(MSUI_CONFIG_PATH, file))
-        else:
-            QtWidgets.QMessageBox.information(None,
-                                              "No Airspace data in file:",
-                                              f"{file}")
+        fpath = os.path.join(osdir, file)
```
joernu76: This looks very different. I do not fully understand the minidom, but I somehow lack the reference to the AIRSPACES tag. Why is this not necessary anymore?

mslib/utils/airdata.py
ReimarBauer: there is an open issue for a refactoring,  https://github.com/Open-MSS/MSS/issues/1332

The changes become neccessary because tests should not write to the users .config/msui dir.

mslib/utils/airdata.py
ReimarBauer: I found after I used minidom a typo in my example.  May be I could also improve it now for ET.  I check this.

The info in that message box is misleading, AIRSPACES Tag is always there, an empty dataset looks like

```
<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
<OPENAIP VERSION="1669769794284" DATAFORMAT="1.1" xmlns="https://www.openaip.net" loremm ipsum ...>
 <AIRSPACES></AIRSPACES>
</OPENAIP>
~         
```
I will add an explicit check I do it currently by enumerate(names) which is empty, when there is no data.

mslib/utils/airdata.py
ReimarBauer: of course no good reason, :)

ReimarBauer:
some_airspace_data

ReimarBauer:
Sometimes I get locally:

FAILED mslib/utils/_tests/test_airdata.py::test_get_airspaces_missing_data - AssertionError: assert [{'bottom': 2...), ...], ...}] == []

For some reason it uses the old version of the data.

ReimarBauer:
cocered

mslib/utils/_tests/test_airdata.py
```diff
@@ -0,0 +1,213 @@
...
+  /assets/common/openaip-schema-v1.xsd">
+<AIRSPACES></AIRSPACES>
+</OPENAIP>
+'''
+    file_path = os.path.join(ROOT_DIR, "bg_asp.xml")
+    with open(file_path, "w") as f:
+        f.write(text)
+
+
+def _cleanup_test_files():
```
joernu76, line 101: fine like this, but I believe this should be a fixture using "pytest guidelines".

mslib/utils/_tests/test_airdata.py
```diff
@@ -0,0 +1,213 @@
...
+42.88388888888889,22.739444444444445 42.88527777777778\
+</POLYGON></GEOMETRY></ASP></AIRSPACES>
+</OPENAIP>
+'''
+    file_path = os.path.join(ROOT_DIR, "bg_asp.xml")
+    with open(file_path, "w") as f:
+        f.write(text)
+
+
+def _download_incomplete_airspace(path, url):
```
joernu76, line 83: should be a pytest fixture.
This would first write the file, then yield None, then remove the file.

mslib/utils/_tests/test_airdata.py
```diff
@@ -0,0 +1,213 @@
...
+"""
+import os
+import mock
+from PyQt5 import QtWidgets
+from mslib.utils.airdata import download_progress, get_airports,\
+    get_available_airspaces, update_airspace, get_airspaces
+from mslib._tests.constants import ROOT_DIR
+
+
+def _download_progress_airports(path, url):
```
joernu76, line 35: this should by a pytest fixture. write file, yield None, remove file.

mslib/utils/_tests/test_airdata.py
```diff
@@ -0,0 +1,213 @@
...
+6523,"00A","heliport","Total Rf Heliport",40.07080078125,-74.93360137939453,11,"NA","US",\
+"US-PA","Bensalem","no","00A",,"00A",,,\
+323361,"00AA","small_airport","Aero B Ranch Airport",38.704022,-101.473911,3435,"NA",\
+"US","US-KS","Leoti","no","00AA",,"00AA",,,'''
+    file_path = os.path.join(ROOT_DIR, "airports.csv")
+    with open(file_path, "w") as f:
+        f.write(text)
+
+
+def _download_progress_airspace(path, url):
```
joernu76, line 51: this should by a pytest fixture. write file, yield None, remove file.

mslib/utils/_tests/test_airdata.py
joernu76: Probably superfluous if functions above are rewritten as fixtures with yield...?

mslib/utils/_tests/test_airdata.py
```diff
@@ -0,0 +1,213 @@
...
+    """
+    In the test environment we start always in a fresh tmp dir, no data is available
+    once it is downloaded it is managed by airdata
+    """
+    _cleanup_test_files()
+    airspaces = get_airspaces(countries=["bg"])
+    assert airspaces == []
+
+
+@mock.patch("mslib.utils.airdata.download_progress", _download_progress_airspace)
```
joernu76, line 163: OK, this particular use makes using a fixture more difficult, *unless* the mocking would be part of the fixture.
Using a fixture had the added benefit that one could "yield" the correct data, ideally using the same data structure for yielding and generating the xml file.

mslib/utils/airdata.py
```diff
@@ -79,19 +82,22 @@ def download_progress(file_path, url, progress_callback=lambda f: logging.info(f
...
 
 
-def get_airports(force_download=False):
+def get_airports(force_download=False, url=None):
     """
     Gets or downloads the airports.csv in ~/.config/msui and returns all airports within
     """
     global _airports, _airports_mtime
-    file_exists = os.path.exists(os.path.join(MSUI_CONFIG_PATH, "airports.csv"))
+    if url is None:
```
joernu76, line 90: a string is not a list. why not add the default string in the function declaration?

mslib/utils/airdata.py
```diff
@@ -79,19 +82,22 @@ def download_progress(file_path, url, progress_callback=lambda f: logging.info(f
...
+        url = "https://ourairports.com/data/airports.csv"
+
+    file_exists = os.path.exists(os.path.join(OSDIR, "airports.csv"))
 
     if _airports and file_exists and \
-            os.path.getmtime(os.path.join(MSUI_CONFIG_PATH, "airports.csv")) == _airports_mtime:
+            os.path.getmtime(os.path.join(OSDIR, "airports.csv")) == _airports_mtime:
```
joernu76, line 96: repeated use of os.path.join(OSDIR, "airports.csv")

joernu76 (approved):

mslib/utils/_tests/test_airdata.py
ReimarBauer: the problem is that it is needed at all. From a users point of view, he did'nt know.

mslib/utils/_tests/test_airdata.py
ReimarBauer: later, currently it is better to see that structure in that test. 

Maybe also later these tests become class members

mslib/utils/_tests/test_airdata.py
ReimarBauer: Currently we see the tests describing the existing code.  That this is needed shows a complexity we should get refactor in develop.

PR #1663, opened 2023-02-04
updated our version string definition

### LINKED ISSUE #1660: update our version string definition, see patch
from https://github.com/conda-forge/mss-feedstock/pull/141
```
diff -ruN -x __pycache__ -x obj -x '*~' -x tags -x TAGS MSS-7.0.7.orig/mslib/version.py MSS-7.0.7/mslib/version.py
--- MSS-7.0.7.orig/mslib/version.py	2022-12-07 18:16:45.000000000 +0100
+++ MSS-7.0.7/mslib/version.py	2023-01-30 15:46:44.944598202 +0100
@@ -24,4 +24,4 @@
 """
-__version__ = u'7.0.7.'
+__version__ = u'7.0.7'
```

Also check all code lines where we use that var.

With solving this we are compatible with recent setuptools.

and undo the fixation of setuptools in our meta.yaml

rootxrishabh:
fixes #1660 

setuptools now requires the __version__ variable to NOT end with a '.'


ReimarBauer:
The test ist now broken too, for what it should test, because the rule had changed. The test needs now to become more clever, see test_milestone_url(self):
It gives no failuren, because it finds now 7.0, which exists.
There should be another check, e.g. endswith(..) 

Please verify that this points to the right version. I assume that needs changes too.
https://github.com/Open-MSS/MSS/blob/stable/mslib/msui/msui.py#L331

lookup all places where that was imported.

ReimarBauer (changes requested):
see comments

rootxrishabh:
I have made some suggestions regarding the test, as for this [link](https://github.com/Open-MSS/MSS/issues?q=is%3Aclosed+milestone%3A{__version__[:-1]}). I found the result, attached as a screenshot.
image

rootxrishabh:
Also, I didn't find "self.milestone_url" anywhere other than the link which you attached.

ReimarBauer:
No but the "version" var you change if I search for that this is on many places used.  E.g. it is also responsible for creating the docs, for installing the right version.



e.g. 

version_string

improve the test in a way that it can't go wrong 

instead of the test code you can do that manually:


```
>>> import sys
>>> from mslib.msui import msui
>>> from PyQt5 import QtWidgets, QtTest
>>> application = QtWidgets.QApplication(sys.argv)
>>> window = msui.MSUI_AboutDialog()
>>> window.milestone_url
'https://github.com/Open-MSS/MSS/issues?q=is%3Aclosed+milestone%3A7.0.7'
```
The pattern matching is wrong for the current release, when you remove the 7, it gets 7.0
https://github.com/Open-MSS/MSS/blob/stable/mslib/msui/_tests/test_msui.py#L47

think about to get a good check for a version string, which is not a float number.

rootxrishabh:
I'll work on finding one

ReimarBauer:
this is one of the many descriptions of pytest around

https://realpython.com/pytest-python-testing/

rootxrishabh:
@ReimarBauer I have added the test please check.

mslib/_tests/test_version.py
```diff
@@ -0,0 +1,11 @@
+import pytest
```
ReimarBauer, line 27: have a look on our template https://github.com/Open-MSS/MSS/blob/develop/new_file_template.py
and use the header.

ReimarBauer (changes requested):
when I search for "version[" by pycharm, I guess any other IDE has also search capabilities, there is another python script which needs a fix.

ReimarBauer:
remove also the version pinning of setuptools as I entered as workaround.
https://github.com/Open-MSS/MSS/commit/db1fd0ae792212af494cc6ea33b296a52f7674cc

rootxrishabh:
@ReimarBauer fixing linters errors in this seem complicated, can you help?

rootxrishabh:
> remove also the version pinning of setuptools as I entered as workaround. [db1fd0a](https://github.com/Open-MSS/MSS/commit/db1fd0ae792212af494cc6ea33b296a52f7674cc)

Do you mean to add back the "." here?
https://github.com/Open-MSS/MSS/pull/1663/files#diff-76811c9b5f89d75f6c9f60f28c5347ba58baa0c8f42854a7b01d2b8648c27758

ReimarBauer:
> 



> > remove also the version pinning of setuptools as I entered as workaround. [db1fd0a](https://github.com/Open-MSS/MSS/commit/db1fd0ae792212af494cc6ea33b296a52f7674cc)
> 
> Do you mean to add back the "." here? https://github.com/Open-MSS/MSS/pull/1663/files#diff-76811c9b5f89d75f6c9f60f28c5347ba58baa0c8f42854a7b01d2b8648c27758

No, definitily not, the fix is to get the last dot removed. 

I mean you can search in pycharm and find where  the changes in slicing a string gets now wrong, when there is one character less than before.

the meaning of version pinning is, that there is used a "specific" version of some module. I changed the module versions in the docker container by adding setuptools.  My workaround was to use older versions of setuptools, 
You can see what I did, when you look at the commit, I showed.
https://github.com/Open-MSS/MSS/commit/db1fd0ae792212af494cc6ea33b296a52f7674cc

ReimarBauer:
btw. you could also run the linter test locally. It describes where the error is and how it should be fixed.

Is that link not working?
https://github.com/Open-MSS/MSS/actions/runs/4286831025/jobs/7466920729

mslib/_tests/test_version.py:34:1: E302 expected 2 blank lines, found 1
1     E302 expected 2 blank lines, found 1

If you want to run the linter test locally, you can take the line from
https://github.com/Open-MSS/MSS/blob/develop/.github/workflows/python-flake8.yml#L30

ReimarBauer:
The previous version of your new test was working, this version shows this failure
https://github.com/Open-MSS/MSS/actions/runs/4286831053/jobs/7466924218#step:9:668

rootxrishabh:
> > 
> 
> > > remove also the version pinning of setuptools as I entered as workaround. [db1fd0a](https://github.com/Open-MSS/MSS/commit/db1fd0ae792212af494cc6ea33b296a52f7674cc)
> > 
> > 
> > Do you mean to add back the "." here? https://github.com/Open-MSS/MSS/pull/1663/files#diff-76811c9b5f89d75f6c9f60f28c5347ba58baa0c8f42854a7b01d2b8648c27758
> 
> No, definitily not, the fix is to get the last dot removed.
> 
> I mean you can search in pycharm and find where the changes in slicing a string gets now wrong, when there is one character less than before.
> 
> the meaning of version pinning is, that there is used a "specific" version of some module. I changed the module versions in the docker container by adding setuptools. My workaround was to use older versions of setuptools, You can see what I did, when you look at the commit, I showed. [db1fd0a](https://github.com/Open-MSS/MSS/commit/db1fd0ae792212af494cc6ea33b296a52f7674cc)

Yes, I saw that you used a specific version of setuptools to make a workaround, I found this file "MSS-stable\docs\conf.py" with incorrect version slicing.

ReimarBauer:
Hi, this looks like you want to merge develop into stable.

Please compare the needed changes in the patch file I send you. It is much too much, changing now 318 files.

rootxrishabh:
Accidentally updated the branch using develop, I will fix this.

ReimarBauer:
please read the patch / diff file again, there is a missing change in the docs dir

mslib/_tests/test_version.py
```diff
@@ -0,0 +1,35 @@
...
+"""
+
+    mslib._tests.utils
+    ~~~~~~~~~~~~~~~~~~
+
+    This module provides common functions for MSS testing
+
+    This file is part of MSS.
+
+    :copyright: Copyright 2017 Reimar Bauer, Joern Ungermann
```
ReimarBauer: What is your name?

mslib/_tests/test_version.py
```diff
@@ -0,0 +1,35 @@
...
+    :copyright: Copyright 2017-2022 by the MSS team, see AUTHORS.
```
ReimarBauer: change to 2023-2023

because in this year it is a new file

mslib/_tests/test_version.py
```diff
@@ -0,0 +1,35 @@
+# -*- coding: utf-8 -*-
+"""
+
+    mslib._tests.utils
```
ReimarBauer: this is not the name of the test module

mslib/_tests/test_version.py
```diff
@@ -0,0 +1,35 @@
+# -*- coding: utf-8 -*-
+"""
+
+    mslib._tests.utils
+    ~~~~~~~~~~~~~~~~~~
+
+    This module provides common functions for MSS testing
```
ReimarBauer: what is the purpose of the test? This should be in this line. 

When you have imported a module, the doc string is accessible by the `__doc__` method

```
import test_version
print(test_version.__doc__)
```

Our IDEs shows than this information too.

mslib/_tests/test_version.py
```diff
@@ -0,0 +1,35 @@
...
+"""
+import pytest
+from packaging.version import Version
+from mslib import __version__
+
+
+def test_version_string():
+    Version(__version__)
+    with pytest.raises():
+        test_version_string()
```
ReimarBauer: have a look on the example in the diff file

ReimarBauer: usually
```
with pytest.raises():
E       TypeError: raises() missing 1 required positional argument: 'expected_exception'
```

ReimarBauer (changes requested):
compare to the example and adjust accordingly, name ;)
[version_patch.diff.txt](https://github.com/Open-MSS/MSS/files/10859955/version_patch.diff.txt)

ReimarBauer (approved):
thx :)

rootxrishabh:
@ReimarBauer Hey, I learned a lot in this PR, thank you supporting every step of the way : )

ReimarBauer:
@rootxrishabh You are welcome :)

PR #1671, opened 2023-02-26
added download folder for openaip

### LINKED ISSUE #1332: download folder in .config/msui for openaip

rootxrishabh:
fixes #1332


ReimarBauer (changes requested):
this didn't look correct

ReimarBauer:
That's an example of the UI how you can trigger a download.
airdata_example

rootxrishabh:
@ReimarBauer I triggered a download using the UI but can't find the file, please help.

rootxrishabh:
As you said in the issue that as of now we directly download into the MSS folder but this is it after downloading a file from the UI. 

image

ReimarBauer:
Have a look into your ~/.config/msui dir. After you did the same as me in the clip. We do want to change the organisation of these downloaded files. (headline of issue)

rootxrishabh:
Moin @ReimarBauer Found it. Also, I think the default path has to be changed from [here](https://github.com/Open-MSS/MSS/blob/develop/mslib/msui/constants.py#L36)

rootxrishabh:
I just checked by changing the path and it works, I'll go ahead and make a commit.

ReimarBauer:
I updated the dir example, because it was written before we renamed the mss to msui for the UI. Because MSS is everything not only an ui ;)

mslib/msui/constants.py
```diff
@@ -33,7 +33,7 @@
 
 # ToDo refactor to generic functions, keep only constants
 HOME = os.path.expanduser(f"~{os.path.sep}")
-MSUI_CONFIG_PATH = os.getenv("MSUI_CONFIG_PATH", os.path.join(HOME, ".config", "msui"))
+MSUI_CONFIG_PATH = os.getenv("MSUI_CONFIG_PATH", os.path.join(HOME, ".config", "msui", "downloads", "aip"))
```
ReimarBauer: this will shift the whole dependencies of everything related to MSUI_CONFIG_PATH to  .config/msui/downloads/aip

```
~/.config/msui/downloads/aip$ ls -rlt
insgesamt 12
-rw-rw-r-- 1 reimar reimar    2 Mar  8 13:39 mssautoplot.json
-rw-rw-r-- 1 reimar reimar 3206 Mar  8 13:39 msui.log
-rw-rw-r-- 1 reimar reimar   24 Mar  8 13:39 msui_settings.json
```


but we want only that the aip files will be stored in a different folder

rootxrishabh:
@ReimarBauer please take a look here.

ReimarBauer:
Hi, @rootxrishabh 
Have you seen my comment? It is not wanted that you change the whole dir, were all user configurations files gets added. The files in the default dir should be sorted into different subdirectories. Currently they are all in '.'.

rootxrishabh:
Moin @ReimarBauer, I have added the downloads/aip directories but not sure what to do with plugins and gravatars directories.

ReimarBauer:
> Moin @ReimarBauer, I have added the downloads/aip directories but not sure what to do with plugins and gravatars directories.

Moin @rootxrishabh make first this openaip working, see failures in tests. We will create new issues to improve the others.
And changes to plugins will get done after we released 8.0.0. This has too much impact to current campaign users.

rootxrishabh:
@ReimarBauer is it possible to setup up these failing tests locally? So that I can debug them properly.

ReimarBauer:
> @ReimarBauer is it possible to setup up these failing tests locally? So that I can debug them properly.

using your mssdev environment, you can call these on commandline 

/2023/rootxrishabh/MSS$ pytest -k test_get_airspaces

or also by your IDE, pycharm recognizes them for pytest. There you can set breakpoints. I assume other IDEs do that similiar.

rootxrishabh:
Moin @ReimarBauer , need some help with debugging. So basically the path(OSDIR) seems to disturb the function (get_airspaces).
image
This results in an empty array and an assertion error [here](https://github.com/Open-MSS/MSS/blob/develop/tests/_test_utils/test_airdata.py#LL170C5-L170C26)

rootxrishabh:
Moin @ReimarBauer, The tests are now solved, please take a look.
image
The files are now being downloaded in the downloads/aip directory.

mslib/msui/constants.py
```diff
@@ -33,6 +33,7 @@
 
 # ToDo refactor to generic functions, keep only constants
 HOME = os.path.expanduser(f"~{os.path.sep}")
+DOWNLOAD_PATH = os.getenv("MSUI_CONFIG_PATH", os.path.join(HOME, ".config", "msui", "downloads", "aip"))
```
ReimarBauer: can you move this into the other module, because it is specific for aip in airdata only. It won't be used anywhere else.

ReimarBauer (changes requested):
see comment, try to minimize required changes.

I do later a practical test.

rootxrishabh:
@ReimarBauer I have made the required changes.

mslib/utils/airdata.py
```diff
@@ -37,10 +37,10 @@
 import time
 
 import defusedxml.ElementTree as etree
-from mslib.msui.constants import MSUI_CONFIG_PATH
+from mslib.msui.constants import HOME
 
-
-OSDIR = fs.open_fs(MSUI_CONFIG_PATH).root_path
+DOWNLOAD_PATH = os.getenv("MSUI_CONFIG_PATH", os.path.join(HOME, ".config", "msui", "downloads", "aip"))
+OSDIR = fs.open_fs(DOWNLOAD_PATH).root_path
```
ReimarBauer: when the downlads dir not exists, it crashes

`~/.config/msui$ rm -rf downloads/`

fs.errors.CreateFailed: root path '/home/reimar/.config/msui/downloads/aip' does not exist

mslib/utils/airdata.py
```diff
@@ -37,10 +37,10 @@
 import time
 
 import defusedxml.ElementTree as etree
-from mslib.msui.constants import MSUI_CONFIG_PATH
```
ReimarBauer, line 40: should be imported

mslib/utils/airdata.py
ReimarBauer: also you better do the join to the OSDIR.  
`OSDIR = fs.open_fs(MSUI_CONFIG_PATH).root_path`

 os.getenv("MSUI_CONFIG_PATH", ...
 
 means that when the env var is set, the second parameter isn't used.

ReimarBauer: for example if I want to use my configurations from a different config dir
 
 export MSUI_CONFIG_PATH=~/tmp

I also want than on this dir a download folder and not what we have in 7.x

ReimarBauer (changes requested):
see comments

mslib/utils/airdata.py
rootxrishabh: > 
Moin @ReimarBauer 

If I don't use the os.getenv function like this 
image

Then I get this error 
image

This error states that the variable [MSUI_CONFIG_PATH] has some values /tmp/tmp*
However when setting the OSDIR value to this(image given below)
image
The test successfully passes.

> when the downlads dir not exists, it crashes
> 
> `~/.config/msui$ rm -rf downloads/`
> 
> fs.errors.CreateFailed: root path '/home/reimar/.config/msui/downloads/aip' does not exist

Moreover this directory issue has been solved

rootxrishabh: > also you better do the join to the OSDIR. `OSDIR = fs.open_fs(MSUI_CONFIG_PATH).root_path`
> 
> os.getenv("MSUI_CONFIG_PATH", ...
> 
> means that when the env var is set, the second parameter isn't used.

I have implemented all the other things. Just one more thing. Do you mean that "os.getenv" should not be used?

ReimarBauer: yes.

because we want to have a download dir always in an OSDIR dir.
when I set an env var MSUI_CONFIG_PATH your default is ignored.

https://docs.python.org/3/library/os.html#os.getenv

ReimarBauer (changes requested):
found a problem, seems we also don't have a test for this case

airdata_problem

You maybe could also add this missing test? 
When you write the testcase first it should fail now and with the fix done it shows that the problem is solved.

rootxrishabh:
Sure. I will get to it.

ReimarBauer:
Wait, it may be a download problem on my site

ReimarBauer:
The previous gifcycle I can't reproduce but:

It is some different problem, now it shows on that "apply" button in my terminal 

```
ERROR: Tried to draw airports but none were found. Try redownloading.
INFO: Downloading to /home/reimar/.config/msui/downloads/aip/be_asp.xml. This might take a while.
INFO: 433KB Downloaded
ERROR: Tried to draw airspaces without asp files.
ERROR: Tried to draw airports but none were found. Try redownloading.
ERROR: Tried to draw airspaces without asp files.
```
seems it did not find the downloaded files?

develop branch did not show this problem

Can you reproduce this?

rootxrishabh:
> The previous gifcycle I can't reproduce but:
> 
> It is some different problem, now it shows on that "apply" button in my terminal
> 
> ```
> ERROR: Tried to draw airports but none were found. Try redownloading.
> INFO: Downloading to /home/reimar/.config/msui/downloads/aip/be_asp.xml. This might take a while.
> INFO: 433KB Downloaded
> ERROR: Tried to draw airspaces without asp files.
> ERROR: Tried to draw airports but none were found. Try redownloading.
> ERROR: Tried to draw airspaces without asp files.
> ```
> 
> seems it did not find the downloaded files?
> 
> develop branch did not show this problem
> 
> Can you reproduce this?

Yes, I can reproduce this.
Looking for the cause.

rootxrishabh:
Moin @ReimarBauer 
I checked the file(/home/rishabh/Desktop/MSS/mslib/msui/mpl_map.py) from which the logging error was generated. In the file at line number 379, a "draw_airspaces" function is being called which calls the "get_airspaces" function from the airdata.py file which returns an array, I checked the array and it was correct, moreover, I tried changing the logging statement from "Tried to draw airspaces without asp files." to "aaaaaaaa" to check if the statement is being executed. But it failed and so I think either the function is not being called which is highly unlikely or the "if" statement on line number 383.

This image shows that the "get_airspaces" function is generating correct information.
image

ReimarBauer (approved):
Thx

PR #1686, opened 2023-03-05
verifying if time.sleep() solves the _test_msui/test_mscolab timeouts

### LINKED ISSUE #1685: tests can have a timeout
random msui tests reach a timeout 

```
  File "/home/reimar/mambaforge/envs/mssdev/lib/python3.10/site-packages/urllib3/connectionpool.py", line 449, in _make_request
    six.raise_from(e, None)
  File "<string>", line 3, in raise_from
  File "/home/reimar/mambaforge/envs/mssdev/lib/python3.10/site-packages/urllib3/connectionpool.py", line 444, in _make_request
    httplib_response = conn.getresponse()
  File "/home/reimar/mambaforge/envs/mssdev/lib/python3.10/http/client.py", line 1374, in getresponse
    response.begin()
  File "/home/reimar/mambaforge/envs/mssdev/lib/python3.10/http/client.py", line 318, in begin
    version, status, reason = self._read_status()
  File "/home/reimar/mambaforge/envs/mssdev/lib/python3.10/http/client.py", line 279, in _read_status
    line = str(self.fp.readline(_MAXLINE + 1), "iso-8859-1")
  File "/home/reimar/mambaforge/envs/mssdev/lib/python3.10/socket.py", line 705, in readinto
    return self._sock.recv_into(b)
TimeoutError: timed out
```
I assume we close the application before socketio has read the whole information or we need a bit more time for the connection.

Because of the discussion on https://stackoverflow.com/questions/67913861/python-multiprocessing-weird-behavior-when-not-using-time-sleep 

I try with time.sleep() instead of qwait

ReimarBauer:
closes #1685

### DIFF
--- a/tests/_test_msui/test_mscolab.py
+++ b/tests/_test_msui/test_mscolab.py
@@ -684,6 +684,9 @@ class Test_Mscolab(object):
         self.connect_window.show()
         QtTest.QTest.mouseClick(self.connect_window.connectBtn, QtCore.Qt.LeftButton)
         QtWidgets.QApplication.processEvents()
+        # socketio multiprocessing can reach a timeout
+        # idea from discussion
+        # https://stackoverflow.com/questions/67913861/python-multiprocessing-weird-behavior-when-not-using-time-sleep
         time.sleep(0.5 * random.random())
 
     def _login(self, emailid, password):
--- a/tests/_test_msui/test_mscolab.py
+++ b/tests/_test_msui/test_mscolab.py
@@ -26,6 +26,9 @@
 """
 import sys
 import os
+import time
+import random
+
 import fs
 import fs.errors
 import fs.opener.errors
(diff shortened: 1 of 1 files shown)


joernu76:
The linked discussion suggests that a "join" like statement is missing somewhere...?

joernu76:
I would suggest from reading up on this to add a "self.process.join()" after every "self.process.terminate()" to wait for the termination to have cleanly occured.  Otherwise this opens up race conditions for multiple servers being present (some in construction, some in termination).
Beyond that, if the sleep makes everything safer, why not?

ReimarBauer:
I had first time that crash locally and it seems to help. 

lets try this.self.process.join() too


I assume it goes in the direction that the event based socketio wants to access the server when it is already terminated or not fully initialized. the timeout usually happens when it wants to read but can't.

joernu76:
Can we (simply) add a queue (or similar) for more communication between the sub process and master? For waiting on proper initialization and command to terminate on its own?

joernu76:
BTW; the cm/fm arguments on start_server are ignored. correctly so??

ReimarBauer:
> Can we (simply) add a queue (or similar) for more communication between the sub process and master? For waiting on proper initialization and command to terminate on its own?

I have no idea yet, how something like this keeps the event based concept without transforming it to a request based one.

ReimarBauer:
> BTW; the cm/fm arguments on start_server are ignored. correctly so??

yes, these are client side tests, not needed in this case, can be cleaned a bit.

PR #1703, opened 2023-03-20
color needs to become initialzed independent from qsettings

### LINKED ISSUE #1701: Multiple Flightpath option resulting in an error
Steps to Reproduce:-
-> python mslib/msui/msui.py
->open top view, we can see that the Web Map Service is open but in the drop-down menu (select to open control) is shown.
   Even on changing the control, the text does not change
   

https://user-images.githubusercontent.com/93277471/225754881-9fe3fa21-1d26-410a-afea-5a9853f2340e.mp4

ReimarBauer:
fixes #1701

### DIFF
--- a/mslib/msui/multiple_flightpath_dockwidget.py
+++ b/mslib/msui/multiple_flightpath_dockwidget.py
@@ -127,7 +127,7 @@ class MultipleFlightpathControlWidget(QtWidgets.QWidget, ui.Ui_MultipleViewWidge
         self.mscolab_server_url = mscolab_server_url
         self.token = token
         self.ft_settings_dict = self.ui.get_settings()
-        self.color = self.ft_settings_dict["colour_ft_vertices"]
+        self.color = (0, 0, 1, 1)
         self.obb = []
 
         self.operation_list = False
--- a/tests/_test_msui/test_multiple_flightpath_dockwidget.py
+++ b/tests/_test_msui/test_multiple_flightpath_dockwidget.py
@@ -1,10 +1,10 @@
 # -*- coding: utf-8 -*-
 """
 
-     tests._test_msui.test_multiple_flightpath_dockwidget
-    ~~~~~~~~~~~~~~~~
+    _tests._test_msui.test_multiple_flightpath_dockwidget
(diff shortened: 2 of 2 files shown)


mslib/msui/multiple_flightpath_dockwidget.py
```diff
@@ -127,7 +127,7 @@ def __init__(self, parent=None, view=None, listFlightTracks=None,
         self.mscolab_server_url = mscolab_server_url
         self.token = token
         self.ft_settings_dict = self.ui.get_settings()
-        self.color = self.ft_settings_dict["colour_ft_vertices"]
+        self.color = (0, 0, 1, 1)
```
joernu76, line 130: I have not understood the issue you encountered, but as I understand it, this dictionary should be intialized properly before this class in the general case...?
Maybe a .get() with this as default value?

joernu76 (approved):

joernu76:
I just had a look into how the settings are handled and it is a mess. I am not sure what it intends to do, but currently SideView and TopView handle this differently. There are 2-4 places where this variable is initialized with a default value, which are too many, and this'd introduce a new one.
I'll try to understand what is going on and provide a decent fix.

ReimarBauer:
@joernu76 the bug is easyily  to reproduce when you set the env var  MSUI_CONFIG_PATH to a new dir and do nothing else as trying to load the widget.

joernu76:
Thanks, I will try.
The major point is that this particular get_settings should maybe not exist and certainly *not* load from disk. This is a bug introduced together with this feature.

mslib/msui/multiple_flightpath_dockwidget.py
ReimarBauer: for a new user there is not that dict defined. Or also not when one deletes the msui-core.conf file. We should not depend on those qsettings created files.

ReimarBauer:
> Thanks, I will try. The major point is that this particular get_settings should maybe not exist and certainly _not_ load from disk. This is a bug introduced together with this feature.

I also wanted to have an initial test definition so that more can be done.

Issue #1649, opened 2023-01-06
satellite prediction writes outside the graph


ReimarBauer:
The tutorial_satellitetrack.py found a problem using sattelite tracks

It seems when data is not in the range of the map text is printed outside, see example image

used data
docs/samples/satellite_tracks/satellite_predictor.txt

![sattelite_prediction_bug](https://user-images.githubusercontent.com/278728/211037222-0050552a-3951-4477-93ab-bdd3107876b1.png)

jalajk24:
Can I work on this issue? I am pretty new to this

ReimarBauer:
you have it, please send a Pull Request to stable branch.

ReimarBauer:
![tutorial_satellitetrack](https://user-images.githubusercontent.com/278728/215855008-55f6f107-0284-4974-9e79-bdc69eb504df.gif)

ReimarBauer:
@jalajk24 have a look on the gifcycle to see where this function is located in the UI.

HarshKhilawala:
@ReimarBauer I would like to work on it. Please assign this issue to me.

Bond010:
This bug is due to incorrect configuration or incorrect data input.

PR #1709, opened 2023-03-21
Fix text data outside map canvas

HarshKhilawala:
This WIP PR fixes issue #1649


HarshKhilawala:
@ReimarBauer Here is the draft PR. 

I suggest: Not to allow labels outside of map if the coordinates of satellite is outside graph region. This can be achieved through Satellite_control_widget.view in satellite_dockwidget.py. 

I am open to feedback/approach if that solves this problem/fixes this bug.

ReimarBauer:
Please continue and show you can fix it.

ReimarBauer:
reopen when you continue

PR #1718, opened 2023-03-22
entrypoint for mssautoplot added

### LINKED ISSUE #1717: entrypoint for mssautoplot
we need an entrypoint for the new script

ReimarBauer:
closes #1717


joernu76 (approved):
Would this be the time to fix the name, when it doesn't hurt, yet, to msautoplot to be consistent?

PR #1724, opened 2023-03-26
v1.2.2 of workflow

### LINKED ISSUE #1723: Parameter token or opts.auth is required
new image creation fails

- dispatch workflow needs update
- token needs to become passed as param

ReimarBauer:
helpful 
- https://github.com/orgs/community/discussions/26705
- https://github.com/actions/toolkit/issues/465

- classic token required 
- secrets are not available to forks


joernu76 (approved):
This change away from 1.2.2 to a cryptic hash might deserve a comment.
In particular with the MR name.

PR #1728, opened 2023-03-31
rename the mscolab server configuration USER_VERIFICATION = False

### LINKED ISSUE #1726: rename the mscolab server configuration  USER_VERIFICATION = False
We have now the option to verify a users email adress and also to reset a password.

The server configuration option USER_VERIFICATION = False is now misleading.

It should be renamed to MAIL_ENABLED, default should be False, because it needs some configurations for sending mails.

Only when it is True the server should send on a new user the confirmation mail and on password reset request the reset password mail

The "MAIL_ENABLED" flag has the function that without mail configured the server can start.

nilupulmanodya:
fixes #1726


mslib/mscolab/conf.py
```diff
@@ -79,7 +79,7 @@ class mscolab_settings(object):
         enable_basic_http_authentication = False
 
         # enable verification by Mail
-        USER_VERIFICATION = False
+        MAIL_ENABLED = False
```
ReimarBauer, line 82: I have a valid email configuration and it sends me the confirmation email also when MAIL_ENABLED is False.

I think it should silently create the account without mailing me the verification token.

ReimarBauer: using that link gives a 500, so when we not send it, this won't happen.

```
ERROR: Exception on /confirm/TOKEN [GET]
  File "/home/icg105/mambaforge/envs/mssenv/lib/python3.10/site-packages/flask/app.py", line 2525, in wsgi_app
    response = self.full_dispatch_request()
  File "/home/icg105/mambaforge/envs/mssenv/lib/python3.10/site-packages/flask/app.py", line 1823, in full_dispatch_request
    return self.finalize_request(rv)
  File "/home/icg105/mambaforge/envs/mssenv/lib/python3.10/site-packages/flask/app.py", line 1842, in finalize_request
    response = self.make_response(rv)
  File "/home/icg105/mambaforge/envs/mssenv/lib/python3.10/site-packages/flask/app.py", line 2134, in make_response
    raise TypeError(
TypeError: The view function for 'confirm_email' did not return a valid response. The function either returned None or ended without a return statement.
```

ReimarBauer: the url for Recover Your Password is protected by the 403 status page. :)  :heart:

mslib/mscolab/server.py
```diff
@@ -259,7 +259,7 @@ def user_register_handler():
     try:
         if result["success"]:
             status_code = 201
-            if mscolab_settings.USER_VERIFICATION:
+            if mscolab_settings.MAIL_ENABLED:
                 status_code = 204
             token = generate_confirmation_token(email)
```
ReimarBauer, line 264: this indenting looks wrong, likly we need only that token until sendmail when MAIL_ENABLED is True

ReimarBauer: from token to sendmail needs to be intended. This is only needed when MAIL_ENABLED is True

on the subject below please use the same prefix as for the other mail.

ReimarBauer (changes requested):
see comments

mslib/mscolab/server.py
```diff
@@ -259,21 +259,21 @@ def user_register_handler():
...
-            token = generate_confirmation_token(email)
-            confirm_url = url_for('confirm_email', token=token, _external=True)
-            html = render_template('user/activate.html', username=username, confirm_url=confirm_url)
-            subject = "Please confirm your email"
-            send_email(email, subject, html)
-        return jsonify(result), status_code
+                token = generate_confirmation_token(email)
+                confirm_url = url_for('confirm_email', token=token, _external=True)
+                html = render_template('user/activate.html', username=username, confirm_url=confirm_url)
+                subject = "Please confirm your email"
```
ReimarBauer: We should use the same prefix as for  the Passwort Reset Mail
MSColab Please confirm your email

Everything else works as expected :)

mslib/mscolab/server.py
ReimarBauer: That works all now as expected, only the MSColab Mail prefix is missing.

ReimarBauer (changes requested):
one last change, prefix name for email subject

ReimarBauer (approved):
Thx

PR #1731, opened 2023-04-01
Refactor keyring

### LINKED ISSUE #1730: refactor usage of keyring
simplify several http-auth authentications

ReimarBauer:
closes #1730

### DIFF
--- a/docs/samples/automation/retriever.py
+++ b/docs/samples/automation/retriever.py
@@ -165,7 +165,7 @@ def main():
             width, height = int(round(ax_bounds[2])), int(round(ax_bounds[3]))
             bbox = params['basemap']
             req = requests.get(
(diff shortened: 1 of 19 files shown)


joernu76:
I think, we should have some kind of backwards-compatibility in the next version, i.e. something like a check on both (old and new) entries, perhaps with a warning in case data is found in the old that you can click away once per start?

ReimarBauer:
We likly need a kind of migration script. I do include this

ReimarBauer:
@joernu76 the new test and the migration is version dependent. We could maybe merge the release preparation PR.

ReimarBauer:
When it recognizes an old configuration it shows
message

ReimarBauer:
the documentation is rendered like this
rendered_docs

joernu76 (approved):

PR #1757, opened 2023-04-27
Issue1754

joernu76:


ReimarBauer (changes requested):
please send this to stable

CHANGES.rst
```diff
@@ -1,6 +1,15 @@
 Changelog
 =========
```
ReimarBauer: there is a missing merge from stable to development. In stable we have this. see PR ;)

ReimarBauer (commented):
tried it, works.

There could be a cosmetic change done. 

The other labels e.g. "Flight Tracks:" and "Open Views:" do endwith a ":". We missed that for "Operations" and "Archived Operations"
 
 
 This could also become a good first issue.

joernu76:
Should have gone to stable. now it goes.

ReimarBauer (dismissed):
Thx

ReimarBauer:
after Unarchiving the bold formatted PH1 gets cleared the styles

problem

ReimarBauer (changes requested):
see comment

ReimarBauer:
this needs a refactoring for next major too.
https://github.com/Open-MSS/MSS/blob/develop/mslib/mscolab/models.py#L136

This has to be done there because this will require a database migration too

joernu76:
I think that bug was there before and the other pull request #1761 might fix it. At least it solves some related issues...

As you can see from the changes, this PR does not do functional changes and I'd like to keep it this way.

joernu76:
> this needs a refactoring for next major too. https://github.com/Open-MSS/MSS/blob/develop/mslib/mscolab/models.py#L136
> 
> This has to be done there because this will require a database migration too

internally, there is a lot of renaming needed, which I mostly defer here to keep the amount of changes to stable minimal.
I will work on fixes in develop as soon as this has been merged and brought into develop.

ReimarBauer:
This can be used for testing, on an created operation

```
>>> from mslib.mscolab.seed import archive_operation
WARNING:root:Couldn't import mscolab_auth (ImportError:'{No module named 'mscolab_auth'), creating dummy config.
>>> archive_operation("v1","aa@aa")
```

that can be later renamed for toggling the state, currently I got the same situation, I've unarchived all I had

ReimarBauer:
ok not readed all, can remove my addition, what do you think?

ReimarBauer:
> I think that bug was there before and the other pull request #1761 might fix it. At least it solves some related issues...
> 
> As you can see from the changes, this PR does not do functional changes and I'd like to keep it this way.

verfied that was before too

ReimarBauer:
that seed function would be needed to add tests in stable which I also prefer for bug fixes

ReimarBauer (approved):
ok, does what is expected verified manually

Issue #1755, opened 2023-04-27
double-clicking archived/inactive operations introduces superfluous secondary selection


joernu76:
inactive/archived operations can only be "activated/unarchived". Currently one can selecte them with the "blue bar" *and* bold by double-clicking. Remove double-clicking.

Issue #1752, opened 2023-04-27
UI bug in bold printing flighttracks/operations


joernu76:
By selecting first an operation, then double-clicking an inactive operation, and then selecting a flight track, both operation and flight-track are bolded. 

Desired state: only flight-track is bold.

joernu76:
![Peek 2023-04-27 08-41](https://user-images.githubusercontent.com/28449201/234787346-5e3a37e0-c188-4569-b0ba-ff0008e04327.gif)

PR #1759, opened 2023-04-27
Small UI bug fixes to MSUI/MSCOlab UI

### LINKED ISSUE #1756: "activate Operation" enabled after starting MSUI
should start disabled.

joernu76:
* Only one flighttrack/operation/inactive operation can be selected.
* "Activate operation" is now disablked by default
* Only one flighttrack/operation can be displayed in bold

Fix #1756, #1755, #1752

### DIFF
--- a/mslib/msui/mscolab.py
+++ b/mslib/msui/mscolab.py
@@ -448,6 +448,7 @@ class MSUIMscolab(QtCore.QObject):
         self.ui.usernameLabel.hide()
         self.ui.userOptionsTb.hide()
         self.ui.actionAddOperation.setEnabled(False)
+        self.ui.actionActivateOperation.setEnabled(False)
         self.hide_operation_options()
         self.ui.activeOperationDesc.setHidden(True)
 
@@ -1482,7 +1483,7 @@ class MSUIMscolab(QtCore.QObject):
                     self.ui.listOperationsMSC.setCurrentItem(selectedOperation)
                     self.ui.listOperationsMSC.itemActivated.emit(selectedOperation)
                 self.ui.listOperationsMSC.itemActivated.connect(self.set_active_op_id)
-                self.ui.listInactiveOperationsMSC.itemActivated.connect(self.select_inactive_operation)
+                self.ui.listInactiveOperationsMSC.itemClicked.connect(self.select_inactive_operation)
             else:
                 show_popup(self.ui, "Error", "Session expired, new login required")
                 self.logout()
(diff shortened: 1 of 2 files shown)


joernu76:
Seems to work, but this is pretty brittle.

The "selection" is also fully useless in operations as the Menu always operates on the "bolded"/active_op_id, not the selected one.
I would classify that as a bug as well. The menu should operate on the selected one (blue bar, if any) while the bolded one is the one used in the views.
I couldn't test the proper enabling/disabling of inactive operations as I have none on the server. Simply for testing an "archive operatiopn" action would be useful.

I propose quick fixes for stable and a proper refactorization for develop...?

ReimarBauer:
show_unarchive

ReimarBauer:
> Seems to work, but this is pretty brittle.
> 
> The "selection" is also fully useless in operations as the Menu always operates on the "bolded"/active_op_id, not the selected one. 

Well that was copied behaviour of Flight Track. In both lists selection have not any functions. And I doubt that it can be used without accessing the data. We use the bold font to indicate what was activated by double click. This can be seen when something else gets pointless selected.
But why don't we activate by selecting? Why was the double click introduced?

Ok the difference to Flight Track is one can close one by selection which doesn't have been activated. And in Operations one can only delete the activated one. When the activated OP is deleted there is a misbehaviour, the controls of the waypoints should become disabled until another OP or Flighttrack becomes selected.
(The Operations in this case use as fall back the stub data from the configuration)
At the moment we have archive for Operations, this is quite close to Close of a Flighttrack.


>I would classify that as a bug as well. The menu should operate on the selected one (blue bar, if any) while the bolded >one is the one used in the views. 

This did not work, we need to get access to the data from the server when we want to change that data. e.g. we could not delete an OP when we don't have logged in to that OP.
We could maybe replace double click by selection.

>>I couldn't test the proper enabling/disabling of inactive operations as I have none on >the server. Simply for testing an "archive operatiopn" action would be useful.

yes, we need an archive menu and a cli command.

> 
> I propose quick fixes for stable and a proper refactorization for develop...?

I would not touch selection and double click in 8, I think we can remove double click and selection is enough.

ReimarBauer (approved):
Thx, see my answers :)

ReimarBauer:
Hmm, Not sure if that is wanted bahavior. The selection is now only in the inactive operation and on open views, before it was also on the bold marked operation or flight track. 

before

before

after

after


When we move inactive Operations to its own widget, I think selecting and keeping what is selected can replace double click

On the Open views we have a similar problem. There can be a view selected which is not the one I see.

joernu76:
Yes, the whole UI is not fully consistent. The selection in views is also fully useless, but I couldn't remove it.

Having multiple selections makes it difficult for the user to find out on which selection the menu action operates. At least only one flight track/operation can be selected now, which makes it a tad simpler to know beforehand what will happen...

AFAIK MSS *must* have an active flighttrack/operation. E.g., this is why you cannot delete the active flight track. I would have assumed the same holds true for operations...? This is another bug provoking a crash. I will file an issue.

We may need a full redesign in 9, but this goes beyond a UI bug fix.
Your description of MSCOlab limitations seem to make this difficult/impossible to achieve.

PR #1761, opened 2023-04-28
Fixed several activation/selection bugs in flight track/operation han…

### LINKED ISSUE #1760: Deleting an "active" operation leaves no active operation
This causes an Exception as MSS must always have an active operation.

I propose as quick fix for stable to simply activate the always being present local flight track.

For 9.x, I propose a redesign, where the selection determines, which operation the actions operate on, similar to the "inactive op id" for inactive operations. If it works for archived operations it should work here as well.

Peek 2023-04-28 08-53

joernu76:
…dling

Fix #1760


mslib/msui/mscolab.py
```diff
@@ -1035,13 +1041,17 @@ def handle_delete_operation(self):
                         res.raise_for_status()
                         self.reload_operations()
                         self.signal_operation_removed.emit(self.active_op_id)
+                        logging.debug("activate local")
+                        self.ui.listFlightTracks.setCurrentRow(0)
```
ReimarBauer, line 1045: what happens if there is None other operation?

ReimarBauer: checked this, it falls back to the flight tracks, that ever exists. :)

That one has None and the other the name is also something I was not able to catch yet

removing_name_thing

ReimarBauer:
There is an edge case not handled, which needs to be checked when we have an "archive" function from the UI. This needs to trigger updating the views on participants using that OP. Currenly they get them into the archived list but the data is still in the UI and can be changed.

(This does not occure now)

ReimarBauer (approved):
Thx

PR #1781, opened 2023-05-16
Add an option to use pcolormesh instead of contourf for generic hsec …

### LINKED ISSUE #1780: MSWMS: Display coarse data with a lot of NaNs correctly
I was trying to add MLS satellite data to my MSWMS Server and encountered the following problem:

- MLS satellite data is provided as level-2 data on orbits, which I gridded on daily maps on a regular lat/lon/pressure grid. Naturally, a lot of grid points are not filled with data (since interpolation and creation of "real" level-3 data is quite challenging and may impose artifacts). As a result, the netCDF variable contains a lot of NaNs. 
- The grid sampling I chose was rather coarse, approximately in the range of the best horizontal resolution of the satellite (approx. 1.5°)

As a result, MSWMS shows this kind of data only very sparsely, I think due to the fact that it tries to interpolate the data on a finer grid, the NaNs are interpolated with the available data, which resulted again in a NaN. So only a few data points are shown in the horizontal map (at locations where orbits crossed and more points are available). 

As a workaround for this issue, I used a finer grid and repeated all of the satellite measurements to the four neighboring lat/lon points. This helps to show the satellite data on kind-of orbits without too many plotting artifacts.

However, I think this issue should be avoided at all in case of coarse and patchy data, doesn't it? Maybe satellite retrospective data is not the primary data used to set up MSWMS, but I find it very handy to have all relevant data at hand for a campaign...

joernu76:
…plots

Fix #1780

### DIFF
--- a/mslib/mswms/mpl_hsec_styles.py
+++ b/mslib/mswms/mpl_hsec_styles.py
@@ -193,7 +193,7 @@ def make_generic_class(name, standard_name, vert, add_data=None, add_contours=No
             Defaults to None.
 
         use_pcolormesh (bool, optional): determines whether to use pcolormesh
-            uor plotting instead of the default "contourf" method. Use
+            or plotting instead of the default "contourf" method. Use
             pcolormesh for data that contains a lot of fill values or NaNs,
             or to show the actual location of data.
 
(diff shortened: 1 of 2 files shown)


joernu76:
@soerenjoh

mslib/mswms/mpl_hsec_styles.py
```diff
@@ -124,47 +127,81 @@ def _plot_style(self):
...
+
+        add_prepare (function, optional): a function to overwrite the
+            _prepare_datafield method. Use this to add derived quantities based
+            on those provided by the modes. For example 'horizontal_wind' could
+            be computed from U and V in here.
+
             Defaults to None.
+
+        use_pcolormesh (bool, optional): determines whether to use pcolormesh
+            uor plotting instead of the default "contourf" method. Use
```
ReimarBauer: cosmetic, typo

ReimarBauer (approved):
thx

PR #1782, opened 2023-05-16
Revamp of mscolab operation menu and addition of "archive" operation

joernu76:


mslib/mscolab/server.py
```diff
@@ -59,6 +59,8 @@
 migrate = Migrate(APP, db, render_as_batch=True)
 auth = HTTPBasicAuth()
 
+ARCHIVE_THRESHOLD = 30
```
ReimarBauer, line 62: we should maybe make this configurable.

mslib/mscolab/server.py
```diff
@@ -489,13 +491,17 @@ def get_operation_details():
...
     temp_operation_active = operation.active
-    operation.active = True
+    if days_ago > ARCHIVE_THRESHOLD:
+        operation.active = False
+    else:
+        operation.active = True
     db.session.commit()
     # Reload Operation List
-    if not temp_operation_active:
+    if temp_operation_active != operation.active:
```
ReimarBauer, line 504: the wording maybe can be changed to archived, unarchived etc.

joernu76: changing operation.active requires changes to the DB, I guess?

mslib/msui/mscolab.py
```diff
@@ -1084,27 +1141,57 @@ def handle_leave_operation(self):
...
-                self.ui.tr(f"{self.active_operation_name}: {self.active_operation_desc}"))
+                self.ui.tr(f"{self.active_operation_name}: {self.active_operation_description}"))
         else:
             self.ui.activeOperationDesc.setText(
                 "Description is too long to show here, for long descriptions go "
                 "to operations menu.")
 
-    def update_description_handler(self):
+    def change_category_handler(self):
+        # only after login
```
ReimarBauer, line 1155: just seen that the comment is on two places in the source

the verification "after login" is done on the server.

verify_user_token is a debugging feature which is skipped by default user configuration. When it is not skipped it forces each request to be checked with the server alive and this slows down any event and helps to debug some problems.

mslib/mscolab/server.py
joernu76: Maybe. 30 days seems a tad short for me as a typical campaign may run longer.
New issue?

ReimarBauer: yes, this needs also a documentation

mslib/mscolab/server.py
ReimarBauer: yes, in the models. 

and on staging you could do this by
https://mss.readthedocs.io/en/stable/mscolab.html#data-base-migration

ReimarBauer: review and also add new parts to the section _Operation based features_ in

https://github.com/Open-MSS/MSS/blob/develop/docs/mscolab.rst#operation-based-features

mslib/msui/qt5/ui_mainwindow.py
```diff
@@ -298,30 +303,30 @@ def retranslateUi(self, MSUIMainWindow):
...
         self.serverOptionsCb.setItemText(1, _translate("MSUIMainWindow", "Fetch From Server"))
         self.serverOptionsCb.setItemText(2, _translate("MSUIMainWindow", "Save To Server"))
-        self.workLocallyCheckbox.setToolTip(_translate("MSUIMainWindow", "Check to work asynchronously from the server"))
-        self.workLocallyCheckbox.setText(_translate("MSUIMainWindow", "Work Asynchronously"))
+        self.categoryLabel.setText(_translate("MSUIMainWindow", "Category:"))
         self.listOperationsMSC.setToolTip(_translate("MSUIMainWindow", "List of mscolab operations.\n"
 "Double click a operation to activate and view its description."))
-        self.activeOperationDesc.setText(_translate("MSUIMainWindow", "Select Operation to View Description"))
-        self.activeOperationsLabel.setText(_translate("MSUIMainWindow", "Operations"))
+        self.pbOpenOperationArchive.setText(_translate("MSUIMainWindow", "Operation Archive"))
```
open-mss-build, line 321: plural?

on the upper list it is also plural and a : is missing on upper one.

ReimarBauer (changes requested):
tried few functions, seems that the archive button does not update the lists or is not working? Likly there an event trigger missing.
archive_problem

joernu76:
> tried few functions, seems that the archive button does not update the lists or is not working? Likly there an event trigger missing. 

This works, when the mscolab server has been updated. I just hot-patched the mss-stage server for testing and it works.

mslib/msui/mscolab.py
joernu76: No idea, I copy pasted that comment in changing category over from the changing of description. I am still not 100% sure about how that authentification and verification works.

mslib/msui/qt5/ui_mainwindow.py
joernu76: singular in front of "archive" seems to be slightly more common than plural according to google

mslib/mscolab/server.py
joernu76: I am not convinced that we should change the DB format to be more consistent in a part that is not user-visible.

ReimarBauer:
When a category is changed, other users have the old entry in theire opened "Manage Users" - "Clone operation" list selection. We maybe should update this too. 
On the one who does the category change the window of "Manage Users" gets closed.

mslib/msui/mscolab.py
ReimarBauer: I'ver overseen that there, the server has to do all that verification.

mslib/mscolab/server.py
ReimarBauer: We can do this when we need another change. It is not needed now, but at the moment we introduce something new we anyways have to do a server database migration.

joernu76:
> When a category is changed, other users have the old entry in theire opened "Manage Users" - "Clone operation" list selection. We maybe should update this too. On the one who does the category change the window of "Manage Users" gets closed.

My Manage User window does not get closed. The behaviour should be the same as with a changed description. With all the category filtering etc. there may be a trigger missing that changes the visible operations. I think that is fine, currently, as this feature will be used seldomly but offers the option to correct for e.g. typo and will sort itself out eventually.

Should be fixed eventually, obviously.

ReimarBauer:
On archive we should not allow further changes, by someone else having the operation open

archive_should_not_allwow_changes

I think in this case we should close the views on other users or go to the local flightpath

ReimarBauer (approved):
This is a step further. We can improve always a few more parts. But it is also a lot changes already.

PR #1785, opened 2023-05-16
Fix bug in login into MSColab after failed authentification

### LINKED ISSUE #1783: MSCOLAB login throws exception after one fails authentification attempt
Peek 2023-05-16 19-16

joernu76:
Fix #1783


ReimarBauer (approved):
verified

thx

PR #1792, opened 2023-05-17
MSColab authenticate/authorize restructuring

joernu76:
This separates "authentification" (ensuring allowed access to server) from authorization (selecting permissions on server).

Authentification happens already upon connect. Failing authentification afterwards can be treated as error causing a disconnect.

login is then authorization.

This seems to simplify things.

Storage of password, user, etc. seems to work, but is not final, yet.

On the server side, the change simply uses a new status that requires authentification for testing.


joernu76:
DO NOT MERGE!
For reasons, stable has been merged into this. Requires rebasing etc. before it can be merged.
Remove this, when it has been tidied up.

ReimarBauer:
We may be can also remove the input field when the server does not provide auth on connect.

```
tried:

1. window

(mssdev) MSS$ export PYTHONPATH=`pwd` 

(mssdev) MSS$ python mslib/mscolab/mscolab.py db --reset # clear old scheme

(mssdev) MSS$ python mslib/mscolab/mscolab.py db --init # apply scheme to db

(mssdev) MSS$ python mslib/mscolab/mscolab.py start  # server start
# this uses builtin service configuration

2. window

(mssdev) MSS$ export PYTHONPATH=`pwd` 

(mssdev) MSS$ python mslib/msui/msui.py

Connect:

url of local server: http://localhost:8083

no auth user
no auth password
```

tried_a_dummy_email


It has not asked me if I want to update my msui_settings.jscon, it just did.

ReimarBauer:
when I disconnected, see status, connect is not possible, deactivated button.

disconnect_connect

mslib/msui/ui/ui_mscolab_connect_dialog.ui
```diff
@@ -303,28 +304,27 @@
...
           <widget class="QLabel" name="httpPasswordLabel">
            <property name="text">
             <string>Password:</string>
            </property>
           </widget>
          </item>
-         <item row="0" column="1">
-          <widget class="QLineEdit" name="httpUsernameLe">
-           <property name="placeholderText">
-            <string>Server Auth Username</string>
```
ReimarBauer, line 316: I would prefer to have an auth user name too. When someone has it in his json file we may not need to show it on the ui. But on serverside we should have it and it is not the same word.

mslib/mscolab/server.py
```diff
@@ -209,13 +216,19 @@ def home():
 def hello():
     return "Mscolab server"
 
+@APP.route("/status_auth")
```
ReimarBauer: we can do this with the origin, don't need a second one.

joernu76: In this way, older clients can still use the server. I think this is a small enough thing to retain backwards-compatibility?

mslib/msui/ui/ui_mscolab_connect_dialog.ui
joernu76: There is a username on the server side, but on the client it is fixed to "mscolab".
For authentification, we do not need a username. 
Why needlessly complicate the user interface?

I simply added a new "mscolab" user with the same password hash to the mss-stage config file.

joernu76:
> We may be can also remove the input field when the server does not provide auth on connect.
> 
> ```
> tried:
> 
> 1. window
> 
> (mssdev) MSS$ export PYTHONPATH=`pwd` 
> 
> (mssdev) MSS$ python mslib/mscolab/mscolab.py db --reset # clear old scheme
> 
> (mssdev) MSS$ python mslib/mscolab/mscolab.py db --init # apply scheme to db
> 
> (mssdev) MSS$ python mslib/mscolab/mscolab.py start  # server start
> # this uses builtin service configuration
> 
> 2. window
> 
> (mssdev) MSS$ export PYTHONPATH=`pwd` 
> 
> (mssdev) MSS$ python mslib/msui/msui.py
> 
> Connect:
> 
> url of local server: http://localhost:8083
> 
> no auth user
> no auth password
> ```
> 
> tried_a_dummy_email
> 
> It has not asked me if I want to update my msui_settings.jscon, it just did.

I am not sure how we can do this without probing the server. This could be done by probing it on every change to the URL field, but I dislike the traffic generated by this.
And I do not want to add a click to the login process (i.e. first connect without auth, then connect with auth).

joernu76:
Will have a look into the Exception and the updating of credentials.

joernu76:
Thinking about that, we could also, maybe, forego the authentification, if we can let this tap into the user database, which should be simple with the current python based approach. This would, for security reasons, not allow new users to get a login. I.e. that step would require administrator intervention, which is annoying. 

So, maybe we need the authentification *only* for the add user stuff? So no authentification anywhere, but for the add user dialogue?
And there, a username is also superfluous.

mslib/msui/ui/ui_mscolab_connect_dialog.ui
ReimarBauer: ok, on client site only a string is usally meant as API key, we could also think on this approach. On server site I would prefer more than one than too.  Currently we would have only for msui, but when there is another client/service I whish to give to that another key.

Compared to BBB there I have to use the same API key for the wiki, nc, greenlight, RC, MM. I dislike this.


By using the HTTPBasicAuth  (https://flask-httpauth.readthedocs.io/en/latest/) as we do (see also wms.py) we can shift the definitions to the web server.  This is also a good argument to keep the behaviour of HTTPBasicAuth.
E.g. if we for example move the socketio Server to a different location path, we could protect the whole location path by the webservers auth definition.  https://github.com/miguelgrinberg/Flask-SocketIO/issues/302

mslib/mscolab/server.py
ReimarBauer: we have new client -> old server, which also does not work. 

I also thought on deprecating the /status check at all. When someone wants to remove the name, currently the service is not working for clients.

ReimarBauer:
> Thinking about that, we could also, maybe, forego the authentification, if we can let this tap into the user database, which should be simple with the current python based approach. This would, for security reasons, not allow new users to get a login. I.e. that step would require administrator intervention, which is annoying.
> 
> So, maybe we need the authentification _only_ for the add user stuff? So no authentification anywhere, but for the add user dialogue? And there, a username is also superfluous.


We are using http_auth inside the python script. When e.g. the service is moved to a different location then someone might configure it similar to the wms server in the web server setup. This means the python auth is disabled and all locations are protected by the webserver definition.  This has a username and a password.

It is a different thing when you want to have always the same name, but I don't feel this a good idea.

joernu76:
I have some bug fixes, which I cannot currently push due to ports being blocked in my wifi.

I find the credential storing very confusing. We use, seemingly, the JSON config file, the keyring, *and* qsettings for storing stuff?

Wouldn't it make sense to align that with WMS server access?
So...
a) have a fixed user name (mscolab) for authentification (modifieable via config for whatever reasons) with the server and store password in keyring per server.
b) have a dict in settings with one username per mscolab server and store password in keyring for authorization.

joernu76:
new client -> old server is less of a  problem as the other way round, as there is, effectively, only one MSCOlab server out there...

joernu76:
This implements most of my proposals from above.
Some beautifying and tests are missing, but I am pretty happy from a usability perspective.

joernu76:
I had a look at testing the mss_auth stuff. As I understand this, this'd require the placement of files into the PYTHONPATH. It is unfortunate that this fully removes the default configuration, as one'd like to change only one parameter and add the auth function (~6 lines of code).

1) Is there a reason why one does not enhance the default configuration by the file?
2) In WMS the config file uses variable=value statements, which are simulated as class in the settings. mscolab requires a class as configuration, which seems needlessly complicated. Is there a particular reason for this?

mslib/mscolab/server.py
ReimarBauer: this is not a good solution, the next one will add a third option.  And when I want to harden the server I don't want to tell something there. And someone not having this new urlpath may show that some modules are not on a recent version.

So people have to update and we should force this.

mslib/mscolab/server.py
```diff
@@ -210,6 +209,12 @@ def hello():
     return "Mscolab server"
 
 
+@APP.route("/status_auth")
+@conditional_decorator(auth.login_required, mscolab_settings.__dict__.get('enable_basic_http_authentication', False))
+def hello2():
```
ReimarBauer: see above

mslib/msui/mscolab.py
```diff
@@ -209,8 +214,15 @@ def enable_login_btn(self):
...
-            r = requests.get(url_join(url, 'status'))
-            if r.text == "Mscolab server":
+            auth = "mscolab", self.httpPasswordLe.text()
+            s = requests.Session()
+            s.auth = auth
+            s.headers.update({'x-test': 'true'})
+            r = s.get(url_join(url, 'status_auth'), timeout=(2, 10))
+            if r.status_code == 401:
+                self.set_status("Error", 'Server authentication were incorrect.')
+            elif r.text == "Mscolab server":
```
ReimarBauer: we should only have a status code here, 200 would be good.

mslib/mscolab/server.py
ReimarBauer: don't add another hello2

mslib/msui/mscolab.py
```diff
@@ -222,23 +234,31 @@ def connect_handler(self):
...
                 self.mscolab_server_url = url
-                # delete mscolab http_auth settings for the url
-                if self.mscolab_server_url in self.settings["auth"].keys():
-                    del self.settings["auth"][self.mscolab_server_url]
-
-                if self.mscolab_server_url not in self.settings["server_settings"].keys():
-                    self.settings["server_settings"].update({self.mscolab_server_url: {}})
-                save_settings_qsettings('mscolab', self.settings)
+                self.auth = auth
+                save_password_to_keyring("MSCOLAB_AUTH_" + url, auth[0], auth[1])
```
ReimarBauer, line 238: :)

mslib/msui/mscolab.py
```diff
@@ -317,77 +323,30 @@ def login_handler(self):
 
         if r.text == "False":
             # show status indicating about wrong credentials
-            self.set_status("Error", 'Oh no, you need to add a user account or '
+            self.set_status("Error", 'Wrong username or password. '
```
ReimarBauer: we don't want to give feedback that a user exists, this could otherwise be used to scan for valid user names.

And not having an account is also a valid possibility.

mslib/msui/mscolab.py
```diff
@@ -317,77 +323,30 @@ def login_handler(self):
...
-
         try:
-            save_password_to_keyring(service_name="MSCOLAB", username=emailid, password=password)
+            save_password_to_keyring(service_name=self.mscolab_server_url, username=emailid, password=password)
         except (NoKeyringError, PasswordSetError, InitError) as ex:
             logging.warning("Can't use Keyring on your system:  %s" % ex)
-        exiting_mscolab_mailid = config_loader(dataset="MSCOLAB_mailid")
-        if exiting_mscolab_mailid != emailid:
+        http_auth = config_loader(dataset="MSS_auth")
+        if http_auth.get(self.mscolab_server_url) != emailid:
```
ReimarBauer: http_auth is not based on an emailid, the users name can be any name.  I think that was destinguished before in auth_username and auth_password, or?

at least the docu has a hint to that https://mss.readthedocs.io/en/stable/usage.html#mscolab-login-and-www-authentication

mslib/msui/mscolab.py
joernu76: This does not give any information about existence of a user. That would be, indeed, bad.
Username, password, or both are "wrong" in the sense that they do not belong to a registered, valid user.

The previous message did not take mistyping into account. And I did not like the "Oh no" part.

What do you propose?

mslib/msui/mscolab.py
joernu76: I think I may be hijacking the http-auth username/password stuff from the WMS stuff for the actual login data of MSColab. Which makes sense IMO, conceptually...?

One could add another config variable (e.g. MSColab_auth) and maybe should rename the variables, which I copy pasted from somewhere?

mslib/msui/mscolab.py
```diff
@@ -317,77 +323,30 @@ def login_handler(self):
...
-                    self.mscolab.after_login(emailid, self.mscolab_server_url, r)
-                else:
-                    self.stackedWidget.setCurrentWidget(self.loginPage)
-                    url_recover_password = f'{self.mscolab_server_url}/reset_request'
-                    self.set_status("Error", 'Oh no, you need to add a user account or '
-                                    f'<a href="{url_recover_password}">Recover Your Password</a>')
-            else:
-                self.set_status("Error", 'Oh no, server authentication were incorrect.')
-                self.stackedWidget.setCurrentWidget(self.loginPage)
+                http_auth[self.mscolab_server_url] = emailid
```
ReimarBauer: emailid sounds not correct

tests/_test_msui/test_mscolab.py
```diff
@@ -51,6 +52,10 @@
 
 class Test_Mscolab_connect_window():
     def setup_method(self):
+        keyring.get_keyring().reset()
+        import mslib.mscolab.conf
+        mslib.mscolab.conf.mscolab_settings.enable_basic_http_authentication = True
+
```
ReimarBauer: we maybe should move these tests, which needs a compilation to a special tests_run_once on same level as tests/ and have a pytest script action which collects the tests below and runs them one after the other in an individual pytest process.

mslib/msui/mscolab.py
joernu76: more likely, http_auth is mislabeled.

tests/_test_msui/test_mscolab.py
```diff
@@ -148,7 +170,7 @@ def test_logout(self):
     def test_add_user(self, mockmessage):
         self._connect_to_mscolab()
         self._create_user("something", "something@something.org", "something")
-        assert config_loader(dataset="MSCOLAB_mailid") == "something@something.org"
+        assert config_loader(dataset="MSS_auth").get(self.url) == "something@something.org"
```
ReimarBauer, line 167: hmm, that will need a documentation.

tests/_test_msui/test_mscolab.py
joernu76: the basic_http_auth has no effect and should have been removed.

tests/_test_msui/test_mscolab.py
joernu76: do we have the keyring stuff documented? Will have a look.

This collects all URL/usernames in the MSS_auth config variable and passwords in the keyring.

joernu76: One could also, likely?, store username *and* password in the keyring and forego this entirely? URLs are otherwise stored in the default_whatever lists.

mslib/msui/mscolab.py
ReimarBauer: there are more "Oh no", may be change them all.

Because in the past a json file with placeholder was exchanged, someone can get to a server but don't have of course this as an account, this data was filled in too. So it was the idea to give also the hint that a user has to become created first.


(did not practical verified this in the new version)

in the past when I exchanged only the default server, it showed me credentials filled in and I can't login.

We can only tell try to create an account or try to recover your password. This is not comfortable for the user having this problem but for all other existing users it is.

ReimarBauer:
> I had a look at testing the mss_auth stuff. As I understand this, this'd require the placement of files into the PYTHONPATH. It is unfortunate that this fully removes the default configuration, as one'd like to change only one parameter and add the auth function (~6 lines of code).
> 
> 1. Is there a reason why one does not enhance the default configuration by the file?
> 2. In WMS the config file uses variable=value statements, which are simulated as class in the settings. mscolab requires a class as configuration, which seems needlessly complicated. Is there a particular reason for this?

We should collect tests which needs a reload(module) in a seperate test dir and run them by individual pytests. 
I've no better idea yet.

tests/_test_msui/test_mscolab.py
ReimarBauer: the keyring needs some get attribute to find informations, you get usually by get username the password

ReimarBauer:
> new client -> old server is less of a problem as the other way round, as there is, effectively, only one MSCOlab server out there...

Well we know then all users, they have to update the client and we update the server. There should be only one of these urls and we should compare by the status code.

When you absolutly want this, we can patch our server but don't have it in the official code.

joernu76:
> > I had a look at testing the mss_auth stuff. As I understand this, this'd require the placement of files into the PYTHONPATH. It is unfortunate that this fully removes the default configuration, as one'd like to change only one parameter and add the auth function (~6 lines of code).
> > 
> > 1. Is there a reason why one does not enhance the default configuration by the file?
> > 2. In WMS the config file uses variable=value statements, which are simulated as class in the settings. mscolab requires a class as configuration, which seems needlessly complicated. Is there a particular reason for this?
> 
> We should collect tests which needs a reload(module) in a seperate test dir and run them by individual pytests. I've no better idea yet.

In this particular case, the mscolab server gets started per testcase, so one *can* change its configuration to test proper client behaviour. However, I mocked the reply so it is tested from the client perspective already now.

joernu76:
> > new client -> old server is less of a problem as the other way round, as there is, effectively, only one MSCOlab server out there...
> 
> Well we know then all users, they have to update the client and we update the server. There should be only one of these urls and we should compare by the status code.
> 
> When you absolutly want this, we can patch our server but don't have it in the official code.

The only clean, non-breaking way would require two targets for the foreseeable time, also in the stable branch. I thus removed the second target, breaking old clients.

tests/_test_msui/test_mscolab.py
joernu76: I added documentation. I am not sure how to "migrate" the changes forward. I would defer that change to another PR.

mslib/msui/mscolab.py
joernu76: renamed to fix problem.

mslib/msui/mscolab.py
joernu76: I removed it. I replaced it by "invalid credentials" and listing all options what can be done to fix it. This does not provide any information exploitable for hacking IMO.

mslib/msui/mscolab.py
joernu76: done.

ReimarBauer:
> I have some bug fixes, which I cannot currently push due to ports being blocked in my wifi.
> 
> I find the credential storing very confusing. We use, seemingly, the JSON config file, the keyring, _and_ qsettings for storing stuff?
> 
> Wouldn't it make sense to align that with WMS server access? So... a) have a fixed user name (mscolab) for authentification (modifieable via config for whatever reasons) with the server and store password in keyring per server. b) have a dict in settings with one username per mscolab server and store password in keyring for authorization.

we should not store plaintext passwords in qsettings. Maybe I missed to remove it from some places?

ReimarBauer:
To look on this may become a follow up task.
connect_on_new_adress

I have no idea how I got it to store the -test url into the pull down, when I exchange this and cycle login, it starts with the given url. I have always to enter all data.

ReimarBauer:
hmm, the behaviour looks like that we need a migration script from 8.x to 9.x or that something is filling into a field where it shouldn't.

The "MSCOLAB_mailid" is now obsolete, or? this got moved to the MSS_auth and the server. Removing it will enable to have different accounts on different servers.

strange_account

When I add my account email instead I also can't login. A wrong password let not show the "reset link" it goes back to connect.

ok, guess I understand, it is able to get also on wrong data to the next page, but there it should not fill in that data.

when the first click gives not a result, can we stay on that frame?  Showing the second and then returning to the first one is a bit confusing. Because it tells in the status  "Successfully connected ..."which is wrong.

it works perfect when you do nothing wrong. But I guess I am not alone whith clicking too fast.

mslib/msui/mscolab.py
```diff
@@ -209,9 +214,16 @@ def enable_login_btn(self):
...
-            r = requests.get(url_join(url, 'status'))
-            if r.text == "Mscolab server":
-                self.set_status("Success", "Successfully connected to MSColab Server")
+            auth = "mscolab", self.httpPasswordLe.text()
+            s = requests.Session()
+            s.auth = auth
+            s.headers.update({'x-test': 'true'})
+            r = s.get(url_join(url, 'status'), timeout=(2, 10))
+            if r.status_code == 401:
+                self.set_status("Error", 'Server authentication data were incorrect.')
```
ReimarBauer, line 223: seems it does not come to this when there was no auth data supplied but needed

ReimarBauer: seems the server is not on same patchlevel.

ReimarBauer:
> hmm, the behaviour looks like that we need a migration script from 8.x to 9.x or that something is filling into a field where it shouldn't.
> 
> The "MSCOLAB_mailid" is now obsolete, or? this got moved to the MSS_auth and the server. Removing it will enable to have different accounts on different servers.
> 
> strange_account
> 
> When I add my account email instead I also can't login. A wrong password let not show the "reset link" it goes back to connect.
> 
> ok, guess I understand, it is able to get also on wrong data to the next page, but there it should not fill in that data.
> 
> when the first click gives not a result, can we stay on that frame? Showing the second and then returning to the first one is a bit confusing. Because it tells in the status "Successfully connected ..."which is wrong.
> 
> it works perfect when you do nothing wrong. But I guess I am not alone whith clicking too fast.

seems the server is not on same patchlevel

joernu76:
> > hmm, the behaviour looks like that we need a migration script from 8.x to 9.x or that something is filling into a field where it shouldn't.
> > The "MSCOLAB_mailid" is now obsolete, or? this got moved to the MSS_auth and the server. Removing it will enable to have different accounts on different servers.
> > strange_account
> >       
> >         [
> >           
> >             strange_account
> >           
> >         ](https://user-images.githubusercontent.com/278728/240656662-7780ccac-d13d-4767-a32d-09083d988ac0.gif)
> >         
> >         
> >           
> >             
> >               
> >             
> >             
> >               
> >               
> >             
> >           
> >           [
> >             
> >               
> >             
> >           ](https://user-images.githubusercontent.com/278728/240656662-7780ccac-d13d-4767-a32d-09083d988ac0.gif)
> >         
> >       
> > When I add my account email instead I also can't login. A wrong password let not show the "reset link" it goes back to connect.
> > ok, guess I understand, it is able to get also on wrong data to the next page, but there it should not fill in that data.
> > when the first click gives not a result, can we stay on that frame? Showing the second and then returning to the first one is a bit confusing. Because it tells in the status "Successfully connected ..."which is wrong.
> > it works perfect when you do nothing wrong. But I guess I am not alone whith clicking too fast.
> 
> seems the server is not on same patchlevel

Nope, it isn't.

joernu76:
> > > hmm, the behaviour looks like that we need a migration script from 8.x to 9.x or that something is filling into a field where it shouldn't.
> > > The "MSCOLAB_mailid" is now obsolete, or? this got moved to the MSS_auth and the server. Removing it will enable to have different accounts on different servers.
> > > strange_account
> > >       
> > >         [
> > >           
> > >             strange_account
> > >           
> > >         ](https://user-images.githubusercontent.com/278728/240656662-7780ccac-d13d-4767-a32d-09083d988ac0.gif)
> > >         
> > >         
> > >           
> > >             
> > >               
> > >             
> > >             
> > >               
> > >               
> > >             
> > >           
> > >           [
> > >             
> > >               
> > >             
> > >           ](https://user-images.githubusercontent.com/278728/240656662-7780ccac-d13d-4767-a32d-09083d988ac0.gif)
> > >         
> > >       
> > > ```
> > >     [
> > >       
> > >         strange_account
> > >       
> > >     ](https://user-images.githubusercontent.com/278728/240656662-7780ccac-d13d-4767-a32d-09083d988ac0.gif)
> > >     
> > >     
> > >       
> > >         
> > >           
> > >         
> > >         
> > >           
> > >           
> > >         
> > >       
> > >       [
> > >         
> > >           
> > >         
> > >       ](https://user-images.githubusercontent.com/278728/240656662-7780ccac-d13d-4767-a32d-09083d988ac0.gif)
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > When I add my account email instead I also can't login. A wrong password let not show the "reset link" it goes back to connect.
> > > ok, guess I understand, it is able to get also on wrong data to the next page, but there it should not fill in that data.
> > > when the first click gives not a result, can we stay on that frame? Showing the second and then returning to the first one is a bit confusing. Because it tells in the status "Successfully connected ..."which is wrong.
> > > it works perfect when you do nothing wrong. But I guess I am not alone whith clicking too fast.
> > 
> > 
> > seems the server is not on same patchlevel
> 
> Nope, it isn't.

And already you run into one of the issues with *not* providing a new "status_auth" target. 
I am not sure how to provide good error messages for inconsistent server/client. With the new, additional target, we'd not have this issue and could remove the status one silently in the next-next release.

ReimarBauer:
> > > > hmm, the behaviour looks like that we need a migration script from 8.x to 9.x or that something is filling into a field where it shouldn't.
> > > > The "MSCOLAB_mailid" is now obsolete, or? this got moved to the MSS_auth and the server. Removing it will enable to have different accounts on different servers.
> > > > strange_account
> > > >       
> > > >         [
> > > >           
> > > >         
> > > >             strange_account
> > > >           ](https://user-images.githubusercontent.com/278728/240656662-7780ccac-d13d-4767-a32d-09083d988ac0.gif)
> > > >         
> > > >         
> > > >           
> > > >             
> > > >               
> > > >             
> > > >             
> > > >               
> > > >               
> > > >             
> > > >           
> > > >           [
> > > >             
> > > >               
> > > >             
> > > >           ](https://user-images.githubusercontent.com/278728/240656662-7780ccac-d13d-4767-a32d-09083d988ac0.gif)
> > > >         
> > > >       
> > > > ```
> > > >     [
> > > >       
> > > >         strange_account
> > > >       
> > > >     ](https://user-images.githubusercontent.com/278728/240656662-7780ccac-d13d-4767-a32d-09083d988ac0.gif)
> > > >     
> > > >     
> > > >       
> > > >         
> > > >           
> > > >         
> > > >         
> > > >           
> > > >           
> > > >         
> > > >       
> > > >       [
> > > >         
> > > >           
> > > >         
> > > >       ](https://user-images.githubusercontent.com/278728/240656662-7780ccac-d13d-4767-a32d-09083d988ac0.gif)
> > > > ```
> > > > 
> > > > 
> > > >     
> > > >       
> > > >     
> > > > 
> > > >       
> > > >     
> > > > 
> > > >     
> > > >   
> > > > ```
> > > >     [
> > > >       
> > > >         strange_account
> > > >       
> > > >     ](https://user-images.githubusercontent.com/278728/240656662-7780ccac-d13d-4767-a32d-09083d988ac0.gif)
> > > >     
> > > >     
> > > >       
> > > >         
> > > >           
> > > >         
> > > >         
> > > >           
> > > >           
> > > >         
> > > >       
> > > >       [
> > > >         
> > > >           
> > > >         
> > > >       ](https://user-images.githubusercontent.com/278728/240656662-7780ccac-d13d-4767-a32d-09083d988ac0.gif)
> > > > ```
> > > > 
> > > > 
> > > >     
> > > >       
> > > >     
> > > > 
> > > >       
> > > >     
> > > > 
> > > >     
> > > >   
> > > > When I add my account email instead I also can't login. A wrong password let not show the "reset link" it goes back to connect.
> > > > ok, guess I understand, it is able to get also on wrong data to the next page, but there it should not fill in that data.
> > > > when the first click gives not a result, can we stay on that frame? Showing the second and then returning to the first one is a bit confusing. Because it tells in the status "Successfully connected ..."which is wrong.
> > > > it works perfect when you do nothing wrong. But I guess I am not alone whith clicking too fast.
> > > 
> > > 
> > > seems the server is not on same patchlevel
> > 
> > 
> > Nope, it isn't.
> 
> And already you run into one of the issues with _not_ providing a new "status_auth" target. I am not sure how to provide good error messages for inconsistent server/client. With the new, additional target, we'd not have this issue and could remove the status one silently in the next-next release.

The problem is that my config is not updated by a migration script. When my config matches, I don't have this problem.

ReimarBauer (approved):
verified to work with a server on same patch level. 
Also started  a migration script as follow up and removing MSCOLAB_mailid

PR #1794, opened 2023-05-24
fill mswms_settings with default values.

joernu76:


joernu76:
@ReimarBauer Do you have an idea why the MSCOlab tests crash and burn? I have no problems on my laptop and do not understand anything about this. My changes should not change the MSColab configuration in the least. They do change the config file format to be consistent with MSWMS config files, but no config file is read in AFAIK.

joernu76:
Found the config file in conftest.py. argh.

mslib/mswms/wms.py
```diff
@@ -79,33 +79,40 @@
...
+    enable_basic_http_authentication = False
+    __file__ = None
+
+
+mswms_settings = default_mswms_settings()
+
 try:
-    import mswms_settings
+    import mswms_settings as user_settings
+    mswms_settings.__dict__.update(user_settings.__dict__)
```
ReimarBauer, line 113: we use the mswms_settings also to patch some parts on our server, how do we do this in the new approach?

ReimarBauer (changes requested):
Currently just questions.

When we don't need to have python code for e.g. patching, introduction further classes in those settings files we can maybe convert their content into yaml, toml, json and have only a textfile for the server configuration. 

When we need these files to mock/patch some python code how does this now been done?

mslib/mswms/wms.py
joernu76: In either case, due to the nature of import, the patching must be done by "importing" the file first and patching it before wms.py is imported. Then the import here will "simply" be refered to the previously imported&patched file.

joernu76:
> Currently just questions.
> 
> When we don't need to have python code for e.g. patching, introduction further classes in those settings files we can maybe convert their content into yaml, toml, json and have only a textfile for the server configuration.
> 
> When we need these files to mock/patch some python code how does this now been done?

We could move a lot of stuff from those config files to, e.g., JSON. Still, I would not remove this as it is veery powerful. Many complex applications embed python or lua for pretty much this purpose.

joernu76:
This PR should do 2 things:
1) make mswms & mscolab configuration consistent
2) allow in both cases to provide default parameters to be overwritten if user-provided.

1) allows us to stay sane
2) allows us to assume existance of certain parameters and provide a default value in *one* place

ReimarBauer (approved):
thx

PR #1795, opened 2023-05-25
Fix #1649

### LINKED ISSUE #1649: satellite prediction writes outside the graph

joernu76:


ReimarBauer:
@joernu76 great :) do you mind to add this to stable, I was not aware that this can be such simple

joernu76:
didn't know either before looking :-)

ReimarBauer (approved):
thx

PR #1797, opened 2023-05-25
make starting up slightly faster

joernu76:


joernu76:
This patch prevents mswms&mscolab being imported, which saves about 2s and anoth half second for QtTest (on my system).
The worst culprit is metpy, which import basically *everything* (full scipy, dask, xarray, etc.)
I will see if this can be prevented somehow.

ReimarBauer:
@joernu76 I don't have a speed difference. I like that the server parts utils are not loaded, but somehow I wish to understand more why it is so different.

mslib/msui/qt5/ui_linearview_window.py
```diff
@@ -9,6 +9,7 @@
 
 from PyQt5 import QtCore, QtGui, QtWidgets
 
+import mslib.msui.mpl_qtwidget
```
ReimarBauer: this might be gone when recreated by pyuic5, or?

ReimarBauer:
for the ui_*.py files, this likly don't get that way created on pyuic5. Will that break in that case?

mslib/mscolab/models.py
```diff
@@ -157,13 +157,6 @@ def __repr__(self):
                f'last_used: {self.last_used}> '
 
 
-class MessageType(enum.IntEnum):
```
ReimarBauer, line 160: that is not so ideal to seperate this from models.py. These classes are used to setup the database.
The migration scripts for the database update are also done on changes of these classes.

That might be better to have a `mscolab/models/__init__` and other modules there, when it needs been splitted.
Then it is better to acknowledge what these definitions are.

mslib/msui/msui.py
```diff
@@ -1056,6 +1056,7 @@ def closeEvent(self, event):
             if self.config_editor is not None:
                 self.config_editor.restart_on_save = False
                 self.config_editor.close()
+                from PyQt5 import QtTest
```
ReimarBauer: we need to replace QtTest by something else. This loads and needs also a framework installed which usually is only needed for testing Qt.
slide 13
https://www.slideshare.net/ICSinc/qt-test-framework

maybe by https://doc.qt.io/qtforpython-5/PySide2/QtCore/QWaitCondition.html

ReimarBauer:
have you looked on this by 

`python -m cProfile  -s time mslib/msui/msui.py  > profile.txt`

starting and closing by ctrl+q and a mouseclick shows for me 
` 6127197 function calls (5900248 primitive calls) in 6.584 seconds` with your branch

 and with develop it is
 ` 6014640 function calls (5787747 primitive calls) in 6.561 seconds`

ReimarBauer (changes requested):
see comments

mslib/msui/qt5/ui_linearview_window.py
joernu76: correct.
Here, I moved the from x import y statement from below to this above to avoid  problems w.r.t. in circular imports.

mslib/mscolab/models.py
joernu76: I do not understand?
The only change is that it is in its own file.

mslib/msui/msui.py
joernu76: Probably, this could be replaced.

joernu76:
> have you looked on this by
> 
> `python -m cProfile -s time mslib/msui/msui.py > profile.txt`
> 
> starting and closing by ctrl+q and a mouseclick shows for me ` 6127197 function calls (5900248 primitive calls) in 6.584 seconds` with your branch
> 
> and with develop it is ` 6014640 function calls (5787747 primitive calls) in 6.561 seconds`

I tried different things and this reduced it from 13s to 10s. Mostly because importing mscolab&mswms is not happening.
However, the variance was rather large.

The biggest delay is caused by metpy, either way.

ReimarBauer:
> > have you looked on this by
> > `python -m cProfile -s time mslib/msui/msui.py > profile.txt`
> > starting and closing by ctrl+q and a mouseclick shows for me ` 6127197 function calls (5900248 primitive calls) in 6.584 seconds` with your branch
> > and with develop it is ` 6014640 function calls (5787747 primitive calls) in 6.561 seconds`
> 
> I tried different things and this reduced it from 13s to 10s. Mostly because importing mscolab&mswms is not happening. However, the variance was rather large.
> 
> The biggest delay is caused by metpy, either way.

Which version of metpy do you have?

ReimarBauer:
sorted by tot time, on my system mostly builtin methods are the expensivest

time

ReimarBauer:
solving the importing of the server stuff helps to create a splitted package on conda-forge a lot. The servers won't need the qt libs, so they can get rather small.

ReimarBauer:
snakeviz has a nice visualisation, the biggest blob is the problem

snakeviz

looked into the details of metpy
looked_into_metpy

ReimarBauer:
```
calling
python -m cProfile  -o msui.prof  mslib/msui/msui.py
after it launches, ctrl+q and Y
then
snakeviz msui.prof
```

mslib/mscolab/models.py
ReimarBauer: when one looks on a grouping folder it is clear everything there belongs to models

now we have a models module and one other models module which is not that easy recognized as before, where all was in one file

joernu76:
Here, my total execution times on battery with a sys.exit in fromt of mainwindow.show():

```
head -n1 pr*txt
==> profile_d1.txt <==
         5397593 function calls (5163936 primitive calls) in 19.189 seconds

==> profile_d2.txt <==
         5398118 function calls (5164451 primitive calls) in 18.635 seconds

==> profile_d3.txt <==
         5397797 function calls (5164139 primitive calls) in 18.942 seconds

==> profile_i1.txt <==
         4995917 function calls (4771917 primitive calls) in 17.338 seconds

==> profile_i2.txt <==
         4995634 function calls (4771642 primitive calls) in 17.246 seconds

==> profile_i3.txt <==
         4995669 function calls (4771678 primitive calls) in 17.208 seconds

```

The _d are the develop branch. The new branch uses consistently 500000 calls and about 1.5-2s less.

It must use less calls, as develop imports fully mscolab and mswms.

joernu76:
Yes, i identified the xarray metpy DataAccessor classes as the biggest problem as well. There is not much we can do about this unless we want to implement the thermodynamics again ourselves... 
It could be that one can restrict metpy use to only the server. But that would be larger changes, I guess.

This change is barely noticeable when my laptop is plugged in, BTW. Still I dislike needlessly importing the whole server packages.

joernu76:
> > > have you looked on this by
> > > `python -m cProfile -s time mslib/msui/msui.py > profile.txt`
> > > starting and closing by ctrl+q and a mouseclick shows for me ` 6127197 function calls (5900248 primitive calls) in 6.584 seconds` with your branch
> > > and with develop it is ` 6014640 function calls (5787747 primitive calls) in 6.561 seconds`
> > 
> > 
> > I tried different things and this reduced it from 13s to 10s. Mostly because importing mscolab&mswms is not happening. However, the variance was rather large.
> > The biggest delay is caused by metpy, either way.
> 
> Which version of metpy do you have?

1.4.0

ReimarBauer:
in battery mode for develop branch, with this sys.exit()

`6125615 function calls (5899792 primitive calls) in 3.660 seconds`

and with the issue1796a branch, with this sys.exit()

`6125570 function calls (5899748 primitive calls) in 3.585 seconds`

I'm on Python 3.10.8 and metpy-1.4.1 and xarray-2023.5.0

ReimarBauer:
let's try with the docker container and the applied patch file `patch -p1 <`
[pyv.patch.txt](https://github.com/Open-MSS/MSS/files/11568614/pyv.patch.txt)

```
docker pull openmss/testing-develop
docker run -it --mount src=`pwd`,target=`pwd`,type=bind -e MSSDIR=`pwd` openmss/testing-develop

conda activate mss-develop-env
cd $MSSDIR
export PYTHONPATH=`pwd`  

python -m cProfile  -s time mslib/msui/msui.py  > profile.txt
```

with the issue1796a branch in battery mode (and wlan), without any graphical output (patch) on my screen, it takes

` 5735300 function calls (5518579 primitive calls) in 3.746 seconds`

ReimarBauer:
got some idea to look on my powermanagment.

when I switch from my default to a max CPU speed between 0.4 to 0.8GHz I get 
`580278 function calls (5364065 primitive calls) in 24.694 seconds`

thermolib needs ~25%

thermolib

When I use the "default" settings which was used for the other results I  have still >8h battery power. So that problem is also related to the notebook used. 


I wonder also a bit about ui_sideview_window.py.  This needs 16.5 seconds.  The ui_topview_windows.py is nearly not to measure.

joernu76:
I do not understand how you number of calls does not drop with the PR.

Please check: When starting msui.py with --debug, I see these message on develop:
```
```
indicating the importing of lots of mscolab&mswms. With my branch, these vanish, reducing the number of executed stuff.

mslib/msui/ui/ui_sideview_window.ui
```diff
@@ -107,7 +107,7 @@
   <customwidget>
    <class>MplSideViewWidget</class>
    <extends>QWidget</extends>
-   <header>mslib.msui.mpl_qtwidget</header>
+   <header location="global">mslib.msui.mpl_qtwidget</header>
```
ReimarBauer: I assume this needs reverted too, or?

ReimarBauer:
used the pyv.patch.txt on each
- 1796a:  5580400 function calls (5364188 primitive calls) in 4.277 seconds
- develop: 6111617 function calls (5887188 primitive calls) in 5.364 seconds
- stable: 6128713 function calls (5902787 primitive calls) in 5.032 seconds

ReimarBauer:
updated metpy to 1.5.0

- stable: 6130496 function calls (5904555 primitive calls) in 4.027 seconds
- develop: 6113242 function calls (5888798 primitive calls) in 3.719 seconds
- remove_qt_globals: 6112535 function calls (5887279 primitive calls) in 3.665 seconds
- issue1796a: 5581308 function calls (5365077 primitive calls) in 3.292 seconds

ReimarBauer:
> I do not understand how you number of calls does not drop with the PR.
> 
> Please check: When starting msui.py with --debug, I see these message on develop:
> 
> ```
> WARNING:root:Couldn't import mscolab_settings (ImportError:'No module named 'mscolab_settings''), using dummy config.
> WARNING:root:No module named 'mswms_settings'. Can't generate gallery.
> ```
> 
> indicating the importing of lots of mscolab&mswms. With my branch, these vanish, reducing the number of executed stuff.

I guess I've not recognized due to the pyvirtualdisplay patch that I've not switched branches correctly. New tests below.

ReimarBauer (approved):
we can maybe group the models in a new folder. But hey good work!

PR #1799, opened 2023-05-25
enable tests for GSOC branches

ReimarBauer:
@joernu76 I think that is the right syntax, may have also a check here

https://docs.github.com/de/actions/using-workflows/workflow-syntax-for-github-actions

### DIFF
--- a/.github/workflows/testing-develop.yml
+++ b/.github/workflows/testing-develop.yml
@@ -4,10 +4,13 @@ on:
   push:
     branches:
     - develop
+    - 'GSOC**'
 
   pull_request:
     branches:
     - develop
+    - 'GSOC**'
+
 
 jobs:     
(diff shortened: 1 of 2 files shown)


.github/workflows/testing-develop.yml
```diff
@@ -4,10 +4,13 @@ on:
   push:
     branches:
     - develop
+    - 'GSOC**'
 
   pull_request:
     branches:
     - develop
+    - 'GSOC**'
```
joernu76: there is a branch name "develop" below...?

ReimarBauer: ah, yes, tried a new fixated solution, because branch dependent we have the env defined.

joernu76 (approved):

PR #1802, opened 2023-05-26
replace globals by locals

ReimarBauer:


mslib/utils/thermolib.py
```diff
@@ -35,7 +35,7 @@
 from metpy.xarray import preprocess_and_wrap
 import metpy.calc as mpcalc
 
-exporter = Exporter(globals())
+exporter = Exporter(locals())
```
joernu76: This doesn't make sense. The Exporter is a metpy helper class that maintains the "\_\_all\_\_" list of module symbols.
Also, is there a difference in globals/locals at this point in the code???
I do not now why we use this anyway. I suggest removing it and its decorator use below.

ReimarBauer: good idea :)

ReimarBauer (commented):
@joernu76 can you try to approve this, as initial author I can't

joernu76 (approved):

PR #1805, opened 2023-06-07
Setup SP and IDP for the SSO

nilupulmanodya:
Setting up the service provider and identity provider for the single sign-on process


ReimarBauer:
I missed to remove conditions in
https://github.com/Open-MSS/MSS/blob/develop/.github/workflows/testing_gsoc.yml#L66 and line 81, 95

I do this soon.

ReimarBauer:
For new files please use the template https://github.com/Open-MSS/MSS/blob/develop/new_file_template.py

using

```
:copyright: Copyright 2023 Nilupul Manodya 
:license: APACHE-2.0, see LICENSE for details.
```

conf_sp_idp/README.md
```diff
@@ -0,0 +1,54 @@
...
+
+The `conf_sp_idp` designed for testing the Single Sign-On (SSO) process using PySAML2. This folder contains both the Identity Provider (IdP) and Service Provider (SP) implementations.
+
+The Identity Provider was set up following the official documentation of [PySAML2](https://pysaml2.readthedocs.io/en/latest/), along with examples provided in the repository. Metadata YAML files will generate using the built-in tools of PySAML2. Actual key and certificate files can be used in when actual implementation. Please note that this project is intended for testing purposes only.
+
+## Getting started
+
+### TLS Setup
+
+** Setting Up Certificates for Local Development **
```
ReimarBauer, line 11: To get a "bold" formatted line remove the blank before/after the `**`

localbuild/meta.yaml
```diff
@@ -92,6 +92,12 @@ requirements:
     - email_validator
     - keyring
     - dbus-python
+    - flask-login
```
ReimarBauer, line 95: these are already installed, they belong to flask.

- flask-sqlalchemy
- flask-migrate (this is in our env already installed)

by alembic  (and alembic comes from flask-migrate)

- mako 


When we already install it by some other package, don't extend the list.

by the repoquery whoneeds you can verfiy "the needed new ones"

`mamba repoquery whoneeds mako`

mslib/utils/thermolib.py
```diff
@@ -27,14 +27,10 @@
...
 import numpy as np
-
-from mslib.utils.units import units, check_units
-
-from metpy.package_tools import Exporter
 from metpy.constants import Rd, g
 from metpy.xarray import preprocess_and_wrap
 
-exporter = Exporter(globals())
+from mslib.utils.units import units, check_units
```
ReimarBauer, line 33: These changes seem not to belong to your PR, can you check if these are vanish with a rebase ?

localbuild/meta.yaml
```diff
@@ -92,6 +92,12 @@ requirements:
...
+    - pysaml2
+    - cherrypy
```
ReimarBauer, line 97: is cherrypy mandatory? We do have already flask.

localbuild/meta.yaml
```diff
@@ -92,6 +92,12 @@ requirements:
...
+    - mako
+    - flask-sqlalchemy
+    - flask-migrate
```
ReimarBauer, line 100: please  check if the xmlsec package is the same as the one which is named xmlsec1 in the debian manager.

https://github.com/conda-forge/xmlsec-feedstock

When we can install it by mamba then we don't have to add it by apt to the docker container

conf_sp_idp/README.md
```diff
@@ -0,0 +1,54 @@
...
+    Make sure to insert the key along with its corresponding certificate.
+
+### Configuring the Service Provider and Identity Provider
+
+First, generate the [metadata](https://pysaml2.readthedocs.io/en/latest/howto/config.html#metadata) file for the service provider. To do that, start the Flask application and download the metadata file by following these steps:
+
+1. Navigate to the directory `MSS/conf_sp_idp/sp/app`.
+2. Start the Flask application by running `flask run`. The application will listen on port `5000`.
+3. Download the metadata file by executing the command: `curl http://localhost:5000/metadata/ -o sp.xml`.
+4. Move generated `sp.xml` to dir `conf_sp_idp/idp/`.
```
ReimarBauer, line 28: on my system it was `curl http://127.0.0.1:5000/metadata -o sp.xml`

conf_sp_idp/README.md
```diff
@@ -0,0 +1,54 @@
...
+
+After that, regenerate the idp.xml file, copy it over to the Service Provider (SP), and restart the SP Flask application:
+
+5. Go to the directory `MSS/conf_sp_idp/idp/`.
+6. Run the command `python make_metadata.py idp_conf.py > idp.xml` in the terminal to generate the `idp.xml` file.
+7. Copy the generated `idp.xml` file and replace the existing `idp.xml` file located at `MSS/conf_sp_idp/sp/idp.xml`.
+8. Restart the Flask application.
```
ReimarBauer, line 35: is 8. needed?

does it mean:

`conf_sp_idp/sp/app$ flask run`

that is a bit confusing, it gets clearer with the next section. May be just end the session here, while in the next section with more context it is better understandable.

conf_sp_idp/README.md
```diff
@@ -0,0 +1,54 @@
...
+1. Start the Service provider:
+ * Navigate to the directory `MSS/conf_sp_idp/sp/app` and run `flask run`.
+2. Start the Identity Provider:
+ * Navigate to the directory `MSS/conf_sp_idp/idp` and run `python idp.py idp_conf`.
+
+By following the provided instructions, you will be able to set up and configure both the Identity Provider and Service Provider for testing the SSO process.
+
+## Testing SSO
+
+Once you have successfully launched the server and identity provider, you can begin testing the Single Sign-On (SSO) process. To log in to the service provider through the identity provider, you can use the credentials specified in the `PASSWD` section of the `MSS/conf_sp_idp/idp/idp.py` file. Look for the relevant section in the file to find the necessary login credentials.
```
ReimarBauer, line 50: it would be good two write here, load in a browser http://127.0.0.1:5000/ and try the example login

ReimarBauer (commented):
This is a good start :)
It is good documented :) I have reproduced all steps. 


We need to develop it in a direction  that we can use the localhost idp later for tests that anything related works, that helps to track changes in our code and changes in the API from what we use.


Also when we install MSS by mamba/conda likly the pathes are not easily to access. so we need to have a settings where we can define pathes.  You can do this similiar by a pattern we used in  https://github.com/Open-MSS/MSS/blob/develop/mslib/mscolab/conf.py

We can define PYTHONPATH and overwrite by this a builtin. conftest can then be used to write a different configuration and add it to the PYTHONPATH.

Have a look on the 
https://stackoverflow.com/questions/10175812/how-to-generate-a-self-signed-ssl-certificate-using-openssl

there is a 3011 point answer I used to create the certs for localhost. The so prepared certs can be later maybe added for testing by e.g. conftest.py. 

or https://serverfault.com/questions/366372/is-it-possible-to-generate-rsa-key-without-pass-phrase with 171 points 

The first uses a passphrase which is for a test setup where we later on each run create new keys/certs not important. Maybe a combination of both, noninteractive.

ReimarBauer (changes requested):
see comments. 

some are for giving a direction to go. Not all needs to be done in that PR at once.

localbuild/meta.yaml
nilupulmanodya: I think the xmlsec package is not the same as the one which is named xmlsec1 in the Debian manager. The xmlsec package is a Python library that provides bindings for the XML Security Library, while the xmlsec1 package is a command line tool for signature and encryption transformations of XML data.

https://pypi.org/project/xmlsec/
https://packages.debian.org/stable/text/xmlsec1

ReimarBauer: when I try the xmlsec package it installs

```
+ libgcrypt      1.10.1  h166bdaf_0       conda-forge/linux-64     Cached
  + libgpg-error     1.46  h620e276_0       conda-forge/linux-64     Cached
  + libtool         2.4.7  h27087fc_0       conda-forge/linux-64     Cached
  + libxmlsec1     1.2.33  hb6668e7_0       conda-forge/linux-64     Cached
  + xmlsec     
```
the debian xmlsec1 does also install  libxmlsec1  

and when I look 
```
try:
    from saml2.sigver import get_xmlsec_binary
except ImportError:
    get_xmlsec_binary = None

if get_xmlsec_binary:
    XMLSEC_PATH = get_xmlsec_binary(["/opt/local/bin"])
else:
    XMLSEC_PATH = '/usr/bin/xmlsec1'
```

the debian package seems to be a hardcoded fallback which ignores the env. (abuses also None in the condition ...)

```
>>> XMLSEC_PATH = get_xmlsec_binary(["/opt/local/bin"])
>>> XMLSEC_PATH
'/home/mss2023test/mambaforge/envs/mssdev/bin/xmlsec1'
```

ReimarBauer: it definitly looks for the xmlsec1 and is able to use the one by xmlsec in the envs path.

ReimarBauer: As it loads this libxmlsec1, we have to look into 
- https://github.com/conda-forge/libxmlsec1-feedstock/blob/main/recipe/meta.yaml
- https://packages.debian.org/bullseye/libxmlsec1

nilupulmanodya:
I opened https://github.com/Open-MSS/MSS/pull/1809 with the correct branch out. Will work on that  with further fixes

PR #1807, opened 2023-06-07
inputs removed

### LINKED ISSUE #1806: running tests in gsoc branches blocked by conditions
conditions have go be removed
https://github.com/Open-MSS/MSS/blob/develop/.github/workflows/testing_gsoc.yml#L66 and line 81, 95

ReimarBauer:
closes #1806

### DIFF
--- a/.github/workflows/testing_gsoc.yml
+++ b/.github/workflows/testing_gsoc.yml
@@ -63,7 +63,7 @@ jobs:
         && mamba list
 
     - name: Run tests
-      if: ${{ success() && inputs.xdist == 'no' }}
+      if: ${{ success() }}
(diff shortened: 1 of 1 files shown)


joernu76:
I do not understand. Why?
This obviously changes the behavior in a meaningful way. parallel/non-parallel testing would need to be handled differently and also the coverage would now always be updated independent on which branch the pipeline was run?

ReimarBauer:
parallel/non-parallel run each after the other but only for GSOC branches.

https://github.com/Open-MSS/MSS/actions/runs/5221881087/jobs/9426695231

coverage didn't work

joernu76 (approved):

PR #1808, opened 2023-06-09
remove inputs from conditions

nilupulmanodya:
Remove inputs from tests in GSoC branches blocked by conditions


ReimarBauer (approved):
:)

PR #1809, opened 2023-06-12
Setup sp and idp for the sso

nilupulmanodya:
fixes of comments https://github.com/Open-MSS/MSS/pull/1805 except https://github.com/Open-MSS/MSS/pull/1805#discussion_r1221637588 . Once this merged will look into work on https://github.com/Open-MSS/MSS/pull/1805#pullrequestreview-1469250404


conf_sp_idp/sp/app/app.py
```diff
@@ -0,0 +1,204 @@
...
+from flask_login.utils import login_required, logout_user
+from flask_login import LoginManager, UserMixin, login_user
+
+from saml2.config import SPConfig
+from saml2.client import Saml2Client
+from saml2.metadata import create_metadata_string
+from saml2 import BINDING_HTTP_REDIRECT, BINDING_HTTP_POST
+from flask import Flask, redirect, request, render_template, url_for
+from flask.wrappers import Response
+import yaml
```
ReimarBauer: move this before the from imports

```
import 
import

from ... import ...

```

see
https://peps.python.org/pep-0008/#imports

conf_sp_idp/sp/app/app.py
```diff
@@ -0,0 +1,204 @@
...
+from yaml.loader import SafeLoader
+
```
ReimarBauer: is this needed or is it in the yaml.loader.SafeLoader already imported?

conf_sp_idp/idp/idp.py
```diff
@@ -0,0 +1,1141 @@
...
+    limitations under the License.
+"""
+# Additional Info:
+"""
+    This file is imported from https://github.com/IdentityPython/pysaml2/blob/master/example/idp2/idp.py
+    and customized as MSS requirements. Pylint has been disabled for this imported file.
+"""
+# Parts of the code
+
+import argparse
```
ReimarBauer, line 33: regroup imports https://peps.python.org/pep-0008/#imports

this improves reading it a bit

not important yet, can be done after the initial commit.

ReimarBauer: There could be added a note about the example codes too
https://github.com/nilupulmanodya/MSS/blob/Setup-SP-and-IDP-for-the-SSO/NOTICE#L29

conf_sp_idp/sp/README.md
```diff
@@ -0,0 +1,11 @@
+# Flask Service Provider with PySAML2 Integration
+
+This is a simple Flask service provider that allows for single sign-on (SSO) authentication using PySAML2.
+
+## Features
+
+- Integration with PySAML2 for SSO authentication.
+- Securely handles SAML assertions and authentication responses.
+- Provides routes for login, logout, and profile endpoints.
+- Uses SQLite database for user management.
```
ReimarBauer: SQLAlchemy database, it is not limited to SQLite

ReimarBauer (approved):
have a look on the comments, there are very minor things, I can merge it and that is done in next round.

conf_sp_idp/README.md
```diff
@@ -0,0 +1,56 @@
...
+## Getting started
+
+### TLS Setup
+
+**Setting Up Certificates for Local Development**
+
+
+To set up the certificates for local development, follow these steps:
+
+1. Generate a primary key `(.key)` and a certificate `(.crt)` files using any certificate authority tool. You will need one for the service provider and another one for the identity provider. Make sure to name certificate of identity provider as `crt_idp.crt` and key as `key_idp.key`. Also name the certificate of service provider as `crt_sp.crt` and key as the `key_sp.key`.
```
matrss, line 16: Might be worth mentioning a copy-paste-able command for creating keys and certificates for testing purposes. E.g. I used
```
openssl req -newkey rsa:4096 -keyout key_sp.key -nodes -x509 -days 365 -out crt_sp.crt
```
and the correspondingly modified command for the idp to get a self-signed certificate and private key.

conf_sp_idp/README.md
```diff
@@ -0,0 +1,56 @@
...
+7. Copy the generated `idp.xml` file and replace the existing `idp.xml` file located at `MSS/conf_sp_idp/sp/idp.xml`.
+
+### Running the Application After Configuration
+
+Once you have successfully configured the Service Provider and the Identity Provider, you don't need to follow the above instructions again. To start the application after the initial configuration, follow these steps:
+
+1. Start the Service provider:
+ * Navigate to the directory `MSS/conf_sp_idp/sp/app` and run `flask run`.
+2. Start the Identity Provider:
+ * Navigate to the directory `MSS/conf_sp_idp/idp` and run `python idp.py idp_conf`.
```
matrss, line 56: At this point I get an error:
```
ModuleNotFoundError: No module named 'cheroot'
```

I guess this happens since cherrypy was removed from the dependencies. Can the parts that are used from cheroot be replaced with something from flask?

conf_sp_idp/README.md
```diff
@@ -0,0 +1,56 @@
...
+
+By following the provided instructions, you will be able to set up and configure both the Identity Provider and Service Provider for testing the SSO process.
+
+## Testing SSO
+
+* Once you have successfully launched the server and identity provider, you can begin testing the Single Sign-On (SSO) process. 
+* Load in a browser `http://127.0.0.1:5000/`.
```
matrss:
```suggestion
* Load in a browser <http://127.0.0.1:5000/>.
```
This makes it an actually click-able link, no need to copy-paste when following the steps then.

conf_sp_idp/idp/idp.py
```diff
@@ -0,0 +1,1135 @@
...
+"""
+    conf_sp_idp.idp.idp.py
+    ~~~~~~~~~~~~~~~~~~~~~~
+
+    Identity provider implementation.
+
+    This file is part of MSS.
+
+    :copyright: Copyright 2023 Nilupul Manodya
+    :copyright: Copyright 2016-2023 by the MSS team, see AUTHORS.
```
matrss: Possibly a bit pedantic, but I don't see how the MSS team holds any copyright (yet). Any changes done until now are most likely so minor that they are not copyright-able. But we can probably leave that in anyway.

On the other hand, I am not sure if the link under "Additional Info" constitutes a copyright notice in the form as required by the upstream license (APACHE-2.0 as well, AFAICT). So we might want to add a copyright section for them here. IANAL though.

This note probably applies to the other files coming from the pysaml2 examples as well, in some form.

I'd like to get @ReimarBauer's opinion on this as well, though.

conf_sp_idp/idp/make_metadata.py
matrss, line 1: Is this file functionally equivalent to the one provided by pysaml2?

The pysaml2 conda package provides this file as well, in my case it is in `~/mambaforge/envs/mssdev/lib/python3.10/site-packages/saml2/tools/make_metadata.py`. Maybe we do not have to include the whole file here and rely on the one distributed with pysaml2 instead?

conf_sp_idp/sp/idp.xml
matrss, line 1: Since this file is newly generated and replaced when following the instructions in the readme, do we actually need it in git?

conf_sp_idp/sp/app/app.py
```diff
@@ -0,0 +1,204 @@
...
+from flask_sqlalchemy import SQLAlchemy
+from flask_migrate import Migrate
+
+
+
+app = Flask(__name__)
+app.config["SQLALCHEMY_DATABASE_URI"] = "sqlite:///db.sqlite"
+app.config[
+    "SECRET_KEY"
+] = "192b9bdd22ab9ed4d12e236c78afcb9a393ec15f71bbf5dc987d54727823bcbf"
```
matrss: We should keep in mind that such secret keys should not end up in git as soon as this is for more than just testing purposes.

matrss (commented):
I followed along with the steps in the readme and was able to set it up easily. It is documented very well!

I left some comments where I see some room for improvement or had something to note.

conf_sp_idp/idp/idp.py
ReimarBauer: this line is not needed
`:copyright: Copyright 2016-2023 by the MSS team, see AUTHORS.`


At the moment one of the AUTHORS changes something later the line likly will be 
`:copyright: Copyright 2023- by the MSS team, see AUTHORS.`

We can't have copyright before your initial contribution ;)

conf_sp_idp/idp/idp_conf.py
```diff
@@ -0,0 +1,213 @@
...
+
+try:
+    from saml2.sigver import get_xmlsec_binary
+except ImportError:
+    get_xmlsec_binary = None
+
+if get_xmlsec_binary:
+    XMLSEC_PATH = get_xmlsec_binary(["/opt/local/bin"])
+else:
+    XMLSEC_PATH = '/usr/bin/xmlsec1'
```
ReimarBauer: this likly get never reached once libxmlsec1 is a dependency on something else or we add it to the meta.yaml

conf_sp_idp/README.md
nilupulmanodya: Apologies for the inconvenience caused. I understand that the reason for the error was that I missed committing the necessary changes. I have now updated the branch with the required modifications.

conf_sp_idp/idp/idp.py
nilupulmanodya: @ReimarBauer @matrss  I have updated the lines according to your comments. Do I need to add anything else regarding the copyrights ?

ReimarBauer: We do use the NOTICE to mention 3rd party libs

https://github.com/Open-MSS/MSS/blob/develop/NOTICE

please it also there. It gives an overview what we have added to our repo.

conf_sp_idp/idp/idp_conf.py
```diff
@@ -37,17 +37,9 @@
...
-try:
-    from saml2.sigver import get_xmlsec_binary
-except ImportError:
-    get_xmlsec_binary = None
-
-if get_xmlsec_binary:
-    XMLSEC_PATH = get_xmlsec_binary(["/opt/local/bin"])
-else:
-    XMLSEC_PATH = '/usr/bin/xmlsec1'
+XMLSEC_PATH = get_xmlsec_binary(["/opt/local/bin"])
```
ReimarBauer: that is not good.

I would rather do it the other way around.  

We would need a conda build of xmlsec1 then, and this should then become an issue in MSS and here a ToDo that it is needed.

Always when we use some system libs this opens much problems. Because this likly different compiled as everything else.

When this stays as dependency, then we can't build for windows.

conf_sp_idp/README.md
```diff
@@ -31,8 +42,13 @@ First, generate the [metadata](https://pysaml2.readthedocs.io/en/latest/howto/co
...
 5. Go to the directory `MSS/conf_sp_idp/idp/`.
-6. Run the command `python make_metadata.py idp_conf.py > idp.xml` in the terminal to generate the `idp.xml` file.
-7. Copy the generated `idp.xml` file and replace the existing `idp.xml` file located at `MSS/conf_sp_idp/sp/idp.xml`.
+6. Run the command 
+    ```
+    python generate_metadata.py
+    ```
+    in the terminal to generate the `idp.xml` file. This function executes the make_metadata tool from pysaml2,
+    captures the output, and extracts the XML content. The XML content is then saved to the specified output file
+    in the service provider dir: `MSS/conf_sp_idp/sp/idp.xml`.
```
matrss: I just now realized from your `generate_metadata.py` script that pysaml2 puts a proper entrypoint for the `make_metadata.py` script into the path. That means it is possible to just use `make_metadata idp_conf.py > ../sp/idp.xml`, without any additional script on our side. That would simplify it a bit further.

conf_sp_idp/sp/app/conf.py
```diff
@@ -0,0 +1,42 @@
...
+import secrets
+
+class DefaultSPSettings:
+    """
+    Default settings for the SP (Service Provider) application.
+    
+    This class provides default configuration settings for the SP application.
+    Modify these settings as needed for your specific application requirements.
+    """
+    # MYSQL CONNECTION STRING:
```
matrss:
```suggestion
    # SQLite CONNECTION STRING:
```

conf_sp_idp/idp/idp.py
```diff
@@ -1120,12 +1127,15 @@ def application(environ, start_response):
         pass
     ds.DefaultSignature(sign_alg, digest_alg)
 
-    SRV = WSGIServer((HOST, PORT), application)
-
+    ssl_context = None
     _https = ""
-    if CONFIG.HTTPS:
+    if CONFIG.HTTPS:
```
matrss:
```suggestion
    if CONFIG.HTTPS:
```

There is some trailing whitespace that probably shouldn't be there.

matrss (approved):
I left a few more comments. Apart from those this looks good to me so far, nice work! I'd say this is merge-able after resolving the comments, unless we have some policy on squashing commits in PRs.

conf_sp_idp/idp/idp.py
nilupulmanodya: @ReimarBauer I have already included it in the notice. Do I need to make any further changes? https://github.com/Open-MSS/MSS/blob/153b402a622ed3f3200725aa0ced57944af3fac6/NOTICE#LL134C1-L143C56

ReimarBauer (approved):
good work.

when we have feedback from 
- https://github.com/IdentityPython/pysaml2/issues/913
- https://github.com/conda-forge/pysaml2-feedstock/issues/10

we can do a follow up adding a module to our meta.yaml (or it belongs to the package)

PR #1811, opened 2023-06-23
Split conf sp idp

nilupulmanodya:
split `conf_sp_idp` into `auth_client_sp` and `idp` to the dir of `mslib`


matrss (approved):
LGTM. I followed the setup instructions again and had no issues. I think the next broad step would be to find out how to make MSS itself into an SP that can authenticate against our test IdP.

matrss:
Thanks to a message by @ReimarBauer I just noticed that the documentation page is not built into the output.

We can add the page to the "Components" section by adding `conf_auth_client_sp_idp` to the bottom of this list:
https://github.com/Open-MSS/MSS/blob/6dc18ef2e71a893d83095c1d557d7bd032c61811/docs/components.rst?plain=1#L4-L13

(For reference, you can build a local version of the docs using `make -C docs html`. The output is then available in `docs/_build/html`.)

nilupulmanodya:
> Thanks to a message by @ReimarBauer I just noticed that the documentation page is not built into the output.
> 
> We can add the page to the "Components" section by adding `conf_auth_client_sp_idp` to the bottom of this list:
> 
> https://github.com/Open-MSS/MSS/blob/6dc18ef2e71a893d83095c1d557d7bd032c61811/docs/components.rst?plain=1#L4-L13
> 
> (For reference, you can build a local version of the docs using `make -C docs html`. The output is then available in `docs/_build/html`.)

Updated.. Thank you :)

PR #1812, opened 2023-06-25
Implement sso test cases

nilupulmanodya:
Test cases implementation for the localhost SSO process
We can review  and merge this when https://github.com/Open-MSS/MSS/pull/1811 merged


localbuild/meta.yaml
```diff
@@ -94,6 +94,7 @@ requirements:
     - dbus-python
     - flask-login
     - pysaml2
+    - beautifulsoup4
```
ReimarBauer, line 97: when this is only needed for tests, please move it to requirements.d/development.txt

tests/_test_sso_localhost/test_sso_localhost.py
```diff
@@ -0,0 +1,390 @@
...
+from urllib.parse import urljoin
+from bs4 import BeautifulSoup
+
+import requests
+import pytest
+
+os.environ['TEST_CASES_ENABLED'] = 'true'
+
+@pytest.fixture(scope='session')
+def client():
```
ReimarBauer, line 14: When client is a helper function this should go into tests/utils

tests/_test_sso_localhost/test_sso_localhost.py
```diff
@@ -0,0 +1,390 @@
...
+    The client fixture provides the test client object within a test context, allowing the tests
+    to interact with the application as a client.
+    """
+
+
+    # Configure flask service provider for testing
+
+    # Generate key and certificate for service provider (SP)
+    process_gen_crt_key_sp = (
+    'openssl req -x509 -newkey rsa:4096 -keyout mslib/auth_client_sp/test_key_sp.key '
```
ReimarBauer, line 38: there we should use a path based on this constant https://github.com/Open-MSS/MSS/blob/develop/tests/constants.py#L48

Maybe you can refactor it to a better identifier name, e.g. just mss ;)

Likly ROOT_DIR is what you want to use then, it is the meaning of root from a pyfilesystem path. 

At all when you use that constant, on each run a new tmp file is used and usually on a reboot that /tmp dir get's cleaned up.

Please don't save files into the mslib dir. Also a relative path is a problem. E.g. when I use the docker container for tests, I can also have a path like: 
`(mss-develop-env) root@d3a9d2e814be:/home/reimar/PycharmProjects/2023/nilupulmanodya/MSS/tests`

tests/_test_sso_localhost/test_sso_localhost.py
```diff
@@ -0,0 +1,390 @@
...
+
+    # Start the service provider (SP)
+    cmd_start_sp = ['python', 'mslib/auth_client_sp/app/app.py']
+    process_sp = subprocess.Popen(cmd_start_sp)
+
+    # Poll for the SP to start by checking if a specific URL is accessible
+    sp_started = False
+    while not sp_started:
+        try:
+            response = subprocess.run(['curl', '-s', '-o', '/dev/null', '-w', '%{http_code}', 'http://localhost:5000/'], capture_output=True)
```
ReimarBauer, line 63: curl maybe can become replaced by the requests module?

tests/_test_sso_localhost/test_sso_localhost.py
```diff
@@ -0,0 +1,390 @@
...
+    # Generate IDP metadata for service provider (SP)
+    cmd_make_metadata_idp = ['make_metadata', 'mslib/idp/idp_conf.py']
+    output_metadata_idp = 'mslib/auth_client_sp/test_idp.xml'
+    with open(output_metadata_idp, 'w', encoding='utf-8') as file:
+        subprocess.run(cmd_make_metadata_idp, stdout=file, check=True)
+
+    # Configure identity provider (IDP)
+
+    # Start the identity provider (IDP)
+    process_start_idp = subprocess.Popen(['python', 'mslib/idp/idp.py', 'idp_conf'])
```
ReimarBauer, line 88: That may can become replaced to something similiar to https://github.com/Open-MSS/MSS/blob/develop/tests/utils.py#L212 ?

tests/_test_sso_localhost/test_sso_localhost.py
```diff
@@ -0,0 +1,390 @@
...
+
+    with app.test_client() as client:
+        with app.app_context():
+            db.create_all()  # Create necessary database tables
+            yield client
+            db.session.remove()  # Remove the session after the test
+            db.drop_all()  # Drop all database tables
+
+    # Forcefully terminate the identity provider (IDP) process
+    process_start_idp.kill()
```
ReimarBauer, line 104: A bit this also looks like it can be part of a Test Class with 
setUp and tearDown

tests/_test_sso_localhost/test_sso_localhost.py
```diff
@@ -0,0 +1,390 @@
...
+
+
+
+    # Delete the generated files
```
ReimarBauer, line 108: see above these files have to go into a defined tmp dir

tests/_test_sso_localhost/test_sso_localhost.py
```diff
@@ -0,0 +1,390 @@
...
+    This test case verifies the functionality of adding a user to the database by
+    checking the result of the commit operation. It ensures that the user is 
+    successfully added to the database without errors.
+    """
+
+    # pylint: disable=C0415
+    from mslib.auth_client_sp.app.app import User, db
+    user = User(email='testemail@address.com')
+    db.session.add(user)
+    assert None is db.session.commit()
```
ReimarBauer, line 146: there should be also a check that the user is in the db.

tests/_test_sso_localhost/test_sso_localhost.py
```diff
@@ -0,0 +1,390 @@
...
+
+    subprocess.run(process_gen_crt_key_sp, shell=True, check=True)
+    # Generate key and certificate for identity provider (IDP)
+    process_gen_crt_key_idp = (
+    'openssl req -x509 -newkey rsa:4096 -keyout mslib/idp/test_key_idp.key '
+    '-out mslib/idp/test_crt_idp.crt -sha256 -days 3650 -nodes '
+    '-subj "/C=XX/ST=State/L=City/O=Company/OU=CompanySection/CN=CommonName"'
+    )
+
+    subprocess.run(process_gen_crt_key_idp, shell=True, check=True)
```
ReimarBauer, line 51: a further step can be to use pyopenssl 
- https://anaconda.org/conda-forge/pyopenssl
- https://www.pyopenssl.org/en/latest/

ReimarBauer (changes requested):
see comments

tests/_test_sso_localhost/test_sso_localhost.py
matrss: A pytest fixture as a generator, like it is done here, seems to be the recommended/idiomatic way to do setup and teardown stuff in pytest. pytest also supports `setup_{module,class,method}` and corresponding teardown functions, but according to the documentation fixtures are recommended over those. `setUp` and `tearDown` seem to be names used in `unittest`, which we should avoid.

Also, this fixture could be used in other tests where we might need the IdP/SP running, but aren't actively testing those. That would be harder with a class-based structure.

Therefore I'd argue to keep this fixture-based structure.

Maybe it would be beneficial to split the fixture up into two fixtures though, one for the IdP and one for the SP. From my understanding it would make sense to merge the SP functionality into the existing mscolab server component at some point, and two fixtures would be easier to deal with then.

matrss (commented):
I don't quite understand the reasoning for the `TEST_CASES_ENABLED` env variable and the special handling of `key_file` and `cert_file` when loading the yaml config. Can't we just have a testing-specific `saml2_backend.yaml` which points to the correct key and cert for testing and ditch the special handling in `get_crt_key()` and when loading yaml configuration file?

ReimarBauer:
@nilupulmanodya please look on the merge conflicts too ;)

tests/_test_sso_localhost/test_sso_localhost.py
```diff
@@ -0,0 +1,390 @@
+import os
```
ReimarBauer, line 1: this needs also a file header, see our template ;)

ReimarBauer (changes requested):
we need all certs and configfiles outside of the mslib dir created and stored. You can use constants or config values to have variables to point on them. We need this also for configuring after the software is installed not only for testing. But testing is then close to a deployed server. 
For example a default dir outside of mslib could be ~/mscolab.  On testing it is a tmp dir see the conftest.py example. 


[conftest.py.txt](https://github.com/Open-MSS/MSS/files/12131707/conftest.py.txt)

Afterwards when you have a working conftest and the pathes of the files needed configurable start with the simple tests. e. g.

 
```
from mslib.auth_client_sp.app.app import rndstr, get_idp_entity_id

def test_rndstr():
    assert rndstr() is not None
    assert len(rndstr()) == 16

def test_get_idp_entity_id():
    assert get_idp_entity_id() is None
    # this crashes currently because it returns an empty list
```

When you use a tests/_test_sso_localhost dir you can setup there an conftest.py, incomplete example above, where you can setup all certs etc. The metadata file could be created by a function test util because we have the cert and we don't need the server to be started to fetch it.

nilupulmanodya:
Closing this PR, since we are no longer using `auth_client_sp` .. will update other functional test cases with upcoming PRs :)

PR #1813, opened 2023-07-05
UI changes in Qt for SSO

nilupulmanodya:
* Disable `Add user` btn
* Add section for login with IDP

image


ReimarBauer:
A better solution is to hide the section, e.g. show only the section needed not both at a time.

ui


When an admin only enables IDP Login, the other path is completly unknown to users, we should hide that completly

The login button label also could be changed to Login by Identity Provider so it won't need the extra label.

You could add a placeholder widget and add the button programatically with its connection.

nilupulmanodya:
> A better solution is to hide the section, e.g. show only the section needed not both at a time.
> 
> ui
> 
> When an admin only enables IDP Login, the other path is completly unknown to users, we should hide that completly
> 
> The login button label also could be changed to Login by Identity Provider so it won't need the extra label.
> 
> You could add a placeholder widget and add the button programatically with its connection.

Updated : 
* Allowed admin to enable or disbale IDP login

* if `IDP_ENABLED = False` UI looks like,

image

* if `IDP_ENABLED = True` UI looks like,

image

mslib/msui/mscolab.py
```diff
@@ -219,6 +219,7 @@ def connect_handler(self):
             s.auth = auth
             s.headers.update({'x-test': 'true'})
             r = s.get(url_join(url, 'status'), timeout=(2, 10))
+            idp_enabled  = json.loads(r.text)["IDP_ENABLED"]
```
matrss:
```suggestion
            try:
                idp_enabled = json.loads(r.text)["IDP_ENABLED"]
            except (json.decoder.JSONDecodeError, KeyError):
                idp_enabled = false
```

The current version would crash if we were using this against an older version of mscolab, since it does not return well-formed json and/or the IDP_ENABLED key would not be present. This should also be moved further down, into the branch checking `r.status_code == 200`, since we can only expect the json response in the case of a successful request.

mslib/msui/mscolab.py
```diff
@@ -219,6 +219,10 @@ def connect_handler(self):
             s.auth = auth
             s.headers.update({'x-test': 'true'})
             r = s.get(url_join(url, 'status'), timeout=(2, 10))
+            try:
+                idp_enabled = json.loads(r.text)["IDP_ENABLED"]
+            except (json.decoder.JSONDecodeError, KeyError):
+                idp_enabled = False
```
matrss: Since this is handling the "happy-path", i.e. a successful request to the `/status` endpoint, this should be further down in the if-branch checking `r.status_code == 200`. A good spot should be right before the `if idp_enabled:` line.

mslib/msui/ui/ui_mscolab_connect_dialog.ui
```diff
@@ -7,7 +7,7 @@
     <x>0</x>
     <y>0</y>
     <width>478</width>
-    <height>255</height>
+    <height>268</height>
```
matrss: The "Login by Identity Provider" button is slightly cutoff on my system. See:

Screenshot from 2023-07-14 12-08-44

Is this happening only for me?

Weirdly if I increase the height value to 270 it is no longer cut off, but if I increase it to 272 it is cut off again. Seems like the vertical layout behaves weird with regards to automatically filling the available space. This might be caused by me using material-shell instead of some more common DE, so maybe this is a non-issue.

matrss (commented):
Some minor comments, otherwise this looks good to me.

Something to keep in mind: with the changes to the `/status` endpoint, it looks like older msui versions cannot connect to a newer mscolab server, i.e. the msui version from stable fails to connect. Not sure if this is an issue for us, users should regularly update anyway.

mslib/msui/ui/ui_mscolab_connect_dialog.ui
nilupulmanodya: it works for me on both 268, and 272..

This is how 272 appears for me.
image

However i set it to 270.

matrss (approved):
LGTM

ReimarBauer:
please add the new config option also to the sample config
https://github.com/Open-MSS/MSS/blob/GSOC2023-NilupulManodya/docs/samples/config/mscolab/mscolab_settings.py.sample

later it needs to become described too.

mslib/msui/mscolab.py
```diff
@@ -233,6 +233,20 @@ def connect_handler(self):
                 self.loginEmailLe.setEnabled(True)
                 self.loginPasswordLe.setEnabled(True)
 
+                try:
+                    idp_enabled = json.loads(r.text)["IDP_ENABLED"]
+                except (json.decoder.JSONDecodeError, KeyError):
+                    idp_enabled = False
+
+                if idp_enabled :
```
ReimarBauer: remove the blank before the ":"

mslib/msui/mscolab.py
```diff
@@ -233,6 +233,20 @@ def connect_handler(self):
...
+                    # Hide user creatiion seccion if IDP login enabled
```
ReimarBauer: typo

tests/_test_mscolab/test_server.py
```diff
@@ -80,7 +80,9 @@ def test_hello(self):
         with self.app.test_client() as test_client:
             response = test_client.get('/status')
             assert response.status_code == 200
-            assert b"Mscolab server" in response.data
+            data = json.loads(response.text)
+            assert "Mscolab server" in data['message']
+            assert True or False in data['IDP_ENABLED']
```
ReimarBauer, line 85: The config var needs to be added into the mscolab_settings.py which gets created by
https://github.com/Open-MSS/MSS/blob/GSOC2023-NilupulManodya/conftest.py#L111
on each test run. 

The more interesting question is, why this test doesn't fail?

ReimarBauer (changes requested):
see comments

ReimarBauer (approved):
good work thx :)

PR #1814, opened 2023-07-07
web browser implementation

nilupulmanodya:
QT web browser UI :

image

### DIFF
--- a/mslib/msui/msui_web_browser.py
+++ b/mslib/msui/msui_web_browser.py
@@ -1,15 +1,22 @@
 # -*- coding: utf-8 -*-
 """
+
     mslib.msui.msui_web_browser.py
-    ~~~~~~~~~~~~~~~~~~~~~~~~~~~
-    QTWebBrowser for the SSO process.
+    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+    MSUIWebBrowser can be used for localhost usage and testing purposes.
+
     This file is part of MSS.
+
     :copyright: Copyright 2023 Nilupul Manodya
     :license: APACHE-2.0, see LICENSE for details.
+
     Licensed under the Apache License, Version 2.0 (the "License");
     you may not use this file except in compliance with the License.
     You may obtain a copy of the License at
+
        http://www.apache.org/licenses/LICENSE-2.0
+
     Unless required by applicable law or agreed to in writing, software
     distributed under the License is distributed on an "AS IS" BASIS,
     WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
--- a/mslib/msui/msui_web_browser.py
+++ b/mslib/msui/msui_web_browser.py
@@ -24,12 +24,11 @@ from PyQt5.QtCore import QUrl, QTimer
 from PyQt5.QtWidgets import QMainWindow, QPushButton, QToolBar, QApplication
 from PyQt5.QtWebEngineWidgets import QWebEngineView, QWebEngineProfile
 
+from mslib.msui.constants import MSUI_CONFIG_PATH
 
-class QTWebBrowser(QMainWindow):
+class MSUIWebBrowser(QMainWindow):
     def __init__(self, url: str):
         super().__init__()
-
-        scriptdir = os.path.dirname(os.path.realpath(__file__))
 
         self.web_view = QWebEngineView(self)
         self.setCentralWidget(self.web_view)
@@ -37,7 +36,7 @@ class QTWebBrowser(QMainWindow):
         self._url = url
         self.profile = QWebEngineProfile().defaultProfile()
         self.profile.setPersistentCookiesPolicy(QWebEngineProfile.ForcePersistentCookies)
-        self.browser_storage_folder = scriptdir + os.path.sep + '/.cookies'
+        self.browser_storage_folder =  os.path.join(MSUI_CONFIG_PATH, '.cookies')
         self.profile.setPersistentStoragePath(self.browser_storage_folder)
 
         self.back_button = QPushButton("← Back", self)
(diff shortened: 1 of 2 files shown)


matrss:
I am not sure if I misunderstand the login flow in some way, but I would have some security concerns if we are using our own browser window instead of the users default browser: as a user using some IdP, e.g. the Helmholtz AAI or even GitHub (though that is not using SAML), I would want my password for that IdP to never be accessible to the SP, i.e. MSS/mscolab in our case. This is especially important if the IdP is also used to login into other applications, which is basically always the case. If we introduce our own QT browser window and ask the user to login to their IdP in that window we would now get at least the possibility to see the users password, even if we do not abuse that. As a user I trust my browser with my login credentials, but not some random window pretending to be a browser. Therefore I'd prefer to use the installed default browser instead of rolling our own.

mslib/msui/msui_web_browser.py
```diff
@@ -0,0 +1,97 @@
...
+
+        scriptdir = os.path.dirname(os.path.realpath(__file__))
+
+        self.web_view = QWebEngineView(self)
+        self.setCentralWidget(self.web_view)
+
+        self._url = url
+        self.profile = QWebEngineProfile().defaultProfile()
+        self.profile.setPersistentCookiesPolicy(QWebEngineProfile.ForcePersistentCookies)
+        self.browser_storage_folder = scriptdir + os.path.sep + '/.cookies'
```
ReimarBauer: this should be based on  constants.MSUI_CONFIG_PATH instead of scriptdir

mslib/msui/msui_web_browser.py
```diff
@@ -0,0 +1,97 @@
...
+from PyQt5.QtCore import QUrl, QTimer
+from PyQt5.QtWidgets import QMainWindow, QPushButton, QToolBar, QApplication
+from PyQt5.QtWebEngineWidgets import QWebEngineView, QWebEngineProfile
+
+
+class QTWebBrowser(QMainWindow):
+    def __init__(self, url: str):
+        super().__init__()
+
+        scriptdir = os.path.dirname(os.path.realpath(__file__))
```
ReimarBauer: use constants.MSUI_CONFIG_PATH

ReimarBauer:
> I am not sure if I misunderstand the login flow in some way, but I would have some security concerns if we are using our own browser window instead of the users default browser: as a user using some IdP, e.g. the Helmholtz AAI or even GitHub (though that is not using SAML), I would want my password for that IdP to never be accessible to the SP, i.e. MSS/mscolab in our case. This is especially important if the IdP is also used to login into other applications, which is basically always the case. If we introduce our own QT browser window and ask the user to login to their IdP in that window we would now get at least the possibility to see the users password, even if we do not abuse that. As a user I trust my browser with my login credentials, but not some random window pretending to be a browser. Therefore I'd prefer to use the installed default browser instead of rolling our own.

It is not planned to send your password to our mscolab SP.  The QT Webbrowser behaves like following examples, besides that we won't show the direct login on the mscolab server site as we have to on nextcloud.

- Example show how to contact over a server the AAI
phileas1

- use the AAI by its web service
phileas2


similiar to our nextcloud implementation we will get some response to see that the authentication was valid.
This part is yet not implemented.

There may be a user config to switch to the platforms browser and a description how to manually "login" by that. But first lets implement the authentication proccess at all. I think switching the browser afterwards is an easier task.

mslib/msui/msui_web_browser.py
```diff
@@ -0,0 +1,97 @@
...
+
+import os
+import sys
+
+from PyQt5.QtCore import QUrl, QTimer
+from PyQt5.QtWidgets import QMainWindow, QPushButton, QToolBar, QApplication
+from PyQt5.QtWebEngineWidgets import QWebEngineView, QWebEngineProfile
+
+
+class QTWebBrowser(QMainWindow):
```
ReimarBauer: the name should get a relation to MSUI, e.g MSUIWebBrowser. compare to class names in msui for the main window

matrss:
> It is not planned to send your password to our mscolab SP.

I am aware of that. My issue is that due to the QT webbrowser being a part of the MSS application running on the users device, it is technically possible to read out the password field of the IdP when it is filled in. If we would use the platforms browser, i.e. an application that we do not directly control, we wouldn't even have the possibility to access that information. As a user I'd prefer if a password leak was impossible (or rather dependent on the security of my preferred browser) instead of needing to trust some application developer to not do something shady. Not that I would be concerned about that with MSS, but in general (I have seen android apps that open PayPal for payments in an in-app browser and that seemed like a very shady thing to do, considering that those app developers could then inject arbitrary JS into the shown PayPal login page AFAIK, just to give an example).

> There may be a user config to switch to the platforms browser and a description how to manually "login" by that. But first lets implement the authentication proccess at all. I think switching the browser afterwards is an easier task.

That seems reasonable. For portability and/or testing purposes I see some value in having our own web browser window, e.g. for programmatic login to our testing IdP when testing this functionality.

ReimarBauer (changes requested):
see my comments

ReimarBauer:
This is a good addition for testing this, as we don't need to install firefox etc. into our docker container and have a clear state of a browser.

It maybe become also interesting for using our documentation locally.

mslib/msui/msui_web_browser.py
```diff
@@ -0,0 +1,96 @@
+# -*- coding: utf-8 -*-
+"""
+    mslib.msui.msui_web_browser.py
+    ~~~~~~~~~~~~~~~~~~~~~~~~~~~
+    QTWebBrowser for the SSO process.
```
ReimarBauer: the formatting is quite condensed but not that important

after a module import you do read it by help()

```
from mslib.msui import msui_web_browser
help(msui_web_browser)
```

ReimarBauer (approved):
Header can be updated when it is used for tests

matrss (approved):

ReimarBauer (approved):
Thx

PR #1818, opened 2023-07-17
Configure mscolab for sso

nilupulmanodya:
Configure the MSS collaboration server  for SSO with testing IDP

Need to merge https://github.com/Open-MSS/MSS/pull/1813 before merging this PR


mslib/mscolab/models.py
```diff
@@ -48,14 +48,16 @@ class User(db.Model):
     confirmed = db.Column(db.Boolean, nullable=False, default=False)
     confirmed_on = db.Column(db.DateTime, nullable=True)
     permissions = db.relationship('Permission', cascade='all,delete,delete-orphan', backref='user')
+    logged_by_idp = db.Column(db.Boolean, nullable=True, default=False)
```
matrss: We could call this field `authentication_backend` and make it a string with "local" identifying non-IdP users, if we ever want to support multiple IdP's. That's a big if though, and the rest of the configuration would need to support that use case as well.

mslib/mscolab/server.py
```diff
@@ -71,6 +80,19 @@ class mscolab_auth:
...
 
+#setup idp login config
+if mscolab_settings.IDP_ENABLED :
+    with open("mslib/mscolab/app/mss_saml2_backend.yaml", encoding="utf-8") as fobj:
+        yaml_data = yaml.safe_load(fobj)
+    if os.path.exists("mslib/mscolab/app/idp.xml"):
+        yaml_data["config"]["sp_config"]["metadata"]["local"] = ["mslib/mscolab/app/idp.xml"]
+    else:
+        yaml_data["config"]["sp_config"]["metadata"]["local"] = []
+        warnings.warn("idp.xml file does not exists !")
```
matrss: Is there any value in continuing to run if the file is missing? If not this could be simplified:

```suggestion
    if not os.path.exists(yaml_data["config"]["sp_config"]["metadata"]["local"][0]):
        sys.exit("idp.xml file does not exists !")
```

This would also remove the hardcoded path here and use the one specified in the configuration.

We should also consider if we want to merge the SP configuration file into the other mscolab configuration, but that can also be done later.

mslib/mscolab/server.py
```diff
@@ -703,6 +754,107 @@ def reset_request():
...
+        None, sp.config, 4, None, None, None, None, None
+    ).decode("utf-8")
+    return Response(metadata_string, mimetype="text/xml")
+
+
+@APP.route("/idp_login/", methods=['GET'])
+def idp_login():
+    """Handle the login process for the user by IDP"""
+    try:
+        # pylint: disable=unused-variable
```
matrss: Do these pylint directives serve any purpose?

mslib/mscolab/server.py
```diff
@@ -703,6 +754,107 @@ def reset_request():
...
+        if not user:
+            user = User(email, username, password=token, confirmed=False, confirmed_on=None, logged_by_idp=True)
+            db.session.add(user)
+            db.session.commit()
+
+        elif user.logged_by_idp is False:
+            user.logged_by_idp = True
+            user.hash_password(token)
+            db.session.add(user)
+            db.session.commit()
```
matrss: We should keep in mind that this can only be safe if the configured IdP confirms ownership of the mail address before allowing any login attempts. Otherwise it would be possible to get access to a local account without knowing its password.

mslib/mscolab/server.py
```diff
@@ -703,6 +754,107 @@ def reset_request():
...
+            request.form["SAMLResponse"], binding, outstanding=outstanding_queries
+        )
+        email = authn_response.ava["email"][0]
+        username = authn_response.ava["givenName"][0]
+
+        user = User.query.filter_by(emailid=email).first()
+        token = generate_confirmation_token(email)
+
+        if not user:
+            user = User(email, username, password=token, confirmed=False, confirmed_on=None, logged_by_idp=True)
```
matrss: ~`password=token` sets the password of the user to a token based on their email, `SECRET_KEY` and `SECURITY_PASSWORD_SALT`. If we ever provide this token to the user (login in the QT application?) this would probably leak `SECRET_KEY` and `SECURITY_PASSWORD_SALT`, which should never happen. Also, this means that every user is effectively getting the same password set, apart from the time based signing done in `generate_confirmation_token`. Can't we use a purely random token instead? I.e. something like `secrets.token_urlsafe(64)`?~

~This is also set again further down in `user.hash_password(token)`.~

Edit: I misread how the `URLSafeTimedSerializer` is called/working, it does not leak those values. Still, I think we should include a randomly generated secret in the data encoded in the token, which could be set as the password and on login be validated against the password field.

Edit Edit: I don't think the salt in `generate_confirmation_token` provides any value to us, since it has the same scope as `SECRET_KEY`. The description of this parameter is "salt ([str](https://docs.python.org/3/library/stdtypes.html#str) | [bytes](https://docs.python.org/3/library/stdtypes.html#bytes) | None) – Extra key to combine with secret_key to distinguish signatures in different contexts.". This would only be beneficial if we had a unique salt per user, to additionally distinguish tokens between them.

mslib/mscolab/server.py
nilupulmanodya: @matrss In the fresh start when [configuring the IDP](https://github.com/nilupulmanodya/MSS/blob/Configure-mscolab-for-SSO/docs/conf_sso_test_msscolab.rst#configuring-the-mss-colab-server-and-identity-provider), it may be necessary to run the `/metadata` . However, if the idp.xml file is missing, the server will encounter an error. This is why I chose to take that approach. Do i need to change it ? If you have any additional suggestions, I would be grateful to hear them ? : )

ReimarBauer:
add libxmlsec1 to the meta.yaml this will solve the missing library

from the description what pysaml2 uses 
https://github.com/IdentityPython/pysaml2/blob/v7.4.2/README.md?plain=1#L134

and from the definition of libxmlsec1
https://github.com/conda-forge/libxmlsec1-feedstock/blob/main/recipe/meta.yaml#L8

it is the same source

mslib/mscolab/server.py
```diff
@@ -71,6 +80,19 @@ class mscolab_auth:
                          ("add_new_user_here", "add_md5_digest_of_PASSWORD_here")]
         __file__ = None
 
+#setup idp login config
+if mscolab_settings.IDP_ENABLED :
+    with open("mslib/mscolab/app/mss_saml2_backend.yaml", encoding="utf-8") as fobj:
+        yaml_data = yaml.safe_load(fobj)
```
ReimarBauer, line 86:
```
On a server instance setup we have dirs differently

INSTANCE# tree -L 1
.
├── config
├── htdocs
├── log
├── mscolab
├── testdata
└── wsgi

ls config/
mscolab_auth.py  mscolab_settings.py  
```
We should be able to configure where this file is "mslib/mscolab/app/mss_saml2_backend.yaml"

We should use the mscolab_settings and a variable  for that file and use this var. default can be ~/mscolab
and on the testing mscolab_settings  it is based on fs.path.join(constants.ROOT_DIR, "mscolab")

mslib/mscolab/server.py
```diff
@@ -71,6 +80,19 @@ class mscolab_auth:
...
+    if os.path.exists("mslib/mscolab/app/idp.xml"):
```
ReimarBauer: also this file needs to be defined in the mscolab_settings

docs/conf_sso_test_msscolab.rst
```diff
@@ -0,0 +1,90 @@
...
+
+Getting started
+---------------
+
+
+Enable IDP login
+----------------
+
+To enable identity provider-based login, set `IDP_ENABLED = True` in the `mslib/mscolab/conf.py` file of the MSS Colab server.
+
```
ReimarBauer, line 25: a better solution is to create in your ~/mscolab dir a file mscolab_settings.py with the content
```

# -*- coding: utf-8 -*-
"""

    mscolab_settings
    ~~~~~~~~~~~~~~~~

    config for mscolab. It has all values different from the defaults.

   
"""
IDP_ENABLED = True
```
Then add this path to the PYTHONPATH as developer you can do this by e.g.
```
cd ~/PycharmProjects/2023/nilupulmanodya/MSS
export PYTHONPATH=`pwd`:~/mscolab
```

docs/conf_sso_test_msscolab.rst
```diff
@@ -0,0 +1,90 @@
...
+    * Generate a self-signed certificate and private key for the MSS Colab server
+    
+        ``openssl req -newkey rsa:4096 -keyout key_sp.key -nodes -x509 -days 365 -out crt_sp.crt``
+    
+    * Generate a self-signed certificate and private key for the Identity Provider (IdP)
+    
+        ``openssl req -newkey rsa:4096 -keyout key_idp.key -nodes -x509 -days 365 -out crt_idp.crt``
+
+2. Copy and paste the certificate and private key into the following file directories:
+
```
ReimarBauer, line 40: for now this is good enough 

later we need config vars in the mscolab_settings.py and these are used in the program

docs/conf_sso_test_msscolab.rst
```diff
@@ -0,0 +1,90 @@
...
+Configuring the MSS Colab server and Identity Provider
+------------------------------------------------------
+
+First, generate the metadata file (https://pysaml2.readthedocs.io/en/latest/howto/config.html#metadata) for the MSS Colab server. To do that, start the Flask application and download the metadata file by following these steps:
+
+1. Navigate to the home directory, ``/MSS/``.
+2. Start the MSS Colab Server by running ``$ python mslib/mscolab/mscolab.py start`` The application will listen on port : 8083.
+3. Download the metadata file by executing the command: ``curl http://localhost:8083/metadata/ -o sp.xml``.
+4. Move generated ``sp.xml`` to dir ``MSS/mslib/idp/``.
+
```
ReimarBauer, line 56: for now this is good enough

later we need config vars in the mscolab_settings.py and these are used in the program

docs/conf_sso_test_msscolab.rst
```diff
@@ -0,0 +1,90 @@
...
+After that, generate the idp.xml file, copy it over to the Service Provider (SP), and restart MSS Coalb Server:
+
+5. Go to the directory ``MSS/``.
+6. Run the command
+    ``$ make_metadata mslib/idp/idp_conf.py > mslib/mscolab/app/idp.xml``
+
```
ReimarBauer, line 59: later the file needs to be in some configurable dir

because idp_conf.py has all the vars we need to be able to change outside of an installation we need maybe to merge its content into the regular mslib/mscolab/conf.py then we can overwrite the values by a mscolab_settings.py.

Afterwards a user can run the command `make_metadata ~/mscolab/mscolab_settings.py > ~/mscolab/idp.xml 
`
or an other solution which enables us to change the file in a different dir by keeping defaults.

mslib/mscolab/server.py
ReimarBauer: see my comments to the documentation.

mslib/mscolab/server.py
ReimarBauer: When the default was not replaced by some "string" then this invalidates all existing tokens by a server restart.
https://github.com/Open-MSS/MSS/blob/GSOC2023-NilupulManodya/mslib/mscolab/conf.py#L55
Because a user can do set a string here, if we want to make it "safer" we can force these are invalid by hardcoding this on some place

docs/conf_sso_test_msscolab.rst
```diff
@@ -0,0 +1,90 @@
...
+To set up the certificates for local development, follow these steps:
+
+1. Generate a primary key `(.key)` and a certificate `(.crt)` files using any certificate authority tool. You will need one for the MSS Colab server and another one for the identity provider. Make sure to name certificate of identity provider as `crt_idp.crt` and key as `key_idp.key`. Also name the certificate of msss colab server as `crt_sp.crt` and key as the `key_sp.key`.
+
+    Here's how you can generate self-signed certificates and private keys using OpenSSL:
+    
+    * Generate a self-signed certificate and private key for the MSS Colab server
+    
+        ``openssl req -newkey rsa:4096 -keyout key_sp.key -nodes -x509 -days 365 -out crt_sp.crt``
```
ReimarBauer: with the option we don't get that Q/A sections

 -batch -subj '/CN=localhost'

mslib/mscolab/app/mss_saml2_backend.yaml
```diff
@@ -0,0 +1,61 @@
+name: Saml2
```
ReimarBauer, line 1: the server must be able to load this file from a config dir. 

on a production system we have to fill a lot data into this.

We maybe can also have a command which fills in some data in such a file.

mslib/mscolab/server.py
```diff
@@ -71,6 +80,19 @@ class mscolab_auth:
                          ("add_new_user_here", "add_md5_digest_of_PASSWORD_here")]
         __file__ = None
 
+#setup idp login config
+if mscolab_settings.IDP_ENABLED :
+    with open("mslib/mscolab/app/mss_saml2_backend.yaml", encoding="utf-8") as fobj:
```
ReimarBauer: this should be a constant so we can use it from a different dir

docs/conf_sso_test_msscolab.rst
ReimarBauer: thought a bit more on this. Our IDP we use only for testing. if possible make its configuration possible for testings. 
When we use it for user managment we need to do more.

docs/conf_sso_test_msscolab.rst
ReimarBauer: IDP_ENABLED = True

maybe should become renamed to USE_SAML2

mslib/mscolab/server.py
```diff
@@ -197,6 +219,34 @@ def wrapper(*args, **kwargs):
...
+    :type alphabet: str
+    :param size: The length of the string
+    :param alphabet: A string with characters.
+    :return: string
+    """
+    rng = random.SystemRandom()
+    if not alphabet:
+        alphabet = string.ascii_letters[0:52] + string.digits
+    return type(alphabet)().join(rng.choice(alphabet) for _ in range(size))
+
```
ReimarBauer, line 249: have also a look on secrets https://docs.python.org/3/library/secrets.html
import secrets

it is maybe a good option

mslib/mscolab/server.py
matrss: You are right, I forgot about the call to `/metadata` and we can not exit there. Still, we should not hardcode a path that is also set in the configuration. Something like this should be better:
```suggestion
    if not os.path.exists(yaml_data["config"]["sp_config"]["metadata"]["local"][0]):
        yaml_data["config"]["sp_config"]["metadata"]["local"] = []
        logging.warning("idp.xml file does not exists !")
```

mslib/static/templates/errors/500.html
```diff
@@ -0,0 +1,5 @@
+<div>
+    <h1>500 - Sorry Unexpected Error</h1>
+    <hr>
+    <p>We are working on fixing the problem be back soon.</p>
+</div>
```
ReimarBauer, line 5: I am not sure if we will recognize the problem, may be the user should file an issue and describe it.

mslib/mscolab/server.py
ReimarBauer: Further thought on that. 

I am not sure if we need the call of the server to fetch the metadata file and create it. It is may be also possible to use a template and fill in all data and return that rendered.  Of course we need to access the metadata by an url. But is it needed to have it saved to a file? When we save it we would need to recreate it on any cert change.

mslib/mscolab/models.py
```diff
@@ -48,16 +48,16 @@ class User(db.Model):
     confirmed = db.Column(db.Boolean, nullable=False, default=False)
     confirmed_on = db.Column(db.DateTime, nullable=True)
     permissions = db.relationship('Permission', cascade='all,delete,delete-orphan', backref='user')
-    logged_by_idp = db.Column(db.Boolean, nullable=True, default=False)
+    authentication_backend = db.Column(db.String(255), nullable=True, default=False)
```
matrss: `default=False` should be `default="local"` instead, I think. Is it necessary to have this field be nullable, or could we also set `nullable=False`? That would avoid an invalid state to be representable at all.

mslib/mscolab/mscolab.py
```diff
@@ -92,6 +94,72 @@ def handle_db_seed():
     seed_data()
     print("Database seeded successfully!")
 
+def handle_mscolab_certificate_init():
+    print('generating CRTs for the mscolab server......')
+    cmd = f"openssl req -newkey rsa:4096 -keyout {mscolab_settings.MSCOLAB_SSO_DIR}/key_mscolab.key -nodes -x509 -days 365 -batch -subj '/CN=localhost' -out {mscolab_settings.MSCOLAB_SSO_DIR}/crt_mscolab.crt"
+    os.system(cmd)
```
matrss: `os.system` should almost always be replaced with `subprocess.run` (<https://docs.python.org/3/library/subprocess.html#subprocess.run>). It is a bit nicer to work with, less platform dependent and e.g. avoids spawning a sub-shell by default, which often is not necessary.

mslib/mscolab/mscolab.py
```diff
@@ -92,6 +94,72 @@ def handle_db_seed():
...
+
+        Before running this function:
+        - Ensure that IDP_ENABLED is set to True.
+        - Generate the necessary keys and certificates and configure them in the .yaml 
+        file for the local IDP.
+    '''
+    print('generating metadata file for the mscolab server')
+
+    cmd ="python mslib/mscolab/mscolab.py start"
+    subprocess.Popen(cmd, shell=True)
```
matrss:
```suggestion
    subprocess.Popen(["python", "mslib/mscolab/mscolab.py", "start"])
```

While I don't see an issue with `shell=True` here, apart from unnecessary overhead, I also don't think it is necessary. In general `shell=True` can quickly become a security issue due to shell injection attacks (<https://docs.python.org/3/library/subprocess.html#security-considerations>), so I would like to avoid it as much as possible.

docs/conf_sso_test_msscolab.rst
```diff
@@ -0,0 +1,77 @@
...
+
+Once you setted certificates and metada files you can start mscolab server and local identity provider. To start local identity provider, simpy execute
+
+`$ python mslib/idp/idp.py idp_conf`
+
+
+Restart the mscolab Server
+--------------------------
+
+Start mscolab server
```
ReimarBauer: Maybe add something like this

When this is the first time you setup a mscolab server, you have to initialize the database by

`$ python mslib/mscolab/mscolab.py db --init`

.. note::
   An existing database maybe needs a migration, have a look for this on our documentation.

for syntax see https://docutils.sourceforge.io/docs/ref/rst/directives.html#hint

docs/conf_sso_test_msscolab.rst
```diff
@@ -0,0 +1,77 @@
...
+
+Start Identity provider
+-----------------------
+
+Once you setted certificates and metada files you can start mscolab server and local identity provider. To start local identity provider, simpy execute
+
+`$ python mslib/idp/idp.py idp_conf`
+
+
+Restart the mscolab Server
```
ReimarBauer: Start instead of Restart

We better describe the "new" state and add hint what to skip or do differently if it is an old instance

ReimarBauer (commented):
comment removed

mslib/mscolab/mscolab.py
```diff
@@ -92,6 +94,72 @@ def handle_db_seed():
...
+    print('generating metadata files.......')
+    handle_mscolab_metadata_init()
+    handle_local_idp_metadata_init()
+    print("ALl necessary metadata file generated successfully")
+
+    # Get the process group ID of the current process
+    pgid = os.getpgid(0)
+
+    # Kill the whole terminal by sending a SIGKILL signal to the process group
+    os.killpg(pgid, 9)
```
matrss: This looks overly intrusive, what is the desired outcome here?

ReimarBauer (changes requested):
in idp_conf.py and conftest.py there is a leftover definition which writes into MSS dir. I am not sure if a file store is needed at all. conftest should add them if needed to the tmp dir and  idp_conf if needed have them on the colabdata. .gitignore has also an entry

 "subject_data": "./idp.subject",
 
 Could we also move the idp/modules to the colabdata? For what are they needed? I mean why is that not just a Jinja2 template used? 
 
 interestingly didn't I find what gets mako into the environment `mamba repoquery search mako` finds only this module.

ReimarBauer:
also in the MSS dir are two file idp.subject.dir idp.subject.dat
there are tokens in it. Is that needed?

matrss (commented):
I tried to follow the instructions and couldn't get this working. I had a hard time figuring out where to set which configuration and when trying to generate the metadata files with `python mslib/mscolab/mscolab.py sso_conf --init_sso_metadata` I ran into a bunch of exceptions from the xml parser due to `~/colabdata/datasso/idp.xml` being empty and `~/colabdata/datasso/metadata_sp.xml` containing:
```
<!doctype html>
<html lang=en>
<title>500 Internal Server Error</title>
<h1>Internal Server Error</h1>
<p>The server encountered an internal error and was unable to complete your request. Either the server is overloaded or there is an error in the application.</p>
```

The output of the `--init_sso_metadata` command still said
```
idp metadata file generated succesfully
ALl necessary metadata file generated successfully
```
so there is something wrong with the error handling of the executed commands.

I think I am missing a step in "Enable IDP login". I just copied over `mslib/mscolab/conf.py` to `~/mss/mscolab_settings.py` and set `IDP_ENABLED = True`. `CONFIGURED_IDPS` seemed to be correctly set already for the testing IdP. I am not sure what, if anything, to do with the yaml SP config.

ReimarBauer:
Try to avoid global as much as possible. I found one in idp_uwsgi.py

These are some ideas on that and some options. For now a ToDo refactor 
before the line is ok me. 

https://betterprogramming.pub/alternatives-to-using-globals-in-python-a3b2a7d5411b

http://google.github.io/styleguide/pyguide.html

nilupulmanodya:
> I tried to follow the instructions and couldn't get this working. I had a hard time figuring out where to set which configuration and when trying to generate the metadata files with `python mslib/mscolab/mscolab.py sso_conf --init_sso_metadata` I ran into a bunch of exceptions from the xml parser due to `~/colabdata/datasso/idp.xml` being empty and `~/colabdata/datasso/metadata_sp.xml` containing:
> 
> ```
> <!doctype html>
> <html lang=en>
> <title>500 Internal Server Error</title>
> <h1>Internal Server Error</h1>
> <p>The server encountered an internal error and was unable to complete your request. Either the server is overloaded or there is an error in the application.</p>
> ```
> 
> The output of the `--init_sso_metadata` command still said
> 
> ```
> idp metadata file generated succesfully
> ALl necessary metadata file generated successfully
> ```
> 
> so there is something wrong with the error handling of the executed commands.
> 
> I think I am missing a step in "Enable IDP login". I just copied over `mslib/mscolab/conf.py` to `~/mss/mscolab_settings.py` and set `IDP_ENABLED = True`. `CONFIGURED_IDPS` seemed to be correctly set already for the testing IdP. I am not sure what, if anything, to do with the yaml SP config.

Thanks for figuring out that point. updated documentation and error handling. 🙂

docs/conf_sso_test_msscolab.rst
```diff
@@ -0,0 +1,92 @@
...
+Before Starting the mscolab server, make sure to do necessary database migrations.
+
+When this is the first time you setup a mscolab server, you have to initialize the database by
+
+`$ python mslib/mscolab/mscolab.py db --init`
+
+.. note::
+An existing database maybe needs a migration, have a look for this on our documentation.
+for syntax see https://docutils.sourceforge.io/docs/ref/rst/directives.html#hint
+
```
ReimarBauer, line 79: That was a misunderstanding. It shows how the "hint" syntax in sphinx is, that is our link in the docs

https://mss.readthedocs.io/en/stable/mscolab.html#data-base-migration

docs/conf_sso_test_msscolab.rst
```diff
@@ -0,0 +1,92 @@
...
+
+When migrations done, you can tart mscolab server by.
```
ReimarBauer: typo:

start

ReimarBauer (changes requested):
small "typos" in the documentation. 

Add ToDos of the not solved comments into the codebase, e.g.  https://github.com/Open-MSS/MSS/pull/1818#discussion_r1270701658

`# ToDo refactor, use common lib secrets? see discussion in https://github.com/Open-MSS/MSS/pull/1818#discussion_r1270701658`

mslib/mscolab/mscolab.py
```diff
@@ -96,69 +96,236 @@ def handle_db_seed():
 
 def handle_mscolab_certificate_init():
     print('generating CRTs for the mscolab server......')
-    cmd = f"openssl req -newkey rsa:4096 -keyout {mscolab_settings.MSCOLAB_SSO_DIR}/key_mscolab.key -nodes -x509 -days 365 -batch -subj '/CN=localhost' -out {mscolab_settings.MSCOLAB_SSO_DIR}/crt_mscolab.crt"
-    os.system(cmd)
-    print('CRTs generated successfully for the mscolab server......')
+
+    try:
+        cmd = f"openssl req -newkey rsa:4096 -keyout {mscolab_settings.MSCOLAB_SSO_DIR}/key_mscolab.key -nodes -x509 -days 365 -batch -subj '/CN=localhost' -out {mscolab_settings.MSCOLAB_SSO_DIR}/crt_mscolab.crt"
+        subprocess.run(cmd, shell=True, check=True)
```
matrss, line 102: I would still prefer getting rid of `shell=True`, which should be possible with:

```suggestion
        cmd = ["openssl", "req", "-newkey", "rsa:4096", "-keyout", f"{mscolab_settings.MSCOLAB_SSO_DIR}/key_mscolab.key", "-nodes", "-x509", "-days", "365", "-batch", "-subj", "/CN=localhost", "-out", f"{mscolab_settings.MSCOLAB_SSO_DIR}/crt_mscolab.crt"]
        subprocess.run(cmd, check=True)
```

Same for the other occurrences of `shell=True`. Only the `make_metadata` calls should be less straightforward due to the stdout redirection, which is a shell feature. That should also be doable in python using subprocess.run's `stdout=<file-object>` though.

nilupulmanodya:
> Try to avoid global as much as possible. I found one in idp_uwsgi.py
> 
> These are some ideas on that and some options. For now a ToDo refactor before the line is ok me.
> 
> https://betterprogramming.pub/alternatives-to-using-globals-in-python-a3b2a7d5411b
> 
> http://google.github.io/styleguide/pyguide.html

Added into ToDo : https://github.com/Open-MSS/MSS/pull/1818/commits/e3a81ba9c1905ed993c56d4bf6c7e21405cf2a6b

nilupulmanodya:
> also in the MSS dir are two file idp.subject.dir idp.subject.dat there are tokens in it. Is that needed?

Added into Todo :https://github.com/Open-MSS/MSS/pull/1818/commits/e3a81ba9c1905ed993c56d4bf6c7e21405cf2a6b

nilupulmanodya:
> in idp_conf.py and conftest.py there is a leftover definition which writes into MSS dir. I am not sure if a file store is needed at all. conftest should add them if needed to the tmp dir and idp_conf if needed have them on the colabdata. .gitignore has also an entry
> 
> "subject_data": "./idp.subject",
> 
> Could we also move the idp/modules to the colabdata? For what are they needed? I mean why is that not just a Jinja2 template used?
> 
> interestingly didn't I find what gets mako into the environment `mamba repoquery search mako` finds only this module.

Added into ToDo : https://github.com/Open-MSS/MSS/pull/1818/commits/e3a81ba9c1905ed993c56d4bf6c7e21405cf2a6b

mslib/mscolab/server.py
nilupulmanodya: Added into Todo : https://github.com/Open-MSS/MSS/pull/1818/commits/e3a81ba9c1905ed993c56d4bf6c7e21405cf2a6b

matrss (commented):
The issue I had was caused by me setting `IDP_ENABLED` (now `USE_SAML2`) in `~/mss/mscolab_settings.py` instead of `mslib/mscolab/conf.py`. Since the former should overwrite the latter there seems to be something wrong with the loading of the configuration. Nevertheless I was able to get this working now, to the point that a login with the IdP in msui works, which is great :)

I saw a few typos in `docs/conf_sso_test_msscolab.rst` though, could you please proofread that?

Apart from this I think we can merge this and improve on it in the next PR.

mslib/mscolab/mscolab.py
ReimarBauer: > 

This needs to be done after the exams too. We get back to this with writing tests.

ReimarBauer (approved):
There are a few ToDo's which will be done in the next PR

mslib/mscolab/mscolab.py
nilupulmanodya: Resolved by #1974

mslib/mscolab/server.py
nilupulmanodya: Resolved by #1974

PR #1843, opened 2023-07-28
renamed endpoint for mscolab authentification

### LINKED ISSUE #1842: mscolab authentfiation backwards compatibility
Currently, the behaviour of the "status" API has been changed. This leads to a problem in operating a server for both old and new clients.
As the behaviour of this endpoint has changed anyway, it makes sense to also rename it such that a server could, potentially, server both versions.

joernu76:
With this change, we can "patch" the mscolab server to serve both 8.x and 9.x but do not need to. This can actually remain in develop, but we can remove it as well. I think it is better this way.

### DIFF
--- a/mslib/mscolab/server.py
+++ b/mslib/mscolab/server.py
@@ -203,7 +203,7 @@ def home():
     return render_template("/index.html")
 
 
-@APP.route("/status")
+@APP.route("/status_auth")
 @conditional_decorator(auth.login_required, mscolab_settings.__dict__.get('enable_basic_http_authentication', False))
 def hello():
     return "Mscolab server"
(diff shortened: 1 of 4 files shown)


mslib/mscolab/server.py
```diff
@@ -203,7 +203,7 @@ def home():
     return render_template("/index.html")
 
 
-@APP.route("/status")
+@APP.route("/status_auth")
 @conditional_decorator(auth.login_required, mscolab_settings.__dict__.get('enable_basic_http_authentication', False))
```
ReimarBauer, line 207: I think this makes upcoming 9 consistent. It means also without /status any older version of msui knows to be outdated

We could also add a hint for that change into the 8.1.x

We could also use a complete different route name. /get_status, /info, /hello

joernu76: Correct, I think it can stay as this. A better name would maybe be "authenticate"? but the current one is fine. Any changes required?

ReimarBauer: We remember this with the merge of the work from @nilupulmanodya

ReimarBauer (approved):

mslib/mscolab/server.py
nilupulmanodya: Sure.. Thanks for remembering :)

PR #1853, opened 2023-08-02
Updated topview.py

ambiguousphoton:
@ReimarBauer I have synced the repo please merge.


ReimarBauer:
Seen I'm in a train. Network ist unstable. Tests running currently. I assume you are done.

ReimarBauer:
Hmm, just seen this is a Bug fix.

We have two Branches stable and develop. Releases are based on stable. That means usually Bug Fixes have to be PR to stable branch and enhancement is always PR to develop.
Only when we need API changes a Bug fix needs to be done in develop.

See also

https://mss.readthedocs.io/en/stable/development.html#development

Do you mind to send the fix to stable?

PR #1854, opened 2023-08-02
Adding GUI mssautoplot CLI tool

swsrkty:
Adding GUI to the mssautoplot CLI tool.

 Following changes are made:
1. Adding the `ui_mssautoplot_gui.py`, `ui_mssautoplot_gui.ui` files corresponding to the GUI.
2. Refactoring the mssautoplot main function so that the GUI can make functional calls to it.
3. changing `QtCore.Signal()` and `QtCore.Slot()` to `QtCore.pyqtSignal()` and `QtCore.pyqtSlot()` as they are no longer supported in PyQt5 (https://github.com/picotech/picosdk-python-examples/issues/2).
4. Minor fix in `wms_control.py` regarding attribute `fig`.

### DIFF
--- a/mslib/msui/mssautoplot_gui.py
+++ b/mslib/msui/mssautoplot_gui.py
@@ -29,7 +29,7 @@ import sys
 import os
 from PyQt5 import QtCore, QtGui, QtWidgets
 from mslib.msui.qt5 import ui_mssautoplot_gui as ui
-from mslib.utils import mssautoplot 
+from mslib.utils import mssautoplot
 from PyQt5.QtWidgets import QApplication, QWidget, QFileDialog
 from mslib.utils.config import config_loader, read_config_file
 from mslib.msui import constants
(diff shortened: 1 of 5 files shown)


joernu76:
Pretty cool!

joernu76:
Some questions:
1) How do I access the GUI? I can call it directly in the qt5/ui folder. Is this intended?
2) There seems to be a mistype with self.config_edit. Should this be self.ConfigPathlineEdit ?
3) I am a bit lost. Could you supply a working example with necessary files for loading in etc?

swsrkty:
Hi @joernu76 ,

>1.  How do I access the GUI? I can call it directly in the qt5/ui folder. Is this intended?

It is supposed to be called as `mssautoplot --gui`. I haven't implemented that yet. Will do soon.

>2. There seems to be a mistype with self.config_edit. Should this be self.ConfigPathlineEdit ?

Ah, I'll look into it.
>3. I am a bit lost. Could you supply a working example with necessary files for loading in etc?

Sure.
I realized that some work needs to be done in the side view part. Side view in the GUI doesn't work as of now. I'll write a detailed documentation after that.

Issue #1835, opened 2023-07-25
traceback on another public WMS server


ReimarBauer:
https://www.geobasisdaten.niedersachsen.de/doorman/noauth/bestand

```
Fatal error in MSS 8.1.1 on Linux-6.2.0-10018-tuxedo-x86_64-with-glibc2.35
Python 3.10.8 | packaged by conda-forge | (main, Nov 22 2022, 08:23:14) [GCC 10.4.0]

Please report bugs in MSS to https://github.com/Open-MSS/MSS

Information about the fatal error:

Traceback (most recent call last):
  File "/home/reimar/mambaforge/envs/mssenv/lib/python3.10/site-packages/mslib/msui/wms_control.py", line 1383, in continue_retrieve_image
    self.display_retrieved_image(img, legend_img, layer, style, init_time, valid_time, complete_level)
  File "/home/reimar/mambaforge/envs/mssenv/lib/python3.10/site-packages/mslib/msui/wms_control.py", line 1639, in display_retrieved_image
    self.view.draw_legend(self.append_multiple_images(legend_imgs))
  File "/home/reimar/mambaforge/envs/mssenv/lib/python3.10/site-packages/mslib/msui/wms_control.py", line 1484, in append_multiple_images
    max_height = int((self.view.fig.get_size_inches() * self.view.fig.get_dpi())[1] * 0.99)
AttributeError: 'MplTopViewCanvas' object has no attribute 'fig'
```

ReimarBauer:
https://geoportal2.duisburg.de/arcgisserver/services/Basisdaten/DGK_hist_Portal/MapServer/WMSServer
https://geoportal2.duisburg.de/arcgisserver/services/Luftbilder/Historische_Luftbilder/MapServer/WMSServer

Fatal error in MSS 8.1.1 on Linux-6.2.0-10018-tuxedo-x86_64-with-glibc2.35
Python 3.10.8 | packaged by conda-forge | (main, Nov 22 2022, 08:23:14) [GCC 10.4.0]

Please report bugs in MSS to https://github.com/Open-MSS/MSS

Information about the fatal error:

```
Traceback (most recent call last):
  File "/home/reimar/mambaforge/envs/mssenv/lib/python3.10/site-packages/mslib/msui/wms_control.py", line 1383, in continue_retrieve_image
    self.display_retrieved_image(img, legend_img, layer, style, init_time, valid_time, complete_level)
[...]

PR #1865, opened 2023-08-08
warn on other needed implementations

ReimarBauer:
minimum for not implemented figure  #1835


ReimarBauer:
@joernu76 that may be also something we could add to stable.  This enables to use some data from e.g.
https://www.geobasisdaten.niedersachsen.de/doorman/noauth/bestand

joernu76 (approved):

PR #1869, opened 2023-08-09
Proper disabling of unticked KML files.

### LINKED ISSUE #1866: tick/untick for multiple kml overlays does not work
if multiple kml overlay files are chosen, on can tick and untick the display of those. If one of two kml files is un-ticked it is anyhow displayed on refreshing the top view. 
The display can be stopped if the tick box is hit twice then (tic, untick

This is not how it should be

joernu76:
Fix #1866

### DIFF
--- a/mslib/msui/kmloverlay_dockwidget.py
+++ b/mslib/msui/kmloverlay_dockwidget.py
@@ -337,7 +337,8 @@ class KMLOverlayControlWidget(QtWidgets.QWidget, ui.Ui_KMLOverlayDockWidget):
 
     def update(self):
         for entry in self.dict_files.values():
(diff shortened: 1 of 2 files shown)


jgrooss:
This seems to work. At least I don't see the problem anymore

jgrooss (dismissed):
This seems to have solved the problem

ReimarBauer (changes requested):
there is another problem, not checked if it is a new one.

When you click on change color on a not check marked kml it crashes.
Same with resize. 

change_color_problem

joernu76:
will have a look

ReimarBauer (approved):
Thx

PR #1872, opened 2023-08-10
improved checks, ToDos added

### LINKED ISSUE #1871: limit for is_member on some calls
We need to limit some functions to members of an operation

ReimarBauer:
fixes #1871


mslib/mscolab/server.py
```diff
@@ -482,12 +495,16 @@ def update_operation():
 def get_operation_details():
     op_id = request.args.get('op_id', request.form.get('op_id', None))
     user = g.user
+    result = str(fm.get_operation_details(int(op_id), user))
+    if result is False:
+        return "False"
     return json.dumps(fm.get_operation_details(int(op_id), user))
```
joernu76: get_operation_details is called twice here in the "normal" case.

joernu76 (dismissed):

mslib/mscolab/server.py
ReimarBauer: one too much, good catch

ReimarBauer:
arg now something is going wrong

PR #1892, opened 2023-08-11
hello2 not called, better solution.

### LINKED ISSUE #1893: fix wrong usage of decorator
hello2() not called and  we need only auth.login_required()

This is also executed on any conditional_decorator in other parts

ReimarBauer:
just don't use a decorator here.

### DIFF
--- a/mslib/mscolab/server.py
+++ b/mslib/mscolab/server.py
@@ -205,15 +205,13 @@ def home():
 # ToDo setup codes in return statements
 @APP.route("/status")
 def hello():
-    try:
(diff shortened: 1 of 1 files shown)


ReimarBauer:
normally indented decorators should give an indention error. However this didn't and also flake8 had no problem.

In our case we just need to have auth.login_required() called when we submit data.

ReimarBauer:
discussed with Andi and Klaus, the indentation was not the problem. I did'nt call hello2.
But as we only have to call auth.login_required() the new solution is much nicer

joernu76 (approved):

Issue #1614, opened 2022-11-23
review and refactor abusing None for boolean


ReimarBauer:
Sometimes None is checked like boolean. This is a problem and needs to become fixed

KrishangSaharia:
If this is still open, can I take this issue, also I am new to open source so it would be a good starting point for me also?

ReimarBauer:
This is not easy. As a startingpoint something with only one line fix would be better to get the setup for local tests verified.  
Some of these problems already fixed. The issue is a reminder to take care of this. I wished we had a linter which can detect these early.

Here is one where it is wrong:

https://github.com/Open-MSS/MSS/blob/develop/docs/samples/plugins/navaid.py#L106

ReimarBauer:
On some places we were not exact, and during fixes or enhancements we will find code locations where we can improve this. 

I am not aware about a test which can help to find these and protect for new ;)

ReimarBauer:
The issue stays open as a reminder.

ReimarBauer:
https://github.com/Open-MSS/MSS/blob/develop/mslib/mscolab/server.py#L186

if user is None:

ReimarBauer:
some of the refactoring is done during GSoC

PR #1911, opened 2023-08-18
Remove usage of werkzeug.urls.url_encode

### LINKED ISSUE #1910: Fix DeprecationWarning from usage of werkzeug.urls.url_encode
From the latest, at time of writing, CI run on develop:
```
  /__w/MSS/MSS/mslib/msui/mscolab_version_history.py:140: DeprecationWarning: 'werkzeug.urls.url_encode' is deprecated and will be removed in Werkzeug 2.4. Use 'urllib.parse.urlencode' instead.
    query_string = url_encode({"named_version": named_version_only})
```

matrss:
Replaces it with `urllib.parse.urlencode`. Fixes #1910.

### DIFF
--- a/mslib/msui/mscolab_version_history.py
+++ b/mslib/msui/mscolab_version_history.py
@@ -133,7 +133,7 @@ class MSColabVersionHistory(QtWidgets.QMainWindow, ui.Ui_MscolabVersionHistory):
                 "token": self.token,
                 "op_id": self.op_id
             }
-            named_version_only = None
+            named_version_only = False
             if self.versionFilterCB.currentIndex() == 0:
                 named_version_only = True
             query_string = urlencode({"named_version": named_version_only})
--- a/mslib/msui/mscolab_version_history.py
+++ b/mslib/msui/mscolab_version_history.py
@@ -29,8 +29,7 @@ from datetime import datetime
 import json
 
 import requests
-from urllib.parse import urljoin
-from werkzeug.urls import url_encode
+from urllib.parse import urljoin, urlencode
 
 from mslib.utils.verify_user_token import verify_user_token
 from mslib.msui.flighttrack import WaypointsTableModel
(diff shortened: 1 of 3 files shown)


matrss:
I don't get why this "AttributeError: module 'PyQt5.QtCore' has no attribute 'Slot'" is happening here, but not on develop. I did not change anything remotely related to that and `PyQt5.QtCore` simply does not have this attribute. See: https://stackoverflow.com/questions/50459672/attributeerror-module-pyqt5-qtcore-has-no-attribute-slot

Both CI runs report the same pyqt version though...

matrss:
OK, so it looks like the `Slot` attribute must be coming from
```
lib/python3.10/site-packages/matplotlib/backends/qt_compat.py
94:        QtCore.Slot = QtCore.pyqtSlot
106:        QtCore.Slot = QtCore.pyqtSlot
```
in the CI run on develop. I am not sure why that is used in develop but not in this PR pipeline though.

Anyway, I think we should change all occurrences of `QtCore.Slot` to `QtCore.pyqtSlot`.

matrss:
```
>>> werkzeug.urls.url_encode({"test": None})
''
>>> urllib.parse.urlencode({"test": None})
'test=None'
```

That's the difference causing our tests to fail.

mslib/msui/mscolab_version_history.py
```diff
@@ -134,10 +133,9 @@ def load_all_changes(self):
                 "token": self.token,
                 "op_id": self.op_id
             }
-            named_version_only = None
```
ReimarBauer, line 137: this was one of these abuse of None for boolean. The minimum change would be here to set it to False.

The coresponding server code should also become fixed, e.g. file_manager.get_all_changes

ReimarBauer (changes requested):
see comment. 

related to https://github.com/Open-MSS/MSS/issues/1614

ReimarBauer:
> ```
> >>> werkzeug.urls.url_encode({"test": None})
> ''
> >>> urllib.parse.urlencode({"test": None})
> 'test=None'
> ```
> 
> That's the difference causing our tests to fail.

Good catch !

mslib/msui/mscolab_version_history.py
matrss: Done

ReimarBauer (approved):
Thx

ReimarBauer:
@joernu76 when this is merged we need to patch our server

PR #1916, opened 2023-08-21
Made MSColab timeout configurable.

### LINKED ISSUE #1915: TimeOut errors on bad internet connection/strange firewall
We cannot connect to the MSColab Server on a guest WiFi with very long connection times (>2s), which may or may not be caused by a deep-inspecting firewall.

The timeout should be configurable to provide a work around.
Also, the current timeouts are defined as "magic number" in a wide range of places and the values should be collected to a single instance (as mentioned above, defined in the json configuration).

joernu76:
Fix #1915


mslib/utils/config.py
```diff
@@ -239,6 +242,7 @@ class MSUIDefaultConfig:
         'num_interpolation_points',
         'new_flighttrack_flightlevel',
         'MSCOLAB_category',
+        'MSCOLAB_timeout',
```
ReimarBauer: This should be added to the list_option_structure, because of two values.

We need to review this at all. But I think it matches better there. 

e.g. file_picker_default is only a checkbox boolean, data_dir is one value.

We don't destinguish from the UI between list and tuple, 2 values -> list

ReimarBauer (changes requested):
see comment, the value is two elements, so list_options fits better

mslib/utils/config.py
joernu76: with the list-options it might complain about two same values?

ReimarBauer: ok, we keep it as it is.

this needs a kind of test later to not doing something unexpected

ReimarBauer:
I agree with this but want to understand what is the difference to the other option first.

ReimarBauer (approved):

PR #1917, opened 2023-08-21
Change QtCore.{Slot,Signal} to QtCore.pyqt{Slot,Signal}

### LINKED ISSUE #1912: Change usage of QtCore.{Slot,Signal} to QtCore.pyqt{Slot,Signal}
`QtCore.Slot` was used in the past, but is not present in pyqt5 as of now. It might come into scope from matplotlib compatibility magic (see https://github.com/Open-MSS/MSS/pull/1911#issuecomment-1684933849), but we should not rely on that.

matrss:
Fixes #1912.


ReimarBauer (approved):
thx

PR #1939, opened 2023-08-24
Added Return as shortcut to connect/login buttons.

### LINKED ISSUE #1937: Logging into MSColab should require less mouse-clicks
When logging into MSColab, typically all information is prefilled. The connect/login boxes should have a simple keybinding such as pressing Return.

joernu76:
Fix #1937


ReimarBauer (commented):
on the initial msui screen "return" does nothing for me, this is intended?

ReimarBauer:
Is there a way to set focus on the buttons? with the tab key this is done and looks like this. The red border around the button. Because this is a common feedback to the user, I would prefer this. The execution happens then with the spacebar.

selected

ReimarBauer (changes requested):
we should use setFocus() of QPushButtons and use the space bar to execute it. 
with the border around the button the user sees what is executed by the keyboard.

joernu76:
On the initial MSUI, I was unsure if "Return" should automatically connect to MSColab. The *next* two windows can be triggered by Return.
One could add Return also to the MainWindow. for three consecutive Returns to login.

I am not sure who tab/spacing relates to keyboard-shortcuts that we have for several functions/entries. E.g. "retrieve" in wms_control also has Return as keyboard shortcut.

ReimarBauer:
I agree not to use space. we can use the focus to signal its function.

We have this focus already on many buttons which react on return.

example_cancel

example_send

joernu76:
OK, instead of having a keyboard shortcut, you propose to set the focus to a sensible button. I am not sure we have used that somewhere. Most people do not use tab/space when they have a mouse. Having a shortcut is the "typical" power-user feature, I think.

ReimarBauer:
it is more both, enable return but "highlight" by setFocus the button which act.

similiar as in picture one. Return/Enter means "cancel", and in the second Return/Enter means "send"

joernu76:
The highlight typically refers to an act of "space" not "return". Also, typing in data, etc. will change the focus? In particular, one cannt enter text in chat, while the send button is focused...?

ReimarBauer:
maybe look also here how return/enter are introduced. Two different keys, but should behave identical.

https://github.com/Open-MSS/MSS/blob/develop/mslib/msui/mscolab_chat.py#L53

joernu76:
Alternatively, we could focus on the server/login text-input-fields and also connect a "return" in there with a connect/login attempt...?

ReimarBauer:
we can do this too,

but here the send example

I always can do A [enter], B [enter] and so on

example

ReimarBauer:
I think we should have a clean behaviour. Differences are typicallly difficult to describe.

Give "value" and "enter" is some common task

I am agreeing with any solution which is then refactored on all the others. Just not a different behaviour.

ReimarBauer:
@joernu76 setFocus for login works, for the ConnectBtn also Disconnect can react on return. I don't know which one wins. 
Not in all testcases I was able to disable setAutoDefault and setFocus()  for "Disconnect" some tests showed always Diconnect focused too. I skipped that.

joernu76:
Ah, the disconnect button should not be the same as the connect button, for a variety of reasons.
I will modify it such as that we have two buttons, one of which is hidden, context-dependent.
I'll play around with the focus, but I do not understand this as most Windows-users I know are not even aware of keyboard based UI interaction.

ReimarBauer:
the button key is your "return", I just want to have the "marking" color ontop of the button as now here.

red_shaded

this is then identical to the other "return" dialogs.

ReimarBauer:
There is one small problem left over

when we are a new user or want to use a different account, then Disconnect is highligthed but shouldn't

disconnect_gets_highlighted

also clicking in the password field, when there is the password added changes the button logic to disconnect.

ReimarBauer (approved):
We can also work on this little (complicated) problem in a follow up.

joernu76:
We need to talk about this, because a "return"-shortcut is not the same as the focus, which defines what "space" is triggering...?
But I agree that we should settle this in a follow-up MR.

ReimarBauer:
Now you can use "space" and you can use "return" on buttons.

I prefer that it is marked what gets done, simlilar to yes/no or the other buttons.

joernu76:
after fixing everything up i have a slight idea about what you mean. The autoDefault is yet a bit beyond me, but setFocus makes sense, but was used a bit too freely in this MR.

PR #1941, opened 2023-08-25
Remove redundant calls to "operations"

### LINKED ISSUE #1940: MSColab login slow due to redundant server requests
When logging in, the client requests the "operations" URL several times. Once should be enough.

joernu76:
Pass on previous results to subsequent functions, disable unnecessary request triggers.

Fix #1940

### DIFF
--- a/mslib/msui/mscolab.py
+++ b/mslib/msui/mscolab.py
@@ -508,7 +508,7 @@ class MSUIMscolab(QtCore.QObject):
         # User email
         self.email = None
         # Display all categories by default
-        self.selected_category = "ANY"
+        self.selected_category = "*ANY*"
         # Gravatar image path
         self.gravatar = None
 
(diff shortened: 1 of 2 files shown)


joernu76:
This seems to work nicely by removing 5-6 redundant requests. Still pretty slow (with our current distance), but better.
I am not very familiar with that part of the code, please review thoroughly.

ReimarBauer:
please verify with a VPN if you are not throttled again.

ReimarBauer (changes requested):
ANY is not a name of a category, it should show ALL existing operations. its meaning is "*". 
The sign only would be on the button invisible.

ReimarBauer:
I have also a category "example" I assume that the OP is archived. We maybe should exclude archived operations for the category scan. I guess this was same behaviour before this change.

joernu76:
I am *not* using the ridiculously slow WLAN, but our much faster house Wifi and/or Georgs Labor-LAN and FZJ-VPN.
A request takes about 1s (compared to >3s for the strange hangar-guest-Wifi).
For the given amount of categories, this Pull request removes ca. 6 redundant "operations" request, removing about 6s from the >14s login-time.

This change introduces some added complexity, but in particular the automatically triggered operation-calls when populating the category list should go in any way (the clear and each add call as well as the final select, each trigger an operations-request).

ReimarBauer:
We need to remove ANY or make it list all operations, or find another way to list all existing.

ReimarBauer (approved):
added ANY functionality and a test for that

Issue #1942, opened 2023-08-25
DeprecationWarnings from dependencies


matrss:
From basemap:
```
../../../opt/conda/envs/mss-develop-env/lib/python3.10/site-packages/mpl_toolkits/basemap/__init__.py:33
  /opt/conda/envs/mss-develop-env/lib/python3.10/site-packages/mpl_toolkits/basemap/__init__.py:33: DeprecationWarning: distutils Version classes are deprecated. Use packaging.version instead.
    _matplotlib_version = LooseVersion(_matplotlib_version)

../../../opt/conda/envs/mss-develop-env/lib/python3.10/site-packages/mpl_toolkits/basemap/__init__.py:34
  /opt/conda/envs/mss-develop-env/lib/python3.10/site-packages/mpl_toolkits/basemap/__init__.py:34: DeprecationWarning: distutils Version classes are deprecated. Use packaging.version instead.
    _mpl_required_version = LooseVersion('0.98')
```

This is addressed upstream, but not yet released: https://github.com/matplotlib/basemap/commit/8c2b3916bb549e921a73e486fd1a08932280c82a

Fix:
- [x] wait for the next release of basemap and update

From Flask-SocketIO:
```
../../../opt/conda/envs/mss-develop-env/lib/python3.10/site-packages/flask_socketio/__init__.py:19
  /opt/conda/envs/mss-develop-env/lib/python3.10/site-packages/flask_socketio/__init__.py:19: DeprecationWarning: '_request_ctx_stack' is deprecated and will be removed in Flask 2.4.
    from flask import _request_ctx_stack, has_request_context, json as flask_json
```

This is addressed upstream since v5.3.0: https://github.com/miguelgrinberg/Flask-SocketIO/commit/6f270eef955e28289285c03138187511bc148547

Fix:
- [x] update Flask-SocketIO dependency

From Flask-WTF:
```
../../../opt/conda/envs/mss-develop-env/lib/python3.10/site-packages/flask_wtf/recaptcha/widgets.py:2
  /opt/conda/envs/mss-develop-env/lib/python3.10/site-packages/flask_wtf/recaptcha/widgets.py:2: DeprecationWarning: 'flask.Markup' is deprecated and will be removed in Flask 2.4. Import 'markupsafe.Markup' instead.
    from flask import Markup
```

Addressed upstream, but not yet released: https://github.com/wtforms/flask-wtf/pull/565/commits/603318bfb31958572ff250fa29fc4e4678d438c8

[...]

Issue #1356, opened 2022-02-15
on new operation  "/ is not a connected namespace" shown


ReimarBauer:
operation is created, seems only the communication fails.

```
CRITICAL: MSS Version: 6.0.2.
CRITICAL: Python Version: 3.9.9 | packaged by conda-forge | (main, Dec 20 2021, 02:41:03) 
[GCC 9.4.0]
CRITICAL: Platform: Linux-5.4.0-99-generic-x86_64-with-glibc2.27 (('64bit', 'ELF'))
CRITICAL: Fatal error: Traceback (most recent call last):
  File "/home/reimar/miniconda3/envs/mssenv/lib/python3.9/site-packages/mslib/msui/mscolab.py", line 742, in add_operation
    self.conn.handle_new_operation(op_id)
  File "/home/reimar/miniconda3/envs/mssenv/lib/python3.9/site-packages/mslib/msui/socket_control.py", line 144, in handle_new_operation
    self.sio.emit('add-user-to-operation', {
  File "/home/reimar/miniconda3/envs/mssenv/lib/python3.9/site-packages/socketio/client.py", line 393, in emit
    raise exceptions.BadNamespaceError(
socketio.exceptions.BadNamespaceError: / is not a connected namespace.

packet queue is empty, aborting
ERROR: packet queue is empty, aborting
packet queue is empty, aborting
ERROR: packet queue is empty, aborting
```

ReimarBauer:
Hmm, error does not happen in my development environment with develop or stable.

ReimarBauer:
This happen with flask-socketio    5.1.1  a downgrade to 5.1.0 solves this.

mamba install flask-socketio==5.1.0

PR #1943, opened 2023-08-25
Remove upper bound on flask-socketio version

matrss:
**Purpose of PR?**:

Part of #1942

Not quite sure. The reason this version was pinned is #1356, but I cannot reproduce this issue even with flask-socketio=5.1.1.

test suite passes

**Checklist:**
- [ ] Commit has unit tests

### DIFF
--- a/localbuild/meta.yaml
+++ b/localbuild/meta.yaml
@@ -66,7 +66,7 @@ requirements:
     - flask-mail
     - flask-migrate
     - werkzeug >=2.2.3
-    - flask-socketio =5.1.0
+    - flask-socketio >=5.1.0
     - flask-sqlalchemy >=3.0.0
     - flask-cors
     - passlib


localbuild/meta.yaml
```diff
@@ -66,7 +66,7 @@ requirements:
     - flask-mail
     - flask-migrate
     - werkzeug >=2.2.3
-    - flask-socketio =5.1.0
+    - flask-socketio
```
ReimarBauer: may be better >=5.1.0

the recent version is 5.3.5

You could try to make a local build, boa is currently the fastest method.

e.g.
```
boa convert meta.yaml > recipe.yaml
boa build .
```

```

ReimarBauer (changes requested):
make the pinning >=5.1.0

we had in the past >5

localbuild/meta.yaml
matrss: Done. A local boa build looks good as well.

ReimarBauer (approved):

Issue #1579, opened 2022-10-25
fixation for matplotlib < 3.6 needed


ReimarBauer:
basemap is not compatible with newest matplotlib. basemap will have soon a new release.
An update of basemap will then fixate matplotlib. We should do this in the next stable of MSS too,

PR #1956, opened 2023-08-28
Replace matplotlib grid parameter 'b' with 'visible'

### LINKED ISSUE #1955: MatplotlibDeprecationWarning from usage of parameter 'b' in grid()
```
  /home/icg149/Projects/MSS/mslib/mswms/mpl_vsec.py:135: MatplotlibDeprecationWarning: The 'b' parameter of grid() has been renamed 'visible' since Matplotlib 3.5; support for the old name will be dropped two minor releases later.
    ax.grid(b=True)
```

Fixing this should just be changing the keyword parameter name.

matrss:
**Purpose of PR?**:

Fixes #1955

No

test suite passes

**Checklist:**
- [X] Bug fix. Fixes #1955
- [ ] Commit has unit tests

### DIFF
--- a/localbuild/meta.yaml
+++ b/localbuild/meta.yaml
@@ -57,7 +57,7 @@ requirements:
     - unicodecsv
     - fs_filepicker
     - cftime >=1.0.1
-    - matplotlib >=3.5.3,<3.6
+    - matplotlib >=3.5.3
     - itsdangerous
     - pyjwt
     - flask >=2.3.2
--- a/localbuild/meta.yaml
+++ b/localbuild/meta.yaml
@@ -57,7 +57,7 @@ requirements:
     - unicodecsv
     - fs_filepicker
     - cftime >=1.0.1
-    - matplotlib >=3.3.2,<3.6
+    - matplotlib >=3.5.3,<3.6
     - itsdangerous
     - pyjwt
     - flask >=2.3.2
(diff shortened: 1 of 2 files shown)


ReimarBauer (changes requested):
pin in the meta.yaml  matplotlib >=3.7

matrss:
We currently have an upper bound of <3.6 on matplotlib, due to #1579. Can this be lifted?

If your concern is that older matplotlib versions would not work with this patch: since we are currently on 3.5.3 as the latest version <3.6 we obviously do not need >=3.7 for this patch to work. But we should at least raise our lower bound to 3.5.3 I think.

ReimarBauer:
current limitation of basemap is < 3.8
https://github.com/conda-forge/basemap-feedstock/blob/main/recipe/meta.yaml

ReimarBauer:
we should have the lowest bound for this patch. Because we will uses basemap this will define what of matplotlib we will use from the upper.

matrss:
Since basemap already specifies the upper bound to be <3.8 this propagates to mss as well, right? That would mean we don't need to specify it again. The reason we added <3.6 here was because there was not yet a release of basemap with the upper bound set on their side, but that reason does not apply anymore.

The current state of this PR with matplotlib>=3.5.3 should be good to go then.

ReimarBauer (approved):

Issue #1960, opened 2023-08-28
import of files into Operations sometimes fails


joernu76:
After double clicking a filename, the dialog closes, a success message appears, the changed flight plan is displayed in Views, but immediately jumps back to the old one.

ReimarBauer:
this problem happens also with the new file and import dialog.

ReimarBauer:
often after the single or double click just the wrong widget is on top. The important one is hidden behind.
![wrong_widget_inf_front](https://github.com/Open-MSS/MSS/assets/278728/4b9a8419-6a8b-45b0-9071-da8751563702)

ReimarBauer:
the initial description is different to what I get.

Issue #1591, opened 2022-11-04
QFileDialog don't start inside of pycharm


ReimarBauer:
I recognized that in the 
options=QtWidgets.QFileDialog.DontUseNativeDialog as in

` filename = QtWidgets.QFileDialog.getOpenFileName(*args, options=QtWidgets.QFileDialog.DontUseNativeDialog)
`
 let me use the dialog inside the PyCharm Environment.

A good hint was on https://stackoverflow.com/questions/59775385/weird-interaction-between-pycharm-and-pyqt-qfiledialog

I am not sure if we can default to this option?

ReimarBauer:
reverted the merge, in None Pycharm Environments this does make a problem
```
-OPTIONS = None
-if os.getenv("PYCHARM_HOSTED") is not None:
-    OPTIONS = QtWidgets.QFileDialog.DontUseNativeDialog
-
-
 def get_open_filename_qt(*args):
-
-    filename = QtWidgets.QFileDialog.getOpenFileName(*args, options=OPTIONS)
+    filename = QtWidgets.QFileDialog.getOpenFileName(*args)
```

PR #1962, opened 2023-08-28
Some tidying up of handling imports in MSCOlab

joernu76:
The connect seems to be wrong, a disconnect seems in order. It might be better to not create a new waypoint_model, but change the content of the old one?
Either way, with these streamlining, the error was not reproducible anymore. Since it was sporadic before, this might be by chance.

See #1960

**Purpose of PR?**:

Fixes #

**Checklist:**
- [ ] Commit has unit tests

### DIFF
--- a/mslib/msui/mscolab.py
+++ b/mslib/msui/mscolab.py
@@ -1829,6 +1829,7 @@ class MSUIMscolab(QtCore.QObject):
         self.reload_view_windows()
 
     def handle_waypoints_changed(self):
+        logging.debug("handle_waypoints_changed")
         if verify_user_token(self.mscolab_server_url, self.token):
             if self.ui.workLocallyCheckbox.isChecked():
                 self.waypoints_model.save_to_ftml(self.local_ftml_file)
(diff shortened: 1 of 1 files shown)


ReimarBauer:
Hmm, I assume you mean this,

same_problem

so it is reproduced and still open

ReimarBauer:
checking without IDE, this might behave different

ReimarBauer:
I assume that some of my other running apps on my desktop blocks this by something similiar to this 
https://github.com/Open-MSS/MSS/issues/1591

ReimarBauer:
I thought first it still exists

On doubleclick, the QT file browser selects a file and closes. Then the main view is on top and the selected import dialog is behind this UI **hidden**

wrong_widget_inf_front

We have to put the widget in front

ReimarBauer (changes requested):
please have a look if you could easily bring the import dialog always in front after a file was selected.

It is not easy to recognize.

joernu76:
Mmmh. This cannot happen on *my* WMS because it is tiling. Thus I cannot test this easily, at least today.
However, I observed the waypoints changing and reverting, which I attribute to some strange timings and the exchange of the waypoint model (I believe it'd be better to change the waypoints within, but this is a bigger change I currently do not have time for).
This merge request also doesn't close the issue, as I think it makes things safer, but does not necessarily fixes (all) problems.

ReimarBauer (approved):
needs a review before release, but good enough for now

PR #1967, opened 2023-08-30
Fix messagebox warnings

### LINKED ISSUE #1957: What to do with "UserWarning: An unhandled message box popped up during your test!"?
There are many of these warnings in the CI logs basically spamming the output and drowning out other more interesting warnings.

These warnings are originating from https://github.com/Open-MSS/MSS/blob/1327ede1dbe3f4eb26bf3889934fa76c74fb428b/conftest.py#L227.

I am not sure why this warning was introduced in the first place, but in my opinion there are two options to handle this better:
1. If it is an issue that there are unhandled message boxes, then this should fail the respective test.
2. If it is not an issue, then we shouldn't flood the output with this warning.

Since I am not too familiar with the internals of the test suite I'd like to discuss how to make this better.

### LINKED ISSUE #1973: Checkbox for renaming operation-popup does not work
After renaming an operation a popup appears informing the user about the successful renaming. This popup has a checkbox about having this same popup in the future.
Unchecking this checkbox seems to have not effect.

matrss:
**Purpose of PR?**:

Fixes #1957. This turns the warning when there are unhandled message boxes left into actual errors. I also started to fix those errors, but that is still WIP.

Fixes: #1973 

No

**Checklist:**
- [X] Bug fix. Fixes #1957
- [ ] Commit has unit tests

### DIFF
--- a/mslib/msui/mscolab.py
+++ b/mslib/msui/mscolab.py
@@ -872,8 +872,11 @@ class MSUIMscolab(QtCore.QObject):
             self.logout()
         else:
             if r.text == "True":
-                self.error_dialog = QtWidgets.QErrorMessage()
-                self.error_dialog.showMessage('Your operation was created successfully')
+                QtWidgets.QMessageBox.information(
+                    self.ui,
+                    "Creation successful",
+                    "Your operation was created successfully.",
+                )
                 op_id = self.get_recent_op_id()
                 self.new_op_id = op_id
                 self.conn.handle_new_operation(op_id)
@@ -1130,8 +1133,11 @@ class MSUIMscolab(QtCore.QObject):
                 if r.text == "True":
                     self.active_operation_category = entered_operation_category
                     self.reload_operation_list()
-                    self.error_dialog = QtWidgets.QErrorMessage()
-                    self.error_dialog.showMessage("Description is updated successfully.")
+                    QtWidgets.QMessageBox.information(
+                        self.ui,
+                        "Update successful",
+                        "Category is updated successfully.",
+                    )
                 else:
                     show_popup(self.ui, "Error", "Your Connection is expired. New Login required!")
                     self.logout()
@@ -1166,8 +1172,11 @@ class MSUIMscolab(QtCore.QObject):
                     self.set_operation_desc_label(entered_operation_desc)
 
                     self.reload_operation_list()
-                    self.error_dialog = QtWidgets.QErrorMessage()
-                    self.error_dialog.showMessage("Description is updated successfully.")
+                    QtWidgets.QMessageBox.information(
+                        self.ui,
+                        "Update successful",
+                        "Description is updated successfully.",
+                    )
                 else:
                     show_popup(self.ui, "Error", "Your Connection is expired. New Login required!")
                     self.logout()
@@ -1207,8 +1216,11 @@ class MSUIMscolab(QtCore.QObject):
                     # Update other user's operation list
                     self.conn.signal_operation_list_updated.connect(self.reload_operation_list)
 
-                    self.error_dialog = QtWidgets.QErrorMessage()
-                    self.error_dialog.showMessage("Operation is renamed successfully.")
+                    QtWidgets.QMessageBox.information(
+                        self.ui,
+                        "Rename successful",
+                        "Operation is renamed successfully.",
+                    )
                 else:
                     show_popup(self.ui, "Error", "Your Connection is expired. New Login required!")
                     self.logout()
--- a/mslib/msui/mscolab.py
+++ b/mslib/msui/mscolab.py
@@ -1986,11 +1986,9 @@ class MSUIMscolab(QtCore.QObject):
 
         self.operation_archive_browser.hide()
 
-        # Don't try to activate local flighttrack while testing
-        if "pytest" not in sys.modules:
-            # activate first local flighttrack after logging out
-            self.ui.listFlightTracks.setCurrentRow(0)
-            self.ui.activate_selected_flight_track()
+        # activate first local flighttrack after logging out
+        self.ui.listFlightTracks.setCurrentRow(0)
+        self.ui.activate_selected_flight_track()
 
 
 class MscolabMergeWaypointsDialog(QtWidgets.QDialog, merge_wp_ui.Ui_MergeWaypointsDialog):
--- a/mslib/msui/mscolab.py
+++ b/mslib/msui/mscolab.py
@@ -1928,6 +1928,12 @@ class MSUIMscolab(QtCore.QObject):
             return
         self.ui.local_active = True
         self.ui.menu_handler()
+
+        # disconnect socket
+        if self.conn is not None:
+            self.conn.disconnect()
+            self.conn = None
+
         # close all hanging window
         self.close_external_windows()
         self.hide_operation_options()
@@ -1960,10 +1966,6 @@ class MSUIMscolab(QtCore.QObject):
         self.ui.activeOperationDesc.setText(self.ui.tr("Select Operation to View Description."))
         # set usernameLabel back to default
         self.ui.usernameLabel.setText("User")
-        # disconnect socket
-        if self.conn is not None:
-            self.conn.disconnect()
-            self.conn = None
         # Turn off work locally toggle
         self.ui.workLocallyCheckbox.blockSignals(True)
         self.ui.workLocallyCheckbox.setChecked(False)
--- a/mslib/utils/airdata.py
+++ b/mslib/utils/airdata.py
@@ -252,7 +252,7 @@ def get_airspaces(countries=None):
                                             for data in airspace_data["polygon"].split(",")]
                 _airspaces.append(airspace_data)
                 _airspaces_mtime[file] = os.path.getmtime(os.path.join(OSDIR, "downloads", "aip", file))
-            else:
-                QtWidgets.QMessageBox.information(None, "No Airspaces data in file:", f"{file}")
+        else:
+            QtWidgets.QMessageBox.information(None, "No Airspaces data in file:", f"{file}")
 
     return _airspaces
--- a/tests/_test_msui/test_mscolab.py
+++ b/tests/_test_msui/test_mscolab.py
@@ -470,32 +470,42 @@ class Test_Mscolab(object):
         assert item.operation_path == "example"
         assert item.access_level == "creator"
 
-    @mock.patch("PyQt5.QtWidgets.QErrorMessage")
-    def test_add_operation(self, mockbox):
+    def test_add_operation(self):
         self._connect_to_mscolab()
         modify_config_file({"MSS_auth": {self.url: "something@something.org"}})
         self._create_user("something", "something@something.org", "something")
         assert self.window.usernameLabel.text() == 'something'
         assert self.window.connectBtn.isVisible() is False
-        self._create_operation("Alpha", "Description Alpha")
-        assert mockbox.return_value.showMessage.call_count == 1
-        with mock.patch("PyQt5.QtWidgets.QLineEdit.text", return_value=None):
+        with mock.patch("PyQt5.QtWidgets.QMessageBox.information", return_value=QtWidgets.QMessageBox.Ok) as m:
+            self._create_operation("Alpha", "Description Alpha")
+            m.assert_called_once_with(
+                self.window,
+                "Creation successful",
+                "Your operation was created successfully.",
+            )
+        with (mock.patch("PyQt5.QtWidgets.QLineEdit.text", return_value=None),
+              mock.patch("PyQt5.QtWidgets.QErrorMessage.showMessage") as m):
             self._create_operation("Alpha2", "Description Alpha")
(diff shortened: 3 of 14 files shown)


matrss:
Started out with 40 errors, down to 20 now.

ReimarBauer:
@matrss maybe merge the refactor_tests_and_setup into yours.

In the mscolab tests and mscolab are a way more broken hidden by the mocking in conftest.

tests/_test_msui/test_mscolab.py
```diff
@@ -132,9 +133,31 @@ def test_login(self):
...
+        assert self.main_window.connectBtn.isVisible() is False
+        assert self.main_window.mscolab.connect_window is None
+        assert self.main_window.local_active is True
+        # test operation listing visibility
+        assert self.main_window.listOperationsMSC.model().rowCount() == 1
+
     def test_logout_action_trigger(self):
         # Login
         self._connect_to_mscolab()
+        modify_config_file({"MSS_auth": {self.url: self.userdata[0]}})
```
ReimarBauer, line 161: less mocking :)

some tests have more than one QMessageBox and we can't mock them differently

matrss:
There is 2 errors left to go. One of those is an unexpected None value, but it happens in the teardown and I did not yet understand why. The other is this `AttributeError: 'NoneType' object has no attribute 'flighttrack_model'`. I tried cherry-picking 35a0ebacad8c7da777792db5011830d1e0b4087b, but that just breaks more tests.

ReimarBauer:
> There is 2 errors left to go. One of those is an unexpected None value, but it happens in the teardown and I did not yet understand why. The other is this `AttributeError: 'NoneType' object has no attribute 'flighttrack_model'`. I tried cherry-picking [35a0eba](https://github.com/Open-MSS/MSS/commit/35a0ebacad8c7da777792db5011830d1e0b4087b), but that just breaks more tests.

Likly the "pytest" check has masked several other incomplete setup of msui.MSUIMainWindow(mscolab_data_dir=mscolab_settings.MSCOLAB_DATA_DIR)

The self.window.create_new_flight_track() was missing in any mscolab ui test.

The NoneType comes likly from this missing switch to local. Because we need always a waypoint table.

https://github.com/Open-MSS/MSS/pull/1968/files#diff-8bb1b53508f9fe252a80649575f92b7aa4756be8dda54c217a7c1d9063d6b0c5R1059

matrss:
OK, looks like it is just one issue left: for some reason in the teardown code of `Test_MscolabAdminWindow`  after `self.window.mscolab.logout()`, which unsets `self.window.mscolab.mscolab_server_url`, there is another call to `render_new_permission` triggered by `QtWidgets.QApplication.processEvents()`. This method then complains about `mscolab_server_url` being None. This issue fails between 0 and 3 tests on my system.

ReimarBauer:
I stopped after some tests completed, the image is about 30% of _test_msui

the main UI is not closed (hide) and at  least 2 rename boxes are kept.

msui_test

the Ui can get removed from the screen by an addition to the teardown https://github.com/Open-MSS/MSS/pull/1968/files#diff-d112e799e3e5bb46f90f69735aa878cb4a2d0cb453816f56edafca3935236fc4R88

One of the "descriptions is updated.." is a c/p bug currently description and category update have the same message,

https://github.com/Open-MSS/MSS/pull/1968/files#diff-8bb1b53508f9fe252a80649575f92b7aa4756be8dda54c217a7c1d9063d6b0c5R1137

these boxes are "information" not "question"

Maybe take the unhandled, I assume three of these boxes, and the rest can be done later.

matrss:
Right, I think these windows remain because I removed the code that just closes all remaining widgets from the fixture in conftest. Ideally, I think this fixture should look like this:
```python
@pytest.fixture(autouse=True)
def fail_if_open_widgets_left():
    """Fail a test if there are any Qt widgets left open at the end
    """
    yield
    widgets = set(QtWidgets.QApplication.topLevelWindows() + QtWidgets.QApplication.topLevelWidgets())
    assert len(widgets) == 0, "There are Qt widgets left open at the end of the test!"
```
but there are many more widgets and windows in that and I think some of those are memory leaks in mss itself (e.g. there is a hidden "Browse archived operations" window that is not closed properly with the main window when I just run `test_login`).

These were previously not catched by the mocking of the message boxes and then just silently closed.

matrss:
> these boxes are "information" not "question"

These boxes are `QErrorMessage`s instead of any `QMessageBox`, that is why they are not currently failing the tests.

ReimarBauer:
> > these boxes are "information" not "question"
> 
> These boxes are `QErrorMessage`s instead of any `QMessageBox`, that is why they are not currently failing the tests.

I think that is the wrong type QErrorMessage for the kind of information.

ReimarBauer:
> Right, I think these windows remain because I removed the code that just closes all remaining widgets from the fixture in conftest. Ideally, I think this fixture should look like this:
> 
> ```python
> @pytest.fixture(autouse=True)
> def fail_if_open_widgets_left():
>     """Fail a test if there are any Qt widgets left open at the end
>     """
>     yield
>     widgets = set(QtWidgets.QApplication.topLevelWindows() + QtWidgets.QApplication.topLevelWidgets())
>     assert len(widgets) == 0, "There are Qt widgets left open at the end of the test!"
> ```
> 
> but there are many more widgets and windows in that and I think some of those are memory leaks in mss itself (e.g. there is a hidden "Browse archived operations" window that is not closed properly with the main window when I just run `test_login`).
> 
> These were previously not catched by the mocking of the message boxes and then just silently closed.

I'm ok with such fixtures, when we don't see these problems we likly won't work on those. 

The teardown of many of the tests got improved in the other branch. At least there remains nothing visible. We have to verify with the monitor in more detailed.

matrss:
> I'm ok with such fixtures, when we don't see these problems we likly won't work on those.

I agree, but this fixture would lead to > 300 failed tests (basically all that involve the UI?). I think that is to much to handle in this one PR.

> The teardown of many of the tests got improved in the other branch. At least there remains nothing visible.

Nothing visible is not enough, unfortunately. In the `test_login` test case for example there are 16 widgets left, but hidden, that did not get removed by the teardown code.

I think I will reintroduce the closing code to the fixture here for now and we can tackle this in a follow-up PR.

matrss:
The remaining issue is this: on teardown of `test_add_permissions` there is a call to `self.window.mscolab.logout()`. This method first sets `self.mscolab_server_url = None` and then at a later point disconnects the socketio socket: `self.conn.disconnect()`. This disconnect queues up a qt event which makes the application call `render_new_permission` the next time `processEvents()` is called. But at that point the logout already unset the mscolab server url, which leads to
```
  File "/__w/MSS/MSS/mslib/msui/mscolab.py", line 1391, in render_new_permission
    r = requests.get(self.mscolab_server_url + '/user', data=data,
TypeError: unsupported operand type(s) for +: 'NoneType' and 'str'
```

Moving the disconnect call to somewhere before un-setting the url and calling `processEvents` immediately after fixes this issue.

ReimarBauer:
This one should be checked if we need a backport to stable. I've not seen it in normal usage yet.
https://github.com/Open-MSS/MSS/pull/1967/commits/0f5c937de9aca3b835e4a16e50308d9ed104cfe9

mslib/msui/mscolab.py
```diff
@@ -1928,6 +1928,13 @@ def logout(self):
             return
         self.ui.local_active = True
         self.ui.menu_handler()
+
+        # disconnect socket
+        if self.conn is not None:
+            self.conn.disconnect()
+            self.conn = None
+        QtWidgets.QApplication.processEvents()
```
ReimarBauer: Please add a comment before the processEvents() line
e.g.

ToDo review why the code gets blocked and refactor this part


The problem with this is that we not know where it hangs and this cause needs to be resolved.
In the "released users code" we should never need QtWidgets.QApplication.processEvents()


good article about qthreadpool, I don't see from our site a thread involved. The article describes the behaviour.
https://www.pythonguis.com/tutorials/multithreading-pyqt-applications-qthreadpool/

ReimarBauer (changes requested):
Please add the comment, so that we later at least for a release check for this again.

ReimarBauer:
> This one should be checked if we need a backport to stable. I've not seen it in normal usage yet. [0f5c937](https://github.com/Open-MSS/MSS/commit/0f5c937de9aca3b835e4a16e50308d9ed104cfe9)

Because of the QApplication.processEvents() we better don't add this to stable.

matrss:
> Because of the QApplication.processEvents() we better don't add this to stable.

Would that be a blocker? We have a call to that method in other places (in `msui/mscolab.py` and `msui/mpl_qtwidget.py`) too, but I have no experience with qt. Anyway, I dug a bit deeper and just moving that section up and adding the `processEvents` call is just hiding the issue, not fixing it.

The issue is a race condition caused by this:
1. the test adds users to an operation in the admin interface
2. this calls `add_selected_users`, which makes a post request to `<mscolab_server_url>/add_bulk_permissions`
3. on the server side on success the handler for that request triggers a `new-permission` signal via socketio
4. on the client side receiving this signal triggers the `signal_new_permission` pyqtSignal
5. in `MSUIMscolab` the `render_new_permission` slot is connected to the signal from 4. and therefore gets called

The problem then is that the test runs so fast that between 3. and 4. it is already in the teardown phase and has called the `logout` method, which unsets the server url. Then in 5. we see this issue where the url is None.

My fix so far just hides this in this particular test, but does not resolve the issue.

It is very unlikely to hit this issue in normal usage, so I don't think we have to look at this on stable. If you add a delay right before emitting the pyqt signal in the socketio handler, then you can recreate the issue interactively though.

I think a proper fix would be to stop the socketio handlers in `ConnectionManager` from emitting pyqt signals after socketio was disconnected, since that does not make much sense anyway. I'll take a look at how that could be done.

tests/_test_msui/test_mscolab.py
matrss: Well, we could, if those message boxes are the result of different lines of code in the test. `mock.patch` can be used as a context manager as well, not just as a decorator to the test function. Still, avoiding the message box entirely if we are not explicitly testing it is the better approach IMHO.

matrss:
I fixed it properly (I think?) now, removing the usage of `processEvents`. This should address the remaining comments.

mslib/utils/airdata.py
```diff
@@ -252,7 +252,7 @@ def get_airspaces(countries=None):
                                             for data in airspace_data["polygon"].split(",")]
                 _airspaces.append(airspace_data)
                 _airspaces_mtime[file] = os.path.getmtime(os.path.join(OSDIR, "downloads", "aip", file))
-            else:
-                QtWidgets.QMessageBox.information(None, "No Airspaces data in file:", f"{file}")
+        else:
```
ReimarBauer, line 255: this prior indenting was correct. The reason is to show to the user that the requested files has no data and is therefore not shown.

we use seldom the for/else clause,  https://book.pythontips.com/en/latest/for_-_else.html

Without this there is no messageBox shown like 
hint

This is not that comfortable for a user, but when this gets improved it maybe should not done in this PR?

tests/_test_utils/test_airdata.py
```diff
@@ -204,10 +204,12 @@ def test_get_airspaces(mockbox):
...
+@mock.patch("PyQt5.QtWidgets.QMessageBox.information")
 @mock.patch("PyQt5.QtWidgets.QMessageBox.question", return_value=QtWidgets.QMessageBox.Yes)
-def test_get_airspaces_missing_data(mockbox):
+def test_get_airspaces_missing_data(mockbox, infobox):
     """ We use a test file without the need for downloading to check handling """
     # update_airspace would only update after 30 days
     _cleanup_test_files()
     airspaces = get_airspaces(countries=["bg"])
     assert airspaces == []
+    infobox.assert_called_once_with(None, 'No Airspaces data in file:', 'bg_asp.xml')
```
ReimarBauer, line 215: once the data gets updated the test will fail.

mslib/utils/airdata.py
ReimarBauer: should have looked into the file downloaded. 

110238 Sep  1 10:36 ae_asp.xml

I agree to the change.

ReimarBauer, line 255: later we need some kind of feedback to the user, that his selection is not what he gets shown on the map.

or valid check needs to be improved. 

first we would need another change in the data

tests/_test_utils/test_airdata.py
matrss, line 215: The only possible test failure I see that shouldn't be one is if the file name of the data changes. So, the last argument could be changed to `mock.ANY`. Is that what you mean?

ReimarBauer (approved):
Thx

tests/_test_utils/test_airdata.py
ReimarBauer: I removed the message because I forgot about, that we have done a mock of the data containing. So that won't happen. 

These days I would just use a name which is not some regular file which we expect to have.

matrss: Ah OK, that makes sense.

PR #1968, opened 2023-08-30
Refactor tests and setup

### LINKED ISSUE #1957: What to do with "UserWarning: An unhandled message box popped up during your test!"?

ReimarBauer:
**Purpose of PR?**:

Fixes #1957 for the mscolab part

the PR reduces the amount of servers needed for mscolab tests. Also a bug in the decription for a category change fixed.
For develop the tests were not adjusted to the UI refactoring. The conftest has removed all QMessage boxes. 
Some errors we had were related that we don't loaded an initial flight track to the window for tests. Other Time out Blockers occure when some of the Questions are not mocked. 

**Checklist:**
- [ ] Commit has unit tests

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -96,9 +96,9 @@ jobs:
         && source /opt/conda/etc/profile.d/conda.sh \
         && source /opt/conda/etc/profile.d/mamba.sh \
         && mamba activate mss-${{ inputs.branch_name }}-env \
-        && pytest -vv -n 6 --dist loadfile --max-worker-restart 0 tests \
+        && pytest -vv -n 6 --dist loadfile --max-worker-restart 4 tests \
         || (for i in {1..5} \
-          ; do pytest -vv -n 6 --dist loadfile --max-worker-restart 4 tests --last-failed --lfnf=none \
+          ; do pytest -vv -n 6 --dist loadfile --max-worker-restart 0 tests --last-failed --lfnf=none \
           && break \
         ; done)
 
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -98,7 +98,7 @@ jobs:
         && mamba activate mss-${{ inputs.branch_name }}-env \
         && pytest -vv -n 6 --dist loadfile --max-worker-restart 0 tests \
         || (for i in {1..5} \
-          ; do pytest -vv -n 6 --dist loadfile --max-worker-restart 0 tests --last-failed --lfnf=none \
+          ; do pytest -vv -n 6 --dist loadfile --max-worker-restart 4 tests --last-failed --lfnf=none \
           && break \
         ; done)
 
--- a/conftest.py
+++ b/conftest.py
@@ -125,6 +125,12 @@ DATA_DIR = fs.path.join(ROOT_DIR, 'colabTestData')
 # mscolab data directory
 MSCOLAB_DATA_DIR = fs.path.join(DATA_DIR, 'filedata')
 
+# To enable logging set to True or pass a logger object to use. 
+SOCKETIO_LOGGER = True
+
+# To enable Engine.IO logging set to True or pass a logger object to use.
+ENGINEIO_LOGGER = True
+
 # used to generate and parse tokens
 SECRET_KEY = secrets.token_urlsafe(16)
 
--- a/conftest.py
+++ b/conftest.py
@@ -210,6 +210,7 @@ _load_module("mswms_settings", constants.SERVER_CONFIG_FILE_PATH)
 _load_module("mscolab_settings", path)
 
 
+'''
 @pytest.fixture(autouse=True)
 def close_open_windows():
     """
@@ -234,7 +235,7 @@ def close_open_windows():
         # Some objects deny permission, pass in that case
         except RuntimeError:
             pass
-
+'''
 
 @pytest.fixture(scope="session", autouse=True)
 def configure_testsetup(request):
--- a/conftest.py
+++ b/conftest.py
@@ -45,7 +45,7 @@ import tests.constants as constants
 sample_path = os.path.join(os.path.dirname(__file__), "tests", "data")
 shutil.copy(os.path.join(sample_path, "example.ftml"), constants.ROOT_DIR)
 
-
+MSCOLAB_PROCESSES = []
 class TestKeyring(keyring.backend.KeyringBackend):
     """A test keyring which always outputs the same password
     from Runtime Configuration
@@ -222,3 +222,7 @@ def configure_testsetup(request):
         VIRT_DISPLAY.stop()
     else:
         yield
+
+def pytest_sessionfinish(session, exitstatus):
+    for process in MSCOLAB_PROCESSES:
+        process.terminate()
--- a/conftest.py
+++ b/conftest.py
@@ -210,33 +210,6 @@ _load_module("mswms_settings", constants.SERVER_CONFIG_FILE_PATH)
 _load_module("mscolab_settings", path)
 
 
-'''
-@pytest.fixture(autouse=True)
-def close_open_windows():
-    """
-    Closes all windows after every test
-    """
-    # Mock every MessageBox widget in the test suite to avoid unwanted freezes on unhandled error popups etc.
-    with mock.patch("PyQt5.QtWidgets.QMessageBox.question") as q, \
-            mock.patch("PyQt5.QtWidgets.QMessageBox.information") as i, \
-            mock.patch("PyQt5.QtWidgets.QMessageBox.critical") as c, \
-            mock.patch("PyQt5.QtWidgets.QMessageBox.warning") as w:
-        yield
-        if any(box.call_count > 0 for box in [q, i, c, w]):
-            summary = "\n".join([f"PyQt5.QtWidgets.QMessageBox.{box()._extract_mock_name()}: {box.mock_calls[:-1]}"
-                                 for box in [q, i, c, w] if box.call_count > 0])
-            warnings.warn(f"An unhandled message box popped up during your test!\n{summary}")
-
-
-    # Try to close all remaining widgets after each test
-    for qobject in set(QtWidgets.QApplication.topLevelWindows() + QtWidgets.QApplication.topLevelWidgets()):
-        try:
-            qobject.destroy()
-        # Some objects deny permission, pass in that case
-        except RuntimeError:
-            pass
-'''
-
 @pytest.fixture(scope="session", autouse=True)
 def configure_testsetup(request):
     if Display is not None:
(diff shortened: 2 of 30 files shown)


ReimarBauer:
this should be synced with work of @matrss https://github.com/Open-MSS/MSS/pull/1967

conftest.py
```diff
@@ -210,32 +210,6 @@ def _load_module(module_name, path):
 _load_module("mscolab_settings", path)
 
 
-@pytest.fixture(autouse=True)
```
ReimarBauer, line 213: removing this will Error all tests with a blocking Yes/No or Ok QMessageBox by a timeout

conftest.py
```diff
@@ -210,32 +210,6 @@ def _load_module(module_name, path):
 _load_module("mscolab_settings", path)
```
ReimarBauer, line 218: Removing this will Error all tests with an unhandled QMessageBox

conftest.py
```diff
@@ -248,3 +222,7 @@ def configure_testsetup(request):
         VIRT_DISPLAY.stop()
     else:
         yield
+
```
ReimarBauer, line 231: the new approach is not to start on any test a mscolab service but once per test class. We need at the end to terminate all processes

.github/workflows/testing.yml
```diff
@@ -96,7 +96,7 @@ jobs:
         && source /opt/conda/etc/profile.d/conda.sh \
         && source /opt/conda/etc/profile.d/mamba.sh \
         && mamba activate mss-${{ inputs.branch_name }}-env \
-        && pytest -vv -n 6 --dist loadfile --max-worker-restart 0 tests \
```
ReimarBauer, line 99: currently will loose all tests assigned to a worker when that crashes

mslib/msui/mscolab.py
```diff
@@ -1056,8 +1056,11 @@ def handle_delete_operation(self):
                         self.reload_operations()
                         self.signal_operation_removed.emit(self.active_op_id)
                         logging.debug("activate local")
-                        self.ui.listFlightTracks.setCurrentRow(0)
-                        self.ui.activate_selected_flight_track()
+                        if self.ui.listFlightTracks.count() == 0:
```
ReimarBauer, line 1066: in mscolab we can remove "all" operations from the list, while we can't in the local flight path (left site of the UI)
When we do this, we have to switch to an active operation, which is done by switch_to_local

mslib/msui/mscolab.py
```diff
@@ -1986,11 +1989,9 @@ def logout(self):
 
         self.operation_archive_browser.hide()
 
-        # Don't try to activate local flighttrack while testing
-        if "pytest" not in sys.modules:
```
ReimarBauer: this is not needed when the tests and code are updated

ReimarBauer:
504 passed, 27 skipped, 208 warnings in 217.48s (0:03:37)

conftest.py
matrss: Is that really a good idea? We could run into issues where one test affects the outcome of another due to the shared mscolab instance, which could be a nightmare to debug. Also, I do not really like the implementation with this global variable here, couldn't this be a fixture sort of like this (pseudocode):
```python
@pytest.fixture()
def mscolab_server():
    server = start_mscolab_server()
    yield server
    server.shutdown()
```
?

Every test that needs a running mscolab server then should just use the fixture. That way we could even use the scope of the fixture to decide how "shared" we want the instance to be.

For what it's worth, I would prefer to get rid of test classes entirely, anyway. A yield fixture is much better suited for setup and teardown code, which leaves classes only as an organizational unit to group tests together. But modules, in turn, do that better.

tests/_test_msui/test_mscolab.py
```diff
@@ -112,15 +119,17 @@ def test_connect_success(self, mockbox, mockset):
...
         QtTest.QTest.mouseClick(self.window.connectBtn, QtCore.Qt.LeftButton)
         assert self.window.mscolab_server_url is None
         # set ui_name_winodw default
         assert self.main_window.usernameLabel.text() == 'User'
 
-    def test_login(self):
+    @mock.patch("PyQt5.QtWidgets.QMessageBox.question", return_value=QtWidgets.QMessageBox.No)
+    def test_login(self, mockquestion):
         self._connect_to_mscolab()
         self._login(self.userdata[0], self.userdata[2])
```
matrss: I disagree with the mocking here, we shouldn't just "click no" and be done with it. The issue here, why the message box pops up at all, is that in the msui config used in the test the "MSS_auth" does not have `self.userdata[0]` set as the username/email for the mscolab server we connect to here. With my approach to this: https://github.com/Open-MSS/MSS/blob/ee1c48c4710d583fdc455c060947af68da48303c/tests/_test_msui/test_mscolab.py#L123-L126 the message box disappears entirely. We could also test that the message box pops up as expected when there is this mismatch in the config, but then we would have to also add a `mockquestion.assert_called_once_with(<args>)` in every test. I think that is something we should just do once in a separate test case.

conftest.py
ReimarBauer: it is not a global variable, it is defined as module local variable. 
That's a great article about https://betterprogramming.pub/alternatives-to-using-globals-in-python-a3b2a7d5411b 

we need for xdist calls servers running on different ports, 

try it.

what do you mean by removing test classes entirely, I think setup and teardown may become complicated. But lets see.

tests/_test_msui/test_mscolab.py
matrss: Case in point: there was a message box appearing in `test_get_airspaces` saying there was no airspace data in the file. That made no sense. It led me to discover this bug: https://github.com/Open-MSS/MSS/pull/1967/files#diff-ce023988bd24cea5863ed80b6ef4058629448da2f64480cd8dc31f06c5186a57R254-R256

On the other hand, in `test_get_airspaces_missing_data` we expected exactly this message box, so there we should mock the message box but then also assert that we get what we expect: https://github.com/Open-MSS/MSS/pull/1967/files#diff-d76a934985b02e5c8cafb478f5e5221eacd80d770c69798e9370ebb2232d8f9dR206-R215

conftest.py
ReimarBauer: each test cleans up the database, and xdist is currently called on --dist loadfile 
The requirement by this is only different port per file. 

I also like to have different servers on each run, which I introduced before. 

Anyway any test needs independent from each other. So not having a test which creates data and another test which tries to vaildate something based on that.

tests/_test_msui/test_mscolab.py
matrss: Obviously this argument applies to most of the other locations where this message box is mocked here as well.

ReimarBauer: that's why I thought when you incorporate changes of this into your branch you can decide this on the merge or cherry pick, afterwards we close this PR and you can verify the other ideas.

conftest.py
matrss: > what do you mean by removing test classes entirely, I think setup and teardown may become complicated. But lets see.

The recommended way (by pytest) to have setup and teardown code is using a yield fixture. You can read more about that here: https://docs.pytest.org/en/6.2.x/fixture.html#yield-fixtures-recommended
They have an example there to demonstrate how user creation and deletion would work with fixtures, in an example testing a hypothetical email server. No class needed.

> Anyway any test needs independent from each other. So not having a test which creates data and another test which tries to vaildate something based on that.

Hypothetically, what would happen with these shared mscolab servers if we had a test for user deletion trying to delete user "UV10@uv10" in `Test_Mscolab_connect_window`, but also try to login to this user in `test_login`? My concern is that we might run into similar, but less obvious, race conditions in the future.

ReimarBauer: xdist tests and reverse order by normal tests likly finds these dependencies. 

There were a lot in the past.

matrss: > it is not a global variable, it is defined as module local variable.

In https://github.com/Open-MSS/MSS/blob/e60909c891f489a7ef1da309a63f85a547fa68bb/conftest.py#L226-L228 we terminate all processes once when we are finished with running the test suite. Isn't `MSCOLAB_PROCESSES` a global variable in the scope of the test session, then? We have one instance of this variable in which we accumulate all the running mscolab processes while executing the tests.

If I understand that right, with this PR we would also keep all the processes running until the end of the session, instead of stopping them as soon as they are not needed anymore.

ReimarBauer: > > what do you mean by removing test classes entirely, I think setup and teardown may become complicated. But lets see.
> 
> The recommended way (by pytest) to have setup and teardown code is using a yield fixture. You can read more about that here: https://docs.pytest.org/en/6.2.x/fixture.html#yield-fixtures-recommended They have an example there to demonstrate how user creation and deletion would work with fixtures, in an example testing a hypothetical email server. No class needed.
> 
> > Anyway any test needs independent from each other. So not having a test which creates data and another test which tries to vaildate something based on that.
> 
> Hypothetically, what would happen with these shared mscolab servers if we had a test for user deletion trying to delete user "UV10@uv10" in `Test_Mscolab_connect_window`, but also try to login to this user in `test_login`? My concern is that we might run into similar, but less obvious, race conditions in the future.


connect_window and 
test_login has to setup the user,
after each test per teardown the database is cleared. 
Any test has to do all the steps for testing an approach.

when the tests are paralell running by xdist, any worker creates new tmp dirs
two workers don't access the same database


We had the described problems before, that's why tests currently not based on some others.

ReimarBauer: I would prefer more functional tests, without using the UI. At least everything what can be tested without the UI should be done.

tests/_test_msui/test_mscolab.py
matrss: I don't think we should combine this larger architectural change with my, I'd argue, "less invasive" PR. I almost have my PR in a state in which I think it would be merge-able, getting rid of these warnings now and cleaning up the CI outputs, and then we could have a look at this more fundamental PR.

ReimarBauer: ok

conftest.py
matrss: I don't think that is true. I just had a test pass with this fixture removed even though it showed a message box and failed with my modified fixture.

conftest.py
matrss: > At least everything what can be tested without the UI should be done.

I have to disagree on that. We should test at the "level" at which users interact with the application as well, and that is the UI. There is no point in testing if a login works if we don't test that the actual login button works also.

Or to phrase it differently: everything that can be done with the UI cannot be tested without it.

ReimarBauer: we do UI tests already in many ways also by the tutorials, But we also have a lot helper functions which also work without the UI and these should become tested if possibile without the UI.

ReimarBauer:
won't be done

PR #1974, opened 2023-08-31
To do fixes #1818

nilupulmanodya:
**Purpose of PR**:
Fix To do fixes of previous PR https://github.com/Open-MSS/MSS/pull/1818

Fixes #
https://github.com/Open-MSS/MSS/pull/1818#discussion_r1270701658
https://github.com/Open-MSS/MSS/pull/1818#pullrequestreview-1554358384
https://github.com/Open-MSS/MSS/pull/1818#issuecomment-1658030366
https://github.com/Open-MSS/MSS/pull/1818#issuecomment-1658068466
https://github.com/Open-MSS/MSS/pull/1818#pullrequestreview-1556723540

No

yes. Fix typos of `conf_sso_test_msscolab.rst`

**Checklist:**
- [ ] Commit has unit tests


ReimarBauer:
Hi @nilupulmanodya welcome back.

nilupulmanodya:
> 
Hi @ReimarBauer..  Thanks.. It's good to be back

docs/conf_sso_test_msscolab.rst
```diff
@@ -1,4 +1,4 @@
-Configuration MSS Colab Server with Testing IDP for SSO
+Configuration MSS Colab Server with Testing IdP for SSO
 =======================================================
 Testing IDP (`mslib/idp`) is designed specifically for testing the Single Sign-On (SSO) process using PySAML2.
```
ReimarBauer, line 4: some lines below, 

`Since mscolab server’s path was set as the default path for the key and certificate, you should manually update the path of SERVER_CERT with the path of the generated .crt file for IDP, and SERVER_KEY with the path of the generated .key file for the IDP in the file MSS/mslib/idp/idp_conf.py`

Since development this is an option. But when this should be an advice for someone to use a company cert and to learn where to store these files, we should always describe "the official" way. 

look from it from the perspective of an installed package.  This section where it is located is for users.

When this is for development, it may be moved to the development document.

docs/conf_sso_test_msscolab.rst
```diff
@@ -42,7 +42,7 @@ Then you can generate files, simply by running,
 Enable USE_SAML2
 ----------------
 
-To enable saml2 based login (identity provider-based login), set `USE_SAML2 = True` in the `mslib/mscolab/conf.py` file of the MSS Colab server.
+To enable SAML2-based login (identity provider-based login), set `USE_SAML2 = True` in the `mslib/mscolab/conf.py` file of the MSS Colab server.
```
ReimarBauer, line 48: a user can't reach this.

You may want to describe the steps differently.

add to the PYTHONPATH where your mscolab_settings.py is defined
To start the process add USE_SAML2 = False

At the enable step set USE_SAML2 = True

ReimarBauer: as developer you can do e.g.  2023/nilupulmanodya/MSS$ ```export PYTHONPATH=`pwd`:/home/reimar/mscolab```

Because of the inheritance of all other vars, as developer it is enough to add USE_SAML2 = False as the only line in this file.

docs/conf_sso_test_msscolab.rst
```diff
@@ -52,15 +52,15 @@ Generate metadata files
...
 `$ python mslib/mscolab/mscolab.py sso_conf --init_sso_metadata`
 
 
 Start Identity provider
 -----------------------
 
-Once you setted certificates and metada files you can start mscolab server and local identity provider. To start local identity provider, simpy execute
+Once you set certificates and metada files you can start mscolab server and local identity provider. To start local identity provider, simply execute
 
 `$ python mslib/idp/idp.py idp_conf`
```
ReimarBauer, line 77: the script needs to become added to the entrypoints e.g. msidp = ...

https://github.com/Open-MSS/MSS/blob/GSOC2023-NilupulManodya/localbuild/meta.yaml#L15

then a user can do after installation

> msidp idp_conf

mslib/idp/idp_conf.py
```diff
@@ -207,13 +203,13 @@ def sso_dir_path(local_file):
 
 # Authentication contexts
 
-    #(r'verify?(.*)$', do_verify),
+# (r'verify?(.*)$', do_verify),
 
 CAS_SERVER = "https://cas.umu.se"
```
ReimarBauer, line 203: The site does not exists, maybe a user wants to have another address there. Can this become a config var? For what is it used?

docs/conf_sso_test_msscolab.rst
ReimarBauer: just thinking more on this, this is only for testing as long as we won't add this to use the "existing" user database.

When it is for tests a user command makes not soo much sense. And also the location as a "main" module  below mslib is not that good. 
- we should think about making it a IDP provider for the existing mscolab databae
- or move it into the _test_utils section

docs/conf_sso_test_msscolab.rst
```diff
@@ -79,14 +79,14 @@ When this is the first time you setup a mscolab server, you have to initialize t
...
 `$ python mslib/mscolab/mscolab.py start`
 
 
 Testing Single Sign-On (SSO) process
 ------------------------------------
 
 * Once you have successfully launched the server and identity provider, you can begin testing the Single Sign-On (SSO) process.
-* Start MSS PyQT application `$ python mslib/msui/msui.py`.
+* Start MSS PyQt application `$ python mslib/msui/msui.py`.
 * Login with identity provider through Qt Client application.
```
ReimarBauer, line 116: the enduser later starts msui by msui.

mslib/idp/idp.py
```diff
@@ -1104,12 +1104,12 @@ def application(environ, start_response):
...
     IDP.ticket = {}
-    
+
     current_directory = os.getcwd()
 
     LOOKUP = TemplateLookup(
-        directories=[current_directory+"/mslib/idp/templates", current_directory+"/mslib/idp/htdocs"],
-        module_directory= current_directory+"/mslib/idp/modules",
+        directories=[current_directory + "/mslib/idp/templates", current_directory + "/mslib/idp/htdocs"],
+        module_directory=current_directory + "/mslib/idp/modules",
```
ReimarBauer: please use os.path.join(...

ReimarBauer (changes requested):
see comments

docs/conf_sso_test_msscolab.rst
nilupulmanodya: updated :)

.gitignore
```diff
@@ -27,4 +26,4 @@ tutorials/recordings
 tutorials/cursor_image.png
 __pycache__/
 instance/
-mslib/idp/modules
+msidp/modules
```
matrss: Maybe I missed a discussion about this, but why is this moved to a top-level directory? If this is meant for production it should stay in mslib/, if it is only for tests it should be in tests/, I think (I strongly favor moving it to tests/, since we should leave production for well-tested and mature IdP's like keycloak).

msidp/idp.py
```diff
@@ -1,10 +1,10 @@
 # pylint: skip-file
 # -*- coding: utf-8 -*-
 """
-    mslib.idp.idp.py
-    ~~~~~~~~~~~~~~~~
+    msidp.msidp.py
```
matrss: This should be `msidp.idp.py` according to the filename.

mslib/mscolab/server.py
```diff
@@ -814,12 +802,10 @@ def idp_login():
         _, response_binding = sp_config.config.getattr("endpoints", "sp")[
             "assertion_consumer_service"
         ][0]
-        relay_state = rndstr()
         entity_id = get_idp_entity_id(selected_idp)
         _, binding, http_args = sp_config.prepare_for_negotiated_authenticate(
             entityid=entity_id,
             response_binding=response_binding,
-            relay_state=relay_state,
```
matrss, line 822: I am not sure what the implications of removing relay_state completely are. pysaml2 seems to have a default value of the empty string for that parameter. This value is internally (in pysaml2) also referred to as `session_id`, which makes it sound important. Then again, naively I would assume that if it has a default value it *should* be safe to omit.

nilupulmanodya: The `relay_state` parameter in requests is optional and serves to maintain the context of an application during SSO. It lets the SP specify where to redirect the user after SSO, such as a particular page or resource. When the user returns to the SP after authentication, the SP can use `relay_state` from the SAML response to decide where to send the user next, ensuring a seamless user experience. By default, as you said in [pysaml2, relay_state](https://github.com/IdentityPython/pysaml2/blob/cae1326ae05bf8039c5990b13ea46b51a1a92aa5/src/saml2/client.py#L113) is often an empty string, but the SP can customize it to convey specific information or instructions as needed. 

I believe that, in our case, since we are navigating from msui to mscolab, the authentication is the sole purpose. Therefore, in this scenario, there's no need to maintain the context of  mscolab during SSO. By the way, I truly appreciate your valuable insights.

.gitignore
nilupulmanodya: Moved the below to the `mslib` directory based on the discussion in the weekly meeting 🙂

.gitignore
```diff
@@ -27,4 +26,4 @@ tutorials/recordings
 tutorials/cursor_image.png
 __pycache__/
 instance/
-mslib/idp/modules
+mslib/msidp/modules
```
ReimarBauer: @nilupulmanodya what is the reason to have this there. I mean, if we have installed MSS and we would change content inside the installed dir, likly installmanagers will run into problems, because the state of installed packages changed. Likly they won't be able to completly uninstall, because dirs are not empty.

mslib/utils/mssautoplot.py
```diff
@@ -289,7 +289,7 @@ def draw(self):
 
                 xmls = wms.getmap(**kwargs)
 
-                if not type(xmls) == 'list':
+                if not isinstance(xmls, list):
```
ReimarBauer, line 292: likly this got merged in?

docs/conf_sso_test_msscolab.rst
```diff
@@ -14,63 +14,72 @@ To set up a local identity provider with the mscolab server, you'll first need t
...
-Before running the command make sure to set
-`USE_SAML2 = False` 
-In some cases, if you set `USE_SAML2 = True` without certificates, this will not execute. So, make sure to set `USE_SAML2 = False` before executing the command.
+.. note::
+    If you set `USE_SAML2 = True` without keys and certificates, this will not execute. So, make sure to set `USE_SAML2 = False` before executing the command.
 
-Then you can generate files, simply by running,
+If everything is correctly set, you can generate keys and certificates simply by running
 
 `$ python mslib/mscolab/mscolab.py sso_conf --init_sso_crts`
```
ReimarBauer, line 45: the documentation is for users, keep in mind, they will use the "entry points" in this case 

mscolab sso_conf --init_sso_crts

docs/conf_sso_test_msscolab.rst
```diff
@@ -14,63 +14,72 @@ To set up a local identity provider with the mscolab server, you'll first need t
...
+--------------------------
 
-This involves generating necessary metadata files for both the identity provider and the service provider. You can generate them by simply running the appropriate command.
+This involves generating necessary metadata files for both the identity provider and the service provider. You can generate them by simply running the below command.
 
-Before executing this, you should set `USE_SAML2=True` as described in the third step(USE_SAML2).
+.. note::
+    Before executing this, you should set `USE_SAML2=True` as described in the third step(Enable USE_SAML2).
 
 `$ python mslib/mscolab/mscolab.py sso_conf --init_sso_metadata`
```
ReimarBauer, line 71: see above


and doing so, shows

```
mscolab sso_conf --init_sso_metadata
WARNING:root:Couldn't import mscolab_auth (ImportError:'{No module named 'mscolab_auth'), creating dummy config.
/home/reimar/mambaforge/envs/alpha2_env/lib/python3.10/site-packages/mslib/mscolab/server.py:99: UserWarning: idp.xml file does not exists ! Ignore this warning when you initializeing metadata.
  warnings.warn("idp.xml file does not exists ! Ignore this warning when you initializeing metadata.")
Are you sure you executed --init_sso_crts before running this? (y/[n]):y

                    This will generate necessary metada data file for sso in mscolab through localhost idp

                    Before running this function:
                    - Ensure that USE_SAML2 is set to True.
                    - Generate the necessary keys and certificates and configure them in the .yaml
                    file for the local IDP.

                    Are you sure you set all correctly as per the documentation? (y/[n]):
                    y


generating metadata files.......
generating metadata file for the mscolab server
python: can't open file '/home/reimar/MSS_nilupul/mslib/mscolab/mscolab.py': [Errno 2] No such file or directory
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
curl: (7) Failed to connect to localhost port 8083 after 0 ms: Couldn't connect to server
Error while generating metadata file for the mscolab server: Command '['curl', 'http://localhost:8083/metadata/', '-o', '/home/reimar/colabdata/datasso/metadata_sp.xml']' returned non-zero exit status 7.
Error while handling mscolab metadata.
```

ReimarBauer: you could for example check if we are in a git repo, then it will work for devs. And when it is installed there is an entry point.

for repo check
https://github.com/Open-MSS/MSS/blob/develop/tests/constants.py#L37

docs/conf_sso_test_msscolab.rst
```diff
@@ -14,63 +14,72 @@ To set up a local identity provider with the mscolab server, you'll first need t
...
-------------------------
+6. Start the mscolab Server
+---------------------------
 
 Before Starting the mscolab server, make sure to do necessary database migrations.
 
-When this is the first time you setup a mscolab server, you have to initialize the database by
+When this is the first time you setup a mscolab server, you have to initialize the database by:
 
 `$ python mslib/mscolab/mscolab.py db --init`
```
ReimarBauer, line 93: the documentation is always meant for the Users, so they use only entrypoints.

Only the section Development ist for us :)

ReimarBauer:
steps I did.

I installed boa into my base environment.

then 
```
cd localbuild

boa convert meta.yaml > recipe.yaml
# I changed alpha to alpha2 in recipe.yaml, to get definitily a new build version name
boa build .
mamba create -n alpha2_env
mamba activate alpha2_env
mamba install mss --use-local
mamba list mss
# packages in environment at /home/reimar/mambaforge/envs/alpha2_env:
#
# Name                    Version                   Build  Channel
mss                       alpha2          py310h8e7c8fa_1000    local
```

mslib/mscolab/server.py
matrss: Right, thanks for the clarification, that makes sense. The parameter can be omitted then, in our case.

mslib/utils/mssautoplot.py
nilupulmanodya: Since we enabled flake8, I got an error when running test cases because of that, I was updated

ReimarBauer: Ok. That explains it. Keep it.
It was also some time ago fixed in stable/develop.

ReimarBauer (approved):
Thx, tried it with a build by boa :)

docs/conf_sso_test_msscolab.rst
```diff
@@ -14,79 +14,92 @@ To set up a local identity provider with the mscolab server, you'll first need t
...
+- Add to the `PYTHONPATH` where your `mscolab_settings.py`.
+- Add `USE_SAML2 = False` in your `mscolab_settings.py` file.
 
-`$ python mslib/mscolab/mscolab.py sso_conf --init_sso_crts`
+.. note::
+    If you set `USE_SAML2 = True` without keys and certificates, this will not execute. So, make sure to set `USE_SAML2 = False` before executing the command.
 
+If everything is correctly set, you can generate keys and certificates simply by running
 
+`$ mscolab sso_conf --init_sso_crts`
```
ReimarBauer: when I do a make html document from the docs, these get rendered into 

`mscolab sso_conf –init_sso_crts`

I'm not sure if we have that also on other places.

matrss: These inline code blocks might need the `:code:` role to be rendered properly, e.g.:
```
:code:`$ mscolab sso_conf --init_sso_crts`
```

docs/conf_sso_test_msscolab.rst
```diff
@@ -14,79 +14,92 @@ To set up a local identity provider with the mscolab server, you'll first need t
...
 
 * Once you have successfully launched the server and identity provider, you can begin testing the Single Sign-On (SSO) process.
-* Start MSS PyQT application `$ python mslib/msui/msui.py`.
+* Start MSS PyQt application:
+
+`$ msui`
+
 * Login with identity provider through Qt Client application.
-* To log in to the mscolab server through the identity provider, you can use the credentials specified in the ``PASSWD`` section of the ``MSS/mslib/idp/idp.py`` file. Look for the relevant section in the file to find the necessary login credentials.
+* To log in to the mscolab server through the identity provider, you can use the credentials specified in the ``PASSWD`` section of the ``MSS/mslib/msidp/idp.py`` file. Look for the relevant section in the file to find the necessary login credentials.
```
matrss, line 117: So, what is our intended use case for this IdP? I am still not sure I have understood it. Right now we have it under mslib/ and expose entrypoints to it in a way that suggests we want to use this in production and expose this functionality to other admins running a mscolab instance. On the other hand, in this line we point to some file in the source code to get credentials for one of the predefined accounts. That sounds more like testing. For production we would need some kind of proper user management, but I think building an entire production-grade IdP is out-of-scope for us.

I am still of the opinion that we should move everything related to setting up and running this IdP below tests/, make the documentation intended for developers only and not expose this to the end-users of MSS. We can then use this IdP in the test suite to have something easy to setup to test against. To benefit from the SSO feature we will hook this up to an externally managed IdP anyway (Helmholtz AAI, or whatever SAML IdP someone may want to connect to) and have no say in what is running on their side.

This comment isn't a blocker functionality-wise though, so we can merge this PR and address this later.

docs/conf_sso_test_msscolab.rst
```diff
@@ -14,79 +14,92 @@ To set up a local identity provider with the mscolab server, you'll first need t
...
+2. Generate Keys, Certificates, and backend_saml files
+------------------------------------------------------
 
-This involves generating both .key files and .crt files for both the Identity provider and mscolab server and backend_saml.yaml file. 
+This involves generating both `.key` files and `.crt` files for both the Identity provider and mscolab server and `backend_saml.yaml` file. 
 
-Before running the command make sure to set
-`USE_SAML2 = False` 
-In some cases, if you set `USE_SAML2 = True` without certificates, this will not execute. So, make sure to set `USE_SAML2 = False` before executing the command.
+Before running the command make sure to set `USE_SAML2 = False` in your `mscolab_settings.py` file,  You can accomplish this by following these steps:
```
matrss, line 32: I've tried to set `USE_SAML2 = True` in ~/mss/mscolab_settings.py, while ~/mss is in `$PYTHONPATH`. That didn't work though, to successfully execute `mscolab sso_conf --init_sso_metadata` I still had to set this in mslib/mscolab/conf.py for some reason.

Do you have this issue as well?

matrss (commented):
I left some comments.

docs/conf_sso_test_msscolab.rst
ReimarBauer: The Syntax for develop compared to stable had changed.

It is only one line no class definition

I have a ~/mscolab dir used. And with the boa installed version I had no issues.

matrss: Ah, that was it, thanks. I didn't catch that format change.

docs/conf_sso_test_msscolab.rst
nilupulmanodya: Thanks

PR #1979, opened 2023-08-31
Round vertical coordinate to closest legal flight level upon insertio…

### LINKED ISSUE #1976: inserting a point in sideview should use flight levels

joernu76:
Round vertical coordinate to closest legal flight level upon insertion in side view.

Fix #1976

**Purpose of PR?**:

Fixes #1976

no

### DIFF
--- a/mslib/msui/mpl_pathinteractor.py
+++ b/mslib/msui/mpl_pathinteractor.py
@@ -977,6 +977,8 @@ class VPathInteractor(PathInteractor):
         y = event.ydata
         wpm = self.waypoints_model
         flightlevel = float(pressure2flightlevel(y * units.Pa).magnitude)
+        # round flightlevel to the nearest multiple of five (legal level)
+        flightlevel = 5.0 * round(flightlevel / 5)
         [lat, lon], best_index = self.plotter.get_lat_lon(event, wpm.all_waypoint_data())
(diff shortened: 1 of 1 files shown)


mslib/msui/mpl_pathinteractor.py
```diff
@@ -977,6 +977,7 @@ def button_release_insert_callback(self, event):
         y = event.ydata
         wpm = self.waypoints_model
         flightlevel = float(pressure2flightlevel(y * units.Pa).magnitude)
+        flightlevel = 5.0 * round(flightlevel / 5)
```
ReimarBauer, line 981: I am not sure if I understand this again in a year , or what happen when someone refactors this

would you please so kind and add a comment above this line.

ReimarBauer:
thx, just one question, we likly will have a 8.3.0 soon because of  that KML fix. Do you want this in stable?

ReimarBauer:
@joernu76 for stable?

joernu76:
This does not reaally fix a bug. -> develop

ReimarBauer:
We are preparing a minor release. This means we can add features which don't need an API change.

Patch level (3rd number) would be Bug fixes only.

ReimarBauer (approved):

PR #1983, opened 2023-09-01
Fixed bug in updating operation names in top view windows

### LINKED ISSUE #1961: MSUI: Window title is very very long and has redundant information
Seen under MSUI 8.1.1 under Windows 11 on a MSColab project. See screenshot below

grafik

joernu76:
Fix #1961

**Purpose of PR?**:

Fixes #1961 

No

### DIFF
--- a/mslib/msui/mpl_map.py
+++ b/mslib/msui/mpl_map.py
@@ -147,17 +147,10 @@ class MapCanvas(basemap.Basemap):
         self.image = None
 
         # Print project name and CRS identifier into figure.
-        crs_text = ""
-        if self.operation_name is not None:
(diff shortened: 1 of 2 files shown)


ReimarBauer (changes requested):
We can have more than 2 entries in fields, because of e.g. used airports widget

this now cumulates then all new operation names

problem

after topview is loaded, select draw airports.

joernu76:
Then this requires a fully redesign instead of the original hack.
I was not aware of more than two lines.
Sigh.
Will have a look.

ReimarBauer (approved):

Issue #1985, opened 2023-09-01
multiple flightpath docking widget too slow


ReimarBauer:
With my current set of operations the UI needs a huge amount of time to load all available also archived operations. 

This can become improved by

- ignore archived operation
- load only by selected category of the main ui

After they are loaded the reaction time is fast

ReimarBauer:
this can be worked on after https://github.com/Open-MSS/MSS/issues/1964

PR #1987, opened 2023-09-01
introduces a config var to skip archived operations on the server

### LINKED ISSUE #1964: option to skip archived operations on server site
I think we should add a config option which is used on start to skip looking up for archived operations.

Currently we sort these only on the UI into a hidden section, but the server has to join each member to the operations.

MSCOLAB_skip_archived_operations = False on default.

This then is used by the sockets_manager queries

Changing the msui settings restarts the app which also means a new login and by this the full list can be seen.

ReimarBauer:
**Purpose of PR?**:

Fixes #1964 
and partly #1985 

**Checklist:**
- [X] Commit has unit tests

### DIFF
--- a/mslib/mscolab/server.py
+++ b/mslib/mscolab/server.py
@@ -471,7 +471,7 @@ def authorized_users():
 @APP.route('/operations', methods=['GET'])
 @verify_user
 def get_operations():
-    skip_archived = request.args.get('skip_archived',  request.form.get('skip_archived', "False"))
+    skip_archived = request.args.get('skip_archived', request.form.get('skip_archived', "False"))
     if skip_archived == "True":
         skip_archived = True
     else:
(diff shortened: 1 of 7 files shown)


mslib/mscolab/server.py
```diff
@@ -397,9 +397,15 @@ def create_operation():
     content = request.form.get('content', None)
     description = request.form.get('description', None)
     category = request.form.get('category', "default")
+    active = request.form.get('active', "True")
+    if active == "True":
```
joernu76: active = (active == "True") ?

mslib/mscolab/server.py
```diff
@@ -465,8 +471,13 @@ def authorized_users():
 @APP.route('/operations', methods=['GET'])
 @verify_user
 def get_operations():
+    skip_archived = request.args.get('skip_archived', request.form.get('skip_archived', "False"))
+    if skip_archived == "True":
```
joernu76: see above

joernu76 (approved):

PR #1990, opened 2023-09-01
I1989

### LINKED ISSUE #1989: source of open airports data offline, needs an alternative url
This morning
    https://ourairports.com/data/airports.csv
was not accessible and it is generally slow.
Here
   https://github.com/davidmegginson/ourairports-data
is an alternate location that is also pointed at by the original website. This should be used as github has much better availability.

joernu76:
**Purpose of PR?**:

Fixes #1989

no

### DIFF
--- a/mslib/utils/airdata.py
+++ b/mslib/utils/airdata.py
@@ -102,7 +102,7 @@ def get_airports(force_download=False, url=None):
     _airports_mtime = Airspace().airports_mtime
 
     if url is None:
-        url = "https://ourairports.com/data/airports.csv"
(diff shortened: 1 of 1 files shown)


ReimarBauer:
when this is only an url change this may become done in stable too.

The other changes in this PR likly comes from a missing merge from stable to develop. Because this merge shows up some merge conflicts I want to do this short before or short after the 8.3.0. release

ReimarBauer (changes requested):
please send the url change to stable

ReimarBauer:
> please send the url change to stable

seen that I can do this in the UI

ReimarBauer (approved):

PR #2038, opened 2023-09-14
topview should not using parent of mainwindow

### LINKED ISSUE #2007: MSUI with Windows: TopView window is always on top of MSUI main window
Under Windows 10, MSUI TopView window is always on top of MSUI main window. In addition, the TopView is not appearing in the Windows Task bar (other views are visible there). Alt +Tab also does not allow to change TopView windows.

grafik

Further, "Server/Layer" window is also always on top of corresponding TopView/SideView

ReimarBauer:
**Purpose of PR?**:

Fixes #2007

API change

**Checklist:**
- [ ] Commit has unit tests

### DIFF
--- a/mslib/msui/topview.py
+++ b/mslib/msui/topview.py
@@ -185,13 +185,14 @@ class MSUITopViewWindow(MSUIMplViewWindow, ui.Ui_TopViewWindow):
     signal_permission_revoked = QtCore.pyqtSignal(int)
     signal_render_new_permission = QtCore.pyqtSignal(int, str)
 
-    def __init__(self, parent=None, model=None, _id=None, active_flighttrack=None, mscolab_server_url=None, token=None):
+    def __init__(self, parent=None, mainwindow=None, model=None, _id=None,
+                 active_flighttrack=None, mscolab_server_url=None, token=None):
         """
         Set up user interface, connect signal/slots.
         """
         super().__init__(parent, model, _id)
         logging.debug(_id)
-        self.ui = parent
+        self.mainwindow = mainwindow
         self.setupUi(self)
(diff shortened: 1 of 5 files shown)


ReimarBauer:
@joernu76 the tests got blocked by a similiar expensive use of slots as you fixed in the kml docking widget.

(Of course speed is much faster now too)

mslib/msui/topview.py
```diff
@@ -185,13 +185,27 @@ class MSUITopViewWindow(MSUIMplViewWindow, ui.Ui_TopViewWindow):
...
 
-    def __init__(self, parent=None, model=None, _id=None, active_flighttrack=None, mscolab_server_url=None, token=None):
+    def __init__(self, parent=None, mainwindow=None, model=None, _id=None,
+                 active_flighttrack=None, mscolab_server_url=None, token=None):
         """
         Set up user interface, connect signal/slots.
         """
         super().__init__(parent, model, _id)
         logging.debug(_id)
-        self.ui = parent
```
joernu76, line 194: wouldn't a reference to mainwindow not "simpler" here?

joernu76 (approved):
I think the same problem affects Side&Linearview? Still a step in the right direction.

mslib/msui/topview.py
ReimarBauer: Later we use parent when the docking widget is called from topview.

That shows currently what we need.

ReimarBauer:
> I think the same problem affects Side&Linearview? Still a step in the right direction.

Sideview and Linearview don't have multiple flightpath dockingwidget. By this the problem was introduced.

PR #2039, opened 2023-09-17
SSL verification enablement for mscolab

nilupulmanodya:
**Purpose of PR?**:

Fixes #
Implementation for SSL verification enablement of mscolab for developers

No

Should need to merge #1974  before merging this

**Checklist:**
- [ ] Commit has unit tests


matrss (approved):
LGTM if we need this.

For what it's worth I see one alternative implementation, where we wouldn't need to change anything under mslib/:
A `requests.Session` object has a `verify` attribute which is used as a default value in case nothing is specified. This attribute is set to true in `Session`s `__init__` (https://github.com/psf/requests/blob/f5a7aefc2dc123c65ec1c6d3e83949999b03876b/src/requests/sessions.py#L424).
All requests, even those that use the `requests.{get,post,...}` functions, do use such a Session object. We could therefore monkeypatch this class in our test suite (probably in conftest.py) to always set the verify attribute to false.

I am indifferent to how we do it. Your approach is more verbose, while what I described above would be a bit "hacky" / less obvious / harder to understand if we ever run into some issue with this.

nilupulmanodya:
based on the discussion https://mss-devel.slack.com/archives/C04VCNH81MJ/p1696760025937339
The implementation is a kind of misunderstanding. Will do SSL verification for SAML later after #2043 is merged.

PR #2043, opened 2023-09-26
improve mscolab for configure multiple Idps

nilupulmanodya:
**Purpose of PR?**:
The purpose of this PR is to improve the mscolab server, enabling it to support multiple IdPs. In this PR, have made necessary functional improvements and error handling enhancements for this purpose

may be better to review Draft Developer doc https://docs.google.com/document/d/1fX7F4KYQ-iIibvMjSqF0oInse-ew2naJ_ataqylNU8Q/edit?usp=sharing before reviewing this PR

Fixes #

**Checklist:**
- [ ] Commit has unit tests


mslib/mscolab/server.py
```diff
@@ -80,31 +80,43 @@ class mscolab_auth:
...
+
+                localhost_test_idp = SPConfig().load(yaml_data["config"]["localhost_test_idp"])
+                sp_localhost_test_idp = Saml2Client(localhost_test_idp)
+
+            except SAMLError:
+                warnings.warn("Invalid Saml2Client Config with localhost_test_idp ! Please configure with valid CRTs\
+                              /metadata and try again.")
+                sys.exit()
+
+        # if multiple IdPs exists, development should need to implement accordingly below
```
ReimarBauer: keep in mind that this file can't be changed when the software gets installed.  When someone would edit the file on the next release all that changes are gone.

When these are kind of configurations, can you move it to conf.py e.g. as function? 
Then a user can replace it in his own mscolab_settings.py

nilupulmanodya: Thank you for figuring it out. I have updated the PR accordingly. : )

nilupulmanodya:
Disabled pytests for GSoC branch with Todo Refactor cause of `Error: Process completed with exit code 137.`

mslib/mscolab/server.py
```diff
@@ -235,23 +207,36 @@ def wrapper(*args, **kwargs):
...
 
-    # The value of 'condition' should be the same as the 'idp_identity_name'\
-    # set in the 'CONFIGURED_IDPS' of conf.py.
-
-    if selected_idp == 'localhost_test_idp':
-        idps = sp_localhost_test_idp.metadata.identity_providers()
 
-    # elif selected_idp == 'idp2':
-    #     idps = sp_idp2.metadata.identity_providers()
+def create_or_udpate_idp_user(email, username, token, authentication_backend):
```
matrss: Typo, `udpate` -> `update`

mslib/mscolab/server.py
```diff
@@ -815,37 +801,65 @@ def idp_login():
...
+                                attribute_name = attribute.name
+                                attribute_value = \
+                                    attribute.attribute_value[0].text if attribute.attribute_value else None
+                                attributes[attribute_name] = attribute_value
+
+                        # Extract the email and givenname attributes
+                        email = attributes.get("email")
+                        username = attributes.get("givenName")
+
+                        token = generate_confirmation_token(email)
```
matrss: `dict`s `.get` returns None if the key does not exist. Therefore it could happen that `generate_confirmation_token` is passed a None value. Is that intended?

In which case is this alternative way to get `email` and `givenName` used, opposed to taking them from `authn_response.ava`?

mslib/mscolab/server.py
```diff
@@ -815,37 +801,65 @@ def idp_login():
...
+
+                    except (NameError, AttributeError, KeyError):
+                        render_template('errors/403.html'), 403
+
+                if email is not None and username is not None:
+                    idp_user_db_state = create_or_udpate_idp_user(email, username, token,
+                                                                  idp_config['idp_identity_name'])
+                    if idp_user_db_state:
+                        return render_template('idp/idp_login_success.html', token=token), 200
```
matrss: Just for my own understanding: in this function, if we receive a `email` and `givenName` the assumption is that those are validated by the IdP (as in, the user successfully authenticated with the IdP and the IdP asserts that the user controls the given email), and therefore we can be sure that the user is indeed the same person that we potentially already have in our database for the given mail address, right?

I just want to make sure that I understand the authentication flow here.

matrss (commented):
I left some comments.

Please re-enable the test workflow. If the PR in its current state does not pass I would prefer to see that, instead of seeing "All checks have passed" with a check missing.

This might have been a fluke anyway, I ran the test suite locally and it passed. I also started a re-run of the job (see: https://github.com/Open-MSS/MSS/actions/runs/6456242735/job/17637454977); let's see if that makes it through. This happens sometimes, unfortunately...

Edit: OK, the re-run didn't make it either, but it failed in a different test function this time...

Edit Edit: exit code 137 indicates that the job is cancelled because of high memory usage, the runner is running out of memory. It looks like there is a memory leak somewhere. That also explains why it did pass on my system.

mslib/mscolab/server.py
nilupulmanodya: @matrss  yeah.. In this function, the first step is to check if the user already exists in the database using their email address. If the user does not exist, a new user is created based on the provided `email`. If the user already exists, the database is updated according to the `authentication_backend`.  and  in both cases `token` is generated for logging in via the msui.

mslib/mscolab/server.py
nilupulmanodya: Thank you :)

mslib/mscolab/server.py
nilupulmanodya: @matrss In certain cases, the IdP response may not contain the `email` or `givenName` within `authn_response.ava`. Instead, these values might be found within `authn_response.assertion.attribute_statement`. This scenario also occurs in Keycloak. I have noted that the IdP response can vary based on individual configurations. The reason behind this behavior is my implementation. As you pointed out, this variation could result in returning a 'None' value within the function. To address this, I have updated by adding a value check statement before calling the function. Thank you very much for figuring out that point. :)

mslib/mscolab/server.py
```diff
@@ -844,13 +845,16 @@ def acs_post_handler(url):
...
                         username = attributes.get("givenName")
 
-                        token = generate_confirmation_token(email)
+                        if email is not None and username is not None:
+                            token = generate_confirmation_token(email)
+                        else:
+                            render_template('errors/403.html'), 403
 
                     except (NameError, AttributeError, KeyError):
                         render_template('errors/403.html'), 403
```
matrss: I think you could just use `attributes["email"]` and `attributes["givenName"]`, since you catch the KeyError exception already. That would shorten this a bit and not duplicate the `render_template` call.

Something like:
```python
                        email = attributes["email"]
                        username = attributes["givenName"]

                        token = generate_confirmation_token(email)
                    except (NameError, AttributeError, KeyError):
                        render_template('errors/403.html'), 403
```

matrss (approved):
Left another comment. Apart from that this looks good to me.

I could not yet figure out why the CI runs fail. Locally I could not reproduce the issue, even if I limit the pytest run to 100M of RAM with:
```
systemd-run --scope -p MemoryMax=100M --user -- pytest -v --durations=20 --reverse --cov=mslib tests
```

According to githubs docs the runner should have 7GB of RAM available. So, no idea what's the issue.

@ReimarBauer advised to not spend to much time debugging this issue, so I'll approve this with the caveat of the one comment. I'll also run the workflow again to make sure it wasn't two flukes in a row...

mslib/mscolab/server.py
nilupulmanodya: Thanks :)

PR #2062, opened 2023-10-17
set SSL certificate verification enablement

nilupulmanodya:
**Purpose of PR?**:

Fixes #
set SSL certificate verification enablement. based on the discussion (https://mss-devel.slack.com/archives/C04VCNH81MJ/p1696760025937339)

**Checklist:**
- [ ] Commit has unit tests

### DIFF
--- a/mslib/mscolab/conf.py
+++ b/mslib/mscolab/conf.py
@@ -174,7 +174,8 @@ except ImportError as ex:
                                        valid CRTs metadata and try again.")
                         sys.exit()
 
-                    # if multiple IdPs exists, development should need to implement accordingly below
+                    # if multiple IdPs exists, development should need to implement accordingly below,
+                    # make sure to set SSL certificates verification enablement.
                     """
                         if 'idp_2'== configured_idp['idp_identity_name']:
                             # rest of code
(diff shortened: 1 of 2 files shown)


mslib/mscolab/conf.py
```diff
@@ -155,6 +158,7 @@ class setup_saml2_backend:
                                            Ignore this warning when you initializeing metadata.")
 
                         localhost_test_idp = SPConfig().load(yaml_data["config"]["localhost_test_idp"])
+                        localhost_test_idp.verify_ssl_cert = mscolab_settings.VERIFY_SSL_CERT
```
ReimarBauer, line 161: unfortuneantly the default is False on the library itselfs. In this solution the config var will only be set on the test_idp, which won't used. Otheres could understand the variable, that this is the only thing to do to have a vlaidated certificat in use.

Please also add a hint on this in the ```# if multiple IdPs exists, ``` sections so it is seen, a user has to take care.


Or is there an other option to pass this value, so that we only have one place to change the constant?

nilupulmanodya: I am not sure about the possibility of adding a single value since we are passing the SAML2Client through the dict, and `idp_identity_name`  also be defined by the user when configuring multiple IdPs.. hint added.

ReimarBauer (approved):

PR #2064, opened 2023-10-22
Create documentation for SSO integration through SAML

nilupulmanodya:
**Purpose of PR?**:

Fixes #

Implementation of documentation for SSO via SAML Integration for MSColab Server

**Does this PR introduce a breaking change?**

No

**If the changes in this PR are manually verified, list down the scenarios covered:**:
Yes

**Additional information for reviewer?** :
_Mention if this PR is part of any design or a continuation of previous PRs_

Created after the suggested improvements of Draft document (https://docs.google.com/document/d/1fX7F4KYQ-iIibvMjSqF0oInse-ew2naJ_ataqylNU8Q/edit?usp=sharing)

**Does this PR results in some Documentation changes?**
_If yes, include the list of Documentation changes_

Added new documentation for SSO

**Checklist:**
- [ ] Bug fix. Fixes #<issue number>
- [ ] New feature (Non-API breaking changes that adds functionality)
- [ ] PR Title follows the convention of  `<type>: <subject>`
- [ ] Commit has unit tests

PR #2068, opened 2023-10-29
improve accessibility saml2 urls

nilupulmanodya:
**Purpose of PR?**:
improve accessibility saml2 urls.

Fixes #
https://github.com/Open-MSS/MSS/pull/2064#discussion_r1374346647

### DIFF
--- a/mslib/mscolab/server.py
+++ b/mslib/mscolab/server.py
@@ -738,14 +738,16 @@ def reset_request():
 @APP.route("/metadata/<idp_identity_name>", methods=['GET'])
 def metadata(idp_identity_name):
     """Return the SAML metadata XML for the requested IDP"""
-    for idp_config in setup_saml2_backend.CONFIGURED_IDPS:
-        if idp_identity_name == idp_config['idp_identity_name']:
-            sp_config = idp_config['idp_data']['saml2client']
-            metadata_string = create_metadata_string(
-                None, sp_config.config, 4, None, None, None, None, None
-            ).decode("utf-8")
-            return Response(metadata_string, mimetype="text/xml")
-    return render_template('errors/404.html'), 404
+    if mscolab_settings.USE_SAML2:
+        for idp_config in setup_saml2_backend.CONFIGURED_IDPS:
+            if idp_identity_name == idp_config['idp_identity_name']:
+                sp_config = idp_config['idp_data']['saml2client']
+                metadata_string = create_metadata_string(
+                    None, sp_config.config, 4, None, None, None, None, None
+                ).decode("utf-8")
+                return Response(metadata_string, mimetype="text/xml")
+        return render_template('errors/404.html'), 404
+    return render_template('errors/403.html'), 403
 
 
 @APP.route('/available_idps/', methods=['GET'])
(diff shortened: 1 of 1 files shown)


ReimarBauer (changes requested):
Hmm, that didn't solve the not needed imports, I thought first we can 

do only the import of setup_saml2_backend
when it is enabled by

 mscolab_settings.USE_SAML2

but the imports are in the conf.py done

mslib/mscolab/server.py
```diff
@@ -738,14 +738,16 @@ def reset_request():
...
+    if mscolab_settings.USE_SAML2:
+        for idp_config in setup_saml2_backend.CONFIGURED_IDPS:
+            if idp_identity_name == idp_config['idp_identity_name']:
+                sp_config = idp_config['idp_data']['saml2client']
+                metadata_string = create_metadata_string(
+                    None, sp_config.config, 4, None, None, None, None, None
+                ).decode("utf-8")
+                return Response(metadata_string, mimetype="text/xml")
+        return render_template('errors/404.html'), 404
+    return render_template('errors/403.html'), 403
```
matrss: `403 Forbidden` is semantically the wrong error value here, since it is not that access to the metadata route is "forbidden" by some means, but it is simply not supported if SAML is disabled. I think it should just be a `404`.

In fact, I think it would be even more elegant to not define the route at all if `USE_SAML2` is false and just let it fallback to the default behaviour for undefined routes. I.e. put the entire function definition into an `if mscolab_settings.USE_SAML2`.

mslib/mscolab/server.py
```diff
@@ -738,14 +738,16 @@ def reset_request():
...
     """Return the SAML metadata XML for the requested IDP"""
-    for idp_config in setup_saml2_backend.CONFIGURED_IDPS:
-        if idp_identity_name == idp_config['idp_identity_name']:
-            sp_config = idp_config['idp_data']['saml2client']
-            metadata_string = create_metadata_string(
-                None, sp_config.config, 4, None, None, None, None, None
-            ).decode("utf-8")
-            return Response(metadata_string, mimetype="text/xml")
-    return render_template('errors/404.html'), 404
+    if mscolab_settings.USE_SAML2:
```
ReimarBauer: please use a decorator for all these if mscolab_settings.USE_SAML2 changes
see 5.3 of https://towardsdatascience.com/5-advanced-tips-on-python-decorators-113307d5a92c 
e.g. use a  verify_use_saml2

mslib/mscolab/server.py
```diff
@@ -764,28 +766,30 @@ def available_idps():
...
+            if binding == BINDING_HTTP_REDIRECT:
+                headers = dict(http_args["headers"])
+                return redirect(str(headers["Location"]), code=303)
+            return Response(http_args["data"], headers=http_args["headers"])
+        except (NameError, AttributeError):
+            return render_template('errors/403.html'), 403
+    return render_template('errors/403.html'), 403
 
 
 @APP.route('/<path:url>', methods=['POST'])
```
matrss: This "catch-all" route makes it so that any request to a non-existing route with a method other than POST now returns a 405 error instead of a 404. E.g. going to `<mscolab-server>/this-route-does-not-exist` in a browser previously returned 404 Not Found and with this change returns 405 Method Not Allowed. I think we should find a way to change that back to a 404.

matrss: I am wondering, since this route effectively only handles POST requests to the URL's specified in `idp_config['idp_data']['assertion_consumer_endpoints']` (right?), is it possible to dynamically define the routes for the specified endpoints? I.e. loop over the endpoints and define `@APP.route('/<assertion_consumer_endpoint>/', methods=["POST"])` functions (or something similar) which handle them, completely removing the catch-all route and restoring flasks default behavior of 404's for GET requests to undefined routes.

matrss (changes requested):

mslib/mscolab/server.py
matrss: I think this shouldn't be a decorator, but the if statement should be outside of the function definitions, so that if `mscolab_settings.USE_SAML2` is false the SAML specific routes are not defined at all. This would mean that only a single if statement is necessary and no explicit error handling is needed, since if the route exists then SAML is enabled.

nilupulmanodya: Thank you.. Resolved

mslib/mscolab/server.py
nilupulmanodya: Thank you.. Resolved :)

matrss (approved):

ReimarBauer (approved):
thx

PR #2069, opened 2023-11-06
merge of Gsoc2023 nilupul manodya

ReimarBauer:
This is a direct merge request for the GSOC 2023 contribution of Nilupul Manodya. Beforehand develop was merged into his branch. We maybe can also squash and merge, what do you think?

The outcome of the project was tested on mss-stage. Further devlopment should be done in develop on that idea.


docs/samples/config/mscolab/mss_saml2_backend.yaml.samlple
joernu76, line 1: I think the ending should be ".sample"?

mslib/mscolab/mscolab.py
```diff
@@ -93,6 +96,264 @@ def handle_db_seed():
     print("Database seeded successfully!")
 
 
+def handle_mscolab_certificate_init():
+    print('generating CRTs for the mscolab server......')
```
joernu76, line 100: There are some print-function calls in here already, but we should generally use the logging for console output.

mslib/mscolab/mscolab.py
```diff
@@ -93,6 +96,264 @@ def handle_db_seed():
...
+
+    try:
+        cmd = ["openssl", "req", "-newkey", "rsa:4096", "-keyout",
+               f"{mscolab_settings.MSCOLAB_SSO_DIR}/key_mscolab.key",
```
joernu76, line 104: please us os.path.join (also below)

mslib/mscolab/mscolab.py
```diff
@@ -93,6 +96,264 @@ def handle_db_seed():
...
+  #         assertion_consumer_service:
+  #           - [http://localhost:8083/idp2/acs/post, 'urn:oasis:names:tc:SAML:2.0:bindings:HTTP-POST']
+  #           - [http://localhost:8083/idp2/acs/redirect, 'urn:oasis:names:tc:SAML:2.0:bindings:HTTP-Redirect']
+  #         discovery_response:
+  #         - [<base_url>/<name>/disco, 'urn:oasis:names:tc:SAML:profiles:SSO:idp-discovery-protocol']
+  #       name_id_format: 'urn:oasis:names:tc:SAML:2.0:nameid-format:transient'
+  #       name_id_format_allow_create: true
+"""
+    try:
+        file_path = f"{mscolab_settings.MSCOLAB_SSO_DIR}/mss_saml2_backend.yaml"
```
joernu76, line 253: os.path.join

mslib/mscolab/mscolab.py
```diff
@@ -93,6 +96,264 @@ def handle_db_seed():
...
+        with open(file_path, "w", encoding="utf-8") as file:
+            file.write(saml_2_backend_yaml_content)
+        return True
+    except (FileNotFoundError, PermissionError) as error:
+        print(f"Error while generated backend .yaml for the local mscolabserver: {error}")
```
joernu76, line 258: logging.error

mslib/mscolab/mscolab.py
```diff
@@ -93,6 +96,264 @@ def handle_db_seed():
...
+
+        Before running this function:
+        - Ensure that USE_SAML2 is set to True.
+        - Generate the necessary keys and certificates and configure them in the .yaml
+        file for the local IDP.
+    '''
+    print('generating metadata file for the mscolab server')
+
+    try:
+        command = ["python", "mslib/mscolab/mscolab.py", "start"] if repo_exists else ["mscolab", "start"]
```
joernu76, line 274: The PYTHONPATH needs to be set properly for the first command to work. How do you ensure this?

mslib/mscolab/mscolab.py
```diff
@@ -93,6 +96,264 @@ def handle_db_seed():
...
+    except subprocess.CalledProcessError as error:
+        print(f"Error while generating metadata file for the mscolab server: {error}")
+        return False
+
+
+def handle_local_idp_metadata_init(repo_exists):
+    print('generating metadata for localhost identity provider')
+
+    try:
+        if os.path.exists(f"{mscolab_settings.MSCOLAB_SSO_DIR}/idp.xml"):
```
joernu76, line 296: os.path.join

mslib/mscolab/mscolab.py
```diff
@@ -93,6 +96,264 @@ def handle_db_seed():
...
+            os.remove(f"{mscolab_settings.MSCOLAB_SSO_DIR}/idp.xml")
+
+        idp_conf_path = "mslib/msidp/idp_conf.py"
```
joernu76, line 299: os.path.join

mslib/mscolab/mscolab.py
```diff
@@ -130,6 +400,13 @@ def main():
         print("Version:", __version__)
         sys.exit()
 
+    try:
+        _ = git.Repo(os.path.dirname(os.path.realpath(__file__)), search_parent_directories=True)
```
joernu76, line 404: I am not sure that this is a reasonable use case. Useful for testing, but adds a  lot of code that needs to be maintained.

joernu76:
Great work!

.gitignore
```diff
@@ -24,4 +24,5 @@ build/
 mss.egg-info/
 tutorials/recordings
 tutorials/cursor_image.png
-
+__pycache__/
+instance/
```
matrss, line 28: Where was that line coming from again? I cannot remember what this path was used for. Is it still required?

docs/samples/config/mscolab/mss_saml2_backend.yaml.samlple
```diff
@@ -0,0 +1,116 @@
...
+  #       want_assertion_signed: true
+  #       allow_unknown_attributes: true
+  #       allow_unsolicited: true
+  #       endpoints:
+  #         assertion_consumer_service:
+  #           - [http://localhost:8083/idp2/acs/post, 'urn:oasis:names:tc:SAML:2.0:bindings:HTTP-POST']
+  #         discovery_response:
+  #         - [<base_url>/<name>/disco, 'urn:oasis:names:tc:SAML:profiles:SSO:idp-discovery-protocol']
+  #       name_id_format: 'urn:oasis:names:tc:SAML:2.0:nameid-format:transient'
+  #       name_id_format_allow_create: true
```
matrss, line 116: This file is missing a newline after the last line.

localbuild/meta.yaml
```diff
@@ -19,6 +19,7 @@ build:
     - mswms_demodata = mslib.mswms.demodata:main
     - mscolab = mslib.mscolab.mscolab:main
     - mssautoplot = mslib.utils.mssautoplot:main
+    - msidp = mslib.msidp.idp:main
```
matrss, line 22: Do we really want to expose the IdP that was implemented? I still consider this to be testing infrastructure that should be moved to tests/ and be used as a lightweight target in integration tests of the SP functionality only.

mslib/mscolab/conf.py
```diff
@@ -107,6 +114,15 @@ class default_mscolab_settings:
     # filepath to md file with gdpr
     GDPR = None
 
+    # enable login by identity provider
+    USE_SAML2 = False
+
+    # SSL certificates verification during SSO.
+    VERIFY_SSL_CERT = True
```
matrss, line 121: This config option name should somehow make clear that it only applies to the certificates used between MSColab as a SAML SP and the IdP, nothing else.

mslib/mscolab/conf.py
```diff
@@ -116,3 +132,73 @@ class default_mscolab_settings:
...
+                        sys.exit()
+
+                    # if multiple IdPs exists, development should need to implement accordingly below,
+                    # make sure to set SSL certificates verification enablement.
+                    """
+                        if 'idp_2'== configured_idp['idp_identity_name']:
+                            # rest of code
+                            # set CRTs and metadata paths for the idp_2
+                            # configuration idp_2 Saml2Client
+                    """
```
matrss, line 204: This is effectively the same as what is in `docs/samples/config/mscolab/setup_saml2_backend.py.sample`. Is there any way to unify that to avoid them becoming out of sync? Or is that not a concern?

mslib/mscolab/mscolab.py
matrss: IIRC this function is only called when initially generating a key and (self-signed) certificate for SAML, after which the program exits. It is not part of a long-running process when MSColab is running as a server. In this case one could argue that this print call is part of the direct user interaction when generating a key and (self-signed) certificate, not really logging.

Granted, in this line the distinction is not that clear, but in general I think everything that is required for a meaningful user interaction with the program should stay as a `print` (or the other way around: the program output should still make sense if someone disables all logging). Some other print statements in this file should therefore really stay that way (or maybe go to stderr), in my opinion.

mslib/mscolab/mscolab.py
matrss: The first command is only used if this is run from a git clone instead of the installed MSS package. Detecting git is used as a proxy to detect if this is run in a development context, I think. The assumption is that you have an adequate python environment enabled in that case.

I'd argue we should only use the second command though, and get rid of the git detection part entirely. In a development context one could then use `pip install -e .` to make the `msui`, `mscolab`, etc. commands use the code in the local checkout, instead of any other version that might be in the PATH. This would require that setup.py declares all entrypoints though, which it currently does not (only msui is in there).

Maybe mamba has a similar "editable install" feature.

mslib/mscolab/mscolab.py
```diff
@@ -93,6 +96,264 @@ def handle_db_seed():
...
+        file for the local IDP.
+    '''
+    print('generating metadata file for the mscolab server')
+
+    try:
+        command = ["python", "mslib/mscolab/mscolab.py", "start"] if repo_exists else ["mscolab", "start"]
+        process = subprocess.Popen(command)
+
+        # Add a small delay to allow the server to start up
+        time.sleep(10)
```
matrss, line 278: Not really optimal to use sleep as a synchronization primitive here (ideally it should wait until `process` signals that it is ready in some way, e.g. when process' stdout contains something that indicates that the server is up and running). This is not a very critical code path though, so not too much of a concern.

mslib/mscolab/mscolab.py
```diff
@@ -93,6 +96,264 @@ def handle_db_seed():
...
+
+        cmd_curl = ["curl", "http://localhost:8083/metadata/localhost_test_idp",
+                    "-o", f"{mscolab_settings.MSCOLAB_SSO_DIR}/metadata_sp.xml"]
+        subprocess.run(cmd_curl, check=True)
+        process.kill()
```
matrss, line 283:
```suggestion
        process.terminate()
```

Sending `SIGKILL` seems a bit excessive.

mslib/mscolab/mscolab.py
```diff
@@ -93,6 +96,264 @@ def handle_db_seed():
...
+
+def handle_sso_crts_init():
+    """
+        This will generate necessary CRTs files for sso in mscolab through localhost idp
+    """
+    print("\n\nmscolab sso conf initiating......")
+    if os.path.exists(mscolab_settings.MSCOLAB_SSO_DIR):
+        shutil.rmtree(mscolab_settings.MSCOLAB_SSO_DIR)
+    create_files()
+    if not handle_mscolab_certificate_init():
```
matrss, line 329: These could also use exceptions instead of boolean return values to indicate success/failure.

mslib/msui/mscolab.py
```diff
@@ -231,6 +234,20 @@ def connect_handler(self):
...
 
+                try:
+                    idp_enabled = json.loads(r.text)["USE_SAML2"]
+                except (json.decoder.JSONDecodeError, KeyError):
+                    idp_enabled = False
+
+                if idp_enabled:
+                    # Hide user creatiion seccion if IDP login enabled
+                    self.addUserBtn.setHidden(True)
+                    self.clickNewUserLabel.setHidden(True)
```
matrss, line 245: Shouldn't it still be possible to register a new non-IdP user on the MSColab instance, even if an IdP is configured? I feel like disabling "local" account registration should be a different config option (that would need to be enforced server-side).

ReimarBauer:
@joernu76 we should merge this and improve afterwards. I already did some changes on the staging system.

mslib/mscolab/server.py
```diff
@@ -684,6 +730,143 @@ def reset_request():
...
+                        "success": True,
+                        'token': random_token,
+                        'user': {'username': user.username, 'id': user.id, 'emailid': user.emailid}
+                    })
+                return jsonify({"success": False}), 401
+            return jsonify({"success": False}), 401
+        except TypeError:
+            return jsonify({"success": False}), 401
+
+    @APP.route("/metadata/<idp_identity_name>", methods=['GET'])
```
matrss, line 857: This route could also be defined in a loop for each IdP (similar to the acs_post_handler), so that it does not match too broadly. That would remove the requirement for an explicit 404 (and the need for the entire 404 template).

joernu76 (approved):

ReimarBauer:
@matrss these are all things we have to improve / and figure out after the merge in develop. The GSOC branch can't have further changes.

mslib/msui/mscolab.py
ReimarBauer: also in the workflow of profile/usercreation by the IDP is not completed. The new profile is not activated, it would need a validation (recover password attempt currently) first. There some changes we have to think about after we merged this in.

localbuild/meta.yaml
ReimarBauer: you are right, we can change this after the merge.

mslib/msui/mscolab.py
ReimarBauer: We have to verify too that a user with a "None" Password because of IDP can't fetch one and update that 'None' over password recovery.  Some of these parts need a coverage by tests.

ReimarBauer:
Merging GSOC project by Nilupul Manodya  - Implement a SAML 2.0 service provider (SP) into mscolab

mslib/mscolab/conf.py
ReimarBauer: In the samples there are later more possibilities. At least H AAI will be added. We should have here the smallest one.

mslib/mscolab/mscolab.py
ReimarBauer: it is quite simple to get the correct path by  ```mscolab.__file__``` 

Contriute to the idea `pip install -e .` on https://github.com/mamba-org/mamba/issues/695

PR #2091, opened 2023-11-13
fixes i2084

### LINKED ISSUE #2084: define xmlsec_binary path dependent to used env
We discussed in https://github.com/IdentityPython/pysaml2/issues/913 using libxmlsec1-feedstock  for the xmlsec1 binary

We can set its location by
https://pysaml2.readthedocs.io/en/latest/howto/config.html#xmlsec-binary

The solution needs to use the conda env var for locating the "env"/bin where it is

Currently we set the PATH var in the WSGI Script.

nilupulmanodya:
**Purpose of PR?**:

Fixes #
#2084


ReimarBauer (approved):
Thx

Issue #2087, opened 2023-11-08
refactor use print or logging


ReimarBauer:
When we have a direct interaction with a user we can use print or logging.info for command line actions. 

Any logging output can  become hidden by configuration. Choose wisely where print is required.

nilupulmanodya:
@ReimarBauer I hope to work on this. please assign to me.

nilupulmanodya:
@ReimarBauer  Do we need further refactoring, or can we close this issue?

PR #2092, opened 2023-11-13
fixes i2087

nilupulmanodya:
**Purpose of PR?**:

Fixes #
#2087


nilupulmanodya:
I am still thinking about this. Matthias also mentioned something similar here (https://github.com/Open-MSS/MSS/pull/2069#discussion_r1384829610). What I was thinking is, do we need to refactor all prints into logging.info and logging.error?

joernu76 (approved):

PR #2095, opened 2023-11-14
remove unnecssary lines from gitignore

### LINKED ISSUE #2070: Unnecessary line in .gitignore?
> https://github.com/Open-MSS/MSS/blob/f6b3ad110dbfaa12d439cadea6476dcd54d9328d/.gitignore#L28
> Where was that line coming from again? I cannot remember what this path was used for. Is it still required?

_Originally posted by @matrss in https://github.com/Open-MSS/MSS/pull/2069#discussion_r1384596047_

nilupulmanodya:
fixes #2070

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,5 +24,3 @@ build/
 mss.egg-info/
 tutorials/recordings
 tutorials/cursor_image.png
-__pycache__/
-instance/
--- a/.gitignore
+++ b/.gitignore
@@ -6,15 +6,9 @@
 *.swp
 *.patch
 *~
-mslib/mss_config.py
 mslib/performance/data/
-mslib/msui/mss.sideview.cfg
-mslib/msui/mss.topview.cfg
-mslib/msui/msui_settings.py
-mslib/msui/wms_cache/
 mslib/mswms/mswms_settings.py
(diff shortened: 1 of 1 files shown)


ReimarBauer:
I think there can be more removed
```
mslib/mss_config.py
mslib/mscolab/colabdata/
mslib/msui/wms_cache/
mslib/msui/mss.sideview.cfg
mslib/msui/mss.topview.cfg
mslib/msui/msui_settings.py
```

ReimarBauer (changes requested):
see comment, 
grep them in the modules, if they are not there defined for these dirs, they can be removed.

nilupulmanodya:
> see comment, grep them in the modules, if they are not there defined for these dirs, they can be removed.

Just to clarify, do you mean to use like `grep -rnw . -e 'mslib/mss_config.py'` ? I don't understand the purpose of using grep for this. Could you please break this down bit?

ReimarBauer:
> grep -rnw . -e 'mslib/mss_config.py

I meant lookup if there are reasons to have these lines. I think they are outdated and were not removed.

ReimarBauer (approved):
Thx

PR #2098, opened 2023-11-15
Update actions used in the workflows

### LINKED ISSUE #2088: use checkout V4 for workflows
we maybe should update our workflows to https://github.com/actions/checkout v4

matrss:
Updates actions/checkout to v4 and actions/setup-python to v4. Some trailing whitespace was removed by my editor as well.

**Purpose of PR?**:

Fixes #2088 

No

### DIFF
--- a/.github/workflows/build_docs_gallery.yml
+++ b/.github/workflows/build_docs_gallery.yml
@@ -1,6 +1,6 @@
 name: Build Gallery
 
-on: 
+on:
    pull_request:
      inputs:
         branch_name:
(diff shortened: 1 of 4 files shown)


ReimarBauer:
@matrss I think that should better go into stable and later merged into develop. I don't know when V4 is required and how long we need until we release 9.0.0.

matrss:
I am pretty sure that actions/checkout uses standard git commands to fetch the repository and v3 will therefore continue to work for a long time. I don't think we have to worry about this if we will release 9.0.0 in the next 10 years. If it ever does stop working we can still do a backport.

ReimarBauer (approved):

PR #2099, opened 2023-11-15
refactor wait until process signals instead pf sleep

### LINKED ISSUE #2074: Do not use `sleep` as a synchronization primitive
> https://github.com/Open-MSS/MSS/blob/f6b3ad110dbfaa12d439cadea6476dcd54d9328d/mslib/mscolab/mscolab.py#L278
> Not really optimal to use sleep as a synchronization primitive here (ideally it should wait until `process` signals that it is ready in some way, e.g. when process' stdout contains something that indicates that the server is up and running). This is not a very critical code path though, so not too much of a concern.

_Originally posted by @matrss in https://github.com/Open-MSS/MSS/pull/2069#discussion_r1384852664_

nilupulmanodya:
fixes #2074

### DIFF
--- a/mslib/mscolab/mscolab.py
+++ b/mslib/mscolab/mscolab.py
@@ -34,7 +34,6 @@ import sys
 import secrets
 import subprocess
 import time
-import socket
 import git
 
 from mslib import __version__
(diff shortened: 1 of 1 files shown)


matrss:
Thanks for the PR! I think this is a bit more complicated than it needs to be though.

What I had in mind initially was to check stdout of the mscolab process for something that tells us that it is ready to accept connections. Unfortunately, it does not do that (the last thing it prints is "INFO: Launching MSColab Server", but that is from before it actually has set up its listening stuff so it does not tell us that it is ready to receive a connection).

Retrying is the next best thing, as you have implemented. curl includes an option for that though (`curl --retry 5 --retry-connrefused`, possibly varying `--retry-delay` as well). I think it would be better to just use that instead of implementing the retry logic again.

nilupulmanodya:
> Thanks for the PR! I think this is a bit more complicated than it needs to be though.
> 
> What I had in mind initially was to check stdout of the mscolab process for something that tells us that it is ready to accept connections. Unfortunately, it does not do that (the last thing it prints is "INFO: Launching MSColab Server", but that is from before it actually has set up its listening stuff so it does not tell us that it is ready to receive a connection).
> 
> Retrying is the next best thing, as you have implemented. curl includes an option for that though (`curl --retry 5 --retry-connrefused`, possibly varying `--retry-delay` as well). I think it would be better to just use that instead of implementing the retry logic again.

Thanks :)

matrss (approved):

PR #2100, opened 2023-11-16
Do not retry failed tests in CI

### LINKED ISSUE #2112: Never rerun failed tests
Currently, the CI test suites (both serial and with xdist) rerun failed tests multiple times. This blanket rerun policy masks a lot of genuine issues in the tests (e.g. I introduced some changes to the mscolab related tests in #1967 which led to those tests consistently ERROR'ing, but that was somehow hidden by the reruns). There are also some segfaults and intermittent issues that go unnoticed. We shouldn't rerun any tests, but accept those that fail on their first run as either genuinely failing tests that need a fix in the application part or as broken tests that need a fix.

When we have this sorted out we could follow this up by not running the tests in reverse order, but in random order (using pytest-randomly or similar). For the xdist tests we could try to get the test suite running without `--dist loadfile` or with `--dist worksteal`, which should avoid more possible inter-test dependencies.

matrss:
Tests that do not consistently pass are broken tests and should be fixed. Failed tests should not be retried to avoid accidentally adding new broken tests.

This is more of an experiment if this is achievable and WIP.


matrss:
I think some of the breakage between tests happens because the paths in `tests.constants` are shared between all tests in a session and some tests write to these paths, influencing other tests. It might make sense to turn that module into a fixture and therefore make these paths test-specific.

matrss:
All tests cleanup any remaining widgets and windows after themselves now and this is enforced through an autouse fixture. It looks like this has fixed issues with seemingly unrelated tests failing due to interfering Qt signals.

There are some heisenbugs remaining and the fact that the non-xdist pytest run hangs until the GitHub actions timeout is reached. I have a guess that the timeout might be related to improper cleanup of multiprocessing.Process'es though.

matrss:
The non-xdist test suite seems to be good now. The xdist-enabled test suite still has some flaky tests and I am not sure why yet.

EDIT: maybe not...

ReimarBauer:
the hdf failures are also strange. 

We check if a port can be used and there should be nothing to get in between when we use it then. You could try if it makes a difference not checking for the serial none xdist tests.

I introduced the xdist tests to get a faster result. When we get the normal tests faster, we can set them optional in the docs.

matrss:
For what it's worth, I had the serial test suite running over night on my laptop and it managed to complete all tests 74 times without any failures, and it is still running (EDIT: 75th run had a failure...). ~I am therefore confident that the test suite is pretty robust now, in the serial and reverse case.~ The CI error seems to be just another case of the GitHub runners being rather slow and running into a timeout that is a bit too tight. Specifically, it ran into the 10 second timeout of https://github.com/Open-MSS/MSS/blob/f8dad9c598bca19ea980cd5d59dcbd9665c18bfd/tests/utils.py#L188
and I think that can be increased a bit. I am also wondering if we even want these additional timeouts when we already wrap everything with pytest-timeout...

matrss:
> We check if a port can be used and there should be nothing to get in between when we use it then. You could try if it makes a difference not checking for the serial none xdist tests.

I am pretty sure that the port selection mechanism is correct under the assumption that no other software is trying to get one of those ports (at least for the serial test suite and with xdist's loadfile, other `--dist` options should have issues). Ideally I would like for it to just use port 0, let the OS pick a suitable free port and discover it programmatically, but I think that would require a few more extensive changes.

tests/utils.py
```diff
@@ -219,16 +219,13 @@ def mscolab_start_server(all_ports, mscolab_settings=mscolab_settings, timeout=1
         args=(_app, sockio, cm, fm,),
         kwargs={'port': port})
     process.start()
```
ReimarBauer: The AI had also an idea on that: 
Consider including some try/except blocks or error checking to provide useful error messages if something goes wrong.
```
try:
    process = multiprocessing.Process(
        target=start_server,
        args=(_app, sockio, cm, fm,),
        kwargs={'port': port})
    process.start()
except Exception as e:
    logging.error(f"Error starting server: {str(e)}")
```
If the server for whatever reason did not start, the while loop is an additional delay. 
We maybe get an idea why it is not starting.

matrss:
~There seems to be something wrong in `mslib/msui/mpl_{map,qtwidget}.py`. Those tend to always turn up in the stacktrace when there is a segfault in the tests, and ultimately it seems to be coming from somewhere in basemap.~

This was an intermittent error introduced by a `del` call in a teardown method.

matrss:
I have a suspicion that many of the remaining intermittent failures are due to the mscolab process running in a multiprocessing.Process, something which the used eventlet wsgi server simply does not support.

matrss:
The serial tests passed four times in CI now and another 75 times overnight on my laptop, so it looks like that is pretty stable now (famous last words after the last time, but whatever).

matrss:
There is a race condition in mslib.mscolab.file_manager.py::FileManager::save_file. If the server is running using multiple threads (e.g. through werkzeug.serving.make_server with threaded=True) then it can happen that two threads try to write to the same "main.ftml" file, corrupting the file in the process. This led to the tests in tests._test_msui.test_mscolab_version_history.py randomly failing, because the corrupted file produced a mslib.mscolab.models.Change in the database pointing to a commit sha that showed a corrupted file that read back as being empty.

This does not happen with the eventlet wsgi server since that one is running single-threaded (at least by default). Depending on how mscolab is deployed in production this issue might appear there though.

matrss:
@ReimarBauer the general idea of the test suite overhaul here is finished and if you want you can review some or all of it.

Almost all tests pass reliably now, both in the serial test suite and the parallel one.

There is a segfault issue remaining that I do not understand yet and which might be the same bug as or at least related to #2107. Do you have any idea on how that could be fixed?

I'll keep this as a draft for now because I am not sure yet if I want to rewrite the commit history to make a little more sense before this is considered to be merged (and I would really like to get this segfault bug fixed as well).

ReimarBauer (commented):
the development docs needs to become adopted, e.g. removed modules
https://github.com/Open-MSS/MSS/blob/develop/docs/development.rst#run-tests

localbuild/meta.yaml
```diff
@@ -77,7 +77,7 @@ requirements:
     - PyMySQL >=0.9.3
     - validate_email
     - multidict
-    - pint
+    - pint <0.23
```
ReimarBauer: do we have an issue what the newer pint does?

matrss: Yes, there were a bunch of test failures after its release last week. I haven't yet looked into what exactly broke though.

This is one reason why https://github.com/conda-forge/mss-feedstock/issues/162 would be a good idea.

tests/utils.py
matrss: I disagree, that try-except would make it worse. This process is started in a pytest fixture, if an exception occurs there then this will be reported as an test error for the test that requested the fixture, with the same output that pytest does for exceptions usually. I.e. similar logging to what was suggested. If we catched it and just logged, then we would run into the loop afterwards and introduce the additional delay.

I am also not sure which exceptions are even possible to come from Process and .start, this would catch none of the possible exceptions in the application code.

matrss: I just realized your comment was from 3 weeks ago and it was not yet a fixture then. It is now though and the same applied then as well, i.e. just letting the exception bubble up to pytest provides sufficient information and logging would make it less visible.

requirements.d/development.txt
```diff
@@ -6,23 +6,16 @@ py
...
 pynco
 conda-verify
-Flask-Testing
 pytest-reverse
-eventlet>0.30.2
 dnspython>=2.0.0, <2.3.0
 gsl==2.7.0
 boa
 xmlschema<2.5.0
-
```
ReimarBauer, line 28: We need any file that is not empty  to end with a new-line character, this is relevant for all POSIX systems and I think it comes from that historical decission http://gcc.gnu.org/ml/gcc/2003-11/msg01568.html

I can't  see hidden chars here, but we have to take care for this.

tests/_test_mscolab/test_files_api.py
```diff
@@ -24,49 +24,25 @@
...
 
-from mslib.mscolab.conf import mscolab_settings
 from mslib.mscolab.models import Operation
-from mslib.mscolab.server import APP
-from mslib.mscolab.file_manager import FileManager
 from mslib.mscolab.seed import add_user, get_user
-from mslib.mscolab.mscolab import handle_db_reset
 
 
-@pytest.mark.skipif(os.name == "nt",
```
ReimarBauer, line 40: I check the tests on the windows tests system next day. I am not sure if the skip is needed or not.

tests/fixtures.py
```diff
@@ -0,0 +1,152 @@
+# -*- coding: utf-8 -*-
+"""
+
+    tests.fixtures
+    ~~~~~~~~~~~
```
ReimarBauer: underline as long as the text (maybe a font problem here on gh)

mslib/msui/msui_mainwindow.py
```diff
@@ -910,7 +912,7 @@ def create_view(self, _type, model):
         view_window = None
         if _type == "topview":
             # Top view.
-            view_window = topview.MSUITopViewWindow(mainwindow=self, model=model,
+            view_window = topview.MSUITopViewWindow(parent=self, mainwindow=self, model=model,
```
ReimarBauer: this must be checked that we don't create a regression with the issue of https://github.com/Open-MSS/MSS/pull/2038

mslib/msui/topview.py
```diff
@@ -191,6 +191,8 @@ def __init__(self, parent=None, mainwindow=None, model=None, _id=None,
         """
         Set up user interface, connect signal/slots.
         """
+        if parent is None:
+            parent = mainwindow
```
ReimarBauer: This looks like undoing the changes for
https://github.com/Open-MSS/MSS/issues/2007

Likly the fix was not good, but it needs to stay solved

ReimarBauer (commented):
I have noticed a few things, I can check tomorrow if windows works with the tests.

localbuild/meta.yaml
ReimarBauer: there will be soon a 8.3.3 including this and then stable merged, so it is matched with the issue for this.

matrss: The commit will be dropped by itself when I rebase the next time then.

requirements.d/development.txt
matrss: This file is not missing a newline character at the end, but it previously had two. My editor automatically drops duplicate newlines at the end of file. GitHub shows a red sign if there is indeed a missing newline, as you can see e.g. here: https://github.com/Open-MSS/MSS/pull/2069/files/a2e10e1ce75daa73306a7e1076f124a3ad8501da#r1384630111 (this instance was fixed later).

tests/_test_mscolab/test_files_api.py
matrss: At least the stated reason for those skips became obsolete (start_method "fork" is no longer used), that's why I removed the skips. We should consider making use of GitHub's windows and macos runners to run the test suite there automatically.

tests/fixtures.py
matrss: Not a font problem, I copied the header from tests.utils and forgot to extend the underline. I'll fix that.

mslib/msui/msui_mainwindow.py
matrss: Thanks for the pointer. It would have been nice if #2038 included a regression test for this, but you catched it anyway.

I think part of the underlying problem stated in #2007 is that MSUITopViewWindow is a QDialog, which will be on top of its parent window since that is a standard behavior for dialogs. Removing the parent disables that standard behavior since there is no parent to "attach" to, but ultimately I think QDialog is just the wrong class to inherit from in this case, since the window is not intended to behave like a dialog then.

That said, MSUITopViewWindow is one of the sources of segfaults that I am still seeing and I'll look into if this can be done in a different way (Adding the parent will ultimately lead to the object being deleted earlier than without a parent and I think that at least increases the frequency of a use-after-free bug somewhere in there. The sideview and linearview classes also do not set a parent, but that does not come out as an error in the test suite since they are only tested by themselves, not together with a main window, and they exhibit a similar use-after-free issue much less frequently.).

mslib/msui/topview.py
matrss: See: https://github.com/Open-MSS/MSS/pull/2100#discussion_r1441545102

tests/_test_mscolab/test_files_api.py
ReimarBauer: yes, I used them also in the https://github.com/Open-MSS/mss-install too.

ReimarBauer (commented):
on windows testing I get

```
tests\_test_mscolab\test_mscolab.py:86:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
mslib\mscolab\mscolab.py:83: in handle_db_reset
    shutil.rmtree(mscolab_settings.DATA_DIR)
..\..\..\..\miniforge\envs\mssdev\lib\shutil.py:749: in rmtree
    return _rmtree_unsafe(path, onerror)
..\..\..\..\miniforge\envs\mssdev\lib\shutil.py:614: in _rmtree_unsafe
    _rmtree_unsafe(fullname, onerror)
..\..\..\..\miniforge\envs\mssdev\lib\shutil.py:614: in _rmtree_unsafe
    _rmtree_unsafe(fullname, onerror)
..\..\..\..\miniforge\envs\mssdev\lib\shutil.py:623: in _rmtree_unsafe
    onerror(os.rmdir, path, sys.exc_info())
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

path = 'C:/Users/LOCAL_~1.BAU/Temp/12/tmp9pf68tuvmsuid1f91249b4/colabTestData\\filedata\\Example'
onerror = <function rmtree.<locals>.onerror at 0x000001F40AA4E9E0>

    def _rmtree_unsafe(path, onerror):
        try:
            with os.scandir(path) as scandir_it:
                entries = list(scandir_it)
        except OSError:
            onerror(os.scandir, path, sys.exc_info())
            entries = []
        for entry in entries:
            fullname = entry.path
            if _rmtree_isdir(entry):
                try:
                    if entry.is_symlink():
                        # This can only happen if someone replaces
                        # a directory with a symlink after the call to
                        # os.scandir or entry.is_dir above.
                        raise OSError("Cannot call rmtree on a symbolic link")
                except OSError:
                    onerror(os.path.islink, fullname, sys.exc_info())
                    continue
                _rmtree_unsafe(fullname, onerror)
            else:
                try:
                    os.unlink(fullname)
                except OSError:
                    onerror(os.unlink, fullname, sys.exc_info())
        try:
>           os.rmdir(path)
E           PermissionError: [WinError 32] Der Prozess kann nicht auf die Datei zugreifen, da sie von einem anderen Prozess verwendet wird: 'C:/Users/LOCAL_~1.BAU/Temp/12/tmp9pf68tuvmsuid1f91249b4/colabTestData\\filedata\\Example'

..\..\..\..\miniforge\envs\mssdev\lib\shutil.py:621: PermissionError
```

tests/_test_mscolab/test_files_api.py
ReimarBauer: test_files_api on windows is a lot better
```
(mssdev) C:\Users\r.bauer\PycharmProjects\2024\matrss\MSS>pytest -k test_files_api
========================================================= test session starts =========================================================
platform win32 -- Python 3.10.8, pytest-7.4.4, pluggy-1.3.0
rootdir: C:\Users\r.bauer\PycharmProjects\2024\matrss\MSS
configfile: pytest.ini
plugins: anyio-4.2.0, cov-4.1.0, reverse-1.7.0, timeout-2.2.0, xdist-3.5.0
timeout: 60.0s
timeout method: thread
timeout func_only: False
collected 541 items / 526 deselected / 1 skipped / 15 selected

tests/_test_mscolab/test_files_api.py::Test_Files::test_create_operation PASSED                                                  [  6%]
tests/_test_mscolab/test_files_api.py::Test_Files::test_list_operations ERROR                                                    [ 13%]
tests/_test_mscolab/test_files_api.py::Test_Files::test_get_operation_details PASSED                                             [ 20%]
tests/_test_mscolab/test_files_api.py::Test_Files::test_get_authorized_users PASSED                                              [ 26%]
tests/_test_mscolab/test_files_api.py::Test_Files::test_fetch_users_without_permission PASSED                                    [ 33%]
tests/_test_mscolab/test_files_api.py::Test_Files::test_fetch_users_with_permission PASSED                                       [ 40%]
tests/_test_mscolab/test_files_api.py::Test_Files::test_add_bulk_permissions ERROR                                               [ 46%]
tests/_test_mscolab/test_files_api.py::Test_Files::test_modify_bulk_permissions PASSED                                           [ 53%]
tests/_test_mscolab/test_files_api.py::Test_Files::test_delete_bulk_permissions PASSED                                           [ 60%]
tests/_test_mscolab/test_files_api.py::Test_Files::test_import_permissions ERROR                                                 [ 66%]
tests/_test_mscolab/test_files_api.py::Test_Files::test_update_operation FAILED                                                  [ 73%]
tests/_test_mscolab/test_files_api.py::Test_Files::test_delete_operation FAILED                                                  [ 80%]
tests/_test_mscolab/test_files_api.py::Test_Files::test_get_all_changes PASSED                                                   [ 86%]
tests/_test_mscolab/test_files_api.py::Test_Files::test_get_change_content PASSED                                                [ 93%]
tests/_test_mscolab/test_files_api.py::Test_Files::test_set_version_name ERROR                                                   [100%]
```
The errors all the same, like.

```
______________________________________ ERROR at setup of Test_MscolabVersionHistory.test_changes ______________________________________

mscolab_session_managers = (<Flask 'mslib.mscolab.app'>, <flask_socketio.SocketIO object at 0x000001FE6A2BDAB0>, <mslib.mscolab.chat_manager.ChatManager object at 0x000001FE6C24CD30>, <mslib.mscolab.file_manager.FileManager object at 0x000001FE6C24E740>)

    @pytest.fixture(scope="session")
    def mscolab_session_server(mscolab_session_managers):
        app, _, _, _ = mscolab_session_managers
        scheme = "http"
        host = "127.0.0.1"
        server = werkzeug.serving.make_server(host=host, port=0, app=app, threaded=True)
        port = server.socket.getsockname()[1]
        process = multiprocessing.Process(target=server.serve_forever, daemon=True)
>       process.start()

tests\fixtures.py:114:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
..\..\..\..\miniforge\envs\mssdev\lib\multiprocessing\process.py:121: in start
    self._popen = self._Popen(self)
..\..\..\..\miniforge\envs\mssdev\lib\multiprocessing\context.py:224: in _Popen
    return _default_context.get_context().Process._Popen(process_obj)
..\..\..\..\miniforge\envs\mssdev\lib\multiprocessing\context.py:336: in _Popen
    return Popen(process_obj)
..\..\..\..\miniforge\envs\mssdev\lib\multiprocessing\popen_spawn_win32.py:93: in __init__
    reduction.dump(process_obj, to_child)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

obj = <Process name='Process-1' parent=13184 initial daemon>, file = <_io.BufferedWriter name=17>, protocol = None

    def dump(obj, file, protocol=None):
        '''Replacement for pickle.dump() using ForkingPickler.'''
>       ForkingPickler(file, protocol).dump(obj)
E       _pickle.PicklingError: Can't pickle <function db at 0x000001FE6A1F8C10>: it's not the same object as flask_migrate.cli.db

..\..\..\..\miniforge\envs\mssdev\lib\multiprocessing\reduction.py:60: PicklingError
_________________________________ ERROR at setup of Test_MscolabVersionHistory.test_set_version_name __________________________________

```

or something like this

```
__________________________________________ ERROR at setup of Test_Files.test_list_operations __________________________________________

self = TempFS(), path = '/colabTestData/filedata/f3'

    def removedir(self, path):
        # type: (Text) -> None
        self.check()
        _path = self.validatepath(path)
        if _path == "/":
            raise errors.RemoveRootError()
        sys_path = self._to_sys_path(path)
        with convert_os_errors("removedir", path, directory=True):
>           os.rmdir(sys_path)
E           PermissionError: [WinError 32] Der Prozess kann nicht auf die Datei zugreifen, da sie von einem anderen Prozess verwendet wird: b'C:\\Users\\LOCAL_~1.BAU\\Temp\\12\\tmpeopjb7sgmsuid1f91249b4\\colabTestData\\filedata\\f3'

..\..\..\..\miniforge\envs\mssdev\lib\site-packages\fs\osfs.py:388: PermissionError
```

ReimarBauer: I think we should first merge and afterwards look on other OS. Here already an amount of good changes.

matrss:
Good to know. This failure should be unrelated to the `skipif` removals, since there was none in that file. Maybe the refactoring to remove flask_testing broke it. Can you confirm that this error does not happen on develop?

ReimarBauer:
> Good to know. This failure should be unrelated to the `skipif` removals, since there was none in that file. Maybe the refactoring to remove flask_testing broke it. Can you confirm that this error does not happen on develop?

In develop are not enough skipped, all which have that  _- ValueError: cannot find context for 'fork'_ can't succeed on windows. 
```
tests/_test_mscolab/test_mscolab.py::test_confirm_action PASSED                                                  [  2%]
tests/_test_mscolab/test_mscolab.py::test_main PASSED                                                            [  4%]
tests/_test_mscolab/test_mscolab.py::Test_Mscolab::test_handle_db_reset FAILED                                   [  6%]
FAILED tests/_test_mscolab/test_mscolab.py::Test_Mscolab::test_handle_db_reset - PermissionError: [WinError 32] Der Prozess kann nicht auf die Datei zugreifen, da sie von einem anderen Prozess ver...
ERROR tests/_test_msui/test_mscolab.py::Test_Mscolab_connect_window::test_url_combo - PermissionError: [WinError 32] Der Prozess kann nicht auf die Datei zugreifen, da sie von einem anderen Prozess ver...
ERROR tests/_test_msui/test_mscolab.py::Test_Mscolab_connect_window::test_connect_except[ConnectionError] - ValueError: cannot find context for 'fork'
ERROR tests/_test_msui/test_mscolab.py::Test_Mscolab_connect_window::test_connect_except[InvalidSchema] - ValueError: cannot find context for 'fork'
ERROR tests/_test_msui/test_mscolab.py::Test_Mscolab_connect_window::test_connect_except[InvalidURL] - ValueError: cannot find context for 'fork'
ERROR tests/_test_msui/test_mscolab.py::Test_Mscolab_connect_window::test_connect_except[SSLError] - ValueError: cannot find context for 'fork'
ERROR tests/_test_msui/test_mscolab.py::Test_Mscolab_connect_window::test_connect_except[Exception] - ValueError: cannot find context for 'fork'
ERROR tests/_test_msui/test_mscolab.py::Test_Mscolab_connect_window::test_connect_denied - ValueError: cannot find context for 'fork'
ERROR tests/_test_msui/test_mscolab.py::Test_Mscolab_connect_window::test_connect_success - ValueError: cannot find context for 'fork'
ERROR tests/_test_msui/test_mscolab.py::Test_Mscolab_connect_window::test_disconnect - ValueError: cannot find context for 'fork'
ERROR tests/_test_msui/test_mscolab.py::Test_Mscolab_connect_window::test_login - ValueError: cannot find context for 'fork'
ERROR tests/_test_msui/test_mscolab.py::Test_Mscolab_connect_window::test_login_with_different_account_shows_update_credentials_popup - ValueError: cannot find context for 'fork'
ERROR tests/_test_msui/test_mscolab.py::Test_Mscolab_connect_window::test_logout_action_trigger - ValueError: cannot find context for 'fork'
ERROR tests/_test_msui/test_mscolab.py::Test_Mscolab_connect_window::test_logout - ValueError: cannot find context for 'fork'
ERROR tests/_test_msui/test_mscolab.py::Test_Mscolab_connect_window::test_add_user - ValueError: cannot find context for 'fork'
ERROR tests/_test_msui/test_mscolab.py::Test_Mscolab_connect_window::test_add_users_without_updating_credentials_in_config_file - ValueError: cannot find context for 'fork'
ERROR tests/_test_msui/test_mscolab.py::Test_Mscolab_connect_window::test_add_users_with_updating_credentials_in_config_file - ValueError: cannot find context for 'fork'
=================== 1 failed, 3 passed, 27 skipped, 491 deselected, 4 warnings, 16 errors in 49.16s ===================
```

tests/_test_mscolab/test_files_api.py
matrss: > At least the stated reason for those skips became obsolete (start_method "fork" is no longer used), that's why I removed the skips. We should consider making use of GitHub's windows and macos runners to run the test suite there automatically.

I just realized that this is not true, "fork" is the default on linux. (I thought it wasn't because it was set explicitly. The default will be changed away from "fork" in python3.14 apparently.)

ReimarBauer:
and test_handle_db runs on windows in develop into the same 

```
    def _rmtree_unsafe(path, onerror):
        try:
            with os.scandir(path) as scandir_it:
                entries = list(scandir_it)
        except OSError:
            onerror(os.scandir, path, sys.exc_info())
            entries = []
        for entry in entries:
            fullname = entry.path
            if _rmtree_isdir(entry):
                try:
                    if entry.is_symlink():
                        # This can only happen if someone replaces
                        # a directory with a symlink after the call to
                        # os.scandir or entry.is_dir above.
                        raise OSError("Cannot call rmtree on a symbolic link")
                except OSError:
                    onerror(os.path.islink, fullname, sys.exc_info())
                    continue
                _rmtree_unsafe(fullname, onerror)
            else:
                try:
                    os.unlink(fullname)
                except OSError:
                    onerror(os.unlink, fullname, sys.exc_info())
        try:
>           os.rmdir(path)
E           PermissionError: [WinError 32] Der Prozess kann nicht auf die Datei zugreifen, da sie von einem anderen Prozess verwendet wird: 'C:/Users/LOCAL_~1.BAU/Temp/17/tmp_fquflnzmsui83b39851af/colabTestData\\filedata\\Example'
```

tests/_test_mscolab/test_files_api.py
ReimarBauer: last time I looked in those docs, spawn is default on MacOS and Windows. And MacOS can also use fork but not Windows.  It is set explicity because of MacOS, there it overwrites the default.

matrss: Yes, that is how it is described in the docs: https://docs.python.org/3/library/multiprocessing.html#contexts-and-start-methods, with the caveat that it will change in 3.14.

I am not sure if the issue with forking multithreaded processes that is mentioned there could affect us.

matrss:
@ReimarBauer Since I am not actually getting closer to having the tests reliably pass without retries I am considering splitting this up into multiple PRs to get at least some improvements merged already, e.g.:
- refactor mscolab and mswms servers into fixtures
- refactor Qt application setup into a fixture
- config reset for each test
- ...

What do you think?

ReimarBauer:
> @ReimarBauer Since I am not actually getting closer to having the tests reliably pass without retries I am considering splitting this up into multiple PRs to get at least some improvements merged already, e.g.:
> 
> * refactor mscolab and mswms servers into fixtures
> * refactor Qt application setup into a fixture
> * config reset for each test
> * ...
> 
> What do you think?

@matrss 
yes, there is already a lot good done in this PR, split it.

This is a big improvement when we can merge parts of it.

Issue #2078, opened 2023-11-08
user/profile creation by SAML2 should not have a password


ReimarBauer:
currently the token is used for the password in our database for a user created by a SAML2 account.
Using a SAML2 authority we should speak from a profile creation connected to an authority.

We should not store a password at all and also indicate that the account is managed by the SAML2 authority.
There should be never a password recovery over our system possible. 

When the authority gets invalid (various cases possible) the user should not get into the system without help of an admin.

nilupulmanodya:
> currently the token is used for the password in our database for a user created by a SAML2 account. Using a SAML2 authority we should speak from a profile creation connected to an authority.
> 
> We should not store a password at all and also indicate that the account is managed by the SAML2 authority. There should be never a password recovery over our system possible.
> 
> When the authority gets invalid (various cases possible) the user should not get into the system without help of an admin.

I trust your explanation regarding scenarios where a user logs into the system via SAML2 authority. However, my concern is this: if a user, who already exists and has transitioned to SAML2-based login, attempts to log in using password-based login again, wouldn't they need to go through a 'password recovery' process? This implies that once a user migrates to SAML2-based login, they will no longer be able to access password-based login. right ?

ReimarBauer:
I think there should become an server Admin involved.  We maybe need a change in a mscolab cli script to be used on the server for this case. 

When we look from the point of the natural person there are just too many possibilities why it can go wrong.
- the user changed the facility. 
- User changed name. 
- saml2 cert was not renewed in time

In all cases we have to know that the user we have has now new data and won't use the old data.

nilupulmanodya:
I would like to work on this. Please assign to me

ReimarBauer:
I expect you can claim it yourself, please try first.

PR #2118, opened 2023-12-13
gives the server control over the clients login options

### LINKED ISSUE #2076: Do not disable local account creation when an IdP is configured
> https://github.com/Open-MSS/MSS/blob/f6b3ad110dbfaa12d439cadea6476dcd54d9328d/mslib/msui/mscolab.py#L242-L245
> Shouldn't it still be possible to register a new non-IdP user on the MSColab instance, even if an IdP is configured? I feel like disabling "local" account registration should be a different config option (that would need to be enforced server-side).

_Originally posted by @matrss in https://github.com/Open-MSS/MSS/pull/2069#discussion_r1384874147_
            

> also in the workflow of profile/usercreation by the IDP is not completed. The new profile is not activated, it would need a validation (recover password attempt currently) first. There some changes we have to think about after we merged this in.

_Originally posted by @ReimarBauer in https://github.com/Open-MSS/MSS/pull/2069#discussion_r1384930546_

> We have to verify too that a user with a "None" Password because of IDP can't fetch one and update that 'None' over password recovery.  Some of these parts need a coverage by tests.

_Originally posted by @ReimarBauer in https://github.com/Open-MSS/MSS/pull/2069#discussion_r1384946717_

ReimarBauer:
**Purpose of PR?**:

Fixes #2076

Dependent of the server configuration a user login to a local database can be archieved.

### DIFF
--- a/mslib/mscolab/server.py
+++ b/mslib/mscolab/server.py
@@ -242,16 +242,19 @@ def hello():
             auth.login_required()
             return json.dumps({
                 'message': "Mscolab server",
-                'USE_SAML2': mscolab_settings.USE_SAML2
+                'USE_SAML2': mscolab_settings.USE_SAML2,
+                'direct_login': mscolab_settings.DIRECT_LOGIN
             })
         return json.dumps({
             'message': "Mscolab server",
-            'USE_SAML2': mscolab_settings.USE_SAML2
+            'USE_SAML2': mscolab_settings.USE_SAML2,
+            'direct_login': mscolab_settings.DIRECT_LOGIN
         })
     else:
         return json.dumps({
             'message': "Mscolab server",
-            'USE_SAML2': mscolab_settings.USE_SAML2
+            'USE_SAML2': mscolab_settings.USE_SAML2,
+            'direct_login': mscolab_settings.DIRECT_LOGIN
         })
 
 
--- a/docs/samples/config/mscolab/mscolab_settings.py.sample
+++ b/docs/samples/config/mscolab/mscolab_settings.py.sample
@@ -85,6 +85,9 @@ STUB_CODE = """<?xml version="1.0" encoding="utf-8"?>
 </FlightTrack>
 """
 
+# accounts on a database on the server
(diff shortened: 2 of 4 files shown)


matrss:
This should be enforced on the server side as well, i.e. disable all endpoints related to local user registration and login.

ReimarBauer:
There is more to do also after this is done https://github.com/Open-MSS/MSS/issues/2078

I added already on some places ToDo's. I think we do this step by step and at the end another one. Some parts are not identified which needs changes too.

mslib/mscolab/server.py
```diff
@@ -242,16 +242,19 @@ def hello():
             auth.login_required()
             return json.dumps({
                 'message': "Mscolab server",
-                'USE_SAML2': mscolab_settings.USE_SAML2
+                'USE_SAML2': mscolab_settings.USE_SAML2,
+                'direct_login': mscolab_settings.DIRECT_LOGIN
```
matrss, line 246: We should decide if keys in this json should be uppercase or lowercase, not mix them. I'd prefer lowercase.

nilupulmanodya: Yeah.. Thank you for figguring out that point. I believe that the `USE_SAML2` configuration pertains to a previous implementation and is not directly relevant to the current issue at hand. Consequently, it may be prudent to consider opening a separate issue, as the matter at hand is distinct and not associated with the present concern. This step is suggested to enhance clarity and maintain a focused approach to problem resolution.

nilupulmanodya:
We have TODOs. Given that the user login to a local database is achievable depending on the server configuration, I am granting approval for this pull request.

nilupulmanodya (approved):

mslib/mscolab/server.py
matrss: If we decide on lowercase then yes, this does not block this PR and should be an issue instead. I just wanted to bring it up, in case someone disagreed with me preferring lowercase.

matrss (approved):

PR #2119, opened 2023-12-13
using an IDP, email is already validated, we activate the profile

### LINKED ISSUE #2079: user/profile creation by SAML2 is valid without verification
When a user gets a profile on the MSColab by using a SAML2 account there should be no further verification from our site for the e-mail address. The account must be connected to the provider. The profile should give access to the service.

Currently a new user has to do a password reset to get that verification. This does not change the only valid password on the SAML2 provider. 

The userhandling needs refactoring and test coverage.

ReimarBauer:
**Purpose of PR?**:

Fixes #2079

When an IDP user accesses an MSColab server first time, the e-mail is verified by the IDP and we create a valid user.
confirmed_on +1 second

### DIFF
--- a/mslib/mscolab/server.py
+++ b/mslib/mscolab/server.py
@@ -220,7 +220,9 @@ def create_or_update_idp_user(email, username, token, authentication_backend):
     """
     user = User.query.filter_by(emailid=email).first()
     if not user:
-        user = User(email, username, password=token, confirmed=False, confirmed_on=None,
+        # using an IDP for a new account/profile, e-mail is already verified by the IDP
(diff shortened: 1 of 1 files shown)


matrss:
Does the assumption that the mail address is verified by the IdP always hold true?

ReimarBauer:
This is discussable, maybe also configurable. Currently I try to do it similiar as nextcloud does.

Login by IDP means also to get an user profile and directly login. For the local user we have add user, configurable validation by email. if that is enabled login is only possible when it is validated.
By the VO configuration we used on the NextCloud, only one who accepted to be in the VO could login.

There may be different use cases - we will see when there is a different backend available.

matrss (approved):

PR #2121, opened 2023-12-13
refactor not to store passwords for none local authentications

### LINKED ISSUE #2078: user/profile creation by SAML2 should not have a password

ReimarBauer:
**Purpose of PR?**:

ideas for 2078

first step for handling userdata differently. Also a few different problems solved.

### DIFF
--- a/mslib/mscolab/models.py
+++ b/mslib/mscolab/models.py
@@ -39,17 +39,28 @@ class User(db.Model):
     id = db.Column(db.Integer, primary_key=True, autoincrement=True)
     username = db.Column(db.String(255))
     emailid = db.Column(db.String(255), unique=True)
-    password = db.Column(db.String(255), unique=True)
+    password = db.Column(db.String(255))
     registered_on = db.Column(db.DateTime, nullable=False)
     confirmed = db.Column(db.Boolean, nullable=False, default=False)
     confirmed_on = db.Column(db.DateTime, nullable=True)
     permissions = db.relationship('Permission', cascade='all,delete,delete-orphan', backref='user')
     authentication_backend = db.Column(db.String(255), nullable=False, default='local')
 
-    def __init__(self, emailid, username, password, confirmed=False, confirmed_on=None, authentication_backend='local'):
+    def __init__(self, emailid, username, password=None, confirmed=False,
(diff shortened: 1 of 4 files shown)


mslib/mscolab/models.py
```diff
@@ -66,25 +77,28 @@ def verify_password(self, password_):
...
-            token = jwt.encode(
-                {
-                    "id": self.id,
-                    "exp": datetime.datetime.now(tz=datetime.timezone.utc) + datetime.timedelta(seconds=expiration)
-                },
-                mscolab_settings.SECRET_KEY,
-                algorithm="HS256"
-            )
-            return token
+        token = jwt.encode(
```
ReimarBauer, line 86: solves,  setting an expiration
the old code returned None

mslib/mscolab/models.py
```diff
@@ -39,17 +39,28 @@ class User(db.Model):
     id = db.Column(db.Integer, primary_key=True, autoincrement=True)
     username = db.Column(db.String(255))
     emailid = db.Column(db.String(255), unique=True)
-    password = db.Column(db.String(255), unique=True)
```
ReimarBauer, line 42: users should not but it can be they have chooen a password identical to some other user. Telling this is not a good idea.

mslib/mscolab/server.py
```diff
@@ -153,19 +153,23 @@ def check_login(emailid, password):
 
 
 def register_user(email, password, username):
-    user = User(email, username, password)
```
ReimarBauer, line 156: we should only create the user object when we have valid data

ReimarBauer:
next is to lookup / understand how the token exchange between server and client can be temporary bound to a user profile.

ReimarBauer:
I do split this into further issues

ReimarBauer:
many of the PR was splitted into smaller ones, the main issue is still open.

PR #2131, opened 2024-01-05
Merge stable to develop

ReimarBauer:
**Purpose of PR?**:

This needs a merge not a squash and merge.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,14 @@
 Changelog
 =========
 
+Version 8.3.2
+~~~~~~~~~~~~~
+
+Bug fix release:
+
+All changes:
+https://github.com/Open-MSS/MSS/milestone/100
+
 Version 8.3.1
 ~~~~~~~~~~~~~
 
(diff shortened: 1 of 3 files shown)


CHANGES.rst
```diff
@@ -1,6 +1,14 @@
 Changelog
 =========
 
+Version 8.3.2
+~~~~~~~~~~~~~
+
+Bug fix release:
+
+All changes:
+https://github.com/Open-MSS/MSS/milestone/100
```
matrss, line 10: Should this include `?closed=1`, as in many previous changelog entries? Also, the milestone is not exhaustive, i.e. the issue/PR for the pint pin is not listed there.

ReimarBauer: yes, but that was a failure which was not seen on the release creation. this is a merge. we don't do changes in a merge.

ReimarBauer: that can be fixed when we do a change on docs, or release later.

matrss: OK, and I just realized that the pint pin change is not part of 8.3.2 but will be in 8.3.3 instead.

matrss (approved):

CHANGES.rst
ReimarBauer: yes, on conda-forge it is already in the build -1, this is to update our code base. 
develop missed a version string merge.

https://anaconda.org/conda-forge/mss/files

PR #2136, opened 2024-01-08
Reset config directory before each test

matrss:
Reset the config directory content before each test to reduce tests that affect each other.

Extracted from #2100.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -40,9 +40,8 @@ import keyring
 from mslib.mswms.demodata import DataFiles
 import tests.constants as constants
 
-# make a copy for mscolab test, so that we read different pathes during parallel tests.
-sample_path = os.path.join(os.path.dirname(__file__), "tests", "data")
-shutil.copy(os.path.join(sample_path, "example.ftml"), constants.ROOT_DIR)
+# This import must come after importing tests.constants due to MSUI_CONFIG_PATH being set there
+from mslib.utils.config import read_config_file
 
 
 class TestKeyring(keyring.backend.KeyringBackend):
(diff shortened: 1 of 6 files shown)


conftest.py
```diff
@@ -185,40 +195,61 @@ def pytest_generate_tests(metafunc):
...
+            # windows needs \\ or / but mixed is terrible. *nix needs /
+            mscolab_fs.writetext(constants.MSCOLAB_CONFIG_FILE, config_string.replace('\\', '/'))
+        path = fs.path.join(constants.ROOT_DIR, 'mscolab', constants.MSCOLAB_CONFIG_FILE)
+        parent_path = fs.path.join(constants.ROOT_DIR, 'mscolab')
+
+    if not constants.SERVER_CONFIG_FS.exists(constants.MSCOLAB_AUTH_FILE):
+        config_string = '''
 import hashlib
 
 class mscolab_auth(object):
```
ReimarBauer, line 211: the (object)  because default, could maybe become removed

tests/_test_msui/test_mscolab.py
```diff
@@ -52,7 +51,6 @@
 class Test_Mscolab_connect_window():
```
ReimarBauer, line 51: the () maybe become removed

conftest.py
matrss: I did a tree-wide removal of those.

tests/_test_msui/test_mscolab.py
matrss: I did a tree-wide removal of those.

matrss:
Maybe I should move the last two commits to their own PR as well, since they aren't really on topic here? At least I think a squash would be inappropriate with those in here.

ReimarBauer:
> Maybe I should move the last two commits to their own PR as well, since they aren't really on topic here? At least I think a squash would be inappropriate with those in here.

hmm, yes, sorry, just seen that here. That could be maybe also a good first issue. Somehow I thought only on the few in the files I saw.

ReimarBauer:
@matrss ignore the ui_*.py files. Don't change. They are autocreated by pyuic5.

conftest.py
```diff
@@ -97,16 +96,27 @@ def pytest_generate_tests(metafunc):
...
+        constants.ROOT_FS.makedirs("msui/testdata")
+
+    # make a copy for mscolab test, so that we read different pathes during parallel tests.
+    sample_path = os.path.join(os.path.dirname(__file__), "tests", "data")
+    shutil.copy(os.path.join(sample_path, "example.ftml"), constants.ROOT_DIR)
+
+    if not constants.SERVER_CONFIG_FS.exists(constants.SERVER_CONFIG_FILE):
+        print('\n configure testdata')
+        # ToDo check pytest tmpdir_factory
+        examples = DataFiles(data_fs=constants.DATA_FS,
```
ReimarBauer, line 113: a note: this is very expensive.

The datafiles are only readed. Once created for each used core may be enough. 

per test we create 68MB, by 515 tests that is 35GB. We should find a better solution

Without using the function create_data we have
`93 failed, 412 passed, 23 skipped, 111 warnings, 11 errors in 182.17s (0:03:02)`
So we need this only for 20% of the tests.

Maybe we should mark tests where we need this data and create it only then.

matrss: > a note: this is very expensive.

It seems to add about ~3 minutes to the test duration in the serial test suite.

I tried to keep the changes relatively minor here to avoid breaking even more, until the test suite is in a somewhat sane state (i.e. passes without retries). Ideally this directory would be split up and e.g. those datafiles should become their own fixture that can then be scoped to the session.

> Once created for each used core may be enough.

This would happen automatically then, if it was a session-scoped fixture.

> Maybe we should mark tests where we need this data and create it only then.

This would happen automatically with a fixture as well, since every test that needs it would have to request the fixture (that is if we don't make it `autouse` and of course, if it is session-scoped it wouldn't really matter how many tests require it).

As it stands I think the impact is not too high on the test duration and refactoring this into a fixture can be a future PR.

ReimarBauer: ok, it is not only the time, we also create and drop a huge amount of data which we should optimize.

ReimarBauer (approved):
Thx

PR #2140, opened 2024-01-09
Remove unnecessary development dependencies

matrss:
We have a GitHub Action that runs flake8, which in turn wraps PyFlakes, pycodestyle and McCabe. Since pep8 was renamed to pycodestyle those two are the same. Adding flake8 to the dependencies therefore means we can remove pep8 and pycodestyle. The pytest-pep8, which is really old, and pytest-flake8 plugins also become redundant since we have the CI action anyway.

Extracted from #2100.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -34,25 +34,7 @@ omit =
  tests/*
 
 [tool:pytest]
-addopts = --flake8
-flake8-max-line-length = 120
-flake8-ignore =
- *.py E402 W504 N801 N802 N803 N805 N806 N813
- conftest.py F821
- setup.py F821
- docs/conf.py ALL
- mslib/__init__.py F401
- mslib/msui/mss_qt.py F401  # lots of imports for importing code
- mslib/msui/mss_pyui.py F401  # nappy imported for testing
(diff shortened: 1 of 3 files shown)


ReimarBauer (commented):
this needs maybe also a change in the developer docs

matrss:
First of all I would like to know why this makes the serial test suite time out... At least it did twice already, trying one more time.

matrss:
With the changes that were already merged this is no longer running into said timeout.

matrss:
Added a removal of pytest-cache, since that was integrated into pytest in 2015, and updated the docs.

setup.cfg
```diff
@@ -34,25 +34,7 @@ omit =
  tests/*
 
 [tool:pytest]
-addopts = --flake8
-flake8-max-line-length = 120
```
ReimarBauer, line 38: is there another place where we set the line length?

matrss: Yes, further down in the flake8 section. See the full file: https://github.com/Open-MSS/MSS/blob/5a8d1c83ef2b763696fe94f93a424f52bfbb2d05/setup.cfg

ReimarBauer (approved):
thx

PR #2141, opened 2024-01-09

### LINKED ISSUE #2132: keys in json shoud be lowercase written

Beeram12:
renamed USE_SAML2 to use_saml2 lowered all .

Fixes #2132

### DIFF
--- a/mslib/mscolab/conf.py
+++ b/mslib/mscolab/conf.py
@@ -115,7 +115,7 @@ class default_mscolab_settings:
     GDPR = None
 
     # enable login by identity provider
-    USE_SAML2 = False
+    use_saml2 = False
 
     # Enable SSL certificate verification during SSO between MSColab and IdP
     ENABLE_SSO_SSL_CERT_VERIFICATION = True
--- a/mslib/mscolab/mscolab.py
+++ b/mslib/mscolab/mscolab.py
@@ -264,7 +264,7 @@ def handle_mscolab_metadata_init(repo_exists):
         This will generate necessary metada data file for sso in mscolab through localhost idp
 
         Before running this function:
-        - Ensure that USE_SAML2 is set to True.
+        - Ensure that use_saml2 is set to True.
         - Generate the necessary keys and certificates and configure them in the .yaml
         file for the local IDP.
     """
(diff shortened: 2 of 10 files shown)


ReimarBauer (changes requested):
Please rename only the USE_SAML2 in the json output. Not any constant we have.

mslib/mscolab/server.py
```diff
@@ -242,16 +242,16 @@ def hello():
             auth.login_required()
             return json.dumps({
                 'message': "Mscolab server",
-                'USE_SAML2': mscolab_settings.USE_SAML2
+                'use_saml2': mscolab_settings.use_saml2
```
ReimarBauer, line 245: this is json :)

mslib/mscolab/server.py
```diff
@@ -732,7 +732,7 @@ def reset_request():
         return render_template('errors/403.html'), 403
 
 
-if mscolab_settings.USE_SAML2:
+if mscolab_settings.use_saml2:
```
ReimarBauer, line 735: this is a constant, this should not be renamed

mslib/mscolab/mscolab.py
```diff
@@ -477,7 +477,7 @@ def main():
                     This will generate necessary metada data file for sso in mscolab through localhost idp
 
                     Before running this function:
-                    - Ensure that USE_SAML2 is set to True.
+                    - Ensure that use_saml2 is set to True.
```
ReimarBauer, line 480: this is not json

mslib/mscolab/conf.py
```diff
@@ -115,7 +115,7 @@ class default_mscolab_settings:
     GDPR = None
 
     # enable login by identity provider
-    USE_SAML2 = False
+    use_saml2 = False
```
ReimarBauer, line 118: this is not json

ReimarBauer: Do you still work on this @Beeram12  ?

Beeram12: @ReimarBauer Sorry for the delay,I am unaware of json format so I couldn't give update regarding this but will come up with right pull request.Kindly forgive me for many mistakes I am a beginner to open source.

ReimarBauer: Only where we use the json module you need to change it. Not on all places. 

https://www.w3schools.com/python/python_json.asp

PR #2144, opened 2024-01-10
Fix collisions with builtins

matrss:
Split off from #2100.

### DIFF
--- a/mslib/mscolab/models.py
+++ b/mslib/mscolab/models.py
@@ -36,7 +36,7 @@ from mslib.mscolab.message_type import MessageType
 class User(db.Model):
 
     __tablename__ = 'users'
-    id = db.Column(db.Integer, primary_key=True, autoincrement=True)
+    id = db.Column(db.Integer, primary_key=True, autoincrement=True)  # noqa: A003
     username = db.Column(db.String(255))
     emailid = db.Column(db.String(255), unique=True)
     password = db.Column(db.String(255), unique=True)
@@ -105,7 +105,7 @@ class User(db.Model):
 class Permission(db.Model):
 
     __tablename__ = 'permissions'
-    id = db.Column(db.Integer, primary_key=True, autoincrement=True)
+    id = db.Column(db.Integer, primary_key=True, autoincrement=True)  # noqa: A003
     op_id = db.Column(db.Integer, db.ForeignKey('operations.id'))
     u_id = db.Column(db.Integer, db.ForeignKey('users.id'))
     access_level = db.Column(db.Enum("admin", "collaborator", "viewer", "creator", name="access_level"))
@@ -127,7 +127,7 @@ class Permission(db.Model):
 class Operation(db.Model):
 
     __tablename__ = "operations"
-    id = db.Column(db.Integer, primary_key=True, autoincrement=True)
+    id = db.Column(db.Integer, primary_key=True, autoincrement=True)  # noqa: A003
     path = db.Column(db.String(255), unique=True)
     category = db.Column(db.String(255))
     description = db.Column(db.String(255))
@@ -158,7 +158,7 @@ class Operation(db.Model):
 class Message(db.Model):
 
     __tablename__ = "messages"
-    id = db.Column(db.Integer, primary_key=True, autoincrement=True)
+    id = db.Column(db.Integer, primary_key=True, autoincrement=True)  # noqa: A003
     op_id = db.Column(db.Integer, db.ForeignKey('operations.id'))
     u_id = db.Column(db.Integer, db.ForeignKey('users.id'))
     text = db.Column(db.Text)
@@ -182,7 +182,7 @@ class Message(db.Model):
 class Change(db.Model):
 
     __tablename__ = "changes"
-    id = db.Column(db.Integer, primary_key=True, autoincrement=True)
+    id = db.Column(db.Integer, primary_key=True, autoincrement=True)  # noqa: A003
     op_id = db.Column(db.Integer, db.ForeignKey('operations.id'))
     u_id = db.Column(db.Integer, db.ForeignKey('users.id'))
     commit_hash = db.Column(db.String(255), default=None)
--- a/.github/workflows/python-flake8.yml
+++ b/.github/workflows/python-flake8.yml
@@ -30,5 +30,5 @@ jobs:
     - name: Lint with flake8
       run: |
         python -m pip install --upgrade pip
-        pip install flake8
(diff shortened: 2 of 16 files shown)


ReimarBauer:
good idea :)
that refactoring is  not the first one. filter was before the favourite.

maybe someone has done a test routine for that, we can add to our project.  

in msui/mscolab.py is a 
`type = self.add_proj_dialog.cb_ImportType.currentText()`

maybe that could become import_type

some other words are very common: filter, type, dir
when not done do please a search for those too.

ReimarBauer:
Question: which builtins usually people accidantly shadow by a redifinition?

The AI Assistant names some more examples

In Python, it is common to accidentally overshadow built-ins by using them as variable names. Here are some examples:

- id: Python's built-in id() function returns a unique identifier for an object. Accidentally overwriting it with your own variable can cause confusion.
- type: The type() function is used to determine the type of an object. It's also a common term and easy to inadvertently overwrite.
- sum: Python's sum() function is used to add items in a sequence. Accidentally overwriting this can break any future call to sum().
- list: Overwriting list would prevent you from converting other data types to lists with list().
- filter: a function to filter items in a list based on a function. Overwriting filter would prevent you from being able to use it.
- str, int, float: All these functions are used for typecasting and it's easy to accidentally use them as variable names.

matrss:
You don't need an AI assistant for that, flake8 could check this with the flake8-builtins plugin. Or we switch to ruff (https://github.com/astral-sh/ruff) which is faster and has these checks built-in already. We should first introduce a test/lint for this, otherwise these issues will come back at some point. Do we want to do that? While at it, do we at some point want to enforce a specific formatting style (my preference would be black, or the code formatter that is now built-in to ruff which is very similar to black)?

ReimarBauer:
> You don't need an AI assistant for that, flake8 could check this with the flake8-builtins plugin. Or we switch to ruff (https://github.com/astral-sh/ruff) which is faster and has these checks built-in already. We should first introduce a test/lint for this, otherwise these issues will come back at some point. Do we want to do that? While at it, do we at some point want to enforce a specific formatting style (my preference would be black, or the code formatter that is now built-in to ruff which is very similar to black)?

We tried black some time ago and it broke MSS.  On the fs_filepicker repo we have a CI which enforces updates, but I am not happy with that extra traffic. I would expect an IDE solves the formatting.

ruff sounds interesting, we should discuss this in an other issue.  

I have few concerns about such "formatting" tools,

they work great if you have test coverage, but if the code is formatted wrong, at least it's noticeable, if it looks right but does wrong and there's no test, it's hard to detect. More tests -> more belief

ReimarBauer:
Do you want to change the type line too?

matrss:
> I would expect an IDE solves the formatting.

I would not expect everyone to use an IDE, let alone expect those IDEs to have the same idea of good formatting :-) Anyway, this was more of a thought and should be discussed in an issue first, if we want that.

> they work great if you have test coverage, but if the code is formatted wrong, at least it's noticeable, if it looks right but does wrong and there's no test, it's hard to detect. More tests -> more belief

I don't quite understand your concern, if it is really formatted wrong then that is a syntax error, everything else is an aesthetic choice (at least black should never change the result of the code, since it checks that the resulting AST stays effectively the same).

For this PR, I'd propose this:
- add flake8-builtins to the dev dependencies and the github action.
  This finds the following issues:
```
$ flake8 mslib tests
mslib/index.py:196:5: A001 variable "help" is shadowing a Python builtin
mslib/mscolab/models.py:39:5: A003 class attribute "id" is shadowing a Python builtin
mslib/mscolab/models.py:108:5: A003 class attribute "id" is shadowing a Python builtin
mslib/mscolab/models.py:130:5: A003 class attribute "id" is shadowing a Python builtin
mslib/mscolab/models.py:161:5: A003 class attribute "id" is shadowing a Python builtin
mslib/mscolab/models.py:185:5: A003 class attribute "id" is shadowing a Python builtin
mslib/msui/mpl_pathinteractor.py:1153:9: A001 variable "map" is shadowing a Python builtin
mslib/msui/mpl_qtwidget.py:1502:5: A003 class attribute "map" is shadowing a Python builtin
mslib/msui/mscolab.py:846:17: A001 variable "type" is shadowing a Python builtin
mslib/msui/msui_mainwindow.py:112:18: A002 argument "type" is shadowing a Python builtin
mslib/msui/wms_control.py:82:16: A002 argument "format" is shadowing a Python builtin
mslib/msui/wms_control.py:1244:66: A002 argument "format" is shadowing a Python builtin
mslib/mswms/gallery_builder.py:726:9: A001 variable "id" is shadowing a Python builtin
mslib/support/qt_json_view/datatypes.py:22:5: A003 class attribute "next" is shadowing a Python builtin
mslib/support/qt_json_view/datatypes.py:165:5: A003 class attribute "next" is shadowing a Python builtin
mslib/support/qt_json_view/datatypes.py:203:5: A003 class attribute "next" is shadowing a Python builtin
mslib/utils/qt.py:283:27: A002 argument "object" is shadowing a Python builtin
tests/_test_mscolab/test_models.py:64:9: A001 variable "id" is shadowing a Python builtin
tests/_test_mscolab/test_utils.py:42:5: A003 class attribute "id" is shadowing a Python builtin
tests/_test_msui/test_mscolab_merge_waypoints.py:140:121: E501 line too long (121 > 120 characters)
tests/_test_utils/test_migration.py:138:13: A001 variable "all" is shadowing a Python builtin
tests/_test_utils/test_migration.py:157:13: A001 variable "all" is shadowing a Python builtin
tests/_test_utils/test_multidict.py:58:24: A002 argument "object" is shadowing a Python builtin
tests/_test_utils/test_multidict.py:59:9: A001 variable "dict" is shadowing a Python builtin
```
- fix those issues
- merge
- discuss ruff and code formatting in general in a different issue

matrss:
I've fixed the cases where the shadowing was purely local and I've added ignores for now for the cases where class attributes or methods/functions shadow a builtin, since those would potentially require changes in multiple different locations.

mslib/mscolab/models.py
```diff
@@ -105,7 +105,7 @@ def verify_auth_token(token):
 class Permission(db.Model):
 
     __tablename__ = 'permissions'
-    id = db.Column(db.Integer, primary_key=True, autoincrement=True)
+    id = db.Column(db.Integer, primary_key=True, autoincrement=True)  # noqa: A003
```
ReimarBauer, line 108: :)

ReimarBauer (approved):
Thx

PR #2146, opened 2024-01-10
Refactor Qt related tests to use pytest-qt

matrss:
Using pytest-qt the QApplication instance can just be taken from its qapp fixture instead of being setup in each test. The fixture to fail if there are message boxes left at the end of a test can then be used by making the qapp fixture request it, which means that only tests that actually use Qt will have this fixture run (as opposed to the previous autouse fixture, which always ran).

Using the pytest-qt plugin also has other benefits:
- catches exceptions in Qt methods that are overloaded (explained here: https://pytest-qt.readthedocs.io/en/latest/virtual_methods.html)
- provides `addWidget` method which will help with proper widget cleanup in tests
- provides `waitSignal`, `waitUntil`, `waitCallback` methods which will help in reducing race conditions in tests where currently a `qWait` is used, i.e. which mistakenly use time for synchronization
- ... and probably more

Split off from #2100.

### DIFF
--- a/tests/_test_msui/test_mscolab_save_merge_points.py
+++ b/tests/_test_msui/test_mscolab_save_merge_points.py
@@ -35,11 +35,10 @@ from PyQt5 import QtCore, QtTest, QtWidgets
 PORTS = list(range(21000, 21500))
 
 
-# ToDo Understand why this needs to be skipped, it runs when direct called
+@pytest.mark.skip("Uses QTimer, which can break other unrelated tests")
 @pytest.mark.skipif(os.name == "nt",
                     reason="multiprocessing needs currently start_method fork")
 class Test_Save_Merge_Points(Test_Mscolab_Merge_Waypoints):
-    @pytest.mark.skip("Uses QTimer, which can break other unrelated tests")
     @mock.patch("PyQt5.QtWidgets.QMessageBox")
     def test_save_merge_points(self, mockbox):
         self.emailid = "mergepoints@alpha.org"
--- a/tests/_test_utils/test_auth.py
+++ b/tests/_test_utils/test_auth.py
@@ -49,7 +49,7 @@ def test_keyring():
 
 def test_get_auth_from_url_and_name():
     # set start condition to prevent definitions from a test earlier
-    constants.AUTH_LOGIN_CACHE == {}
+    constants.AUTH_LOGIN_CACHE = {}
     # empty http_auth definition
     server_url = "http://example.com"
     http_auth = config_loader(dataset="MSS_auth")
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -62,7 +62,6 @@ jobs:
         | sed -e "s/menuinst.*//" \
         | sed -e "s/.*://" > reqs.txt \
         && cat requirements.d/development.txt >> reqs.txt \
-        && echo pyvirtualdisplay >> reqs.txt \
         && cat reqs.txt \
         && mamba env remove -n mss-${{ inputs.branch_name }}-env \
         && mamba create -y -n mss-${{ inputs.branch_name }}-env --file reqs.txt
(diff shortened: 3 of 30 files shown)


tests/_test_msui/test_mscolab_save_merge_points.py
```diff
@@ -35,11 +35,10 @@
 PORTS = list(range(21000, 21500))
 
 
-# ToDo Understand why this needs to be skipped, it runs when direct called
+@pytest.mark.skip("Uses QTimer, which can break other unrelated tests")
```
ReimarBauer, line 38: skipping ok,

why does QTimer or can break other tests, what is the problem with that?

tests/_test_utils/test_auth.py
```diff
@@ -49,7 +49,7 @@ def test_keyring():
 
 def test_get_auth_from_url_and_name():
     # set start condition to prevent definitions from a test earlier
-    constants.AUTH_LOGIN_CACHE == {}
+    constants.AUTH_LOGIN_CACHE = {}
     # empty http_auth definition
```
ReimarBauer, line 53: good catch,  is there more broken, that was not changing to empty .

ReimarBauer (commented):
looks good, few questions

tests/_test_msui/test_mscolab_save_merge_points.py
matrss: The problem is that the QTimer apparently can outlive the test and then run its function when another test already started. This then obviously affects that other test. At least this is what happened in `test_mscolab_merge_waypoints.py` where I was able to get rid of the QTimers (see: https://github.com/Open-MSS/MSS/commit/1e98de5187977fe0e140cf1649776f918b2bf693#diff-5e7ce9c062dc89ba1755f9d8100cb6473831e4575df18b728e922afd641964dc).

Here in `test_mscolab_save_merge_points.py` I was not (yet) able to get rid of it (but the test was skipped anyway).

tests/_test_utils/test_auth.py
matrss: If this is a question I am not sure I quite get it. This was the only instance of such a mistake that I saw, but I didn't specifically scan the codebase for more.

tests/_test_msui/test_mscolab_save_merge_points.py
ReimarBauer: My thoughts on this: Ideally the test should be independent of other tests when and how it runs. Each test has its own tmp dir and each test has a different server on a different port. They should not actually interfere with each other. We are looking for something that causes a problem, although we make sure that the tests are independent of each other.

tests/_test_utils/test_auth.py
ReimarBauer: I assume it was added because an other test had added data  to the constants.AUTH_LOGIN_CACHE in the past, which could do a conflict.

The tests seem not have to failed without the start condition. Is the test requiring an empty AUTH_LOGIN_CACHE, is the test wrong? is that line at all needed, because it had no effect.

tests/_test_msui/test_mscolab_save_merge_points.py
matrss: As I said, my guess is that this is caused by the creation of the QTimer, which then outlives the test in the QApplication and will run at a later point when the Qt event loop is handling events the next time (through processEvents or qWait). At least this is what I saw with the tests in `test_mscolab_merge_waypoints.py`, where the QTimer triggered in the teardown code of a different test.

This test was also skipped for 3 years already, see: https://github.com/Open-MSS/MSS/blame/c1f6e86ce3cd45605825486d1f5a1192d6897503/tests/_test_msui/test_mscolab_save_merge_points.py#L61-L62 (I never saw it not skip based on the if condition there).

tests/_test_utils/test_auth.py
matrss: The test is failing on latest develop: https://github.com/Open-MSS/MSS/actions/runs/7477525278/job/20350414652#step:9:103

It passes on first try with this change: https://github.com/Open-MSS/MSS/actions/runs/7476710573/job/20347734552?pr=2146#step:9:104

So I assume that this line is required (for now at least, maybe this should also be moved to the reset fixture? That can be judged better in a later PR, when I will try to get completely rid of the retries again).

ReimarBauer (approved):
Thx

PR #2147, opened 2024-01-10
Add per-operation locks to FileManager

matrss:
Without these locks there is a race condition when writing to the operations "main.ftml" file, which happens in multiple different methods, e.g. save_file.  When the mscolab application is running in a multi-threaded wsgi server it can happen that two threads try to write to the same file, corrupting it in the process.

Split off from #2100.

### DIFF
--- a/mslib/mscolab/file_manager.py
+++ b/mslib/mscolab/file_manager.py
@@ -29,6 +29,7 @@ import fs
 import difflib
 import logging
 import git
+import threading
 from sqlalchemy.exc import IntegrityError
 from mslib.mscolab.models import db, Operation, Permission, User, Change, Message
 from mslib.mscolab.conf import mscolab_settings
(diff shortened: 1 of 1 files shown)


mslib/mscolab/file_manager.py
```diff
@@ -29,6 +29,7 @@
 import difflib
 import logging
 import git
+import threading
```
ReimarBauer, line 32: please import only what we need from threading 
  
 from threading import Lock

matrss: Why? I prefer having it namespace'd properly where I use it and it is not like `threading` is excessively long. Having it just be called `Lock` where it is used makes it ambiguous if it is coming from threading or from multiprocessing for example.

ReimarBauer: good point, we keep it. 
We have no generel style for this, usually we use import when we use lots of its methods.

in both cases the module is always fully imported (no speedup by a selection, python parses the whole module).

ReimarBauer (approved):
Thx

PR #2149, opened 2024-01-11
Refactor servers into fixtures

matrss:
This refactors the mscolab and mswms servers into pytest fixtures and also gets rid of some remnants from unittest (TestCase classes). The advantage is that this consolidates server setup and teardown code into one place, so that the tests requesting the fixtures no long have to care about e.g. resetting the database before tests or stopping the process afterwards. This also removes the need for the global `PORTS` variables.

Split off from #2100.

### DIFF
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -38,10 +38,7 @@ from urllib.parse import urljoin
 from mslib.mscolab.server import register_user
 from flask import json
 from tests.constants import MSUI_CONFIG_PATH
-from mslib.mscolab.conf import mscolab_settings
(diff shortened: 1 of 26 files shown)


tests/utils.py
```diff
@@ -28,20 +28,12 @@
 import requests
 import time
 import fs
-import socket
-import multiprocessing
-
-from flask_testing import LiveServerTestCase
```
ReimarBauer, line 34: good, that we can remove this.

ReimarBauer (approved):
Thx

PR #2152, opened 2024-01-16
Remove wrong config file modification

matrss:
The functionality under test here is that MSS does update the config file with the new credentials when creating a user, therefore modifying the file in the test defeated its purpose.


ReimarBauer (approved):
good catch

PR #2153, opened 2024-01-16
Remove test retries in CI

### LINKED ISSUE #2112: Never rerun failed tests

matrss:
Tests that do not consistently pass are broken tests and should be fixed. Failed tests should not be retried to avoid accidentally adding new broken tests.

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -81,11 +81,7 @@ jobs:
         && source /opt/conda/etc/profile.d/conda.sh \
         && source /opt/conda/etc/profile.d/mamba.sh \
         && mamba activate mss-${{ inputs.branch_name }}-env \
-        && xvfb-run pytest -v --durations=20 --reverse --cov=mslib tests \
-        || (for i in {1..5} \
-          ; do xvfb-run pytest tests -v --durations=0 --reverse --last-failed --lfnf=none \
-          && break \
-        ; done)
+        && xvfb-run pytest -v --durations=20 --reverse --cov=mslib tests
 
     - name: Run tests in parallel
       if: ${{ success() && inputs.xdist == 'yes' }}
@@ -95,11 +91,7 @@ jobs:
         && source /opt/conda/etc/profile.d/conda.sh \
         && source /opt/conda/etc/profile.d/mamba.sh \
         && mamba activate mss-${{ inputs.branch_name }}-env \
-        && xvfb-run pytest -vv -n 6 --dist loadfile --max-worker-restart 4 tests \
-        || (for i in {1..5} \
-          ; do xvfb-run pytest -vv -n 6 --dist loadfile --max-worker-restart 0 tests --last-failed --lfnf=none \
-          && break \
-        ; done)
+        && xvfb-run pytest -vv -n 6 --dist loadfile --max-worker-restart 4 tests
 
     - name: Collect coverage
       if: ${{ success() && inputs.event_name == 'push' && inputs.branch_name == 'develop' && inputs.xdist == 'no'}}
--- a/tests/_test_msui/test_mscolab.py
+++ b/tests/_test_msui/test_mscolab.py
@@ -288,6 +288,8 @@ class Test_Mscolab:
         self.window = msui.MSUIMainWindow(mscolab_data_dir=mscolab_settings.MSCOLAB_DATA_DIR)
         self.window.create_new_flight_track()
         self.window.show()
+
+        self.total_created_operations = 0
         yield
         self.window.mscolab.logout()
         if self.window.mscolab.version_window:
@@ -406,10 +408,10 @@ class Test_Mscolab:
     @pytest.mark.skip("fails often on github on a timeout >60s")
     @mock.patch("mslib.msui.mscolab.QtWidgets.QErrorMessage.showMessage")
     @mock.patch("mslib.msui.mscolab.get_open_filename", return_value=os.path.join(sample_path, u"example.ftml"))
-    def test_browse_add_operation(self, mockopen, mockmessage):
+    def test_browse_add_operation(self, mockopen, mockmessage, qtbot):
         self._connect_to_mscolab()
         modify_config_file({"MSS_auth": {self.url: "something@something.org"}})
-        self._create_user("something", "something@something.org", "something")
+        self._create_user(qtbot, "something", "something@something.org", "something")
         assert self.window.listOperationsMSC.model().rowCount() == 0
         self.window.actionAddOperation.trigger()
         QtWidgets.QApplication.processEvents()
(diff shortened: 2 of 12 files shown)


matrss:
@ReimarBauer It looks like with these changes the test suite is in a state in which it can reliably pass (almost?) always in CI. The latest 100 runs (https://github.com/Open-MSS/MSS/actions/runs/7567970125?pr=2153) succeeded without issue. I am pretty sure that there are many more potential places that have race conditions and *could* fail the test suite, but those that came up most often are fixed now. I can still get some fails if I execute the test suite locally, but only if I severely limit the CPU quota it can use.

If you think this is good to merge I'll remove the "TEMP" commit and mark this as ready.

Sidenote: if this CI configuration with 100 jobs is executed too often without enough time between them it seems like GitHub is rate-limiting it in some way. Yesterday in the afternoon at least half of the tests started to timeout for some inexplicable reason.

tests/_test_msui/test_mscolab.py
```diff
@@ -585,13 +572,13 @@ def test_update_category(self, mocktext):
...
+    def test_any_special_category(self, mockbox, qtbot):
         self._connect_to_mscolab()
         modify_config_file({"MSS_auth": {self.url: "something@something.org"}})
-        self._create_user("something", "something@something.org", "something")
-        self._create_operation("flight1234", "Description flight1234")
+        self._create_user(qtbot, "something", "something@something.org", "something")
+        self._create_operation_successfully(qtbot, "flight1234", "Description flight1234")
         QtTest.QTest.qWait(0)
-        self._create_operation("flight5678", "Description flight5678", category="furtherexample")
+        self._create_operation_successfully(qtbot, "flight5678", "Description flight5678", category="furtherexample")
```
ReimarBauer: not sure if I understand the reason of the postfix "successfully". can we be unsuccessfull too?

.github/workflows/testing.yml
```diff
@@ -81,25 +88,17 @@ jobs:
...
         cd $GITHUB_WORKSPACE \
         && source /opt/conda/etc/profile.d/conda.sh \
         && source /opt/conda/etc/profile.d/mamba.sh \
         && mamba activate mss-${{ inputs.branch_name }}-env \
-        && xvfb-run pytest -vv -n 6 --dist loadfile --max-worker-restart 4 tests \
-        || (for i in {1..5} \
-          ; do xvfb-run pytest -vv -n 6 --dist loadfile --max-worker-restart 0 tests --last-failed --lfnf=none \
-          && break \
-        ; done)
+        && xvfb-run pytest -vv -n 6 --dist loadfile --max-worker-restart 4 tests
```
ReimarBauer, line 94: the 4 is for four times restarting, do we keep this?

ReimarBauer:
in the past not more than 6 cores on my laptop could be used to decrease the runtime of tests. 8 now speeds it a bit up. 
but 6 is still a good choice. We maybe should split some files. When you use only 2 cores, the distribution over the cores does not take in account which are the expensive ones.

- 8 cores: 514 passed, 26 skipped, 229 warnings in 284.21s (0:04:44)
- 6 cores 514 passed, 26 skipped, 213 warnings in 300.63s (0:05:00)
- 4 cores 514 passed, 26 skipped, 197 warnings in 345.45s (0:05:45)
- 2 cores 514 passed, 26 skipped, 181 warnings in 510.90s (0:08:30)

tests/_test_msui/test_mscolab.py
matrss: The `_create_operation_successfully` method ensures, through additional asserts and waiting, that the operation is successfully created and visible in the UI, as opposed to the plain `_create_operation`, which does not have these checks. I.e. due to the asynchronous nature of Qt it is possible that after a `_create_operation` the operation is not yet visible in the operations list while after a `_create_operation_successfully` it will be because it waits for it (or it will fail the test if creating the operation failed).

That is why I chose this suffix. Do you have a better name for this method?

.github/workflows/testing.yml
matrss: I kept it for now because I didn't want to change the xdist related flags. I have a follow up PR to explore if those can be improved.

matrss:
> in the past not more than 6 cores on my laptop could be used to decrease the runtime of tests. 8 now speeds it a bit up. but 6 is still a good choice. We maybe should split some files. When you use only 2 cores, the distribution over the cores does not take in account which are the expensive ones.
> 
>     * 8 cores: 514 passed, 26 skipped, 229 warnings in 284.21s (0:04:44)
> 
>     * 6 cores 514 passed, 26 skipped, 213 warnings in 300.63s (0:05:00)
> 
>     * 4 cores 514 passed, 26 skipped, 197 warnings in 345.45s (0:05:45)
> 
>     * 2 cores 514 passed, 26 skipped, 181 warnings in 510.90s (0:08:30)

This is very much system dependent. pytest-xdist supports `-n auto` to use as many processes as physical cores, or `-n logical` to use as many processes as logical cores exist (i.e. hyperthreading). I think changing those values is out of scope for this PR and can be done later (I already have a PR open to explore the effects of this setting as well as other values for the `--dist` flag).

tests/_test_msui/test_mscolab.py
ReimarBauer: that is ok, but do we need both functions? when the _successfully is the prefered one, we should remove the other and keep its name. We would always expect we do something which is successful.

matrss: There are three uses of `_create_operation` left which try to create invalid operations and assert that a corresponding error is emitted. They are here:
https://github.com/Open-MSS/MSS/blob/9a9345c50edf50bace6f75c359d54e4071d0f7da/tests/_test_msui/test_mscolab.py#L442-L452

I think we could rename `_create_operation_successfully` back into `_create_operation` and name this other one `_create_operation_unchecked` or something like that, to have a shorter name for the more common operation (and be more inline with `_create_user`, which does similar checks but doesn't have an additional suffix since there is just one version of it).

EDIT: I've renamed it as described and did a rebase.

ReimarBauer:
> > in the past not more than 6 cores on my laptop could be used to decrease the runtime of tests. 8 now speeds it a bit up. but 6 is still a good choice. We maybe should split some files. When you use only 2 cores, the distribution over the cores does not take in account which are the expensive ones.
> > ```
> > * 8 cores: 514 passed, 26 skipped, 229 warnings in 284.21s (0:04:44)
> > 
> > * 6 cores 514 passed, 26 skipped, 213 warnings in 300.63s (0:05:00)
> > 
> > * 4 cores 514 passed, 26 skipped, 197 warnings in 345.45s (0:05:45)
> > 
> > * 2 cores 514 passed, 26 skipped, 181 warnings in 510.90s (0:08:30)
> > ```
> 
> This is very much system dependent. pytest-xdist supports `-n auto` to use as many processes as physical cores, or `-n logical` to use as many processes as logical cores exist (i.e. hyperthreading). I think changing those values is out of scope for this PR and can be done later (I already have a PR open to explore the effects of this setting as well as other values for the `--dist` flag).

-n auto on github in the past used regulalrly 2 cores. That was not much better as wanted

matrss:
> -n auto on github in the past used regulalrly 2 cores. That was not much better as wanted

Yes, obviously, because the normal ubuntu github runners only have 2 cores. Using `-n logical` instead would use 4 workers, which achieves similar times as `-n 6`, as you can see here: https://github.com/Open-MSS/MSS/actions/runs/7570380297?pr=2159.

matrss:
I've tried multiple times over the day to have this pass a hundred times again, but there were always a few jobs that timed out for some reason. At this point I am pretty certain that this is not an issue with the test suite itself, but with the github runners (maybe general load on the GitHub Actions has some influence here based on the time of day, but I am just speculating).

Anyway, I've reduced the number of jobs a bit and as you can see in the latest run (https://github.com/Open-MSS/MSS/actions/runs/7626758571?pr=2153) it passed 50 times, half of those in normal order and the other half in reverse order. I feel like this is sufficiently reliable to be merged, so I'll remove the TEMP commit and mark this as ready.

ReimarBauer (approved):
Thx

PR #2154, opened 2024-01-16

### LINKED ISSUE #2132: keys in json shoud be lowercase written

Beeram12:
Renamed USE_SAML2 to use_saml2 

Fixes #2132

### DIFF
--- a/mslib/mscolab/server.py
+++ b/mslib/mscolab/server.py
@@ -242,18 +242,18 @@ def hello():
             auth.login_required()
             return json.dumps({
                 'message': "Mscolab server",
-                'USE_SAML2': mscolab_settings.USE_SAML2,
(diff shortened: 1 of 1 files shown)


mslib/mscolab/server.py
```diff
@@ -242,18 +242,18 @@ def hello():
             auth.login_required()
             return json.dumps({
                 'message': "Mscolab server",
-                'USE_SAML2': mscolab_settings.USE_SAML2,
+                'use_saml2': mscolab_settings.use_saml2,
```
ReimarBauer, line 245: change only the left side, not the constant

ReimarBauer (changes requested):
There it comes into the UI and needs also become changed
https://github.com/Beeram12/MSS/blob/develop/mslib/msui/mscolab.py#L238

PR #2155, opened 2024-01-17

### LINKED ISSUE #2132: keys in json shoud be lowercase written

Beeram12:
renamed USE_SAML2 to use_saml2

Fixes #2132


ReimarBauer (approved):
thx

PR #2158, opened 2024-01-18
Remove the serial test execution from CI

### LINKED ISSUE #2191: Remove the serial test execution from CI
There is no necessity to have a serial and a parallel execution of the test suite. This would speed up the CI run times by a factor of almost 3.

matrss:
There is no necessity to have a serial and a parallel execution of the test suite. Pytest-xdist and pytest-cov can be combined. Since the retries get removed when #2153 (which this PR depends on) lands there should no longer be a difference between the test coverages due to the crashes and retries that existed previously.

Fixes #2191.

### DIFF
--- a/.github/workflows/testing-develop.yml
+++ b/.github/workflows/testing-develop.yml
@@ -11,22 +11,10 @@ on:
 
 jobs:
   test-develop:
-    uses: 
+    uses:
       ./.github/workflows/testing.yml
     with:
-      xdist: no
-      branch_name: develop
-      event_name: ${{ github.event_name }}
-    secrets: 
-      PAT: ${{ secrets.PAT }}
-
-  test-develop-xdist:
-    uses: 
-      ./.github/workflows/testing.yml
-    with:
-      xdist: yes
       branch_name: develop
       event_name: ${{ github.event_name }}
     secrets:
       PAT: ${{ secrets.PAT }}
-
--- a/.github/workflows/testing-scheduled.yml
+++ b/.github/workflows/testing-scheduled.yml
@@ -2,29 +2,24 @@ name: new dependency test (scheduled)
 
 on:
   schedule:
-    - cron: '30 5 * * 1' 
+    - cron: '30 5 * * 1'
 
 
-jobs:     
+jobs:
   test-stable-scheduled:
-    uses: 
+    uses:
       ./.github/workflows/testing.yml
     with:
-      xdist: no
       branch_name: stable
       event_name: ${{ github.event_name }}
     secrets:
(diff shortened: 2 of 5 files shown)


ReimarBauer:
the parallel tests where in order and divided by different processes. The serial tests where used in reverse order to get the tests which depend on one other.  This tests which these problemes are not equivalent. But the dependency from a test to some other should be solved aready.
We need fast results, so yes the serial tests are not good for that. We kept them because of the inconsitent failing of tests.
This is solved too.
We maybe should look if we can use a different amount of processes dependent from e.g. days to check for some runtime dependencies, but not take too much processes.

matrss:
> the parallel tests where in order and divided by different processes. The serial tests where used in reverse order to get the tests which depend on one other. This tests which these problemes are not equivalent. But the dependency from a test to some other should be solved aready.

Yes, with `--dist loadfile` the tests in each file are executed in order, but the order of the files is non-deterministic. If we additionally switch to `--dist load` or `--dist worksteal` the order of all tests would become non-deterministic. This should address your concern. We could also throw in pytest-randomly to explicitly randomize the test order further.

> We maybe should look if we can use a different amount of processes dependent from e.g. days to check for some runtime dependencies, but not take too much processes.

I do not understand what you mean. Are you talking about changing the `-n` option of xdist based on what day it is? I don't see a point in that.

matrss:
> We could also throw in pytest-randomly to explicitly randomize the test order further.

It looks like the test suite is not yet in a state in which it consistently passes with pytest-randomly. But we could have two jobs which run with xdist, one in the normal order and one in reverse, until it works with pytest-randomly as well.

ReimarBauer:
> > the parallel tests where in order and divided by different processes. The serial tests where used in reverse order to get the tests which depend on one other. This tests which these problemes are not equivalent. But the dependency from a test to some other should be solved aready.
> 
> Yes, with `--dist loadfile` the tests in each file are executed in order, but the order of the files is non-deterministic. If we additionally switch to `--dist load` or `--dist worksteal` the order of all tests would become non-deterministic. This should address your concern. We could also throw in pytest-randomly to explicitly randomize the test order further.
> 
> > We maybe should look if we can use a different amount of processes dependent from e.g. days to check for some runtime dependencies, but not take too much processes.
> 
> I do not understand what you mean. Are you talking about changing the `-n` option of xdist based on what day it is? I don't see a point in that.

I was thinking that had also influence about what runs before. But it is also cleaned. So we can not catch with that change dependencies which we should not have. No point any more.

ReimarBauer (approved):
Thx

PR #2159, opened 2024-01-18
Improve xdist flags in CI

### LINKED ISSUE #2194: Improve xdist flags in CI
Right now the number of workers is hardcoded. This could be improved by letting xdist choose the number automatically (`-n auto` or `-n logical`). This might be useful when changes to the GitHub Actions runners happen, like recently when it was announced that they get 4 cores instead of 2.

The `--dist loadfile` option might also not be necessary and could be replaced with the default or `--dist worksteal`. These options would no longer enforce the (I think unnecessary) grouping by file, which is somewhat related to finding inter-test dependencies.

We should evaluate multiple combinations of flags to see if we can improve over what we have now.

matrss:
This is an exploration of if we can pick better settings for xdist in the CI.

Fixes #2194.

Depends on #2158.

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -32,6 +32,8 @@ jobs:
       fail-fast: false
       matrix:
         order: ["normal", "reverse"]
+        xdist-n: ["6", "auto", "logical"]
+        xdist-dist: ["loadfile", "load", "worksteal"]
 
     steps:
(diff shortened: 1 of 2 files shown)


matrss:
This got mostly irrelevant, and the testing-all-oses.yml workflow actually uses `-n logical` to pick the right amount of processes for each platform. Closing.

PR #2134, opened 2024-01-06
fix: the flake8 max line failure

rohit2p:
**Purpose of PR?**:

Fixes #2111

**Checklist:**
- [X] Bug fix. Fixes #2111 

**I removed the max-line-length = 127 from the python-flake8 because it was overwriting setup.cfg. These changes ensure that the codebase adheres to the project's coding standards and resolves the reported issues.**

PR #2163, opened 2024-01-19
Fix: ci issues and cherry-pick from stable branch.

### LINKED ISSUE #2111: python-flake8 max-line-length failure
The setup.cfg defines a max-line-length = 120 and this is overwritten in the python-flake8 with 127.

The extra definition in the worker script should be removed so that the settings from setup.cfg are used

rohit2p:
**Purpose of PR?**:
This pull request is a follow-up to PR #2134 , which addressed the flake8 max line failure. in this PR i am trying to get the code in the develop as well by cherry-pick the previous one and also made some changes in `tests/_test_msui/test_mscolab_merge_waypoints.py` so the ci test will run successfully. 

Fixes #2111 

**Checklist:**
- [X] Bug fix. Fixes #2111

### DIFF
--- a/.github/workflows/python-flake8.yml
+++ b/.github/workflows/python-flake8.yml
@@ -30,5 +30,5 @@ jobs:
     - name: Lint with flake8
       run: |
         python -m pip install --upgrade pip
-        pip install flake8 flake8-builtins
-        flake8 --count --max-line-length=127 --statistics mslib tests
+        pip install flake8
+        flake8 --count --statistics mslib tests
--- a/.github/workflows/python-flake8.yml
+++ b/.github/workflows/python-flake8.yml
@@ -30,5 +30,5 @@ jobs:
     - name: Lint with flake8
       run: |
         python -m pip install --upgrade pip
-        pip install flake8 flake8-builtin
+        pip install flake8 flake8-builtins
         flake8 --count --statistics mslib tests
--- a/.github/workflows/python-flake8.yml
+++ b/.github/workflows/python-flake8.yml
@@ -30,5 +30,5 @@ jobs:
     - name: Lint with flake8
       run: |
         python -m pip install --upgrade pip
-        pip install flake8
+        pip install flake8 flake8-builtin
         flake8 --count --statistics mslib tests
--- a/tests/_test_msui/test_mscolab_merge_waypoints.py
+++ b/tests/_test_msui/test_mscolab_merge_waypoints.py
@@ -132,7 +132,9 @@ class Test_Overwrite_To_Server(Test_Mscolab_Merge_Waypoints):
         assert wp_server_before.lat != wp_local_before.lat
 
         # trigger save to server action from server options combobox
-        with mock.patch("mslib.msui.mscolab.MscolabMergeWaypointsDialog", AutoClickOverwriteMscolabMergeWaypointsDialog):
+        with mock.patch(
+            "mslib.msui.mscolab.MscolabMergeWaypointsDialog",
+                AutoClickOverwriteMscolabMergeWaypointsDialog):
             self.window.serverOptionsCb.setCurrentIndex(2)
         QtWidgets.QApplication.processEvents()
         # get the updated waypoints model from the server


.github/workflows/python-flake8.yml
```diff
@@ -30,5 +30,5 @@ jobs:
     - name: Lint with flake8
       run: |
         python -m pip install --upgrade pip
-        pip install flake8 flake8-builtins
```
matrss, line 33: This line should stay as it is, the flake8-builtins package adds additional checks that we want to have.

tests/_test_msui/test_mscolab_merge_waypoints.py
```diff
@@ -132,7 +132,9 @@ def test_save_overwrite_to_server(self, mockbox):
         assert wp_server_before.lat != wp_local_before.lat
 
         # trigger save to server action from server options combobox
-        with mock.patch("mslib.msui.mscolab.MscolabMergeWaypointsDialog", AutoClickOverwriteMscolabMergeWaypointsDialog):
+        with mock.patch(
+            "mslib.msui.mscolab.MscolabMergeWaypointsDialog",
+                AutoClickOverwriteMscolabMergeWaypointsDialog):
```
matrss, line 137: I think arguments to the same function should be aligned with the same indentation, i.e. the A should be under the double quote. This is usually done to make them more easily readable at a glance.

matrss (changes requested):
Thanks for the PR! I left some comments. Also, the PR contains the cherry-picked commit twice, one of which is empty. I think that should be deduplicated.

tests/_test_msui/test_mscolab_merge_waypoints.py
rohit2p: yea but when i do that i get this error `.\tests\_test_msui\test_mscolab_merge_waypoints.py:137:13: E129 visually indented line with same indent as next logical line  `

.github/workflows/python-flake8.yml
rohit2p: ok I'll do that

rohit2p:
sorry for the late response i was trying to figure out on my own but i tried deduplicating this by using revert, rebase everything i can but but i cant seem to do it. could to help me out a little bit here

tests/_test_msui/test_mscolab_merge_waypoints.py
matrss: You are right, I didn't think of that. We can leave it as it is then.

matrss:
> sorry for the late response i was trying to figure out on my own but i tried deduplicating this by using revert, rebase everything i can but but i cant seem to do it. could to help me out a little bit here

The command you are looking for is `git reset HEAD~1`. What this will do is set HEAD, i.e. the latest commit of your branch, to be the commit right before your current HEAD, effectively removing the latest commit.

Alternatively you could also do an interactive rebase onto develop: `git rebase -i develop`. You should then see an editor in which you can change the command for the empty commit (last line) to be drop instead of pick. After saving and closing in the editor git should apply the rebase, dropping the empty commit.

I hope this helps!

.github/workflows/python-flake8.yml
```diff
@@ -30,5 +30,5 @@ jobs:
...
-        flake8 --count --max-line-length=127 --statistics mslib tests
+        pip install flake8 flake8-builtin
```
matrss:
```suggestion
        pip install flake8 flake8-builtins
```

The package name is missing an "s" at the end. You can also see that in the failing CI runs output: https://github.com/Open-MSS/MSS/actions/runs/7629972323/job/20784618748?pr=2163.

rohit2p:
Thankyou for you help i think i have dropped the empty one.
now, It seems like there is an issue with the` flake8-builtin` package. maybe if we specify a specific version of it like this ` flake8-builtin==<version>` this may fix this

.github/workflows/python-flake8.yml
rohit2p: i am so sorry i missed that my bad

matrss (approved):
LGTM, Thanks!

PR #2165, opened 2024-01-24
Remove unnecessary skips

### LINKED ISSUE #2188: Remove unnecessary test skips
There are a bunch of tests that get skipped which should pass with no or minimal changes (at least since the effort to reduce tests influencing each other). We should re-evaluate each skip for if it is still necessary or if the test can be enabled again.

matrss:
This PR removes many unnecessary skips for tests that either work fine or needed just a small amount of changes.

Fixes #2188.

Split off from #2100.

I want to wait with this until after #2158 is merged.

### DIFF
--- a/tests/_test_msui/test_msui.py
+++ b/tests/_test_msui/test_msui.py
@@ -237,10 +237,10 @@ class Test_MSSSideViewWindow:
         QtWidgets.QApplication.processEvents()
 
     def test_open_config(self):
-        pytest.skip("To be done")
-        self.window.actionConfigurationEditor.trigger()
+        self.window.actionConfiguration.trigger()
         QtWidgets.QApplication.processEvents()
-        self.window.config_editor.close()
+        with mock.patch("PyQt5.QtWidgets.QMessageBox.warning", return_value=QtWidgets.QMessageBox.Yes):
+            self.window.config_editor.close()
 
     def test_open_shortcut(self):
         self.window.actionShortcuts.trigger()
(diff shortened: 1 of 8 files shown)


tests/_test_msui/test_msui.py
```diff
@@ -315,22 +314,21 @@ def test_plugin_export(self, save_file):
             assert os.path.exists(save_file[0])
             os.remove(save_file[0])
 
-    @pytest.mark.skip("needs to be refactored to become independent")
     @mock.patch("PyQt5.QtWidgets.QMessageBox")
     @mock.patch("mslib.msui.msui_mainwindow.config_loader", return_value=export_plugins)
     def test_add_plugins(self, mockopen, mockbox):
```
ReimarBauer: The default is not to have further import / export plugins. 
The name is misleading or the result should show that plugins can become added

ReimarBauer (commented):
more working tests :)

there is one test which I marked, which I think does nothing.

tests/_test_msui/test_msui.py
matrss: I don't understand what you mean. The test adds import and export plugins and then checks that they (or at least "one more") shows up in the respective lists. It also removes them again and checks that they no longer show up then, so I guess it could be renamed to `test_add_and_remove_plugins` or split into two tests.

ReimarBauer (approved):

tests/_test_msui/test_msui.py
ReimarBauer: seems I readed it not completly. Thanks for asking.

PR #2166, opened 2024-01-24
Consolidate skips due to start method "fork"

matrss:
There is just one place where the start method "fork" is used, which is in tests.fixtures. By adding a conditional skip at that location all other skips for this reason become obsolete, since the tests will be automatically skipped if they request one of the fixtures that require this start method from multiprocessing and the start method is unsupported on the given system.

This PR addresses an issue that was found in https://github.com/Open-MSS/MSS/pull/2100#issuecomment-1877243231, where some tests that needed to be skipped actually weren't and therefore failed on windows.

### DIFF
--- a/tests/fixtures.py
+++ b/tests/fixtures.py
@@ -175,6 +175,8 @@ def _running_eventlet_server(app):
     port = socket.getsockname()[1]
     url = f"{scheme}://{host}:{port}"
     app.config['URL'] = url
+    if "fork" not in multiprocessing.get_all_start_methods():
+        pytest.skip("requires the multiprocessing start_method 'fork', which is unavailable on this system")
     ctx = multiprocessing.get_context("fork")
     process = ctx.Process(target=eventlet.wsgi.server, args=(socket, app), daemon=True)
     try:
--- a/tests/_test_mscolab/test_file_manager.py
+++ b/tests/_test_mscolab/test_file_manager.py
@@ -24,7 +24,6 @@
     See the License for the specific language governing permissions and
     limitations under the License.
 """
-import os
 import datetime
 import pytest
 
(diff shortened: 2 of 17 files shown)


tests/fixtures.py
```diff
@@ -175,6 +175,8 @@ def _running_eventlet_server(app):
     port = socket.getsockname()[1]
     url = f"{scheme}://{host}:{port}"
     app.config['URL'] = url
+    if "fork" not in multiprocessing.get_all_start_methods():
+        pytest.skip("requires the multiprocessing start_method 'fork', which is unavailable on this system")
```
ReimarBauer, line 179: pytest.mark may have influence to the tests sorted to the different processes. It is used in the test collection phase. it never runs a marked test.

pytest.skip is used in the call phase, it stops a running tests.

Maybe we can use the allow_module_level flag here?
https://docs.pytest.org/en/7.1.x/reference/reference.html#pytest-skip

matrss: > pytest.mark may have influence to the tests sorted to the different processes. It is used in the test collection phase. it never runs a marked test.
> 
> pytest.skip is used in the call phase, it stops a running tests.

Yes, I think that is true, but what is the concern with it? The fixture setup happens relatively early and in this case the `_running_eventlet_server` context manager is even executed just twice per test session (once for mscolab, once for mswms). The pytest.skip call is therefore reached just twice to skip all dependent tests.

Does it have any issues on windows? I do not have a setup to test this properly.

> Maybe we can use the allow_module_level flag here?
> https://docs.pytest.org/en/7.1.x/reference/reference.html#pytest-skip

I don't see how this would help in any way.

ReimarBauer: that was more like a remark, I didn't request a change. 

mark.skip was in the past prefered used intestead .skip.

ReimarBauer (approved):
Thx

tests/fixtures.py
matrss: > mark.skip was in the past prefered used intestead .skip.

I agree, as long as pytest.mark.skip can be used. But, to my knowledge, it is impossible to skip all tests that depend on a fixture with just pytest.mark.skip.

PR #2170, opened 2024-01-30
Remove pytest option to pass msui settings

### LINKED ISSUE #2190: Remove the custom pytest option to pass msui settings to tests
This option is, to my knowledge, unused and it does not work anymore since the configuration is reset before each test.

matrss:
This option is, to my knowledge, unused and it does not work anymore since the configuration is reset before each test.

Fixes #2190.

Follow up to #2136.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -74,18 +74,6 @@ def keyring_reset():
     keyring.get_keyring().reset()
 
 
-def pytest_addoption(parser):
-    parser.addoption("--msui_settings", action="store")
-
-
-def pytest_generate_tests(metafunc):
-    option_value = metafunc.config.option.msui_settings
-    if option_value is not None:
-        msui_settings_file_fs = fs.open_fs(constants.MSUI_CONFIG_PATH)
-        msui_settings_file_fs.writetext("msui_settings.json", option_value)
-        msui_settings_file_fs.close()
-
-
 def generate_initial_config():
     """Generate an initial state for the configuration directory in tests.constants.ROOT_FS
     """


ReimarBauer (commented):
this was added to simulate an update of the msui_settings.json former mss_settings.json.

We restart the app when the config is changed by the config editor. This was not an automated test, because changes of the default values changes the behaviour of all other tests.

Maybe that can now differently added.

matrss:
> this was added to simulate an update of the msui_settings.json former mss_settings.json.

Why couldn't that happen inside of a test?

> We restart the app when the config is changed by the config editor. This was not an automated test, because changes of the default values changes the behaviour of all other tests.

I don't see how this necessitates setting a config file as a pytest option. This should just happen in an ordinary test.

 Also, affecting other tests should no longer be an issue since the config file is reset before each test. And also, if a test isn't automated it might as well not exist, because no one will regularly run it.

> Maybe that can now differently added.

As its own normal test case, if I understand your issue correctly. I might not, though.

ReimarBauer:
I don't know if we have currently a test active which verifies the reload of the configuration by the restart of the application, https://github.com/Open-MSS/MSS/blob/develop/mslib/msui/msui_mainwindow.py#L998

In the past that made trouble and I think all these were skipped

ReimarBauer:
I don't disagree with removing it. At the moment I'm more concerned that we don't add it again afterwards and possibly set up a runner.

matrss:
> I don't know if we have currently a test active which verifies the reload of the configuration by the restart of the application, https://github.com/Open-MSS/MSS/blob/develop/mslib/msui/msui_mainwindow.py#L998
> 
> In the past that made trouble and I think all these were skipped

I think the test you refer to is this one: https://github.com/Open-MSS/MSS/blob/1089e365b6144c68771785ab9793986f83f08395/tests/_test_msui/test_editor.py#L60-L67

All tests in that file are unconditionally skipped because no one bothered to update them after there was a rework to the config editor. 

> I don't disagree with removing it. At the moment I'm more concerned that we don't add it again afterwards and possibly set up a runner.

The problem you describe is that the above test was never reworked. Actually fixing that test is a different issue than removing an unused and probably broken pytest option.

ReimarBauer (approved):
thx

PR #2171, opened 2024-02-01
Remove unnecessary config settings

matrss:
Since pycodestyle is not used directly, but only as part of flake8, the flake8 settings apply already. Therefore it is unnecessary to duplicate these settings.


ReimarBauer (approved):
thx

PR #2173, opened 2024-02-02
Simplify GitHub Actions workflows

### LINKED ISSUE #2189: Simplify GitHub Actions workflows
There is a bunch of potential to simplify the workflows. Some of those are:
- remove `success()` in `if:` blocks, since that is the implicit default
- replace `source` and `activate` calls with just `mamba run`, because that is more concise and readable
- remove ineffective `inputs:` definition for pull_request event, because it does nothing
- remove the specification of bash as the shell, since the default of `sh` works just fine
- rearrange shell invocations to remove unnecessary backslashes, because I find that to be much more pleasant to read
- replace `always()` with `!cancelled()` in `if:`, since the first would even run if the workflow is manually cancelled
- move trusting the git repository to the location where it is needed
- do not hardcode specific GSoC branches
- remove unused env variables

matrss:
This removes a bit of unnecessary code. The resulting workflows should do effectively the same as before. The changes and reasoning are these:
- remove `success()` in `if:` blocks, since that is the implicit default
- replace `source` and `activate` calls with just `mamba run`, because that is more concise and readable
- remove ineffective `inputs:` definition for pull_request event, because it does nothing
- remove the specification of bash as the shell, since the default of `sh` works just fine
- rearrange shell invocations to remove unnecessary backslashes, because I find that to be much more pleasant to read
- replace `always()` with `!cancelled()` in `if:`, since the first would even run if the workflow is manually cancelled
- move trusting the git repository to the location where it is needed
- do not hardcode specific GSoC branches
- remove unused env variables

Fixes #2189.

Depends on #2158.

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -14,16 +14,11 @@ on:
         required: true
 
 env:
-  PAT: ${{ secrets.PAT }}
-  EVENT: ${{ inputs.event_name }}
+  mamba-env: mss-${{ inputs.branch_name }}-env
 
 jobs:
   Test-MSS:
     runs-on: ubuntu-latest
-
-    defaults:
-      run:
-        shell: bash
 
     container:
       image: openmss/testing-${{ inputs.branch_name }}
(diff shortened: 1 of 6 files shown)


.github/workflows/testing.yml
```diff
@@ -14,17 +14,12 @@ on:
         required: true
 
 env:
-  PAT: ${{ secrets.PAT }}
```
joernu76, line 17: sure that we do not need the PAT anymore? I think that was necessary for the triggering of the docker-build in the other repo?

matrss: For the workflow-dispatch action that triggers the docker build it is set directly from `secrets.PAT`, not through `env`. See here: https://github.com/Open-MSS/MSS/blob/dc4db0bde011c559bb5ed0eff2530f47876d8a26/.github/workflows/testing.yml#L76-L80

Way back when this was introduced `env.PAT` was used in an `if:` clause here: https://github.com/Open-MSS/MSS/blob/177bef6cdd1d9dd92c32f8d1be1bf9b795f2054a/.github/workflows/testing.yml#L42-L46

I am pretty sure that no other step implicitly relies (or even should rely) on the `PAT` environment variable to be set and there is no longer an explicit reference to `env.PAT`, so I think it can be removed.

ReimarBauer (approved):
let us try

PR #2177, opened 2024-02-05
Fix : UTC based timezone

### LINKED ISSUE #2122: use UTC based timestamps
We use in mscolab `datetime.datetime.now() `
This can have a timezone issue, we should replace it by a timezone based on UTC
`datetime.datetime.now(tz=datetime.timezone.utc)`

Preetam-Das26:
**Purpose of PR?**:

Fixes #2122

NO

**Checklist:**
- [ ] Commit has unit tests


ReimarBauer:
please remove the pytest.log

.github/workflows/python-flake8.yml
```diff
@@ -31,4 +31,5 @@ jobs:
       run: |
         python -m pip install --upgrade pip
         pip install flake8 flake8-builtins
+
```
ReimarBauer: ths change is not related to the issue

tests/_test_mscolab/test_file_manager.py
```diff
@@ -65,7 +65,7 @@ def test_modify_user(self):
             user = User("user@example.com", "user", "password")
             assert user.id is None
             assert User.query.filter_by(emailid=user.emailid).first() is None
-            # creeat the user
+            # create the user
```
ReimarBauer, line 68: :) good catch

ReimarBauer:
on "future" fixes you may be want to create a branch for an issue you work on.  git -b name

our stable branch means released on conda-forge.

ReimarBauer (changes requested):
remove accidantly committed files, line added to /python-flake8.yml

Preetam-Das26:
Ok sure

Preetam-Das26:
@ReimarBauer Kindly review the recent changes made .

ReimarBauer (approved):
Thx

ReimarBauer:
not seen that I had to enable tests ;) let us see what they tell.

instance/mscolab.db
matrss, line 1: This does not belong in the repository.

.github/workflows/python-flake8.yml
```diff
@@ -30,5 +30,5 @@ jobs:
     - name: Lint with flake8
       run: |
         python -m pip install --upgrade pip
-        pip install flake8 flake8-builtins
+        pip install flake8 flake8-builtins
```
matrss: I think this happened by mistake.

matrss (changes requested):

tests/_test_mscolab/test_file_manager.py
```diff
@@ -74,12 +74,12 @@ def test_modify_user(self):
...
             assert self.fm.modify_user(user, action="create") is False
             # confirming the user
-            confirm_time = datetime.datetime.now() + datetime.timedelta(days=1)
+            confirm_time = datetime.datetime.now(tz=datetime.timezone.utc) + datetime.timedelta(days=1)
             self.fm.modify_user(user_query, attribute="confirmed_on", value=confirm_time)
             self.fm.modify_user(user_query, attribute="confirmed", value=True)
             user_query = User.query.filter_by(id=user.id).first()
             assert user_query.confirmed is True
-            assert user_query.confirmed_on == confirm_time
+            assert user_query.confirmed_on.replace(tzinfo=None) == confirm_time.replace(tzinfo=None)
```
matrss, line 82: I think this change hides a bug.

When I revert this I get the following error:
```
FAILED tests/_test_mscolab/test_file_manager.py::Test_FileManager::test_modify_user - assert datetime.datetime(2024, 2, 9, 8, 49, 39, 968400) == datetime.datetime(2024, 2, 9, 8, 49, 39, 968400, tzinfo=datetime.timezone.utc)
```

This means that the database query returns a naive datetime object, which is then not equal to the timezone aware object. I think the database query should return an aware datetime object instead, but I am not yet sure how to make it do that.

matrss:
I also just noticed that we have a bunch of calls to `datetime.datetime.utcnow()`, which is even worse than `.now()` without a timezone. We should address those as well, but that could also be a follow-up PR.

tests/_test_mscolab/test_file_manager.py
Preetam-Das26: Ok I didn't realise this situation. I will look into it.

matrss: This can also be handled in a follow-up PR because it is not exactly related to the initial issue that is addressed here. But I would prefer if this line stays as is (i.e. `assert user_query.confirmed_on == confirm_time`), and the test be marked as xfail instead to acknowledge that this is an issue.

matrss (approved):

PR #2180, opened 2024-02-08
Remove executable bit from text files

### LINKED ISSUE #2181: Some text files have the executable bit set
I.e. NOTICE, CHANGES.rst, mslib/msui/flighttrack.py and mslib/msui/satellite_dockwidget.py.

matrss:
Text files definitely don't need it.

Fixes #2181.


ReimarBauer:
This is likly also in stable, can you do it there please. We will merge stable afterwards to develop. 
Also please add an issue.  The issues are listed in a new revision.

matrss:
I think I don't understand yet when something should go to develop and when it should go to stable. This issue is definitely not urgent or critical in any sense and doesn't require to hit a release any time soon, it is just something I stumbled across and quickly threw in a PR. Shouldn't stuff like that go to develop, and reach stable at some point through a release?

Anyway, I can re-target this to stable as well, if you want.

ReimarBauer:
stable would be good. 

so in principle when something is in stable which was the past from develop and we can solve it there we do it there. Then we get it with a merge (less conflicts) into develop.

When something is further developed already so it is different to what we have in stable, we do it in develop. Currently develop is quite different from stable but this is also there an issiue.

matrss:
OK. I don't think I like this approach as there are effectively two branches that are actively being worked on and there is a lot of potential for merge issues between them, but it is what it is I guess.

matrss:
> Also please add an issue. The issues are listed in a new revision.

I've added an issue, but it is also possible to add milestones to PRs (PRs and issues are treated very similar by GitHub, they even share the same auto-increment counter. A PR is basically just an issue with code attached.). If the PR has a milestone attached it would be listed in the issues closed by a milestone, which is then referenced by a release. For some things (e.g. a small code change without a bigger issue attached, like here) I just think that creating both an issue and a PR adds unnecessary overhead.

ReimarBauer:
We can maybe use also the Milestone in the PR for such things. An issue has a different position. First agree that it is an issue and discuss a scope how to fix it.  

Issues have also the option to declare them as "good first issue".

matrss:
> We can maybe use also the Milestone in the PR for such things.

If we added milestones to both a PR and an issue for the same topic then this would become redundant. So either an issue with a milestone attached and then a PR for that issue or a PR with a milestone but without an issue. Having everything show up twice in the changelog would be bad.

> An issue has a different position. First agree that it is an issue and discuss a scope how to fix it.

I agree, but only if the change even necessitates a discussion. I think there is no discussion necessary for removing the executable bit from e.g. the changelog, so I didn't open an issue. Also, if I first opened an issue and waited long enough for anyone who wants to to chime in then a 5 minute task suddenly gets stretched over days, if I don't immediately open the PR as well. If I open then PR immediately then first creating an issue is just overhead, discussion can happen in PRs as well and if the change is unwanted those can just be closed.

TL;DR if the change is rather uncontroversial and a ~5 minute task to code up then I think an issue is just unnecessary. But that is just my opinion, if you want I can also start opening issues for those things.

ReimarBauer:
A newcomer to open source,  github and MSS needs very simple issues.  A 5 minute task for us is an ideally good first issue for someone learning all the tools and workflows. We only have to take care that we have a few of these.

So simply consider whether you should do it yourself now or make a good first issue out of it and wait a while.  You did this already. We tend to have hardly any good first issues, because we also do everything that is easy immediately.  So this is just a suggestion or a reminder of very simple task for others.
 
For a release, you can also implement it yourself, you just don't have to do it at the moment when the task catches your eye. This is only a suggestion to explain how I try to do these things.

ReimarBauer (approved):
Thx

matrss:
> A newcomer to open source, github and MSS needs very simple issues. A 5 minute task for us is an ideally good first issue for someone learning all the tools and workflows. We only have to take care that we have a few of these.
> 
> So simply consider whether you should do it yourself now or make a good first issue out of it and wait a while. You did this already. We tend to have hardly any good first issues, because we also do everything that is easy immediately. So this is just a suggestion or a reminder of very simple task for others.

I didn't think about it that way. I'll keep it in mind.

PR #2195, opened 2024-02-10
[Change] datetime.datetime.utcnow()

### LINKED ISSUE #2183: Do not use `datetime.datetime.utcnow()`
There are multiple locations where this function is used. It is documented as being potentially dangerous and should be replaced with `datetime.datetime.now(tz=datetime.timezone.utc)`: https://docs.python.org/3/library/datetime.html#datetime.datetime.utcnow

This came up in https://github.com/Open-MSS/MSS/pull/2177#issuecomment-1933712007.

Related to #2122.

vikas-chaudhary-2802:
Fix #2183 

This pull request addresses the usage of the datetime.datetime.utcnow() function, which is documented as potentially dangerous due to its reliance on the system's timezone settings. Instead, this PR replaces all occurrences of datetime.datetime.utcnow() with datetime.datetime.now(tz=datetime.timezone.utc)

### DIFF
--- a/mslib/mscolab/file_manager.py
+++ b/mslib/mscolab/file_manager.py
@@ -64,7 +64,7 @@ class FileManager:
         if proj_available is not None:
             return False
         if last_used is None:
-            last_used = datetime.datetime.utcnow()
+            last_used = datetime.datetime.now(tz=datetime.timezone.utc)
         operation = Operation(path, description, last_used, category, active=active)
         db.session.add(operation)
         db.session.flush()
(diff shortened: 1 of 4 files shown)


vikas-chaudhary-2802:
@ReimarBauer Kindly, Review my PR

ReimarBauer:
Hi @vikas-chaudhary-2802 
below the comment fields is a section which looks like this

looks_like_this

please look yourself on the findings of the test suite and fix those. You can do this by clicking on Details.

ReimarBauer:
solved in some other PR

PR #2196, opened 2024-02-10
Fix: Scheduled test now triggers testing-stable.yml using workflow_dispatch

### LINKED ISSUE #2172: The scheduled test does not test stable properly
The scheduled test runs the workflow from develop with the code from develop, but with the container image for stable. This is wrong, it should run the workflow from stable with the code from stable and its corresponding container image instead.

This is not so trivial to fix, because scheduled workflows only run for the default branch. I think the most straight forward approach would be to use workflow_dispatch and trigger a workflow run on stable from a scheduled workflow on develop.

rohit2p:
**Purpose of PR?**: This pull request addresses the issue where the scheduled test was running the `testing.yml` workflow from the develop branch but using the container image for the stable branch. This was incorrect behavior.

Fixes #2172 

Modified the` testing-scheduled.yml` workflow on the develop branch to directly trigger the `testing-stable.yml` workflow using the `benc-uk/workflow-dispatch@v1.2.2 action.`

This change was made to work directly on the develop branch.

**Checklist:**
- [ X] Bug fix. Fixes #2172

### DIFF
--- a/.github/workflows/testing-scheduled.yml
+++ b/.github/workflows/testing-scheduled.yml
@@ -9,7 +9,7 @@ jobs:
   trigger-testing-stable:
     runs-on: ubuntu-latest
     permissions:
-        actions: write
+      actions: write
     steps:
       - uses: benc-uk/workflow-dispatch@v1.2.2
         with:
--- a/.github/workflows/testing-scheduled.yml
+++ b/.github/workflows/testing-scheduled.yml
@@ -6,12 +6,14 @@ on:
 
 
 jobs:
-  trigger_stable_test:
+  trigger-testing-stable:
     runs-on: ubuntu-latest
+    permissions:
+        actions: write
     steps:
       - uses: benc-uk/workflow-dispatch@v1.2.2
         with:
-          workflow: testing-stable
+          workflow: testing-stable.yml
           ref: stable
 
   test-develop-scheduled:
--- a/.github/workflows/testing-scheduled.yml
+++ b/.github/workflows/testing-scheduled.yml
@@ -5,16 +5,14 @@ on:
     - cron: '30 5 * * 1' 
 
 
-jobs:     
-  test-stable-scheduled:
-    uses: 
-      ./.github/workflows/testing.yml
-    with:
-      xdist: no
-      branch_name: stable
-      event_name: ${{ github.event_name }}
-    secrets:
-      PAT: ${{ secrets.PAT }}
+jobs:
+  trigger_stable_test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: benc-uk/workflow-dispatch@v1.2.2
+        with:
+          workflow: testing-stable
+          ref: stable
 
   test-develop-scheduled:
     uses: 


.github/workflows/testing-scheduled.yml
```diff
@@ -5,16 +5,14 @@ on:
...
-    uses: 
-      ./.github/workflows/testing.yml
-    with:
-      xdist: no
-      branch_name: stable
-      event_name: ${{ github.event_name }}
-    secrets:
-      PAT: ${{ secrets.PAT }}
+jobs:
+  trigger_stable_test:
```
matrss:
```suggestion
  trigger-testing-stable:
```

Just to keep a somewhat consistent naming scheme.

.github/workflows/testing-scheduled.yml
```diff
@@ -5,16 +5,14 @@ on:
...
+    runs-on: ubuntu-latest
```
matrss, line 10:
```suggestion
    runs-on: ubuntu-latest
    permissions:
      actions: write
```

While testing I found out that this permission is required to be able to use the runner token to trigger another workflow.

.github/workflows/testing-scheduled.yml
```diff
@@ -5,16 +5,14 @@ on:
...
+    steps:
+      - uses: benc-uk/workflow-dispatch@v1.2.2
+        with:
+          workflow: testing-stable
```
matrss:
```suggestion
          workflow: testing-stable.yml
```

This can be either the filename of the workflow or the full name specified at the top of the workflow file. I think I prefer using the filename.

matrss (changes requested):
I left a few comments. In addition to those the testing-stable.yml workflow on the stable branch would need to include the `workflow_dispatch:` trigger. Since that is on a different branch it must happen in an additional PR, unfortunately.

rohit2p:
@matrss i have some question, Do you prefer me to create a separate branch form the stable branch for this new PR or should i directly work on stable. Because stable branch defines the workflow for stable tests. Modifying this file directly could break the existing stable tests that's way i was confused.
the last tym when you said it need to go to develop branch i think misunderstood you and i thought you are saying me to directly work on develop branch. my apologies 
also while adding the trigger is this how the code should look like?? 
```
on:
  schedule:
    - cron: '30 5 * * 1'  # Existing schedule trigger
  workflow_dispatch:      # Add the workflow_dispatch trigger
    jobs:     
      test-stable-scheduled:
        uses: 
          ./.github/workflows/testing.yml
# Rest of the code as it is
```
or do i need to add permission or any requirement  like this?
```
on:
  schedule:
    - cron: '30 5 * * 1'
  workflow_dispatch:
    permissions:
      actions: write  
```

matrss:
> Do you prefer me to create a separate branch form the stable branch for this new PR or should i directly work on stable. Because stable branch defines the workflow for stable tests. Modifying this file directly could break the existing stable tests that's way i was confused.
> the last tym when you said it need to go to develop branch i think misunderstood you and i thought you are saying me to directly work on develop branch.

What I meant with "it needs to go to develop" is that this change to the scheduled workflow ultimately needs to go into the develop branch of Open-MSS/MSS (i.e. this repository). This would happen through this PR here. What the source branch of the PR is (in this case the develop branch of your fork of this repository) is usually not important, although for my own stuff I prefer to use feature branches named after what they implement.

So what I meant regarding stable is that ultimately the `testing-stable.yml` workflow on the stable branch of Open-MSS/MSS needs to have the workflow_dispatch event added so that it can be triggered by the scheduled workflow running on develop.

> also while adding the trigger is this how the code should look like??
> 
> ```
> on:
>   schedule:
>     - cron: '30 5 * * 1'  # Existing schedule trigger
>   workflow_dispatch:      # Add the workflow_dispatch trigger
> 
> jobs:     
>   test-stable-scheduled:
>     uses: 
>       ./.github/workflows/testing.yml
> # Rest of the code as it is
> ```

This, but in `testing-stable.yml` instead of `testing-scheduled.yml`.

To explain what it does:
The
```
on:
  workflow_dispatch:
```
part tells GitHub Actions that the workflow that it is in can be started through a workflow_dispatch event. The benc-uk/workflow-dispatch action uses this event type to trigger an execution of that workflow. Therefore the workflow that is being started needs to "react" to this type of event.

That is also why it needs to go into stable: the `testing-scheduled.yml` workflow triggers the `testing-stable.yml` workflow on the stable branch to run. For this to successfully happen the `testing-stable.yml` workflow on the stable branch needs to be startable through the workflow_dispatch event.

> or do i need to add permission or any requirement like this?
> 
> ```
> on:
>   schedule:
>     - cron: '30 5 * * 1'
>   workflow_dispatch:
>     permissions:
>       actions: write  
> ```

No, the permission just need to be set on the job that uses the benc-uk/workflow-dispatch action (as it now already is), so that this action is allowed to start another workflow run through the workflow_dispatch event.

Does that clarify things a bit?

.github/workflows/testing-scheduled.yml
```diff
@@ -5,16 +5,16 @@ on:
...
-      xdist: no
-      branch_name: stable
-      event_name: ${{ github.event_name }}
-    secrets:
-      PAT: ${{ secrets.PAT }}
+jobs:
+  trigger-testing-stable:
+    runs-on: ubuntu-latest
+    permissions:
+        actions: write
```
matrss:
```suggestion
      actions: write
```

This should just be indented with two spaces more compared to the previous line, to stay consistent with the rest of the file.

matrss (changes requested):

rohit2p:
Yes, Thank you for your guidance!
Does now everything look cool?

matrss (approved):
Looks good to me. Thanks a lot for the PR! There is a conflict between this PR and Open-MSS:develop, which is caused by a recent commit on develop (3a0eb968516da8714361e1f284bd955715833cf4) that also touched the testing-scheduled.yml workflow. If you want to fix that conflict yourself feel free to do so, otherwise this could be done when merging as well.

I've tested this change on my fork of MSS and it works so far:
https://github.com/matrss/MSS/actions/runs/7902483658
This workflow run triggered a run of testing-stable.yml that ran on stable as it should:
https://github.com/matrss/MSS/actions/runs/7902485837

@ReimarBauer what do you think about this change? It would need to be merged together with #2197.

rohit2p:
I've resolved the conflict in the testing-scheduled.yml file by keeping the "trigger-testing-stable" job from the develop branch.

I removed the unused section related to the "test-stable-scheduled" job and cleaned up the conflict markers. Please review the changes and let me know if everything looks good.

ReimarBauer:
#2197 shows a coredump

matrss:
> #2197 shows a coredump

See my comment in that PR.

PR #2197, opened 2024-02-13
Add workflow_dispatch trigger to enable manual stable tests

### LINKED ISSUE #2172: The scheduled test does not test stable properly

rohit2p:
**Purpose of PR?**:
This pull request builds upon my previous change in PR #2196 
I've added the `workflow_dispatch:` trigger to the `testing-stable.yml` file on the stable branch. This allows manual triggering of stable tests, improving testing flexibility and control.

Fixes #2172 

**Checklist:**
- [x] Bug fix. Fixes #2172


rohit2p:
The error message I'm seeing is `Fatal Python error: Aborted`. I've checked the relevant code section but haven't been able to identify the issue.
Could you please help me understand what might be causing this failure and how to fix it?

rohit2p:
not sure, but is it because any update  regarding the `actions/checkout` version
image

matrss:
> The error message I'm seeing is `Fatal Python error: Aborted`. I've checked the relevant code section but haven't been able to identify the issue. Could you please help me understand what might be causing this failure and how to fix it?

This is happening on the stable branch already without this PR, and I don't know yet what is broken there. This PR certainly did not cause it.

> not sure, but is it because any update regarding the `actions/checkout` version image

This warning is also unrelated to this PR, the actions/checkout version needs to be updated.

matrss (approved):
LGTM

rohit2p:
Thanks for your comment. Please let me know if you need anything further from me.

ReimarBauer:
Hmm, ideas why we now have these coredumps?

matrss:
@ReimarBauer they are happening on the stable branch already, e.g. here: https://github.com/Open-MSS/MSS/actions/runs/7854540812/job/21569532674. So that is unrelated to this PR. I am very much sure that it is because of the missing pyvirtualdisplay/xvfb-run. The issue for that is here: #2199.

ReimarBauer:
> @ReimarBauer they are happening on the stable branch already, e.g. here: https://github.com/Open-MSS/MSS/actions/runs/7854540812/job/21569532674. So that is unrelated to this PR. I am very much sure that it is because of the missing pyvirtualdisplay/xvfb-run. The issue for that is here: #2199.

ok, this fix must then be first merged then we can recheck this one. With "repeating" failing tests other merge requests will be more difficult to merge.

matrss:
In general I'd agree, but this PR is literally a single line change that makes the workflow manually start-able, nothing else.

Slightly OT: the stable/develop split makes it so that failures on stable (i.e. the non-default branch of the two) are not as easily visible, since they show up nowhere on the repo landing page. I don't like that. Also, the change that broke the CI happened in a different repository. I also don't like that that is possible.

Anyway, I'll send a fix for the broken CI.

PR #2205, opened 2024-02-15
Install pyvirtualdisplay in testing workflow

### LINKED ISSUE #2199: Fix CI test runs on stable
The CI test runs on stable fail consistently right now. Looking through the list of installed packages I saw that pyvirtualdisplay is now missing in the list for the recent runs of the tests. This might be the reason for the failures.

matrss:
The testing-stable container image no longer includes pyvirtualdisplay due to a change that made it unnecessary for develop that also affected the stable image. This installs the package as part of the testing workflow instead. This is a band-aid to fix the stable CI until all of the recent changes to the tests are backported from develop to stable.

Fixes #2199.

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -43,6 +43,13 @@ jobs:
         || (echo Dependencies differ \
             && echo "triggerdockerbuild=yes" >> $GITHUB_ENV )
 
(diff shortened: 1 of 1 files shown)


ReimarBauer (approved):
ok

PR #2209, opened 2024-02-15
Return aware datetime objects

### LINKED ISSUE #2182: Database queries should return aware datetime objects
Right now queries for e.g. the `User.confirmed_on` field in mscolab, but also other datetime database fields, return naive datetime objects in UTC. This is bad for the same reason as documented with the `datetime.utcnow()` function here: https://docs.python.org/3/library/datetime.html#datetime.datetime.utcnow

This came up in https://github.com/Open-MSS/MSS/pull/2177#discussion_r1482685642.

Related to #2122 and #2183.

Preetam-Das26:
**Purpose of PR?**:

Fixes #2182 

**Checklist:**
- [ ] Commit has unit tests

### DIFF
--- a/mslib/mscolab/models.py
+++ b/mslib/mscolab/models.py
@@ -34,7 +34,7 @@ from mslib.mscolab.message_type import MessageType
 import sqlalchemy.types as types
 
 
-class awareDateTime(types.TypeDecorator):
+class AwareDateTime(types.TypeDecorator):
     impl = types.DateTime
 
     def process_bind_param(self, value, dialect):
@@ -55,9 +55,9 @@ class User(db.Model):
     username = db.Column(db.String(255))
     emailid = db.Column(db.String(255), unique=True)
     password = db.Column(db.String(255), unique=True)
-    registered_on = db.Column(awareDateTime, nullable=False)
+    registered_on = db.Column(AwareDateTime, nullable=False)
     confirmed = db.Column(db.Boolean, nullable=False, default=False)
-    confirmed_on = db.Column(awareDateTime, nullable=True)
+    confirmed_on = db.Column(AwareDateTime, nullable=True)
     permissions = db.relationship('Permission', cascade='all,delete,delete-orphan', backref='user')
     authentication_backend = db.Column(db.String(255), nullable=False, default='local')
 
--- a/mslib/mscolab/models.py
+++ b/mslib/mscolab/models.py
@@ -42,7 +42,7 @@ class User(db.Model):
     password = db.Column(db.String(255), unique=True)
     registered_on = db.Column(db.DateTime, nullable=False)
     confirmed = db.Column(db.Boolean, nullable=False, default=False)
-    confirmed_on = db.Column(TIMESTAMP(timezone=True), nullable=True, default=None)
+    confirmed_on = db.Column(db.TIMESTAMP(timezone=True), nullable=True, default=None)
     permissions = db.relationship('Permission', cascade='all,delete,delete-orphan', backref='user')
     authentication_backend = db.Column(db.String(255), nullable=False, default='local')
 
--- a/mslib/mscolab/models.py
+++ b/mslib/mscolab/models.py
@@ -28,10 +28,12 @@
 import datetime
 import logging
 import jwt
+
 from passlib.apps import custom_app_context as pwd_context
+import sqlalchemy.types
+
 from mslib.mscolab.app import db
 from mslib.mscolab.message_type import MessageType
-import sqlalchemy.types
 
 
 class AwareDateTime(sqlalchemy.types.TypeDecorator):
--- a/mslib/mscolab/models.py
+++ b/mslib/mscolab/models.py
@@ -42,7 +42,9 @@ class User(db.Model):
     password = db.Column(db.String(255), unique=True)
     registered_on = db.Column(db.DateTime, nullable=False)
     confirmed = db.Column(db.Boolean, nullable=False, default=False)
-    confirmed_on = db.Column(db.DateTime, nullable=True)
+    confirmed_on = db.Column(db.DateTime(timezone=True),
+                             nullable=True,
+                             default=datetime.datetime.now(tz=datetime.timezone.utc))
     permissions = db.relationship('Permission', cascade='all,delete,delete-orphan', backref='user')
     authentication_backend = db.Column(db.String(255), nullable=False, default='local')
 
--- a/mslib/mscolab/models.py
+++ b/mslib/mscolab/models.py
@@ -42,9 +42,7 @@ class User(db.Model):
     password = db.Column(db.String(255), unique=True)
     registered_on = db.Column(db.DateTime, nullable=False)
     confirmed = db.Column(db.Boolean, nullable=False, default=False)
-    confirmed_on = db.Column(db.DateTime(timezone=True),
-                             nullable=True,
-                             default=datetime.datetime.now(tz=datetime.timezone.utc))
+    confirmed_on = db.Column(db.DateTime(timezone=True), nullable=True, default=None)
     permissions = db.relationship('Permission', cascade='all,delete,delete-orphan', backref='user')
     authentication_backend = db.Column(db.String(255), nullable=False, default='local')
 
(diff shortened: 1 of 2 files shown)


Preetam-Das26:
@matrss kindly checkout the PR .

mslib/mscolab/models.py
```diff
@@ -42,7 +42,9 @@ class User(db.Model):
     password = db.Column(db.String(255), unique=True)
     registered_on = db.Column(db.DateTime, nullable=False)
     confirmed = db.Column(db.Boolean, nullable=False, default=False)
-    confirmed_on = db.Column(db.DateTime, nullable=True)
+    confirmed_on = db.Column(db.DateTime(timezone=True),
+                             nullable=True,
+                             default=datetime.datetime.now(tz=datetime.timezone.utc))
```
matrss: This shouldn't have a default: when creating a user their mail address is usually not confirmed immediately at creation, but at a later point in time. (Or the default should be None, which I think it is with `nullable=True` but no default set).

tests/_test_mscolab/test_file_manager.py
```diff
@@ -79,7 +79,7 @@ def test_modify_user(self):
             self.fm.modify_user(user_query, attribute="confirmed", value=True)
             user_query = User.query.filter_by(id=user.id).first()
             assert user_query.confirmed is True
-            assert user_query.confirmed_on.replace(tzinfo=None) == confirm_time.replace(tzinfo=None)
+            assert user_query.confirmed_on.replace(tzinfo=datetime.timezone.utc) == confirm_time
```
matrss:
```suggestion
            assert user_query.confirmed_on == confirm_time
```

I think the query itself should already return an aware datetime object. From what I've read this might require a custom column type in the database model.

Preetam-Das26: @matrss Can you please elaborate on this

Preetam-Das26:
Kindly review the latest commit and help me resolve this issue.@matrss

mslib/mscolab/models.py
```diff
@@ -31,6 +31,21 @@
 from passlib.apps import custom_app_context as pwd_context
 from mslib.mscolab.app import db
 from mslib.mscolab.message_type import MessageType
+import sqlalchemy.types as types
+
+
+class Awaredatetime(types.TypeDecorator):
```
matrss:
```suggestion
class AwareDateTime(types.TypeDecorator):
```

The class name should stay consistent with the usual CamelCase naming convention for classes.

mslib/mscolab/models.py
```diff
@@ -31,6 +31,21 @@
...
+    impl = types.DateTime(timezone=True)
```
matrss:
```suggestion
    impl = types.DateTime
```

Removing the timezone=True parameter has the advantage that the underlying database representation stays the same as before in all cases, i.e. no database migration is necessary. This is nicer for backwards-compatibility. The database column then would continue to contain the timestamp without any timezone information. But this is fine, because on the application side we get a guarantee that all timestamps are first converted to UTC and also get returned as UTC. This means the database will only ever contain UTC timestamps.

mslib/mscolab/models.py
```diff
@@ -42,7 +57,7 @@ class User(db.Model):
     password = db.Column(db.String(255), unique=True)
     registered_on = db.Column(db.DateTime, nullable=False)
     confirmed = db.Column(db.Boolean, nullable=False, default=False)
-    confirmed_on = db.Column(db.DateTime, nullable=True)
+    confirmed_on = db.Column(Awaredatetime, nullable=True, default=None)
```
matrss:
```suggestion
    confirmed_on = db.Column(AwareDateTime, nullable=True)
```

I think we should just leave out the default entirely, as before.

matrss (changes requested):
Yes, this is what I had in mind :)

You can see here what exactly the testing issue is: https://github.com/Open-MSS/MSS/actions/runs/7932202441/job/21658127074?pr=2209#step:9:1133

As you can see, the confirmed_on datetime is compared with the registered_on datetime. The latter is still stored in a db.DateTime column, so it is returned as a naive datetime object and then compared to an aware datetime object. To fix this the registered_on database column just needs to be an AwareDateTime column as well. Then the tests should pass.

tests/_test_mscolab/test_file_manager.py
matrss: What I meant is exactly what you found yourself already: defining a custom type by augmenting the standard DateTime type with a conversion to UTC and interpreting the value in that database column as UTC. It is documented more in-depth here: https://docs.sqlalchemy.org/en/20/core/custom_types.html#augmenting-existing-types

Preetam-Das26:
Thank you @matrss .All the changes made .Kindly verify the changes made.

mslib/mscolab/models.py
```diff
@@ -31,6 +31,21 @@
 from passlib.apps import custom_app_context as pwd_context
 from mslib.mscolab.app import db
 from mslib.mscolab.message_type import MessageType
+import sqlalchemy.types as types
+
+
+class awareDateTime(types.TypeDecorator):
```
matrss: The class name should still start with a capital A.

Preetam-Das26:
Changes done @matrss

matrss (approved):
LGTM, thanks for the PR!

@ReimarBauer you are more familiar with the database stuff. What do you think about this? The same change should probably go to all other db.DateTime columns as well, but that is a bit more involved and can be another PR, I think.

ReimarBauer:
This needs a migrations script added. 

At least at release time.

ReimarBauer:
> LGTM, thanks for the PR!
> 
> @ReimarBauer you are more familiar with the database stuff. What do you think about this? The same change should probably go to all other db.DateTime columns as well, but that is a bit more involved and can be another PR, I think.

I look later on this. Just added a note because seeing model was changed.

matrss:
> This needs a migrations script added.
> 
> At least at release time.

See my comment here: https://github.com/Open-MSS/MSS/pull/2209#discussion_r1492639168

Since the underlying column type does not change I don't think we need a migration. This change is based on the assumption that, until now, we have also written UTC timestamps to the database, but just didn't explicitly ensure that.

mslib/mscolab/models.py
```diff
@@ -31,6 +31,21 @@
 from passlib.apps import custom_app_context as pwd_context
 from mslib.mscolab.app import db
 from mslib.mscolab.message_type import MessageType
+import sqlalchemy.types as types
```
ReimarBauer: this shadows the builtin types, rename  to e.g. sqlatypes

ReimarBauer (changes requested):
see comment

mslib/mscolab/models.py
matrss: There is no builtin `types`, I think you mean the singular `type`? If it did shadow a builtin flake8 should complain :)

Anyway, I wouldn't object to using `import sqlalchemy.types` and using the fully-qualified name in the code.

Preetam-Das26: Should I use import sqlalchemy.types?        
@ReimarBauer @matrss

ReimarBauer: It is not builtin but an utility function since python 3.3

It also defines names for some object types that are used by the standard Python interpreter, but not exposed as builtins like int or str are.

https://docs.python.org/3/library/types.html

fully qualified will  help us to refactor less when python occupies more names

Preetam-Das26:
Kindly check the changes @ReimarBauer

ReimarBauer:
Hi 
so far everything fine. we can merge it as is and have a review issue afterwards, to look up how we do the order of imports. A refactoring for FQN I would shift to 10.0.0 because it is another step of refactoring needed for server users. And I want to have this in smaller steps too.  Our Local application/library specific imports should be the last. 

https://peps.python.org/pep-0008/#imports

Last change for this PR if you like.

Preetam-Das26:
Changes done @ReimarBauer

ReimarBauer (approved):
Thx

PR #2217, opened 2024-02-16
cycle login/logout while using in more than one topview the multiple flightpath widget

### LINKED ISSUE #2037: multiple flightpath docking widget does not clear all topviews
when the multiple flightpath docking widget is used in two or more topviews on logout only one gets cleared.

on a new connect strange failure occure e.g.

```
Fatal error in MSS 8.3.0 on Linux-6.2.0-10022-tuxedo-x86_64-with-glibc2.35
Python 3.10.8 | packaged by conda-forge | (main, Nov 22 2022, 08:23:14) [GCC 10.4.0]

  File "/home/reimar/MAIN/MSS/mslib/msui/mscolab.py", line 586, in after_login
  File "/home/reimar/MAIN/MSS/mslib/msui/socket_control.py", line 58, in __init__
    self.sio.connect(self.mscolab_server_url)
  File "/home/reimar/mambaforge/envs/mssdev/lib/python3.10/site-packages/socketio/client.py", line 347, in connect
    raise exceptions.ConnectionError(
socketio.exceptions.ConnectionError: One or more namespaces failed to connect

  File "/home/reimar/MAIN/MSS/mslib/msui/multiple_flightpath_dockwidget.py", line 181, in logout
    self.operations.logout_mscolab()
AttributeError: 'NoneType' object has no attribute 'logout_mscolab'
```

```
  File "/home/reimar/MAIN/MSS/mslib/msui/topview.py", line 365, in closed
    self.ui.signal_login_mscolab.disconnect()
TypeError: disconnect() failed between 'signal_login_mscolab' and all its connections
CRITICAL: MSS Version: 8.3.0
CRITICAL: Python Version: 3.10.8 | packaged by conda-forge | (main, Nov 22 2022, 08:23:14) [GCC 10.4.0]
CRITICAL: Platform: Linux-6.2.0-10022-tuxedo-x86_64-with-glibc2.35 (('64bit', 'ELF'))
  File "/home/reimar/MAIN/MSS/mslib/msui/topview.py", line 365, in closed
    self.ui.signal_login_mscolab.disconnect()
TypeError: disconnect() failed between 'signal_login_mscolab' and all its connections
```

ReimarBauer:
first step a test which reproduces parts of issue 2037
test for login/logout issue on a combination of topviews and multiple… flightpath while logged in

**Purpose of PR?**:

Fixes #2037


mslib/msui/topview.py
```diff
@@ -379,12 +379,16 @@ def openTool(self, index):
...
+        # A disconnect can be performed only once.
+        try:
+            self.mainwindow_signal_login_mscolab.disconnect()
+            self.mainwindow_signal_logout_mscolab.disconnect()
+            self.mainwindow_signal_listFlighttrack_doubleClicked.disconnect()
+            self.mainwindow_signal_activate_operation.disconnect()
+            self.mainwindow_signal_permission_revoked.disconnect()
+            self.mainwindow_signal_render_new_permission.disconnect()
+        except TypeError as ex:
+            logging.debug("mscolab connection already closed: %s", ex)
```
matrss: These `.disconnect()` calls disconnect *all* slots that are connected to the respective signals, which belong to a different object. I.e. when the topview window is closed all slots, even from different windows, that are connected to the mainwindow's signals get disconnected. This seems wrong to me. What is the intention of this code?

mslib/msui/multiple_flightpath_dockwidget.py
```diff
@@ -177,16 +178,21 @@ def __init__(self, parent=None, view=None, listFlightTracks=None,
...
-        self.ui.signal_render_new_permission.disconnect()
-        self.operations = None
-        self.flighttrack_list = True
-        self.operation_list = False
-        for idx in range(len(self.obb)):
-            del self.obb[idx]
+        try:
+            self.ui.signal_listFlighttrack_doubleClicked.disconnect()
+            self.ui.signal_permission_revoked.disconnect()
+            self.ui.signal_render_new_permission.disconnect()
```
matrss, line 186: Same as https://github.com/Open-MSS/MSS/pull/2217#discussion_r1494618935; I think this doesn't serve its intended purpose, which might be the underlying bug.

tests/_test_msui/test_mscolab.py
```diff
@@ -344,6 +344,36 @@ def test_view_open(self):
...
+        and we are able to cycle a login/logout
+        """
+        self._connect_to_mscolab()
+        modify_config_file({"MSS_auth": {self.url: self.userdata[0]}})
+        self._login(emailid=self.userdata[0], password=self.userdata[2])
+        # test after activating operation
+        self._activate_operation_at_index(0)
+        self.window.actionTopView.trigger()
+        QtWidgets.QApplication.processEvents()
+        assert len(self.window.get_active_views()) == 1
```
matrss: Qt is an inherently asynchronous system (at least once signals and slots come into play). This means that there is no guarantee (even with the processEvents call) that all events pertaining to creating the view are already handled once this assert is reached. I.e. there is a potential race condition between this assert and creating the view, which can fail the test.

A better approach is to use pytest-qt's `wait*` methods, specifically `wait_until`: https://pytest-qt.readthedocs.io/en/latest/wait_until.html

I've used it in other locations already to fix potential races that made the test suite unreliable, you can take a look there as well.

This applies to the other asserts as well.

matrss (changes requested):

mslib/msui/topview.py
ReimarBauer: when you use the multiple flightpath docking widget with mscolab, you have one connect done  to one server. There you use then different operations over this connection.

At the moment the logout in the profile is used, a disconnect to the server is driven and this can only be done once.

Without this exception handler we get a problem in the teardown which looks like this.

```
_ ERROR at teardown of Test_Mscolab.test_multiple_views_and_multiple_flightpath _
INFO:root:disconnected
INFO:root:dasdas
127.0.0.1 - - [19/Feb/2024 20:25:51] "GET /socket.io/?transport=websocket&EIO=4&sid=-blb&t=1708370751.269579 HTTP/1.1" 200 0 0.700590
INFO:root:Already disconnected from the server and switched to Fligh Tracks: True
INFO:root:Already disconnected from the server and switched to Fligh Tracks: True
DEBUG:root:switch_to_local
DEBUG:root:update_active_flight_track
Exceptions caught in Qt event loop:
________________________________________________________________________________
  File "/home/reimar/PycharmProjects/MSS/mslib/msui/topview.py", line 383, in closed
    self.mainwindow_signal_login_mscolab.disconnect()
TypeError: disconnect() failed between 'signal_login_mscolab' and all its connections

```
So yes it may hide a different problem in the teardown.  For the multiflight docking widget in a real example it is maybe not needed.

With the section removed the example shows that the second change may be enough:

without_topview

Output:
```
INFO: MSS Version: 8.3.3
INFO: Python Version: 3.10.8 | packaged by conda-forge | (main, Nov 22 2022, 08:23:14) [GCC 10.4.0]
INFO: Platform: Linux-6.5.0-10022-tuxedo-x86_64-with-glibc2.35 (('64bit', 'ELF'))
INFO: Already disconnected from the server and switched to Fligh Tracks: True
INFO: Already disconnected from the server and switched to Fligh Tracks: True
```

ReimarBauer: the function is not meant for closing the topview.

matrss: > With the section removed the example shows that the second change may be enough:

Do I understand that right, that the entire method can be removed and the test passes?

> the function is not meant for closing the topview.

Right, I miss-interpreted its name. It's called when the parent view closes, at which point it also makes little sense to disconnect all slots connected to signals on the mscolab mainwindow, IMO.

ReimarBauer: _Do I understand that right, that the entire method can be removed and the test passes?_

good idea, I try later.  I meant when I remove my changes, I can create the peek gifcycle but the test fails.

ReimarBauer:
Another test needs to verify what happens when multiple operations are drawn and then cycle login.

ReimarBauer:
hmm, redoing

PR #2218, opened 2024-02-17
Hardcoded img convention for route

### LINKED ISSUE #2178: Simplify route definition for mss_theme
In the following route definition the `<name>` parameter must always be "img", so we might as well hardcode it:

https://github.com/Open-MSS/MSS/blob/6bf94d35b4d49f1c24a161ff61a96542f9e6ac8b/mslib/index.py#L98-L104

Also, I think the route with a default is unnecessary.

workaryangupta:
**Purpose of PR?**:

Fixes #2178


ReimarBauer (approved):
Thx

PR #2220, opened 2024-02-18
Fix: Refined QMessageBox mocking

### LINKED ISSUE #2210: Remove too broad mocking of QMessageBox
There are some tests in `test_wms_control.py` which still have a too broad mock on QMessageBox. One of them is this one: https://github.com/Open-MSS/MSS/blob/167af85de1ae22293bacb423886991af0e70c7d3/tests/_test_msui/test_wms_control.py#L115-L121

Ideally I'd want this:
- since the assert only checks `.critical`, only that part should be mocked
- the mock should be moved into the function as a context manager in a `with:` statement around the `query_server` call
  - this isn't really that important in this instance, but in general for larger tests this makes it easier to understand which part a mock is needed for and to avoid accidentally affecting other parts of a test
- replace `.call_count == 1` with `.assert_called_once()`, since the mock already provides that as a method
  - or even `.assert_called_once_with(...)` and making sure that the critical message box has an error message that we expect

rohit2p:
**Purpose of PR?**:

This PR refactors the `test_no_server` function in `test_wms_control.py` to mock only the critical method of QMessageBox and replace `.call_count == 1 `with `.assert_called_once()`. This improves test clarity.

Fixes #2210

### DIFF
--- a/tests/_test_msui/test_wms_control.py
+++ b/tests/_test_msui/test_wms_control.py
@@ -112,13 +112,13 @@ class Test_HSecWMSControlWidget(WMSControlWidgetSetup):
         yield
         self._teardown()
 
-    @mock.patch("PyQt5.QtWidgets.QMessageBox")
-    def test_no_server(self, mockbox):
+    def test_no_server(self):
         """
         assert that a message box informs about server troubles
         """
-        self.query_server(f"{self.scheme}://{self.host}:{self.port-1}")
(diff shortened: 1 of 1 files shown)


matrss (changes requested):
This looks good. But there are 4 more test functions below this one that should get the same treatment.

matrss (approved):
LGTM, thanks for the PR!

PR #2225, opened 2024-02-20
Do not use datetime.datetime.utcnow()

### LINKED ISSUE #2183: Do not use `datetime.datetime.utcnow()`

Preetam-Das26:
**Purpose of PR?**:
datetime.datetime.utcnow()  replaced by  datetime.datetime.now(tz=datetime.timezone.utc)
Fixes #2183

**Checklist:**
- [ ] Commit has unit tests

### DIFF
--- a/mslib/mscolab/seed.py
+++ b/mslib/mscolab/seed.py
@@ -218,7 +218,10 @@ def archive_operation(path=None, emailid=None):
                 elif perm.access_level != "creator":
                     return False
                 operation.active = False
-                operation.last_used = datetime.datetime.utcnow() - dateutil.relativedelta.relativedelta(months=2)
+                operation.last_used = (
+                    datetime.datetime.now(tz=datetime.timezone.utc) -
+                    dateutil.relativedelta.relativedelta(months=2)
+                )
                 db.session.commit()
 
 
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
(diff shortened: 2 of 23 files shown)


Preetam-Das26:
@ReimarBauer @matrss 
Kindly review this PR .

mslib/mscolab/seed.py
```diff
@@ -218,7 +218,10 @@ def archive_operation(path=None, emailid=None):
                 elif perm.access_level != "creator":
                     return False
                 operation.active = False
-                operation.last_used = datetime.datetime.utcnow() - dateutil.relativedelta.relativedelta(months=2)
+                operation.last_used = (
```
ReimarBauer, line 221: add here a 
`#` ToDo use mscolab_settings.ARCHIVE_THRESHOLD instead of months=2

This remembers us, or will become a next issue. There may be more places. 
A change of this is in the current PR is not a good idea 

When changing something we have always to have in mind we could rollback, just one thing. But when things are mixed up it gets difficult.

In this case a `#` ToDo is the solution

ReimarBauer: I file a bug report for stable.

ReimarBauer: Now you don't need to add the ToDo's

ReimarBauer (approved):
Thx

PR #2230, opened 2024-02-22
Fix : Updated seed module to use "mscolab_settings.ARCHIVE_THRESHOLD"

### LINKED ISSUE #2227: Develop - seed module needs always to use mscolab_settings.ARCHIVE_THRESHOLD
In the seed module, there are still some hardcooded,  dateutil.relativedelta.relativedelta(months=2) 
the 2 must be exchanged by  mscolab_settings.ARCHIVE_THRESHOLD

workaryangupta:
Purpose of PR?

Fixes #2227 

Updated the seed module to always use `mscolab_settings.ARCHIVE_THRESHOLD` instead of hardcoded `dateutil.relativedelta.relativedelta(months=**2**)`.

### DIFF
--- a/docs/samples/config/mscolab/mscolab_settings.py.sample
+++ b/docs/samples/config/mscolab/mscolab_settings.py.sample
@@ -28,6 +28,9 @@ import os
 import logging
 
 class mscolab_settings:
+    # In the unit "months" when Operations get archived because not used
+    ARCHIVE_THRESHOLD = 2
+
     # Set which origins are allowed to communicate with your server
     CORS_ORIGINS = ["*"]
 
--- a/docs/samples/config/mscolab/mscolab_settings.py.sample
+++ b/docs/samples/config/mscolab/mscolab_settings.py.sample
@@ -27,6 +27,7 @@
 import os
 import logging
 
+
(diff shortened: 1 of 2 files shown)


docs/samples/config/mscolab/mscolab_settings.py.sample
```diff
@@ -27,7 +27,11 @@
 import os
 import logging
 
+
 class mscolab_settings:
+    # In the unit months when Operations get archived because not used
```
ReimarBauer, line 32: good to add it here too. i just looked how it is configured in the https://github.com/Open-MSS/MSS/blob/develop/mslib/mscolab/conf.py#L44

I think we should keep it in days. I did not described it correctly in the issue.

Please change it to days so that it matches with the other occurencies where it is already used.

Another problem is, that I should have looked if this is because of refactoring in develop.

ReimarBauer:
@workaryangupta please look on this in develop. My failure to think we can fix this in stable.

workaryangupta:
> @workaryangupta please look on this in develop. My failure to think we can fix this in stable.

Alright. I'll be closing this pull request in that case, will link a new one for the develop branch.

PR #2237, opened 2024-02-23
Remove all occurrences of processEvents in tests/

### LINKED ISSUE #2232: Eliminate all calls to `processEvents` in tests/
Probably all instances of `PyQt5.QtWidgets.QApplication.processEvents` are potential race conditions because they are used as a proxy to ensure that something has happened in Qt without actually checking.

In the test suite those should be replaced by using the `qtbot.wait_*` functions from pytest-qt. E.g. asserts that something has happened in Qt should be wrapped in `wait_until` as documented here: https://pytest-qt.readthedocs.io/en/latest/wait_until.html.

matrss:
Fixes #2232.

### DIFF
--- a/tests/test_meta.py
+++ b/tests/test_meta.py
@@ -1,0 +1,39 @@
+# -*- coding: utf-8 -*-
+"""
+
+    tests.test_meta
+    ~~~~~~~~~~~~~~~
+
+    This module provides tests that are "meta" in some way, i.e. that don't test
+    application code but test that e.g. other tests follow some convention.
+
+    This file is part of MSS.
+
+    :copyright: Copyright 2024 by the MSS team, see AUTHORS.
+    :license: APACHE-2.0, see LICENSE for details.
+
+    Licensed under the Apache License, Version 2.0 (the "License");
+    you may not use this file except in compliance with the License.
+    You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+"""
+import pathlib
+
+
+def test_processEvents_is_not_used_in_tests(request):
+    """Check that no test is calling PyQt5.QtWidgets.QApplication.processEvents() explicitly."""
+    tests_path = pathlib.Path(request.config.rootdir) / "tests"
+    for test_file in tests_path.rglob("*.py"):
+        if str(test_file) == request.fspath:
+            # Skip the current file
+            continue
+        assert (
+            "processEvents" not in test_file.read_text()
+        ), "processEvents is mentioned in {}".format(test_file.relative_to(request.config.rootdir))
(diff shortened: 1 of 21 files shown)


ReimarBauer:
@matrss that's pretty good work, I'll take a closer look tomorrow.

matrss:
I didn't actually do much, I think that this works is a by-product of the other changes to the test suite. All I did in this PR was to use `sed` to remove all calls to `processEvents` under tests/ and it surprisingly worked x)

ReimarBauer:
When this should be kept, please add a comment
https://github.com/matrss/MSS/blob/remove-processEvents-in-tests/tests/_test_msui/test_mscolab.py#L116

ReimarBauer:
We maybe can also add a test for that. That will keep it off. 

```
FAILED tests/_test_utils/test_our_style.py::test_not_using_processevents_in_tests - AssertionError: Function processEvents() found in file test_mscolab.py
```

```
import os
import pathlib
import conftest


def test_not_using_processevents_in_tests():
    forbidden_function = 'processEvents()'
    tests_root = os.path.join(os.path.dirname(conftest.__file__), 'tests')
    whitelist = ['test_forbidden.py']
    for subdir, dirs, files in os.walk(tests_root):
        for file in files:
            if file.endswith('.py') and not pathlib.Path(file).name in whitelist:
                with open(os.path.join(subdir, file), 'r') as f:
                    contents = f.read()
                    assert forbidden_function not in contents, \
                        f"Function {forbidden_function} found in file {file}"

```

matrss:
> When this should be kept, please add a comment https://github.com/matrss/MSS/blob/remove-processEvents-in-tests/tests/_test_msui/test_mscolab.py#L116

I must have missed that one, not sure why. I'll update this later.

> We maybe can also add a test for that. That will keep it off.
> 
> ```
> FAILED tests/_test_utils/test_our_style.py::test_not_using_processevents_in_tests - AssertionError: Function processEvents() found in file test_mscolab.py
> ```
> 
> ```
> import os
> import pathlib
> import conftest
> 
> 
> def test_not_using_processevents_in_tests():
>     forbidden_function = 'processEvents()'
>     tests_root = os.path.join(os.path.dirname(conftest.__file__), 'tests')
>     whitelist = ['test_forbidden.py']
>     for subdir, dirs, files in os.walk(tests_root):
>         for file in files:
>             if file.endswith('.py') and not pathlib.Path(file).name in whitelist:
>                 with open(os.path.join(subdir, file), 'r') as f:
>                     contents = f.read()
>                     assert forbidden_function not in contents, \
>                         f"Function {forbidden_function} found in file {file}"
> ```

Good idea.

matrss:
@ReimarBauer I've removed the last use of processEvents and added a test that checks for its usage.

tests/test_meta.py
```diff
@@ -0,0 +1,13 @@
+import pathlib
```
ReimarBauer, line 27: please add the https://github.com/Open-MSS/MSS/blob/develop/new_file_template.py

matrss: done

tests/test_meta.py
```diff
@@ -0,0 +1,39 @@
...
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+"""
+import pathlib
+
+
+def test_processEvents_is_not_used_in_tests(request):
+    """Check that no test is calling PyQt5.QtWidgets.QApplication.processEvents() explicitly."""
```
ReimarBauer, line 31: I had not thought on this request :) Much nicer.

ReimarBauer (approved):
Thx

PR #2239, opened 2024-02-23
Use of Aware datetime objects 

### LINKED ISSUE #2122: use UTC based timestamps

Preetam-Das26:
**Purpose of PR?**:

Fixes #2122 #2183 #2182 

No

**Checklist:**
- [ ] Commit has unit tests


Preetam-Das26:
I have made the changes in the following files as per the issue. Kindly help me resolve the errors I am facing with this regard.
@ReimarBauer @matrss

ReimarBauer:
Hi,

sometimes it is surprising.  test_messages needs first a waypoint added that it generate a chat message. This test needs an improvement. 
But also the default timestamp value of 1970-01-01, 00:00:00 does not match to the format 
`timestamp = datetime.datetime.strptime(timestamp, "%Y-%m-%d, %H:%M:%S %Z")` This generates currently the 500 error.

ReimarBauer:
It is a lower case "z" and the default format needs changes. This also shows where we don't have a good testcoverage. 
```
diff --git a/mslib/mscolab/chat_manager.py b/mslib/mscolab/chat_manager.py
index b12cdd9a..a5580c5a 100644
--- a/mslib/mscolab/chat_manager.py
+++ b/mslib/mscolab/chat_manager.py
@@ -64,7 +64,7 @@ class ChatManager:
         if timestamp is None:
             timestamp = datetime.datetime(1970, 1, 1, tzinfo=datetime.timezone.utc)
         else:
-            timestamp = datetime.datetime.strptime(timestamp, "%Y-%m-%d, %H:%M:%S %Z")
+            timestamp = datetime.datetime.strptime(timestamp, "%Y-%m-%d, %H:%M:%S %z")
         messages = Message.query \
             .filter(Message.op_id == op_id) \
             .filter(Message.reply_id.is_(None)) \
diff --git a/mslib/mscolab/server.py b/mslib/mscolab/server.py
index 77b6b41a..56d54c1e 100644
--- a/mslib/mscolab/server.py
+++ b/mslib/mscolab/server.py
@@ -367,7 +367,7 @@ def messages():
     user = g.user
     op_id = request.args.get("op_id", request.form.get("op_id", None))
     if fm.is_member(user.id, op_id):
-        timestamp = request.args.get("timestamp", request.form.get("timestamp", "1970-01-01, 00:00:00"))
+        timestamp = request.args.get("timestamp", request.form.get("timestamp", "1970-01-01, 00:00:00 +00:00"))
         chat_messages = cm.get_messages(op_id, timestamp)
         return jsonify({"messages": chat_messages})
     return "False"
```

ReimarBauer:
I would prefer to have first the mscolab db tz issues solved and the wms and msui changes in a different PR.  Because currently mscolab is broken by the other PR. And when we need to do a rollback of this PR because of the other parts it gets broken again.

For wms and its UI there is likly more needed than just adopting the timezone information. And for a refactoring there the testcoverage should be checked / improved. 




```
ERROR    root:wms_control.py:858 Can't understand time string '2014-10-17T12:00:00Z/current/P1Y'. Please check the implementation.
ERROR    root:multilayers.py:644 Cannot determine valid time format of  for Temperature (degC) and Geopotential Height (m) | ecmwf_EUR_LL
```

ReimarBauer:
This looks better, but it is likly a [design problem](https://github.com/Open-MSS/MSS/blob/develop/mslib/mscolab/file_manager.py#L394)  which migh get visible. 
On my system all entries have same time. 


```
git diff tests/_test_mscolab/test_files_api.py
diff --git a/tests/_test_mscolab/test_files_api.py b/tests/_test_mscolab/test_files_api.py
index 1523eb18..16a07033 100644
--- a/tests/_test_mscolab/test_files_api.py
+++ b/tests/_test_mscolab/test_files_api.py
@@ -182,10 +182,12 @@ class Test_Files:
             assert self.fm.save_file(operation.id, "content2", self.user)
             assert self.fm.save_file(operation.id, "content3", self.user)
             all_changes = self.fm.get_all_changes(operation.id, self.user)
-            previous_change = self.fm.get_change_content(all_changes[2]["id"], self.user)
+            previous_change = self.fm.get_change_content(all_changes[0]["id"], self.user)
             assert previous_change == "content1"
             previous_change = self.fm.get_change_content(all_changes[1]["id"], self.user)
             assert previous_change == "content2"
+            previous_change = self.fm.get_change_content(all_changes[2]["id"], self.user)
+            assert previous_change == "content3"
 
     def test_set_version_name(self):
         with self.app.test_client():

```

mslib/mscolab/file_manager.py
```diff
@@ -399,7 +399,7 @@ def get_all_changes(self, op_id, user, named_version=False):
             'comment': change.comment,
             'version_name': change.version_name,
             'username': change.user.username,
-            'created_at': change.created_at.strftime("%Y-%m-%d, %H:%M:%S")
+            'created_at': change.created_at.strftime("%Y-%m-%d, %H:%M:%S %Z")
```
ReimarBauer: lowercase z

mslib/mscolab/utils.py
```diff
@@ -55,7 +55,7 @@ def get_message_dict(message):
         "message_type": message.message_type,
         "reply_id": message.reply_id,
         "replies": [],
-        "time": message.created_at.strftime("%Y-%m-%d, %H:%M:%S")
+        "time": message.created_at.strftime("%Y-%m-%d, %H:%M:%S %Z")
     }
```
ReimarBauer, line 59: lowercase z

mslib/msui/mscolab_chat.py
```diff
@@ -349,7 +349,7 @@ def load_all_messages(self):
         data = {
             "token": self.token,
             "op_id": self.op_id,
-            "timestamp": datetime.datetime(1970, 1, 1).strftime("%Y-%m-%d, %H:%M:%S")
+            "timestamp": datetime.datetime(1970, 1, 1, tzinfo=datetime.timezone.utc).strftime("%Y-%m-%d, %H:%M:%S %Z")
```
ReimarBauer: lowercase z

ReimarBauer: and so on

Preetam-Das26: ok all corresponding changes will be made.

mslib/msui/wms_control.py
```diff
@@ -27,7 +27,7 @@
 """
```
ReimarBauer: The query string has a fixed format given by the WMS standard. That is not that easy to change as it looks like.
I think it is better to look on this in a seperate Pull Request or to keep it as it is. A change in the format will make us likly incompatible. At least it makes older servers unusable for us.

Preetam-Das26:
Should I now create a new PR for the msui and wms changes?

ReimarBauer:
> Should I now create a new PR for the msui and wms changes?

remove the wms_control.py change for now.  Add a comment # ToDo before refactoring to an aware datetime object add a test to verify the WMS part. 
No other PR is needed. Without the new test(s) I am not sure  about the consequences and rather won't merge that change.

Preetam-Das26:
Changes done

ReimarBauer:
Just visualizing the problem of the two tests. Tests are now too fast.  Or created_at gets not changed.

indicating_the_problem

Looking how I can make "created_at" again different. Update: stable has it also not different.

But we have also to look on the test itselfs if they are correct.

ReimarBauer:
This suggestion

https://github.com/Open-MSS/MSS/pull/2239#issuecomment-1962310401

was wrong, because the newest entry should have  index 0 because of the "biggest" created_at time value. It is ordered by created_at time. For some reason it changed with your changes the order and we were able to see the problem.

mslib/msui/wms_control.py
```diff
@@ -474,6 +474,7 @@ def __init__(self, parent=None, default_WMS=None, wms_cache=None, view=None):
             self.wms_cache = None
 
         # Initialise date/time fields with current day, 00 UTC.
+        # To do Before refactoring to an aware datetime object add a test to verify the WMS part.
```
ReimarBauer: Without a blank  ToDo, my IDE can list those. 

by PyCharm e.g. 
without_a_blank

We have a repeating issue to look at ToDo's. - check still ToDo's - filing issues etc.

mslib/msui/wms_control.py
```diff
@@ -474,6 +474,7 @@ def __init__(self, parent=None, default_WMS=None, wms_cache=None, view=None):
...
         self.dteInitTime.setDateTime(QtCore.QDateTime(
             datetime.utcnow().replace(hour=0, minute=0, second=0, microsecond=0)))
```
ReimarBauer, line 479: this might also become datetime.now() with UTC 


nice article by Miguel
https://blog.miguelgrinberg.com/post/it-s-time-for-a-change-datetime-utcnow-is-now-deprecated

mslib/msui/wms_control.py
Preetam-Das26: Ok.

mslib/msui/wms_control.py
Preetam-Das26: @ReimarBauer will using aware_utcnow() function with the same datetime.now(timezone.utc) make any difference ?
Because we were getting the error due to the same issue.
I think this might be case of type mismatch for the aware datetime object.

ReimarBauer: I've overseen that this again is wms_control. We don't touch it.

ReimarBauer: AwareDateTime is fine. The wms_control is not related to mscolab.  msui uses different servers.  This is for exchange data with the wms server. 

When it is not touched we can merge this first.

Preetam-Das26: ok

ReimarBauer:
have a look if this fixes the tests
```
index 0aeb8947..d85deeff 100644
--- a/tests/_test_mscolab/test_files_api.py
+++ b/tests/_test_mscolab/test_files_api.py
@@ -25,6 +25,7 @@
 """
 from flask_testing import TestCase
+import time
 import os
 import fs
 import pytest
@@ -195,9 +196,15 @@ class Test_Files(TestCase):
         with self.app.test_client():
             flight_path, operation = self._create_operation(flight_path="V11")
             assert self.fm.save_file(operation.id, "content1", self.user)
+            # we need to wait to get an updated created_at
+            time.sleep(1)
             assert self.fm.save_file(operation.id, "content2", self.user)
             all_changes = self.fm.get_all_changes(operation.id, self.user)
+            # the newest change is on index 0, because it has a recent created_at time
             assert len(all_changes) == 2
+            assert all_changes[0]["id"] == 2
+            assert all_changes[0]["id"] > all_changes[1]["id"]
+            assert all_changes[0]["created_at"] > all_changes[1]["created_at"]
 
     def test_get_change_content(self):
         with self.app.test_client():
diff --git a/tests/_test_mscolab/test_server.py b/tests/_test_mscolab/test_server.py
index 63ddf484..411536d9 100644
--- a/tests/_test_mscolab/test_server.py
+++ b/tests/_test_mscolab/test_server.py
@@ -24,7 +24,7 @@
 """
-
+import time
 from flask_testing import TestCase
 import os
 import pytest
@@ -244,13 +244,19 @@ class Test_Server(TestCase):
             operation, token = self._create_operation(test_client, self.userdata)
             fm, user = self._save_content(operation, self.userdata)
+            # we need to wait to get an updated created_at
+            time.sleep(1)
             fm.save_file(operation.id, "content2", user)
             all_changes = fm.get_all_changes(operation.id, user)
+            # the newest change is on index 0, because it has a recent created_at time
             response = test_client.get('/get_change_content', data={"token": token,
                                                                     "ch_id": all_changes[1]["id"]})
             data = json.loads(response.data.decode('utf-8'))
             assert data == {'content': 'content1'}
+            assert all_changes[0]["id"] == 2
+            assert all_changes[0]["id"] > all_changes[1]["id"]
+            assert all_changes[0]["created_at"] > all_changes[1]["created_at"]
 
     def test_set_version_name(self):
         assert add_user(self.userdata[0], self.userdata[1], self.userdata[2])
```

Preetam-Das26:
Done @ReimarBauer

ReimarBauer:
ah, that happens when I copy a diff from the stable branch with the old test syntax

ReimarBauer:
@Preetam-Das26 please add only the 

`+ changes `

to the existing test, not the whole test.

I tried to comment what needs removed

tests/_test_mscolab/test_files_api.py
```diff
@@ -24,14 +24,16 @@
     See the License for the specific language governing permissions and
     limitations under the License.
 """
+from flask_testing import TestCase
```
ReimarBauer: that was not a `+` change in the diff. 

the lines of the whole diff can't be used, because I created that first for stable and here it was already refactored

tests/_test_mscolab/test_files_api.py
```diff
@@ -24,14 +24,16 @@
...
+import time
 import fs
 import pytest
 
 from mslib.mscolab.models import Operation
 from mslib.mscolab.seed import add_user, get_user
 
 
-class Test_Files:
+class Test_Files(TestCase):
```
ReimarBauer: keep the old line

tests/_test_mscolab/test_server.py
```diff
@@ -24,6 +24,8 @@
     See the License for the specific language governing permissions and
     limitations under the License.
 """
+import time
+from flask_testing import TestCase
```
ReimarBauer: remove `from flask_testing import TestCase`

tests/_test_mscolab/test_server.py
```diff
@@ -35,7 +37,7 @@
 from mslib.mscolab.seed import add_user, get_user
 
 
-class Test_Server:
+class Test_Server(TestCase):
```
ReimarBauer: keep the old line

Preetam-Das26:
Alright @ReimarBauer I thought those lines were also necessary.I will make the changes.

ReimarBauer:
let us try if this solves the last failing test
```
--- a/tests/_test_mscolab/test_file_manager.py
+++ b/tests/_test_mscolab/test_file_manager.py
@@ -25,6 +25,8 @@
 """
 import datetime
+import time
 import pytest
 
 from mslib.mscolab.models import Operation, User
@@ -265,8 +267,12 @@ class Test_FileManager:
             flight_path, operation = self._create_operation(flight_path="operation8")
             assert self.fm.get_all_changes(operation.id, self.user) == []
             assert self.fm.save_file(operation.id, self.content1, self.user)
+            time.sleep(1)
             assert self.fm.save_file(operation.id, self.content2, self.user)
             all_changes = self.fm.get_all_changes(operation.id, self.user)
+            assert all_changes[0]["created_at"] > all_changes[1]["created_at"]
+            assert all_changes[0]["id"] == 2
+            assert all_changes[0]["id"] > all_changes[1]["id"]
             assert self.fm.get_change_content(all_changes[1]["id"], self.user) == self.content1
 
     def test_set_version_name(self):

```

mslib/mscolab/models.py
```diff
@@ -191,6 +191,7 @@ def __init__(self, op_id, u_id, text, message_type=MessageType.TEXT, reply_id=No
         self.text = text
         self.message_type = message_type
         self.reply_id = reply_id
+        self.created_at = datetime.datetime.now(tz=datetime.timezone.utc)
```
ReimarBauer: we need to check if this is needed, or why it makes a difference

develop

ReimarBauer:
In develop Matthias has the speed of the tests much improved. And currently created_at has only a time reolution of 1 Second.

always when it fails 
self.fm.save_file(operation.id, "content1", self.user) 
because of a next
assert self.fm.save_file(operation.id, "content1", self.user)

add a time.sleep(1) in between

 
e.g. for the tests below.

https://github.com/Open-MSS/MSS/actions/runs/8053264211/job/21995247117?pr=2239#step:9:1115
 ```
def test_get_change_content(self):
        with self.app.test_client():
            flight_path, operation = self._create_operation(flight_path="V12", content='initial')
            assert self.fm.save_file(operation.id, "content1", self.user)
            assert self.fm.save_file(operation.id, "content2", self.user)
            assert self.fm.save_file(operation.id, "content3", self.user)
            all_changes = self.fm.get_all_changes(operation.id, self.user)
            previous_change = self.fm.get_change_content(all_changes[0]["id"], self.user)
>           assert previous_change == "content1"
E           AssertionError: assert 'content3' == 'content1'
E             
E             - content1
E             ?        ^
E             + content3
E             ?        ^
```

File: mslib/mscolab/models.py [hunk 12]
```diff
@@ -181,7 +181,7 @@ class Message(db.Model):
     text = db.Column(db.Text)
     message_type = db.Column(db.Enum(MessageType), default=MessageType.TEXT)
     reply_id = db.Column(db.Integer, db.ForeignKey('messages.id'))
-    created_at = db.Column(db.DateTime, default=datetime.datetime.utcnow)
+    created_at = db.Column(AwareDateTime, default=datetime.datetime.now(tz=datetime.timezone.utc))
```
ReimarBauer: try if 
db.Column(AwareDateTime, default=lambda: datetime.datetime.now(tz=datetime.timezone.utc))

makes it not necessary to have the `self.created_at =` line.

also for the other.

tests/_test_mscolab/test_files_api.py
```diff
@@ -171,21 +172,31 @@ def test_get_all_changes(self):
...
         with self.app.test_client():
             flight_path, operation = self._create_operation(flight_path="V12", content='initial')
             assert self.fm.save_file(operation.id, "content1", self.user)
+            time.sleep(1)
             assert self.fm.save_file(operation.id, "content2", self.user)
+            time.sleep(1)
             assert self.fm.save_file(operation.id, "content3", self.user)
             all_changes = self.fm.get_all_changes(operation.id, self.user)
-            previous_change = self.fm.get_change_content(all_changes[2]["id"], self.user)
+            previous_change = self.fm.get_change_content(all_changes[0]["id"], self.user)
```
ReimarBauer: undo these assert changed, I have introduced as I hadn't understand the problem.

when time setting works always the most recent change has index 0.

tests/_test_mscolab/test_files_api.py
```diff
@@ -171,21 +172,31 @@ def test_get_all_changes(self):
...
             assert previous_change == "content1"
             previous_change = self.fm.get_change_content(all_changes[1]["id"], self.user)
             assert previous_change == "content2"
+            previous_change = self.fm.get_change_content(all_changes[2]["id"], self.user)
```
ReimarBauer: undo this two lines too

Preetam-Das26: Do I have to change the complete method to as it was before?

ReimarBauer: the time.sleep(1) are needed, the other changes can be undone

Preetam-Das26:
Do we need the time.sleep(1) change? Because it seems that it works fine even without it.

ReimarBauer:
> Do we need the time.sleep(1) change? Because it seems that it works fine even without it.

It works only without if we compare fractions of seconds where those are ordered. 

all_changes = self.fm.get_all_changes(operation.id, self.user)

check this. There is a follow up issue. The web interface gots already slow.

Preetam-Das26:
Ok.
Sure.

tests/_test_mscolab/test_files_api.py
```diff
@@ -24,14 +24,15 @@
...
+import time
 import fs
 import pytest
 
 from mslib.mscolab.models import Operation
 from mslib.mscolab.seed import add_user, get_user
 
 
-class Test_Files:
+class Test_Files():
```
matrss: The parenthesis are unnecessary.

tests/_test_mscolab/test_server.py
```diff
@@ -35,7 +36,7 @@
 from mslib.mscolab.seed import add_user, get_user
 
 
-class Test_Server:
+class Test_Server():
```
matrss: The parenthesis are unnecessary.

tests/_test_mscolab/test_server.py
```diff
@@ -245,13 +246,19 @@ def test_get_change_content(self):
...
             all_changes = fm.get_all_changes(operation.id, user)
+            # the newest change is on index 0, because it has a recent created_at time
             response = test_client.get('/get_change_content', data={"token": token,
                                                                     "ch_id": all_changes[1]["id"]})
             assert response.status_code == 200
             data = json.loads(response.data.decode('utf-8'))
             assert data == {'content': 'content1'}
+            assert all_changes[0]["id"] == 2
+            assert all_changes[0]["id"] > all_changes[1]["id"]
+            assert all_changes[0]["created_at"] > all_changes[1]["created_at"]
```
matrss: This test now tests two things: the get_change_content endpoint and that two consecutive changes are in the right order. The test for the correct order should be moved to test_get_all_changes, similar to how it is in test_files_api.py.

matrss (changes requested):
With this PR merged I think it will become impossible to use a future v9 version of msui with a pre-v9 version of mscolab. Is that an issue @ReimarBauer?

tests/_test_mscolab/test_files_api.py
Preetam-Das26: ok

Preetam-Das26:
Changes done @matrss

ReimarBauer:
> With this PR merged I think it will become impossible to use a future v9 version of msui with a pre-v9 version of mscolab. Is that an issue @ReimarBauer?

No issue, because we are already on that point, because of the changes of the status information.
https://github.com/Open-MSS/MSS/blob/develop/mslib/mscolab/server.py#L242

ReimarBauer:
it need to do the order by fractions of seconds or the test needs a time.sleep(1)
```
def test_get_all_changes(self):

>           assert all_changes[0]["created_at"] > all_changes[1]["created_at"]
E           AssertionError: assert '2024-02-27, 12:33:23 +0000' > '2024-02-27, 12:33:23 +0000'
```
ahhh, for the follow up
hmm, the order does not solve it, we need all_changes[0]["created_at"] have to output fractions too. But we don't want to output the fractions on any place in the UI.

something like this. 
```
s = t.strftime('%Y-%m-%d %H:%M:%S.%f')

```

Preetam-Das26:
Have I added the time.sleep(1) correctly? @ReimarBauer

tests/_test_mscolab/test_server.py
```diff
@@ -234,19 +235,27 @@ def test_get_all_changes(self):
             operation, token = self._create_operation(test_client, self.userdata)
             fm, user = self._save_content(operation, self.userdata)
```
matrss, line 236: The sleep should be between this line and the following, instead of further down.

tests/_test_mscolab/test_server.py
```diff
@@ -234,19 +235,27 @@ def test_get_all_changes(self):
...
     def test_get_change_content(self):
         assert add_user(self.userdata[0], self.userdata[1], self.userdata[2])
         with self.app.test_client() as test_client:
             operation, token = self._create_operation(test_client, self.userdata)
             fm, user = self._save_content(operation, self.userdata)
+            # we need to wait to get an updated created_at
+            time.sleep(1)
             fm.save_file(operation.id, "content2", user)
             all_changes = fm.get_all_changes(operation.id, user)
+            # the newest change is on index 0, because it has a recent created_at time
```
matrss: This comment is in the wrong place now, it belongs to the new asserts in the other test.

tests/_test_mscolab/test_server.py
```diff
@@ -234,19 +235,27 @@ def test_get_all_changes(self):
             operation, token = self._create_operation(test_client, self.userdata)
             fm, user = self._save_content(operation, self.userdata)
             fm.save_file(operation.id, "content2", user)
+            all_changes = fm.get_all_changes(operation.id, user)
```
matrss, line 239: This shouldn't be necessary to get all changes, the line below this one uses the API endpoint to get them so they should be extractable from `response`.

mslib/mscolab/file_manager.py
```diff
@@ -399,7 +399,7 @@ def get_all_changes(self, op_id, user, named_version=False):
             'comment': change.comment,
             'version_name': change.version_name,
             'username': change.user.username,
-            'created_at': change.created_at.strftime("%Y-%m-%d, %H:%M:%S")
+            'created_at': change.created_at.strftime("%Y-%m-%d, %H:%M:%S %z")
```
ReimarBauer, line 402: Not sure if this is the only what we need to do to get fractions used
"%Y-%m-%d, %H:%M:%S.%f %z"

tests/_test_mscolab/test_server.py
Preetam-Das26: Removing this shows error as it is not defined.

mslib/mscolab/file_manager.py
Preetam-Das26: Should this be changed at all places?

ReimarBauer:
@Preetam-Das26 I decided long time ago to use pycharm because it helps me a lot by its design. It is over 10 years community driven developed. It has a lot features like this one. I am not familiar with other IDEs but when something like this is missing, I can show you that pycharm has it.

pycharm_helper

mslib/mscolab/file_manager.py
ReimarBauer: not yet. I suggest. Get  the parts which you have now passing the tests. 
Then we can merge this. Afterward work on getting the time.sleep(1) away.

Preetam-Das26:
So should I keep this line as it is to resolve the error?
 all_changes = fm.get_all_changes(operation.id, user) in test_server.py

ReimarBauer:
> So should I keep this line as it is to resolve the error? all_changes = fm.get_all_changes(operation.id, user) in test_server.py

yes, otherwise you have an unresolved reference and this 5 times.

tests/_test_mscolab/test_server.py
matrss: Yes, that is expected since the `all_changes` variable is no longer defined then. But the same data is already contained in `response` and should be taken from there in this test. The asserts need to be adapted for that.

matrss:
> > So should I keep this line as it is to resolve the error? all_changes = fm.get_all_changes(operation.id, user) in test_server.py
> 
> yes, otherwise you have an unresolved reference and this 5 times.

No, it is the wrong way to get this data in this test. It is already contained in `response`.

ReimarBauer:
> > > So should I keep this line as it is to resolve the error? all_changes = fm.get_all_changes(operation.id, user) in test_server.py
> > 
> > 
> > yes, otherwise you have an unresolved reference and this 5 times.
> 
> No, it is the wrong way to get this data in this test. It is already contained in `response`.

ah ok, but the unresolved reference should be solved

ReimarBauer:
please undo the fraction change, that is not for now. Look if you added the comment and the time.sleep to the correct line.

ReimarBauer:
To get further, it is always first make the tests satisfying then maybe improving, and merge. Then the next issue.  Not all issues we have in one PR.

tests/_test_mscolab/test_server.py
ReimarBauer: @Preetam-Das26 have a look here. You want to wait after you did saved something so that the next save has a delay of 1 second.

Preetam-Das26:
Changes done @ReimarBauer @matrss

tests/_test_mscolab/test_server.py
matrss: I've talked to @ReimarBauer and we came to the conclusion that this is not a blocker for this PR. Other tests in test_server.py could be improved in this regard as well and it can be addressed in the future.

matrss (approved):
Looks good to me now. Thanks for the PR! :)

PR #2242, opened 2024-02-25
update created_at for Changes and Messages

### LINKED ISSUE #2241: mscolab does not update created_at for Changes and Messages
We define in the models for Message and Change a "created_at" and keep the initial creation time. 

This needs to become updated on each new entry.

We see this problem in the chat by the service messages.

service_message

but we can't see it in the current tests (or we see it randomly), because for some reason it seems to order correctly. 

get_all_changes did a query

```
changes = Change.query.\
                filter_by(op_id=op_id)\
                .all()
```

A database entry for changes looks like. 

indicating_the_problem

ReimarBauer:
and improved tests.

**Purpose of PR?**:

Fixes #2241 

We recognized this issue because of https://github.com/Open-MSS/MSS/pull/2239

### DIFF
--- a/mslib/mscolab/models.py
+++ b/mslib/mscolab/models.py
@@ -183,6 +183,7 @@ class Message(db.Model):
         self.text = text
         self.message_type = message_type
         self.reply_id = reply_id
+        self.created_at = datetime.datetime.utcnow()
 
     def __repr__(self):
         return f'<Message text: {self.text}, u_id: {self.u_id}, op_id: {self.op_id}>, message_type: {self.message_type}'
@@ -206,3 +207,4 @@ class Change(db.Model):
         self.commit_hash = commit_hash
         self.version_name = version_name
         self.comment = comment
+        self.created_at = datetime.datetime.utcnow()
--- a/tests/_test_mscolab/test_files_api.py
+++ b/tests/_test_mscolab/test_files_api.py
@@ -25,6 +25,7 @@
     limitations under the License.
 """
 from flask_testing import TestCase
+import time
 import os
 import fs
 import pytest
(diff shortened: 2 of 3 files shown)


mslib/mscolab/models.py
```diff
@@ -183,6 +183,7 @@ def __init__(self, op_id, u_id, text, message_type=MessageType.TEXT, reply_id=No
         self.text = text
         self.message_type = message_type
         self.reply_id = reply_id
+        self.created_at = datetime.datetime.utcnow()
```
matrss, line 186: Is this intentionally utcnow?  `datetime.datetime.now(tz=datetime.timezone.utc)` should behave the same in this case, I think.

matrss: Also, since the created_at column has a default of `datetime.datetime.utcnow` I think this entire line is redundant.

mslib/mscolab/models.py
```diff
@@ -206,3 +207,4 @@ def __init__(self, op_id, u_id, commit_hash, version_name=None, comment=None):
         self.commit_hash = commit_hash
         self.version_name = version_name
         self.comment = comment
+        self.created_at = datetime.datetime.utcnow()
```
matrss, line 210: Same as with the other created_at line.

mslib/mscolab/models.py
ReimarBauer: > Is this intentionally utcnow? `datetime.datetime.now(tz=datetime.timezone.utc)` should behave the same in this case, I think.

it is a bug in MSS-8.3.3 which has the old writings on all places. And I don't want to mix that now up. Because it gets refatored in develop.

ReimarBauer: > Also, since the created_at column has a default of `datetime.datetime.utcnow` I think this entire line is redundant.

I thougt this too, but it only gets instanciated a new value, when I added the line. See the updated tests.

ReimarBauer:
hmm, can't reproduce today.

mslib/mscolab/models.py
matrss: > I thougt this too, but it only gets instanciated a new value, when I added the line.

That would be a bug in sqlalchemy then. But I cannot reproduce that behavior.

> See the updated tests.

The tests pass without the changes to mslib/mscolab/models.py.

ReimarBauer:
lets move the tests to the redactoring https://github.com/Open-MSS/MSS/pull/2239 
Then we may get that merged.

PR #2244, opened 2024-02-25
created_at bug (while creation in the same second)

### LINKED ISSUE #2241: mscolab does not update created_at for Changes and Messages

### LINKED ISSUE #2245: changes query needs compare .created_at with the stored milliseconds
After https://github.com/Open-MSS/MSS/issues/2241 is merged the query needs to become improved. This PR has the needed tests to check for a solution of the problem.

This compares currently only seconds. We need also the milliseconds which are in the database
```
changes = Change.query.\
                filter_by(op_id=op_id)\
                .all()
```
In the changed tests of the mentioned PR we need to remove the time.sleep(1) so that the change can be tested.

I am currently not sure if this can be done without an API change. When it needs an API change we can do this only in develop.
Because it is a bug we try for stable first.

mahmoudnabil133:
**Purpose of PR?**:
i see that bug  happened when we get changes
`           changes = Change.query\
                .filter_by(op_id=op_id)\
                .order_by(Change.created_at.desc()))\
                .all()
`
when we get 2 changes with the same timestamp ordered by creation time
and in Changes class:
`created_at = db.Column(db.DateTime, default=datetime.datetime.utcnow`
the created at parameter is defoult with `datetime.utcnow()` function in Python provides the current UTC time, but it doesn't include microseconds by default. Therefore, if two changes are recorded within the same second, they would have the same timestamp down to the second.
**My changes**
`           changes = Change.query\
                .filter_by(op_id=op_id)\
                .order_by(Change.created_at.desc(), Change.id.desc())\
                .all()
`
when ordering by created_at parameter there is a second level of filtering to filter 2 changes with there id
also,
`created_at = db.Column(db.DateTime, default=datetime.datetime.utcnow.replace(microsecond=0))`
With this modification, the created_at column will include the current timestamp with milliseconds or microseconds precision, ensuring that each change has a unique timestamp even if they are recorded within the same second. This would help in distinguishing between changes that occur very close to each other in time.

Fixes #2241

### DIFF
--- a/mslib/mscolab/models.py
+++ b/mslib/mscolab/models.py
@@ -205,7 +205,7 @@ class Change(db.Model):
     commit_hash = db.Column(db.String(255), default=None)
     version_name = db.Column(db.String(255), default=None)
     comment = db.Column(db.String(255), default=None)
-    created_at = db.Column(db.DateTime, default=datetime.datetime.utcnow)
+    created_at = db.Column(db.DateTime, default=datetime.datetime.utcnow.replace(microsecond=0))
     user = db.relationship('User')
 
     def __init__(self, op_id, u_id, commit_hash, version_name=None, comment=None):
--- a/mslib/mscolab/models.py
+++ b/mslib/mscolab/models.py
@@ -205,7 +205,7 @@ class Change(db.Model):
     commit_hash = db.Column(db.String(255), default=None)
     version_name = db.Column(db.String(255), default=None)
     comment = db.Column(db.String(255), default=None)
-    created_at = db.Column(db.DateTime, default=datetime.datetime.utcnow.replace(microsecond=0))
+    created_at = db.Column(db.DateTime, default=datetime.datetime.utcnow().replace(microsecond=0))
     user = db.relationship('User')
 
     def __init__(self, op_id, u_id, commit_hash, version_name=None, comment=None):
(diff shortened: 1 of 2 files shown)


ReimarBauer (changes requested):
on hold until https://github.com/Open-MSS/MSS/pull/2239
is merged. This will include tests for this problem.


The follow up is to compare with milliseconds

mslib/mscolab/models.py
```diff
@@ -205,7 +205,7 @@ class Change(db.Model):
     commit_hash = db.Column(db.String(255), default=None)
     version_name = db.Column(db.String(255), default=None)
     comment = db.Column(db.String(255), default=None)
-    created_at = db.Column(db.DateTime, default=datetime.datetime.utcnow.replace(microsecond=0))
+    created_at = db.Column(db.DateTime, default=datetime.datetime.utcnow().replace(microsecond=0))
```
ReimarBauer, line 208: this sounds not correct.  We need fractions of seconds which are stored in the database.

We have to use them

ReimarBauer (changes requested):

ReimarBauer:
fixed by https://github.com/Open-MSS/MSS/pull/2308

PR #2246, opened 2024-02-26
Remove all occurrences of qWait in tests/

### LINKED ISSUE #2236: Eliminate all calls to `qWait` in tests/
Probably all instances of `PyQt5.QtTest.QTest.qWait` are potential race conditions because they are used as a proxy to ensure that something has happened in Qt without actually checking.

In the test suite those should be replaced by using the `qtbot.wait_*` functions from pytest-qt. E.g. asserts that something has happened in Qt should be wrapped in `wait_until` as documented here: https://pytest-qt.readthedocs.io/en/latest/wait_until.html.

matrss:
Fixes #2236.

### DIFF
--- a/tests/test_meta.py
+++ b/tests/test_meta.py
@@ -37,3 +37,19 @@ def test_processEvents_is_not_used_in_tests(request):
         assert (
             "processEvents" not in test_file.read_text()
         ), "processEvents is mentioned in {}".format(test_file.relative_to(request.config.rootdir))
+
+
+def test_qWait_is_not_used_in_tests(request):
+    """Check that no test is calling PyQt5.QtTest.QTest.qWait explicitly."""
+    tests_path = pathlib.Path(request.config.rootdir) / "tests"
+    for test_file in tests_path.rglob("*.py"):
+        if str(test_file) == request.fspath:
+            # Skip the current file
+            continue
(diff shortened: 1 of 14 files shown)


tests/test_meta.py
```diff
@@ -37,3 +37,19 @@ def test_processEvents_is_not_used_in_tests(request):
         assert (
             "processEvents" not in test_file.read_text()
         ), "processEvents is mentioned in {}".format(test_file.relative_to(request.config.rootdir))
+
```
ReimarBauer, line 40: much likes :)

ReimarBauer (commented):
I grabbed for self._connect_to_mscolab() 

There are a few not using qtbot,
test_mscolab_version_history 52, test_mscolab_admin_window 65, test_mscolab_merge_waypoints 67, test_mscolab_operations 61

I guess they have always the same problem, should they also become changed?

matrss:
So far I've taken the pragmatic approach of "only fix the tests that actually exhibit issues in practice". There are probably many more places that could fail under the right conditions. That said, I'll change those 4 methods as well, since that's pretty simple.

Maybe in a future refactor those methods could be unified into one, since they all do effectively the same.

ReimarBauer (approved):
Thx

Issue #2145, opened 2024-01-10
Consider replacing flake8 with ruff


matrss:
Ruff is "An extremely fast Python linter and code formatter, written in Rust." (https://github.com/astral-sh/ruff)

It effectively does what flake8 does, with some of its plugins (flake8-builtins, flake8-bugbear, and many more) added and it claims to be (and probably is) faster. Optionally it can also do import sorting like isort and code formatting like black (but we don't have to use those).

This came up in #2144.

ReimarBauer:
@joernu76 do you remember the issue we had with black?

joernu76:
black was rather aggressive and, let's say controversial, with its formatting.

the setup.py/flake8 integration isn't working anymore, but flake8 itself is a nice *checker*.

Reformatting source code cannot be done in the pipeline, but would have to be done on the users systems before commit via hooks.

We can certainly add ruff as a checker in addition to the current flow and compare the results.

matrss:
I've found this older discussions related to black, which I think Reimar initially referenced in #2144 regarding black breaking MSS: https://github.com/Open-MSS/MSS/pull/618#discussion_r553611510.

> black was rather aggressive and, let's say controversial, with its formatting.

I've found the black formatting style to be very pleasant to read, compared to more inconsistent formatting like we have here. Yes, it is very aggressive and very much opinionated, but I think that is its greatest strength (i.e. no discussion about how to format a line of code, apply black and be done with it).

I was also able to format the entirety of MSS with ruff's formatter and also with its import sorter and have the test suite pass without issue. So maybe the concerns that this breaks code no longer apply.

> Reformatting source code cannot be done in the pipeline, but would have to be done on the users systems before commit via hooks.

Not necessarily via hooks, but yes, CI should just check that the formatting is OK and not apply changes itself (just like what the flake8 action does).

In general I like to apply a linter like flake8 or ruff mostly in their standard configuration, because that removes a lot of potential for bikeshedding about what code should look like. That means I would prefer the CI action to apply linting to the entirety of this repository (not just mslib/ and tests/ like right now; but the excludes for autogenerated stuff are reasonable) and I'd like to not have the `ignore = E124,E125,E402,W504` we have right now for flake8, because I think all of those linting rules are valid and the code would look better if they were applied (of course there might be some locations where a linting rule does not apply, e.g. E402 with some of the imports in conftest.py because they cannot be at the top, but those should be ignored for those specific locations and not everywhere, in my opinion).

[...]

PR #2247, opened 2024-02-27
Replace: occurrences of tests.utils.wait_until_signal with qtbot.wait_signal

### LINKED ISSUE #2156: Replace occurrences of `tests.utils.wait_until_signal` with `qtbot.wait_signal`
The `wait_until_signal` function has the limitation that it starts listening for a signal only *after* the code that should emit that signal was called. This means there is a time window in between these calls in which the signal could be emitted but is not seen by the wait. I.e. a race condition.

The `wait_signal` context manager from pytest-qt does not have this issue because it wraps the code that is expected to emit the given signal in a `with` block and therefore starts listening *before* the signal has a chance to be emitted.

rohit2p:
**Purpose of PR?**: This issue concerns a potential race condition when using the `wait_until_signal` function for Qt testing.  The function has a limitation where it only starts listening for a specific signal after the code that should emit that signal has been called. so, we'll Replace occurrences of `tests.utils.wait_until_signal` with `qtbot.wait_signal`. Because `wait_signal` is a context manager from pytest-qt that does not have this issue.

Fixes #2156

### DIFF
--- a/tests/_test_msui/test_wms_control.py
+++ b/tests/_test_msui/test_wms_control.py
@@ -88,13 +88,12 @@ class WMSControlWidgetSetup:
         self.window.hide()
         shutil.rmtree(self.tempdir)
 
-    def query_server(self, url):
+    def query_server(self, url, qtbot):
         while len(self.window.multilayers.cbWMS_URL.currentText()) > 0:
             QtTest.QTest.keyClick(self.window.multilayers.cbWMS_URL, QtCore.Qt.Key_Backspace)
         QtTest.QTest.keyClicks(self.window.multilayers.cbWMS_URL, url)
-        QtTest.QTest.mouseClick(self.window.multilayers.btGetCapabilities, QtCore.Qt.LeftButton)
-        wait_until_signal(self.window.cpdlg.canceled)
-
+        with qtbot.wait_signal(self.window.cpdlg.canceled):
+            QtTest.QTest.mouseClick(self.window.multilayers.btGetCapabilities, QtCore.Qt.LeftButton)
 
 class Test_HSecWMSControlWidget(WMSControlWidgetSetup):
     @pytest.fixture(autouse=True)
@@ -162,15 +161,14 @@ class Test_HSecWMSControlWidget(WMSControlWidgetSetup):
             pass
 
     @pytest.mark.skip("Has a race condition where the abort might not happen fast enough")
-    def test_server_abort_getmap(self):
+    def test_server_abort_getmap(self, qtbot):
         """
         assert that an aborted getmap call does not change the displayed image
         """
         self.query_server(self.url)
-        QtTest.QTest.mouseClick(self.window.btGetMap, QtCore.Qt.LeftButton)
         QtTest.QTest.keyClick(self.window.pdlg, QtCore.Qt.Key_Enter)
-        wait_until_signal(self.window.image_displayed)
-
+        with qtbot.wait_signal(self.window.image_displayed):
+            QtTest.QTest.mouseClick(self.window.btGetMap, QtCore.Qt.LeftButton)
         assert self.view.draw_image.call_count == 0
         assert self.view.draw_legend.call_count == 0
         assert self.view.draw_metadata.call_count == 0
@@ -204,8 +202,8 @@ class Test_HSecWMSControlWidget(WMSControlWidgetSetup):
         self.view.reset_mock()
 
         QtTest.QTest.mouseClick(self.window.cbCacheEnabled, QtCore.Qt.LeftButton)
-        QtTest.QTest.mouseClick(self.window.btGetMap, QtCore.Qt.LeftButton)
-        wait_until_signal(self.window.image_displayed)
+        with qtbot.wait_signal(self.window.image_displayed):
+            QtTest.QTest.mouseClick(self.window.btGetMap, QtCore.Qt.LeftButton)
 
         assert self.view.draw_image.call_count == 1
         assert self.view.draw_legend.call_count == 1
@@ -239,7 +237,7 @@ class Test_HSecWMSControlWidget(WMSControlWidgetSetup):
         assert self.view.draw_legend.call_count == 1
         assert self.view.draw_metadata.call_count == 1
 
-    def test_multilayer_handling(self):
+    def test_multilayer_handling(self, qtbot):
         """
         assert that multilayers get created, handled and drawn properly
         """
@@ -273,8 +271,8 @@ class Test_HSecWMSControlWidget(WMSControlWidgetSetup):
         assert self.window.multilayers.listLayers.itemWidget(server.child(0), 2).currentText() == "1"
 
         # Check drawing not causing errors
-        QtTest.QTest.mouseClick(self.window.btGetMap, QtCore.Qt.LeftButton)
-        wait_until_signal(self.window.image_displayed)
+        with qtbot.wait_signal(self.window.image_displayed):
+            QtTest.QTest.mouseClick(self.window.btGetMap, QtCore.Qt.LeftButton)
 
         assert self.view.draw_image.call_count == 1
         assert self.view.draw_legend.call_count == 1
@@ -321,7 +319,7 @@ class Test_HSecWMSControlWidget(WMSControlWidgetSetup):
         assert len(self.window.multilayers.listLayers.findItems(f"{self.url}/",
                                                                 QtCore.Qt.MatchFixedString)) == 0
 
-    def test_singlelayer_handling(self):
+    def test_singlelayer_handling(self, qtbot):
         """
         assert that singlelayer mode behaves as expected
         """
@@ -348,8 +346,8 @@ class Test_HSecWMSControlWidget(WMSControlWidgetSetup):
         assert self.window.lLayerName.text().endswith(server.child(1).text(0))
 
         # Check drawing not causing errors
-        QtTest.QTest.mouseClick(self.window.btGetMap, QtCore.Qt.LeftButton)
-        wait_until_signal(self.window.image_displayed)
+        with qtbot.wait_signal(self.window.image_displayed):
+            QtTest.QTest.mouseClick(self.window.btGetMap, QtCore.Qt.LeftButton)
 
         assert self.view.draw_image.call_count == 1
         assert self.view.draw_legend.call_count == 1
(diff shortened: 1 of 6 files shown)


rohit2p:
@matrss so,  I've created this PR to address and have attempted to modify the `query_server` function. However, I'm encountering some roadblocks related to the purpose and usage of certain elements:
1. Uncertainty about `self.wms_control.cpdlg.canceled: ` : I'm unsure what this part of the code represents and its purpose within the query_server function.
2. Confusion about `with qtbot.wait_signal(self.wms_control.cpdlg.canceled):` block: like what will actually go inside this block.

While I've made changes from what i understood like the primary goal of the test in this change i did is to verify that the application handles user cancellation appropriately. i don't think i made the correct changes. Ik that their are some other changes that is required but i held off thinking i should make things clear before going further. could you guide me a bit.

matrss:
> 1. Uncertainty about `self.wms_control.cpdlg.canceled: ` : I'm unsure what this part of the code represents and its purpose within the query_server function.

`self.wms_control.cpdlg.canceled` is a signal of the QProgressDialog that is defined here: https://github.com/Open-MSS/MSS/blob/b8a200d53bb3fd1948278cadfef4cc1465594626/mslib/msui/wms_control.py#L519-L521

Waiting for this signal to be emitted is used as an indicator that the retrieval that is associated with this progress dialog has finished. I.e. it is used in the query_server method to wait until all actions have been fully processed, essentially.

> 2\. Confusion about `with qtbot.wait_signal(self.wms_control.cpdlg.canceled):` block: like what will actually go inside this block.

In the case of the method that you have already modified it would be this line that needs to be wrapped with the context manager: https://github.com/Open-MSS/MSS/blob/2adcb27ba5da481c3a804996bfa449072216babc/tests/_test_msui/test_sideview.py#L152
(Maybe also the line before that one, but I don't think so, would need to see if tests pass or fail)

> While I've made changes from what i understood like the primary goal of the test in this change i did is to verify that the application handles user cancellation appropriately. i don't think i made the correct changes.

No, the query_server method in itself is not actually a test (you can see that because it doesn't start with `test_`), but a method that is used in other tests. The canceled signal is also emitted when the retrieval completed successfully (at least that is what I think, and I understand where your confusion is coming from, it is not that well named in my opinion). That means the cancel button doesn't need to be clicked there, the emission of the cancel signal is an expected effect of clicking btGetCapabilities instead.

The query_server method not being a test in itself also means that all call sites of this method (i.e. the actual tests) need to pass the qtbot fixture to it.

rohit2p:
i think the failure is because of Fixture Management can you take a look

matrss:
All of the test failures look like this:
```
FAILED tests/_test_msui/test_wms_control.py::Test_HSecWMSControlWidget::test_no_server - TypeError: WMSControlWidgetSetup.query_server() missing 1 required positional argument: 'qtbot'
```
What it means is that all tests that use query_server have to pass along the qtbot fixture that they receive. I.e. the calls to query_server have to be adapted as well, when its signature changes.

Also, there are some minor linting issues from flake8, just click on "Details" of that check to see what they are.

Other than that it is looking good already. Just a minor nitpick: I would like `qtbot` to be the first argument to `query_server`, instead of the last. Just to be consistent with other places where this kind of thing is done, e.g. here: https://github.com/Open-MSS/MSS/blob/d4e8d6b39ea8ca83c1aed405df0d49c8d19ddb4f/tests/_test_msui/test_mscolab.py#L229

rohit2p:
@matrss the CI failed again can you take a look. this time is it because the missing `url` arguments are within the context of the `test_no_server`.

ReimarBauer:
When you look on the error message 
problem
It is pretty verbose, or?

Please have a look on the definition of query_server

ReimarBauer:
Please look on the definition of the method query_server. I don't know which IDE you use. In pycharm I would do this by 

CTRL + click on the query_server word. 

you could also debug this, set a breakpoint and go into the line where it is defined.

rohit2p:
I use PyCharm and have examined the `query_server` definition. I see it does require a url argument, and I've updated my `test_no_server `function to provide a `mock_url` thinking while calling `query_server`  with a dynamic URL might causing the error.

However, the error persists. I'll try debug the code further to pinpoint the issue – any specific debugging tips from your experience would be helpfull 
 this is what i see why i CTRL+Click:

rohit2p:
i was trying to debug to find the issue but, unfortunately it seems like the "fork" start method in multiprocessing is not available in my system that why i am encountering the "SKIPPEDe" message because the test requires the "fork" start method. so i looked up i found out i need to switch to 'spawn' so i am trying to figure out how to do that. 
image

ReimarBauer:
when I do this, I see the declaration a few lines above, 

myresults

ReimarBauer:
> i was trying to debug to find the issue but, unfortunately it seems like the "fork" start method in multiprocessing is not available in my system that why i am encountering the "SKIPPEDe" message because the test requires the "fork" start method. so i looked up i found out i need to switch to 'spawn' so i am trying to figure out how to do that. image

I know, we have a ticket already to change this.

rohit2p:
> when I do this, I see the declaration a few lines above,
> 
> myresults

can you explain a bit i don't get it @ReimarBauer

rohit2p:
> > i was trying to debug to find the issue but, unfortunately it seems like the "fork" start method in multiprocessing is not available in my system that why i am encountering the "SKIPPEDe" message because the test requires the "fork" start method. so i looked up i found out i need to switch to 'spawn' so i am trying to figure out how to do that. image
> 
> I know, we have a ticket already to change this.

does this means i can't test my code for now on my system

matrss:
> > > i was trying to debug to find the issue but, unfortunately it seems like the "fork" start method in multiprocessing is not available in my system that why i am encountering the "SKIPPEDe" message because the test requires the "fork" start method. so i looked up i found out i need to switch to 'spawn' so i am trying to figure out how to do that. image
> > 
> > 
> > I know, we have a ticket already to change this.
> 
> does this means i can't test my code for now on my system

Yes, unfortunately the tests that require a running mscolab or mswms server can not be ran on windows at the moment. But the test suite runs in GitHub Actions for every PR, so you can see the test output there.

> > when I do this, I see the declaration a few lines above,
> > myresults
> 
> can you explain a bit i don't get it @ReimarBauer

You have changed the signature of the query_server methods to include an additional parameter. Obviously all code that calls this method now needs to be adapted accordingly and pass in the correct arguments in the correct order.

rohit2p:
@matrss @ReimarBauer can you take a look does everything look good ? and i don't see `wait_until_signal` function in utils.py is being used anywhere so can i remove it now. and what about this part in MSS\tests\test_meta.py

 ```
  if str(test_file.relative_to(request.config.rootdir)) == "tests/utils.py":
           # Skip tests/utils.py
           # This skip can be removed once wait_until_signal is no longer used
           continue
```
also i was thinking if removing it is not necessary  right now. i could create a new issue out of it and this can be solved through a new PR. It could be a nice Good first issue for maybe newcomers.

File: tests/_test_msui/test_wms_control.py [hunk 1]
```diff
@@ -103,49 +102,54 @@ def setup(self, qapp):
         yield
         self._teardown()
 
-    def test_no_server(self):
+    def test_no_server(self, qtbot):
         """
         assert that a message box informs about server troubles
         """
+        mock_url = f"{self.scheme}://{self.host}:{self.port-1}"
```
matrss: Please move these back to where they were, there is no need to have the URLs in separate variables.

tests/_test_msui/test_wms_control.py
```diff
@@ -162,15 +166,14 @@ def test_forward_backward_clicks(self):
             pass
 
     @pytest.mark.skip("Has a race condition where the abort might not happen fast enough")
-    def test_server_abort_getmap(self):
+    def test_server_abort_getmap(self, qtbot):
         """
         assert that an aborted getmap call does not change the displayed image
         """
-        self.query_server(self.url)
```
matrss, line 169: It doesn't look quite right in this test. The order of the mouseClick and keyClick calls is reversed now. I think the keyClick should be in the with: block as well.

ReimarBauer:
> @matrss @ReimarBauer can you take a look does everything look good ? and i don't see `wait_until_signal` function in utils.py is being used anywhere so can i remove it now. and what about this part in MSS\tests\test_meta.py
> 
> ```
>  if str(test_file.relative_to(request.config.rootdir)) == "tests/utils.py":
>           # Skip tests/utils.py
>           # This skip can be removed once wait_until_signal is no longer used
>           continue
> ```
> 
> also i was thinking if removing it is not necessary right now. i could create a new issue out of it and this can be solved through a new PR. It could be a nice Good first issue for maybe newcomers.

I don't think it is a good idea, when we already know the two steps to finish this issue. It may happen that we oversee sometimes something. But in this case it would mean we have to reopen the issue immeaditly because incomplete. We do this also with others when we recognize that it is not done.

matrss:
> and i don't see `wait_until_signal` function in utils.py is being used anywhere so can i remove it now.

Please remove it.

> and what about this part in MSS\tests\test_meta.py

As explained in the comment there this conditional skip can be removed as well, once wait_until_signal is gone.

> also i was thinking if removing it is not necessary right now.

I think it is within the scope for this PR to clean up the code that becomes obsolete due to the other changes. I don't want to have lingering dead code remain if it could be cleaned now as well.

rohit2p:
okey got it thankyou

rohit2p:
I saw several linting issues in the conf.py and some other files i am curious are there existing known issues or pull requests related to addressing these Flake8 warnings specifically in these files? this issues have been their since January i guess.

matrss (approved):
Looks good to me now. Thanks for the PR!

matrss:
> I saw several linting issues in the conf.py and some other files i am curious are there existing known issues or pull requests related to addressing these Flake8 warnings specifically in these files? this issues have been their since January i guess.

I assume you mean docs/conf.py, right? Currently we only lint the files under mslib/ and under tests/. There is no specific issue for this yet, but I would prefer if we would lint the entirety of the repository (modulo some auto-generated files). There is a somewhat related issue in #2145, where I wrote a bit about what I would like to have: https://github.com/Open-MSS/MSS/issues/2145#issuecomment-1921162556. This isn't high up on my list of priorities right now, though.

rohit2p:
Thanks for clarifying the context.

PR #2207, opened 2024-02-15
Test on Windows and MacOS

matrss:
WIP

Fixes #2129.

Issue #1919, opened 2023-08-23
Ignore deprecation warnings for pkg_resources.declare_namespace usage


matrss:
These deprecation warnings in the CI runs come from dependencies which still use this deprecated method. There is not much we can do about this, apart from making sure they have a ticket about it in their respective issue tracker.

The following is the list of declared namespaces we see this warning coming from:
- [X] `fs` (issue opened: https://github.com/PyFilesystem/pyfilesystem2/issues/577)
- [X] `fs.opener` (see above)
- [ ] `xstatic` (the github repo contains usage of this deprecated method, the installed package does not (?))
- [ ] `xstatic.pkg`
- [X] `mpl_toolkits` (fixed upstream: https://github.com/matplotlib/matplotlib/blob/a4dca24d04f928a9e614db403c716237446140b2/doc/users/next_whats_new/mpl_toolkit_pep420.rst)
- [ ] `mpl_toolkits.basemap_data`
- [ ] `sphinxcontrib`
- [X] `zope` (tracked here: https://github.com/zopefoundation/meta/issues/194)

I'll tick these off as I make sure that the issue is handled or at least made aware of.

Apart from that I think we should just ignore this type of warning. That is also what others ([xarray](https://github.com/pydata/xarray/blob/42d42bab5811702e56c638b9489665d3c505a0c1/xarray/tests/__init__.py#L40), [ibis](https://github.com/ibis-project/ibis/pull/5980), [SCT](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4127)) are doing.

ReimarBauer:
after this is done we will have also updated versions
https://github.com/Open-MSS/MSS/issues/1750

matrss:
Actually this issue seems to be happening only due to pyfilesystem2 using `pkg_resources`, which is entirely deprecated.

If you just use xstatic for example, there is no issue:
```
>>> import warnings
>>> warnings.resetwarnings()
>>> from xstatic.main import XStatic
```

but as soon as pkg_resources is imported with xstatic installed in the same environment:
```
>>> import warnings
>>> warnings.resetwarnings()
>>> import pkg_resources
<stdin>:1: DeprecationWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html
[...]

PR #2249, opened 2024-02-27
Test on macOS 13 and 14

### LINKED ISSUE #2284: Add a GitHub Actions workflow to test on ARM macOS

### LINKED ISSUE #2285: Add a GitHub Actions workflow to test on Intel macOS

matrss:
Adds a GitHub Actions workflow to test on macOS 13 and 14 runners, i.e. Intel-based and Apple silicon-based runners.

Split of from #2207. Fixes #2284, fixes #2285.

### DIFF
--- a/.github/workflows/testing-all-oses.yml
+++ b/.github/workflows/testing-all-oses.yml
@@ -1,0 +1,53 @@
+name: Test MSS
+
+on:
+  push:
+    branches:
+      - develop
+      - stable
+      - 'GSOC**'
+  pull_request:
+    branches:
+      - develop
+      - stable
+      - 'GSOC**'
+
+jobs:
+  test:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: ["macos-13", "macos-14", "ubuntu-latest"]
+        order: ["normal", "reverse"]
+    steps:
+    - uses: actions/checkout@v4
+    - name: Build requirements.txt file
+      run: |
+        sed -n '/^requirements:/,/^test:/p' localbuild/meta.yaml |
+          sed -e "s/.*- //" |
+          sed -e "s/menuinst.*//" |
+          sed -e "s/.*://" > requirements.tmp.txt
+        cat requirements.d/development.txt >> requirements.tmp.txt
+        sed -e '/^$/d' -e '/^#.*$/d' requirements.tmp.txt > requirements.txt
+        rm requirements.tmp.txt
+        cat requirements.txt
+    - name: Get current year and calendar week
+      id: year-and-week
+      run: echo "year-and-week=$(date +%Y-%V)" >> "$GITHUB_OUTPUT"
+    - uses: mamba-org/setup-micromamba@v1
+      with:
+        environment-file: requirements.txt
+        environment-name: ci
+        cache-environment: true
+        # Set the cache key in a way that the cache is invalidated every week on monday
+        cache-environment-key: environment-${{ steps.year-and-week.outputs.year-and-week }}
+    - name: Run tests
+      timeout-minutes: 20
+      # The ignored files can somehow cause the test suite to timeout.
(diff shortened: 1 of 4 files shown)


matrss:
I had to disable a few tests and three whole files to get the test suite to pass consistently, but the tests that do run still add up to \~64% coverage. To ensure that this is somewhat reliable I've again ran the test suite 50 times, half in normal order and the other half in reverse, which all passed without issue (see: https://github.com/Open-MSS/MSS/actions/runs/8340807936/job/22825515984).

I think this is ready to get merged as a starting point for testing on different platforms. I would like this workflow to become one unified testing workflow for all platforms, which is why I've included running on ubuntu-latest as well. I will try to add Intel-based macOS and Windows to it in future PRs.

.github/workflows/testing-all-oses.yml
```diff
@@ -0,0 +1,53 @@
...
+        # Set the cache key in a way that the cache is invalidated every week on monday
+        cache-environment-key: environment-${{ steps.year-and-week.outputs.year-and-week }}
+    - name: Run tests
+      timeout-minutes: 10
+      # The ignored files can somehow cause the test suite to timeout.
+      # I have no idea yet on why this happens and how to fix it.
+      # Even a module level skip is not enough, they need to be completely ignored.
+      # TODO: fix those tests and drop the ignores
+      run: micromamba run -n ci env QT_QPA_PLATFORM=offscreen pytest -v -n logical --durations=20 --cov=mslib
+        --ignore=tests/_test_msui/test_sideview.py --ignore=tests/_test_msui/test_topview.py --ignore=tests/_test_msui/test_wms_control.py
```
ReimarBauer, line 52: is this 20 a total time? and what happens when the tests timeout of 10 is reached?

ReimarBauer: While debugging a test I recently had seen that it sometimes gets into the http_auth dialog but that is not defined. It is not reproducable, it then enters the password there and fails ;)

pytest.ini
```diff
@@ -6,7 +6,7 @@ log_file = pytest.log
 log_file_level = DEBUG
 log_file_format = %(asctime)s %(levelname)s %(message)s
 log_file_date_format = %Y-%m-%d %H:%M:%S
-timeout = 30
+timeout = 60
 filterwarnings =
     # These namespaces are declared in a way not conformant with PEP420. Not much we can do about that here, we should keep an eye on when this is fixed in our dependencies though.
```
ReimarBauer, line 11: see https://github.com/xstatic-py/xstatic/issues/12

matrss: That is unrelated to this PR and there is an issue for it already: #1919. As mentioned there it seems like the real culprit is PyFilesystem2.

.github/workflows/testing-all-oses.yml
matrss: The 20 just means "print the duration of the 20 slowest tests". I copied that over from the other testing workflow. If the 10 minute step timeout is reached the workflow is aborted and marked as failed.

Yes, there are still some flaky tests left. The files that are disabled here seem to cause at least some of that, since I have seen the same symptoms that they exhibit on macOS also happen on Linux, just much less frequently which is why it isn't much of a problem in the other workflow.

matrss:
I am considering merging #2288 into this PR, since that is just a single line change. I initially thought that Intel-based macOS would require more changes, but it turned out that it doesn't.

.github/workflows/testing-all-oses.yml
matrss: Note: I've increased the step timeout because the macOS 13 runners are a bit slower than the macOS 14 ones.

matrss:
I've added Intel-based macOS to this PR now. It is a bit slower, but seems to run reliably as well. This is ready for review again.

ReimarBauer (approved):
LGTM :)

PR #2257, opened 2024-02-29
fix:set data to NaN at right border of non-global data sets to avoid extrapolation issues.

### LINKED ISSUE #2226: sideview out of valid data range causes strange extrapolated values
When having a flight path that extends out of the geographical range of the data, the sideview plots include a strange extrapolation. There should be no data displayed for way points outside the valid range.

This shoes the (correctly) displayed data  range and the flight path:
topview_chile
and this the corresponding side view
sideview_chile
From about half way between the first and second waypoint on the left to the right end of the plot, there are no available data.

joernu76:
**Purpose of PR?**:

Fixes #2226

no

BIANKA server works.

**Checklist:**
- [ ] Commit has unit tests

### DIFF
--- a/mslib/mswms/mss_plot_driver.py
+++ b/mslib/mswms/mss_plot_driver.py
@@ -467,6 +467,10 @@ class VerticalSectionDriver(MSSPlotDriver):
         lon_data = ((self.lon_data - left_longitude) % 360) + left_longitude
         lon_indices = lon_data.argsort()
         lon_data = lon_data[lon_indices]
+        # Identify jump in longitudes due to non-global dataset
+        dlon_data = np.diff(lon_data)
+        jump = np.where(dlon_data > 2 * dlon)[0]
+
         lons = ((self.lons - left_longitude) % 360) + left_longitude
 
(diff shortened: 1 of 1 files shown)


open-mss-build (approved):
verified on an actual system.

ReimarBauer (approved):
such problems can be visual good identified.

Thx

jgrooss:
I did now manage the mswms locally running and it seems that the error is not resolved.
below are tow examples that should be identical
(1) from (patched) MSS server 
sideview_ECMWF_PV_bianka
(2) from locally running server using branch i2226 (or current stable)
sideview_ECMWF_PV_i2226

ReimarBauer:
@jgrooss you have to wait for the release of 8.3.4 or patch manually.

Have you patched?

joernu76:
I propose to upgrade the bianka server to stable and reevaluate the issue.
Setting up a test server is not trivial.

jgrooss:
sorry, my fault. I accidently used mss 8.3.3. instaead of this branch.
So, the issue is resolved.

PR #2266, opened 2024-03-04
Add new tests for mscolab.model

### LINKED ISSUE #2253: add missing tests for mscolab.model
Currently we have in test_models only tests for User.

We need tests for AwareDateTime, Permission, Operation, Message, Change

Preetam-Das26:
**Purpose of PR?**:

Fixes #2253

**Checklist:**
- [ ] Commit has unit tests

### DIFF
--- a/tests/_test_mscolab/test_models.py
+++ b/tests/_test_mscolab/test_models.py
@@ -27,6 +27,7 @@
 import pytest
 import datetime
 import textwrap
+import pytz
 
 from mslib.mscolab.server import register_user
 from mslib.mscolab.models import AwareDateTime, User, Permission, Operation, Message, Change
@@ -47,6 +48,13 @@ class Test_AwareDateTime:
         result_result = aware_datetime_type.process_result_value(self.aware_datetime, None)
         assert result_result is not None
         assert result_result == self.aware_datetime
+
+        result_none = aware_datetime_type.process_bind_param(None, None)
+        assert result_none is None
+
+        cet_time = datetime.datetime.now(tz=pytz.timezone("CET"))
+        result_cet = aware_datetime_type.process_bind_param(cet_time, None)
+        assert result_cet is not None
 
 
 class Test_User:
--- a/tests/_test_mscolab/test_models.py
+++ b/tests/_test_mscolab/test_models.py
@@ -33,97 +33,109 @@ from mslib.mscolab.server import register_user
 from mslib.mscolab.models import AwareDateTime, User, Permission, Operation, Message, Change
 
 
-class Test_User:
-    @pytest.fixture(autouse=True)
-    def setup(self, mscolab_app):
-        self.userdata = 'UV10@uv10', 'UV10', 'uv10'
-        with mscolab_app.app_context():
-            result = register_user(self.userdata[0], self.userdata[1], self.userdata[2])
-            assert result["success"] is True
-            yield
+@pytest.fixture(autouse=True)
+def setup(mscolab_app):
+    userdata = 'UV10@uv10', 'UV10', 'uv10'
+    with mscolab_app.app_context():
+        result = register_user(userdata[0], userdata[1], userdata[2])
+        assert result["success"] is True
+        yield
 
-    def test_generate_auth_token(self):
-        user = User(self.userdata[0], self.userdata[1], self.userdata[2])
-        token = user.generate_auth_token()
-        assert token is not None
-        assert len(token) > 20
 
-    def test_verify_auth_token(self):
-        user = User(self.userdata[0], self.userdata[1], self.userdata[2])
-        token = user.generate_auth_token()
-        uid = user.verify_auth_token(token)
-        assert user.id == uid
+def test_generate_auth_token():
+    userdata = 'UV10@uv10', 'UV10', 'uv10'
+    user = User(userdata[0], userdata[1], userdata[2])
+    token = user.generate_auth_token()
+    assert token is not None
+    assert len(token) > 20
 
-    def test_verify_password(self):
-        user = User(self.userdata[0], self.userdata[1], self.userdata[2])
-        assert user.verify_password("fail") is False
-        assert user.verify_password(self.userdata[2]) is True
 
-    def test_aware_datetime_conversion(self):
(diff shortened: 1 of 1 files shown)


Preetam-Das26:
@ReimarBauer Kindly review this PR.

tests/_test_mscolab/test_models.py
```diff
@@ -25,9 +25,25 @@
...
+
+
+class Test_AwareDateTime:
+    @pytest.fixture(autouse=True)
+    def setup(self):
+        self.aware_datetime = datetime.datetime.now(tz=datetime.timezone.utc)
+
+    def test_aware_datetime_conversion(self):
+        aware_datetime_type = AwareDateTime()
+
```
ReimarBauer, line 91: check the value is None case too

tests/_test_mscolab/test_models.py
```diff
@@ -55,3 +71,83 @@ def test_verify_password(self):
...
+        self.access_level = "admin"
+
+    def test_permission_creation(self):
+        permission = Permission(self.u_id, self.op_id, self.access_level)
+
+        assert permission.u_id == self.u_id
+        assert permission.op_id == self.op_id
+        assert permission.access_level == self.access_level
+
+
```
ReimarBauer: please add also a test for
 test_permission_repr_values

maybe you can use @pytest.mark.parametrize and use all other than admin

tests/_test_mscolab/test_models.py
```diff
@@ -55,3 +71,83 @@ def test_verify_password(self):
...
+        self.category = "test_category"
+
+    def test_operation_creation(self):
+        operation = Operation(self.path, self.description, category=self.category)
+
+        assert operation.path == self.path
+        assert operation.description == self.description
+        assert operation.category == self.category
+        assert operation.active is True
+
```
ReimarBauer: also repr needs a test

ReimarBauer (commented):
thx, good start. see my comments

Preetam-Das26:
Welcome.
Ok I will go through the comments.

Preetam-Das26:
@ReimarBauer Changes done.

tests/_test_mscolab/test_models.py
joernu76: This is a good start and a definite improvement of the tests. To achieve a 100% coverage (a typical goal), one needs to write tests that trigger all LoC of the relevant functions. Here, the next step would be to supply None as datetime and expect None back to test the if-branch in the code.

More important, the functionality of the code should be tested. Here, the relevant functionality is a conversion from (potentially) local time to UTC, either with/without actual conversion.

Here, one could add tests with a time in, e.g., CET, generated by using  ```import pytz; pytz.timezone("CET")```

Preetam-Das26:
Kindly review @ReimarBauer @joernu76

joernu76 (approved):

tests/_test_mscolab/test_models.py
```diff
@@ -55,3 +71,83 @@ def test_verify_password(self):
...
+        self.u_id = 1
+        self.op_id = 1
+        self.access_level = "admin"
+
+    def test_permission_creation(self):
+        permission = Permission(self.u_id, self.op_id, self.access_level)
+
+        assert permission.u_id == self.u_id
+        assert permission.op_id == self.op_id
+        assert permission.access_level == self.access_level
```
matrss: While other tests are structured this way, I would like to not introduce new classes when they are not required (and they are never actually required). This test could just be:
```python
def test_permission_creation(self):
    permission = Permission(1, 1, "admin")

    assert permission.u_id == 1
    assert permission.op_id == 1
    assert permission.access_level == "admin"
```
and be more legible.

The `setup` autouse fixtures are more of a crutch to convert old unittest-style setup and teardown methods of classes into pytest-compatible fixtures, without larger changes to the tests. Fixtures should be reserved for setup and teardown of resources that a test might require, like setting up a database connection. If a test really requires a fixture and that fixture is shared between multiple tests I would prefer if those tests still were plain functions and each of them requested the fixture themselves, instead of using a class with an autouse fixture.

This comment applies to the other (newly introduced) classes as well.

Preetam-Das26: Ok I will make the necessary changes.

Preetam-Das26:
Changes done @matrss

tests/_test_mscolab/test_models.py
matrss: The new tests are now nested under the Test_User class. What I meant is that I would like them to be top-level function definitions, i.e. not nested under any class.

tests/_test_mscolab/test_models.py
```diff
@@ -55,3 +58,72 @@ def test_verify_password(self):
...
+
+    def test_operation_creation(self):
+        operation = Operation("/path/to/operation", "Description of the operation", category="test_category")
+
+        assert operation.path == "/path/to/operation"
+        assert operation.description == "Description of the operation"
+        assert operation.category == "test_category"
+        assert operation.active is True
+
+    def test_repr(self):
```
matrss: Please rename this to `test_operation_repr`, so that the test name is unambiguous as to which repr is tested.

tests/_test_mscolab/test_models.py
Preetam-Das26: Alright I will do the changes accordingly.

Preetam-Das26:
Changes done.  @matrss

tests/_test_mscolab/test_models.py
matrss: Now you have also moved the pre-existing tests in the Test_User class, with the effect that the setup autouse fixture now applies to all tests in the file. This is not what I meant. What I meant is that the newly added tests should be at the top-level, but the Test_User class with the tests that you haven't added (and the autouse fixture) should stay as they were (refactoring those could be a future effort over the entire test suite).

Preetam-Das26: Sorry for the misunderstanding. I thought everything was supposed to be independent of the class. I will make the necessary changes.

matrss (approved):
LGTM now. Thanks for the PR!

PR #2272, opened 2024-03-09
coverage for all branches and PRs in Coveralls

### LINKED ISSUE #2174: Send coverage for all branches and PRs to coveralls
Right now coverage is only send for the develop branch. I don't see any reason for why we should limit this. If we send coverage for e.g. PRs as well we would get a status report about coverage changes directly in the PR.

Preetam-Das26:
**Purpose of PR?**:

Fixes #2174 

**Checklist:**
- [ ] Commit has unit tests

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -69,7 +69,7 @@ jobs:
         ${{ (matrix.order == 'normal' && ' ') || (matrix.order == 'reverse' && '--reverse') }} tests
 
     - name: Collect coverage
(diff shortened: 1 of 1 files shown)


matrss (approved):
LGTM.

In the future we could consider if we want to collect coverage for all test runs (i.e. normal and reverse order) as well, and if we maybe want to use the coveralls github action (https://github.com/coverallsapp/github-action), since that would give a deeper integration with GitHub through additional reports as comments in PRs.

But this is already an adequate way to get coverage info in PRs.

PR #2278, opened 2024-03-12
cycle login with multiple_flightpath_dockwidget

### LINKED ISSUE #2037: multiple flightpath docking widget does not clear all topviews

ReimarBauer:
**Purpose of PR?**:

Fixes #2037

**Checklist:**
- [X] Bug fix. Fixes #2037
- [ ] Commit has unit tests

### DIFF
--- a/tests/_test_msui/test_mscolab.py
+++ b/tests/_test_msui/test_mscolab.py
@@ -367,22 +367,26 @@ class Test_Mscolab:
             assert len(self.window.get_active_views()) == 2
         qtbot.wait_until(assert_active_views)
 
-        def select_widget():
-            # open multiple flightpath first window
-            topview_0.window.cbTools.currentIndexChanged.emit(6)
+        # open multiple flightpath first window
+        topview_0.window.cbTools.currentIndexChanged.emit(6)
+
+        def assert_dock_loaded():
             assert topview_0.window.docks[5] is not None
-        qtbot.wait_until(select_widget)
+        qtbot.wait_until(assert_dock_loaded)
+
         # activate all operation, this enables them in the docking widget too
         self._activate_operation_at_index(1)
         self._activate_operation_at_index(2)
         self._activate_operation_at_index(0)
         # ToDo refactor to be able to activate/deactivate by the docking widget and that it can be checked
 
-        def select_widget():
-            # open multiple flightpath second window
-            topview_1.window.cbTools.currentIndexChanged.emit(6)
+        # open multiple flightpath second window
+        topview_1.window.cbTools.currentIndexChanged.emit(6)
+
+        def assert_dock_loaded():
             assert topview_1.window.docks[5] is not None
-        qtbot.wait_until(select_widget)
+        qtbot.wait_until(assert_dock_loaded)
+
         # activate all operation, this enables them in the docking widget too
         self._activate_operation_at_index(1)
         self._activate_operation_at_index(2)
@@ -395,6 +399,11 @@ class Test_Mscolab:
         qtbot.wait_until(assert_label_text)
 
         self.window.mscolab.logout()
+
+        def assert_logout_text():
+            assert self.window.usernameLabel.text() == "User"
+        qtbot.wait_until(assert_logout_text)
+
         self._connect_to_mscolab(qtbot)
         self._login(qtbot, emailid=self.userdata[0], password=self.userdata[2])
         # verify logged in again
@@ -805,7 +814,7 @@ class Test_Mscolab:
 
     def _activate_operation_at_index(self, index):
         # The main window must be on top
-        QtTest.QTest.keyClick(self.window, QtCore.Qt.Key_Up, QtCore.Qt.ControlModifier)
+        self.window.activateWindow()
         # get the item by its index
         item = self.window.listOperationsMSC.item(index)
         point = self.window.listOperationsMSC.visualItemRect(item).center()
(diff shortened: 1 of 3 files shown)


ReimarBauer:
A further test which enables multiple operations is needed

ReimarBauer:
> A further test which enables multiple operations is needed

seems that would need a wider refactoring. This could be part of a GSOC project or later a new issue.

We should have a class for collecting the data and all QT elements and related methods in the  MultipleFlightpathControlWidget for using the data. Currently that is mixed.

By the current design the control qt elements of the different operations can't be accessed over the parent. That means also they can't be used for tests or tutorials, because they can't be searched.

The gifcycle shows the problem.
problem_for_a_tutorial

rohit2p:
> seems that would need a wider refactoring. This could be part of a GSOC project or later a new issue.

So, by this you mean for now we don't need to worry about the test and for now just focus on having a class for collecting the data and all QT elements and related methods in the `MultipleFlightpathControlWidget` for using the data ??

ReimarBauer:
We need more tests. The tests define the later solution. When we know the reason a test now fails we set it to xfail.
When the code gets improved it can be checked with the tests.  

The tests should describe first what we expect. Then when we see the picture we can do the refactoring.
I assume we find some more problems or get ideas what is missing.

ReimarBauer:
Visual the test results on a "slow" CPU verified.  Some ToDo's added

tests/_test_msui/test_mscolab.py
```diff
@@ -334,6 +334,73 @@ def test_view_open(self, qtbot):
...
+        topview_1 = self.window.listViews.item(1)
+
+        def assert_active_views():
+            # check 2 view opened
+            assert len(self.window.get_active_views()) == 2
+        qtbot.wait_until(assert_active_views)
+
+        def select_widget():
+            # open multiple flightpath first window
+            topview_0.window.cbTools.currentIndexChanged.emit(6)
```
matrss: This line will potentially be executed multiple times. Is that intended or should it just happen once and then wait_until should wait for the assert that ensures this lines effect has happened?

tests/_test_msui/test_mscolab.py
```diff
@@ -334,6 +334,73 @@ def test_view_open(self, qtbot):
...
+        qtbot.wait_until(select_widget)
+        # activate all operation, this enables them in the docking widget too
+        self._activate_operation_at_index(1)
+        self._activate_operation_at_index(2)
+        self._activate_operation_at_index(0)
+        # ToDo refactor to be able to activate/deactivate by the docking widget and that it can be checked
+
+        def select_widget():
+            # open multiple flightpath second window
+            topview_1.window.cbTools.currentIndexChanged.emit(6)
```
matrss: Same as the other comment.

tests/_test_msui/test_mscolab.py
```diff
@@ -334,6 +334,73 @@ def test_view_open(self, qtbot):
...
+        self._activate_operation_at_index(2)
+        self._activate_operation_at_index(0)
+        # ToDo refactor to be able to activate/deactivate by the docking widget and that it can be checked
+
+        def assert_label_text():
+            # verify logged in
+            assert self.window.usernameLabel.text() == self.userdata[1]
+        qtbot.wait_until(assert_label_text)
+
+        self.window.mscolab.logout()
```
matrss, line 401: Maybe there should be an explicit assert after this to check that the logout was successful.

tests/_test_msui/test_mscolab.py
```diff
@@ -737,6 +804,9 @@ def assert_operation_is_listed():
         qtbot.wait_until(assert_operation_is_listed)
 
     def _activate_operation_at_index(self, index):
+        # The main window must be on top
+        QtTest.QTest.keyClick(self.window, QtCore.Qt.Key_Up, QtCore.Qt.ControlModifier)
```
matrss: This line is a bit cryptic and I don't know if this would be another potential race condition. In my mental model of Qt every keyClick is always an asynchronous event, so I think it would be.

Could this be replaced by https://doc.qt.io/qt-5/qwidget.html#activateWindow?

matrss (approved):

PR #2287, opened 2024-03-19
Run the test suite in a random order

### LINKED ISSUE #2192: Add pytest-randomly to randomize test execution
Right now we run tests in normal order and in reverse order in CI. This finds some issues, but not all of them. Ideally we would run tests in a random order to avoid any implicit dependencies between tests.

Right now the test suite is not in a state in which this would be possible, AFAICT.

matrss:
Fixes #2192.

Depends on #2249.

### DIFF
--- a/.github/workflows/testing-all-oses.yml
+++ b/.github/workflows/testing-all-oses.yml
@@ -19,7 +19,6 @@ jobs:
       fail-fast: false
       matrix:
         os: ["macos-13", "macos-14", "ubuntu-latest"]
-        order: ["normal", "reverse"]
     steps:
     - uses: actions/checkout@v4
     - name: Build requirements.txt file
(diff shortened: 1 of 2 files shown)


matrss:
I've again tested how reliable this is: https://github.com/Open-MSS/MSS/actions/runs/8376494045/job/22936321858. Most runs passed, but two didn't. One of them ran into the 60 second pytest timeout, the other reached the 20 minute step timeout. I *think* this is just a symptom of the macOS 13 runners being very much all over the place in terms of performance (runtimes vary between \~8 and \~14 minutes for the jobs that passed as well). I also think that this problem is unrelated to this PR, and could happen on current develop also.

This does decrease coverage a bit, because it disables problematic tests that previously passed by chance, when running in normal and reverse order on linux.

.github/workflows/testing-all-oses.yml
```diff
@@ -29,6 +28,7 @@ jobs:
           sed -e "s/menuinst.*//" |
           sed -e "s/.*://" > requirements.tmp.txt
         cat requirements.d/development.txt >> requirements.tmp.txt
+        echo "pytest-randomly" >> requirements.tmp.txt
```
ReimarBauer, line 31: We maybe should add this to the https://github.com/Open-MSS/MSS/blob/develop/requirements.d/development.txt

or does it once it is installed immeadiatly change to randomly?

matrss: Yes, once it is installed it will randomize the test order by default, unless you pass `--randomly-dont-reorganize`. That is why I decided to only add it in this workflow, until the test suite fully works with it.

ReimarBauer (approved):
Thx

PR #2289, opened 2024-03-20
msidp : Update Docs, user settings and warn msgs

### LINKED ISSUE #2072: Do not expose the testing IdP implementation
> https://github.com/Open-MSS/MSS/blob/f6b3ad110dbfaa12d439cadea6476dcd54d9328d/localbuild/meta.yaml#L22
> Do we really want to expose the IdP that was implemented? I still consider this to be testing infrastructure that should be moved to tests/ and be used as a lightweight target in integration tests of the SP functionality only.

_Originally posted by @matrss in https://github.com/Open-MSS/MSS/pull/2069#discussion_r1384668617_

nilupulmanodya:
**Purpose of PR?**:

fixes #2072

### DIFF
--- a/docs/conf_sso_test_msscolab.rst
+++ b/docs/conf_sso_test_msscolab.rst
@@ -114,4 +114,4 @@ When migrations finished, you can start mscolab server  using the following comm
     $ msui
 
 * Login with identity provider through Qt Client application.
-* To log in to the mscolab server through the identity provider, you can use the credentials specified in the ``PASSWD`` section of the ``MSS/mslib/msidp/idp.py`` file. Look for the relevant section in the file to find the necessary login credentials.
+* To log in to the mscolab server through the identity provider, you can use the credentials specified in the ``USERS`` and ``PASSWD`` section of the ``MSS/mslib/msidp/idp_user.py`` file. Look for the relevant section in the file to find the necessary login credentials.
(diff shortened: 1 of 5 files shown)


docs/conf_sso_test_msscolab.rst
```diff
@@ -1,6 +1,6 @@
 Configuration MSS Colab Server with Testing IdP for SSO
 =======================================================
-Testing IDP (`mslib/msidp`) is specifically designed for testing the Single Sign-On (SSO) process with the mscolab server using PySAML2.
+Testing IDP (`mslib/msidp`) is specifically designed for testing the Single Sign-On (SSO) process with the mscolab server using PySAML2. This is solely for development and testing purposes, do not use in production environments.
```
ReimarBauer, line 4: use the caution or warning attribute, https://sphinxawesome.xyz/demo/notes/

That hightlights it better.

e.g. 
```
.. warning::
   When running publicly rather than in development, you should not use the built-in development server ( msidp / idp.py ).

   The development server is provided by MSS for convenience, but is not designed to be particularly efficient, stable, or secure.
```

mslib/msidp/idp_user.py
```diff
@@ -133,3 +58,7 @@
         "schacUserPresenceID": "skype:pepe.perez",
```
ReimarBauer: when that not be needed to be named "skype"  I would prefer "sky"

The background of  that request is, that we recently got requests by scanning tools finding strings of proprietary software on our systems. This tools create noise because of possible license issues.

ReimarBauer (approved):
LGTM

PR #2290, opened 2024-03-20
Run the test suite on Windows runners

### LINKED ISSUE #2286: Add a GitHub Actions workflow to test on Windows

matrss:
Fixes #2286.

Depends on #2249.

### DIFF
--- a/.github/workflows/testing-all-oses.yml
+++ b/.github/workflows/testing-all-oses.yml
@@ -18,9 +18,18 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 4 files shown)


matrss:
The test suite is completely broken on Windows at this point (>200 errors), and I don't want to spend too much time on a bad OS that I dislike. So I am closing this PR and if anyone else wants to try to fix these Windows problems feel free to pick the issue up.

I think the issues are mostly related to how Windows handles the filesystem differently, e.g. you can't delete files that are still in use while on Linux/macOS that is not a problem. There is probably quite a bit in the cleanup code for tests that needs to be adapted to work on Windows.

PR #2291, opened 2024-03-21
Stop ignoring some test files

matrss:
This PR is to keep an eye on if some other fix allows us to re-enable the test files that are skipped here: https://github.com/Open-MSS/MSS/blob/daf8f2ae64f9907c9e27114660659a5f5d1cb6d5/.github/workflows/testing-all-oses.yml#L52


matrss:
It seems like something changed and this mostly works now.

ReimarBauer:
running again to see if the result recures

matrss:
The one test run that failed I've also seen fail on current stable and develop (the macOS test runs in general seem to have occasional issues). The test that failed isn't even one of those that were re-enabled here.

ReimarBauer (approved):
more tests :)

Issue #2258, opened 2024-02-29
Scheduled builds do not recreate the mamba environment


matrss:
This step is intended to trigger a recreation of the mamba environment every time a scheduled test is done: https://github.com/Open-MSS/MSS/blob/f5034357cef3602afdb59ece9be683a990896c76/.github/workflows/testing.yml#L48-L51

There are some issues with it:
1. The condition is always false for develop, so scheduled tests on develop would not always recreate the environment
2. Since #2196 was merged the condition is also always false for scheduled tests on stable because the event_name is not "schedule", but "workflow_dispatch"

Also, the scheduled tests never trigger a docker build because their event type is never "push", which might be another bug unless I misunderstand when the docker image builds should happen.

matrss:
@ReimarBauer do you have some input on this?

matrss:
@joernu76 this is the related issue to docker image creation.

ReimarBauer:
@matrss are the two merges related and solving this?

PR #2296, opened 2024-03-22
Fix docker image creation from develop branch

matrss:
This changes the testing-develop.yml workflow to be triggered via workflow_dispatch, as is also already done for testing-stable.yml, and changes the logic in testing.yml to always rebuild the environment for runs triggered by workflow_dispatch, while only triggering image rebuilds on push events.

Part of a fix for #2258.

### DIFF
--- a/.github/workflows/testing-develop.yml
+++ b/.github/workflows/testing-develop.yml
@@ -7,6 +7,7 @@ on:
   pull_request:
     branches:
     - develop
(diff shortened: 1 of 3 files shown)


matrss:
The coverage check is red because the coverage is unfortunately still not deterministic, for some reason... So it is more or less random if it succeeds or fails.

joernu76 (approved):

PR #2117, opened 2023-12-12
removing pinned xmlschema

ReimarBauer:
**Purpose of PR?**:

Fixes #2056 

When the xmlschema works here we can update our dockerimages for testing to miniforge3 too

PR #2298, opened 2024-03-26
CI and development environment overhaul

### LINKED ISSUE #2160: Ideas for a CI overhaul
I think the CI setup could be simplified in a lot of ways and I would like to get some input on what you think would be desirable.

#### 1. Removing the docker containers from the CI

Instead of using this bespoke contraption of `testing-{stable,develop}` docker containers, we could instead switch to a more standard setup with e.g. the micromamba setup action: https://github.com/mamba-org/setup-micromamba.

My understanding is that we have the docker containers mainly because then we don't have to reinstall the mamba environment with every test run. The micromamba action can do the same by caching the environment in GitHub's caching mechanism, which could potentially be even faster than the docker approach (because it is directly a part of the GitHub Actions offering).

If we dropped container usage from the CI this would also open up a (hopefully easy) path to testing on the macos and windows runners (those do not support the `container:` options, AFAIK) as well, without much difference in the actions themselves.

This change would entail removing the localbuild directory (for which I don't see a point anyway, conda packaging matters should happen in mss-feedstock) and instead having a more standard environment.yml file.

#### 2. Drop the testing-stable.yml and testing-develop.yml files

It should be possible to declare the given `on:` options in `testing.yml` directly and drop the reusable workflow usage that is there right now.

#### 3. Move the scheduled tests into the mss-feedstock repository

The purpose of the scheduled tests is to check that the package as it gets installed with mamba at that moment still works. AFAICT the scheduled tests in this repository do not actually do this properly, e.g. they always use the test suite from `develop` instead of the released package.

I think this workflow should instead be part of the mss-feedstock repository and it should open issues there, since it is a packaging matter if the published package in conda-forge suddenly breaks.

We could still have a scheduled test in this repository though, but just to prime the environment cache described in 1. every time that it expires. This would, again, just be another `on:` clause in testing.yml.

#### 4. Unify testing.yml and testing_gsoc.yml

Those two workflows have a lot in common and I think they don't have to be separate files.

---

@ReimarBauer I would like to get some input on this before I start to change anything in this regard.

matrss:
This removes the docker image logic from the CI setup and instead replaces it with a setup based on pixi. As part of that it moves the dependency specification from localbuild/meta.yaml to pixi.toml and pixi.lock. This turns the MSS repository into a single source of truth for both the application code as well as the development environment (whereas the latter was previously only specified in the docker images, and not reproducible in any way).

Setting up a development environment is as simple as installing pixi and running `pixi shell` (or `pixi run <cmd>`, or `pixi install` to just create the environment, etc.). This environment will, by construction, be the same that is used in the CI as well (modulo platform differences).

There is a new workflow that periodically (once a week on Monday) recreates the pixi lockfile and opens a PR for that update. The checks in that PR essentially serve as a replacement for the previous scheduled runs to ensure that no dependency update breaks MSS. Merging that PR is a manual step that can be done just as with any other PR and would then update the environment on the given target branch. This is essentially what was previously the (manual) trigger of a docker image creation.

Including new dependencies can be done with `pixi add`, which will also automatically add the dependency to the (pre-existing) lockfile. This means dependency additions can be part of the PR that necessitate them and they won't affect the entire environment (as they previously did, where they would trigger a full image rebuild) but instead just add that new package to the existing specification.

---

This (mostly) implements the ideas I've outlined in #2160. I wanted to substantiate what I had in mind, because I think there was some confusion there about what these ideas would entail.

I consider this to be a massive simplification of the CI setup, while retaining all required functionality (or even improving it). The additions count looks large due to the pixi.lock file, but that file is automatically generated. Outside of it this is a net negative in code size.

Fixes #2160.


joernu76:
What the heck is pixi? This https://pixi.sh/latest/ ? Is this mature/stable? It doesn't have a wikipedia page and I have never heard from it...
So it is a backport of cargo for conda-forge??
I mean, frankly, *every* package manager is better than conda, so it must be better...
I assume this does only affect our development etc. and has no affect on our users?

matrss:
> What the heck is pixi? This https://pixi.sh/latest/ ? Is this mature/stable? It doesn't have a wikipedia page and I have never heard from it...

Yes, that link is correct. It is a relatively new package manager, it was released in summer last year (they had a release party at EuroSciPy 2023, Reimar and I attended it). It is developed by the same people that were/are behind mamba, so for what it's worth there is some credibility behind it.

> So it is a backport of cargo for conda-forge??

I haven't worked with cargo/rust in a while but yes, I think they took a lot of inspiration from that and there are many similarities in how they work.

> I mean, frankly, _every_ package manager is better than conda, so it must be better...

Hard agree.

> I assume this does only affect our development etc. and has no affect on our users?

Yes, the mss-feedstock and therefore the conda-forge package remains the same (and installable through conda, mamba, pixi or whatever else might work with conda-forge packages, with the usual caveat that conda-forge packages can not be tested and might break at any point in time due to the install-time dependency resolution - nothing new from this PR and a gripe I have with the entire conda packaging model).

ReimarBauer:
I recently gave a lighning talk at the Barcamp in Karlsruhe.

Sources:
https://prefix.dev/blog/pixi_a_fast_conda_alternative

There is also a talk by Wolf on pyconde in mid april

ReimarBauer:
with an uptodate conda you have same speed as with mamba, see 

https://conda.org/blog/2023-11-06-conda-23-10-0-release/

matrss:
> with an uptodate conda you have same speed as with mamba, see
> 
> https://conda.org/blog/2023-11-06-conda-23-10-0-release/

Speed isn't the issue that conda has, in my opinion. The issue is that it is impossible to define a development environment in its entirety right next to the source code (i.e. inside of a git repository). Mamba has the same issue, because its interface is basically the same as conda's. Cargo fixes this for the rust world with its Cargo.lock file pinning dependencies from crates.io, PDM or poetry (or others) can do the same for python on top of PyPI, nix can do it independently of the programming language using pinned references to nixpkgs, and pixi does it on top of the conda packaging ecosystem (also basically language independent). Conda or mamba simply don't provide this feature, speed is irrelevant in that case. (Although, using a lock file moves the dependency resolution to the step of creating the lock file, so that at install time there is no dependency resolution at all necessary. A no-op is strictly faster than any solver conda could possibly integrate.)

ReimarBauer:
With trying to differentiate the packages https://github.com/Open-MSS/MSS/pull/2390 I will likly get to a conda requirements.d/mswms.txt, requirements.d/mscolab.txt and likly a requirements.txt for the whole.

I am currently looking for the jinja syntax to read a requirements.txt files into the current meta.yaml which I prefer to use until we can migrate to e.g. rattler-build etc. on conda-forge too. The one in the feedstock is not identical because of the cross compilation. Without any extra hassle I want to be able to redo a build by conda build locally.

But keeiping that dir is then independent for switching in a first step to a requirements.txt instead of current transformation from a meta.yaml.  With the jinja2 Syntax there is then also no duplication.

ReimarBauer (commented):
nice!
Wolf recently announced a speedup using conda-forge from prefix-dev.
https://x.com/wolfvollprecht/status/1863590394992193803
We maybe should move all docs etc to pixi too, there is also an issue for that.

pixi.toml
```diff
@@ -0,0 +1,105 @@
...
+python = "*"
+python-engineio = ">=4"
+python-slugify = "*"
+python-socketio = ">=5"
+qt = ">=5.15.0"
+requests = ">=2.31.0"
+scipy = "*"
+shapely = ">=2.0.0"
+skyfield = ">=1.12"
+skyfield-data = ">=5"
```
ReimarBauer: `>=6`

when it not always uses the newest, we would need to enumber it each year

pixi.toml
```diff
@@ -0,0 +1,105 @@
...
+libxmlsec1 = "*"
+
+[target.osx-64.dependencies]
+libxmlsec1 = "*"
+
+[target.osx-arm64.dependencies]
+libxmlsec1 = "*"
+
+[feature.dev.dependencies]
+boa = "*"
```
ReimarBauer: we should remove this. 

This was also only an intermediate on the way to rattler-build

pixi.toml
```diff
@@ -0,0 +1,105 @@
...
+pytest = "*"
+pytest-cov = "*"
+pytest-qt = "*"
+pytest-reverse = "*"
+pytest-timeout = "*"
+pytest-xdist = "*"
+sphinx = "*"
+sphinx_rtd_theme = "*"
+xmlschema = "<2.5.0"
+
```
ReimarBauer, line 121: this sections are a nice feature, so we could also add a section for the tutorials.

pixi.toml
```diff
@@ -0,0 +1,105 @@
...
+pynco = "*"
+pytest = "*"
+pytest-cov = "*"
+pytest-qt = "*"
+pytest-reverse = "*"
+pytest-timeout = "*"
+pytest-xdist = "*"
+sphinx = "*"
+sphinx_rtd_theme = "*"
+xmlschema = "<2.5.0"
```
ReimarBauer: the xmlschema and its pinning was removed some time ago
https://github.com/Open-MSS/MSS/pull/2117/files

pixi.toml
```diff
@@ -0,0 +1,105 @@
...
+gpxpy = ">=1.4.2"
+hdf4 = "*"
+isodate = "*"
+itsdangerous = "*"
+keyring = "*"
+libtiff = "<4.5.0"
+lxml = "*"
+markdown = "*"
+matplotlib = ">=3.5.3"
+menuinst = "*"
```
ReimarBauer: menuinst is not implemented for pixi, https://github.com/prefix-dev/pixi/discussions/1894

matrss:
@ReimarBauer this is still WIP, I have only rebased so far and didn't yet update the pixi.toml file to reflect all of the changes since end of March or work on the documentation or anything. I am also looking into `pixi upgrade` instead of specifying `=*` version constraints; that option has been newly added since I started this PR.

ReimarBauer:
> @ReimarBauer this is still WIP, I have only rebased so far and didn't yet update the pixi.toml file to reflect all of the changes since end of March or work on the documentation or anything. I am also looking into `pixi upgrade` instead of specifying `=*` version constraints; that option has been newly added since I started this PR.

I know that it has WIP state, but I like it especially because of the recent problems which perfectly illustrate what disadvantages we have if we don't do it. These are benevolent comments ;)

.github/workflows/update-pixi-lockfile.yml
```diff
@@ -0,0 +1,39 @@
...
+          pixi-version: latest
+          run-install: false
+      - name: Update pixi lockfile
+        run: |
+          set -o pipefail
+          pixi update --json | pixi exec pixi-diff-to-markdown >> diff.md
+      - name: Create or update pull request
+        uses: peter-evans/create-pull-request@v7
+        with:
+          token: ${{ secrets.PAT }}
```
matrss, line 33: This is assuming that the currently-configured PAT has the right permissions to create a PR against this repository.

setup.py
```diff
@@ -44,7 +44,7 @@
     console_scripts.append('msidp = mslib.msidp.idp:main')
 
 setup(
-    name="mss",
+    name="mslib",
```
matrss: This name change was necessary because of a name collision with the mss package on PyPI, which we also depend on. Without it the editable install of this MSS into the pixi environment was impossible.

ReimarBauer: I am not clear about the consequences. Does this mean that we would have to call ourselves MSLIB on conda-forge and within the project? I would then need to initiate a migration; it also has implications for publications.

If it all comes down to a renaming everywhere, then open-mss fits well with gh and open-mss.org.

The other mss project is called python-mss on conda-forge. If we choose a new name, we should do so carefully because "Mission Support System" is a very generic term.

open-mss?

.github/workflows/update-pixi-lockfile.yml
ReimarBauer: ok, I updated it on some places, end of year, we will see.

.github/workflows/update-pixi-lockfile.yml
```diff
@@ -0,0 +1,39 @@
...
+on:
+  schedule:
+  # At 04:00 on Monday
+    - cron: "0 4 * * 1"
+  # ...or manually
+  workflow_dispatch:
+
+jobs:
+  pixi-update:
+    runs-on: ubuntu-latest
```
ReimarBauer, line 12: there maybe a chiseled or slim package available which we can also use, but its not important, only if that takes time

pixi.lock
```diff
@@ -0,0 +1,17956 @@
...
+      - conda: https://conda.anaconda.org/conda-forge/noarch/pyopenssl-24.2.1-pyhd8ed1ab_2.conda
+      - conda: https://conda.anaconda.org/conda-forge/noarch/pyparsing-3.2.1-pyhd8ed1ab_0.conda
+      - conda: https://conda.anaconda.org/conda-forge/linux-64/pyproj-3.6.1-py311h0f98d5a_10.conda
+      - conda: https://conda.anaconda.org/conda-forge/linux-64/pyqt-5.15.9-py311hf0fb5b6_5.conda
+      - conda: https://conda.anaconda.org/conda-forge/linux-64/pyqt5-sip-12.12.2-py311hb755f60_5.conda
+      - conda: https://conda.anaconda.org/conda-forge/noarch/pysaml2-7.5.2-pyhd8ed1ab_0.conda
+      - conda: https://conda.anaconda.org/conda-forge/noarch/pyshp-2.3.1-pyhd8ed1ab_1.conda
+      - conda: https://conda.anaconda.org/conda-forge/noarch/pysocks-1.7.1-pyha55dd90_7.conda
+      - conda: https://conda.anaconda.org/conda-forge/noarch/pytest-8.3.4-pyhd8ed1ab_1.conda
+      - conda: https://conda.anaconda.org/conda-forge/noarch/pytest-cov-6.0.0-pyhd8ed1ab_1.conda
```
ReimarBauer, line 2451: does this mean we have all testing, devoloping tools always installed on a developer and users system?

pixi.toml
```diff
@@ -0,0 +1,123 @@
...
+itsdangerous = "*"
+keyring = "*"
+libtiff = "*"
+lxml = "*"
+markdown = "*"
+matplotlib = ">=3.5.3"
+metpy = "*"
+multidict = "*"
+netcdf4 = "*"
+numpy = "<2"
```
ReimarBauer, line 43: after merge on develop it is >2

pixi.toml
```diff
@@ -0,0 +1,123 @@
...
+xstatic-bootstrap = "*"
+xstatic-jquery = "*"
+
+[target.linux-64.dependencies]
+dbus-python = "*"
+libxmlsec1 = "*"
+
+[target.osx-64.dependencies]
+libxmlsec1 = "*"
+
```
ReimarBauer, line 81: we maybe need dbus-python also on osx-64. I have this on my mss-9.3.1 installation env.

better adding it.

The tests have a mocked TestKeyring which don't verify this need.

ReimarBauer (commented):
I added some ideas/questions.

setup.py
matrss: This is only the name of the python package defined by this setup.py file. It does not mean that we have to change the package name anywhere else AFAIU. PyPI should be the only package registry that actually enforces matching package names in the python package and the project on the index, if it even does that.

I've called it mslib because that is the name of the main module inside, but I agree it isn't a nice name. open-mss would sound good to me. `python-mss` is called that on conda-forge because all pure-python packages should get a `python-` prefix there AFAIK.

.github/workflows/update-pixi-lockfile.yml
matrss: I think you are confusing this with docker images? These are GitHub actions runner labels and all of the available ones (excluding the possibility of self-hosted runners) are listed here: https://docs.github.com/en/actions/writing-workflows/choosing-where-your-workflow-runs/choosing-the-runner-for-a-job#standard-github-hosted-runners-for-public-repositories

pixi.lock
matrss: Not really. `pixi {install,shell,run,...}` should only actually download the packages for the default environment, if you don't request the dev environment. Likewise for the other ones (addons, tutorials). The lock file only contains the combined environment resolution result.

Users aren't affected anyway since they do not interact with this pixi project at all.

pixi.toml
matrss: `>=2` actually

pixi.toml
matrss: I somehow remembered that the dbus-python package did not exist for osx at all, but it looks like it does. Weird.

setup.py
ReimarBauer: let us change that name already to open-mss. Then we have used that already there and at some point someone again tries to create a pypi package it needs not another renaming.

We can change the inner later anyway.

.github/workflows/update-pixi-lockfile.yml
ReimarBauer: yes, catched

docs/development.rst
```diff
@@ -131,102 +114,31 @@ Requirements
...
-This can be used to run tests on an invisible virtual display by prepending the pytest call with `xvfb-run`, e.g.::
-
-  $ xvfb-run pytest ...
-
-We have implemented demodata as data base for testing. On first call of pytest a set of demodata becomes stored
-in a /tmp/mss* folder. If you have installed gitpython a postfix of the revision head is added.
+    pixi run -e dev msui
 
 
 Setup MSWMS server
```
ReimarBauer, line 144: Please also change in the MSWMS and MSColab sections also the calls.
I wanted to send a patch but somehow I do something wrong, "There was an error committing your changes: File could not be edited" 
Next time we meet I need to become "aufschlaut"

https://iffmd.fz-juelich.de/IENAfBSeRhuFDO5VIN2s3Q#
Then we have all relevant here updated

ReimarBauer (changes requested):
I tried the whole process on my Mac also for the servers not only the ui.
a few commands should be changed in the development.rst for the servers

https://iffmd.fz-juelich.de/IENAfBSeRhuFDO5VIN2s3Q#

we should use open-mss from now on in the setup.py.

docs/development.rst
matrss: The error doesn't read like anything I've ever encountered from git. At which point did you get it?

Anyway, I've made those changes.

matrss:
Off-topic: I like the new grouping of the CI results, makes it much easier to see what failed at a glance...

setup.py
matrss: I changed it. We should reserve the open-mss name on PyPI though, so we don't run into this again in the future.

docs/development.rst
ReimarBauer: I´d not tried again today, this was the path


https://github.com/user-attachments/assets/a66b4c7f-3965-4fbc-ae28-5b585afd97c5

below that button I had the option to send a PR or direct edit, the PR gaves that error.

setup.py
ReimarBauer: yes, I do

ReimarBauer (approved):
Great !
Thx

setup.py
ReimarBauer: https://test.pypi.org/project/open-mss/

just learned a global installed pixi has precedence

```
(venv) (base) reimarbauer@MacBook-Pro-von-Reimar venv % bin/mss
open-mss is a conda-forge package.You can install it with pixi.
Get pixi from https://pixi.sh/latest/ for your operation system.

pixi global install mss

Usage:
    msui -h
    mswms -h
    mscolab -h

```
[our pypi project
](https://github.com/Open-MSS/open-mss)

joernu76 (approved):

joernu76:
I have several technical questions, which we may adress in todays meeting. Generally this seems like a step in a good direction!

PR #2303, opened 2024-03-26
Removed Unique Constraint for Password Field in User Model

### LINKED ISSUE #2299: mscolab model password does not need to be unique
Currently we define in the model 

password = db.Column(db.String(255), unique=True)

We should not do that, this allows guessing of a password.

Because of a db change this has to be done in develop

workaryangupta:
Fixes #2299


nilupulmanodya (approved):
LGTM, Thx

ReimarBauer (approved):

PR #2304, opened 2024-03-26
Enhance user registration validation

### LINKED ISSUE #2301: we should only create the user object when we have valid data
In register_user we currently create the user object before we validated to have none empty email or username

A check like, 
```
if len(str(email.strip())) == 0 or len(str(username.strip())) == 0:
    return {"success": False, "message": "Your username or email cannot be empty"}
```
is missing

For a new user the User object should be created just before 

`result = fm.modify_user(user, action="create")`

workaryangupta:
Fixes #2301

This PR addresses issue where the user registration process did not sufficiently validate user input(username and email) before creating a new user.

### DIFF
--- a/mslib/mscolab/server.py
+++ b/mslib/mscolab/server.py
@@ -154,7 +154,8 @@ def check_login(emailid, password):
 
 
 def register_user(email, password, username):
-    user = User(email, username, password)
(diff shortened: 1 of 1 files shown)


ReimarBauer (approved):
Thx

PR #2310, opened 2024-03-28
multipe flightpath docking widget on logout

### LINKED ISSUE #2309: multipe flightpath docking widget on logout traceback
when you have worked with the docking widget then closed topview and did logout.

click some operations, click a local flightpath, click again on a operation then logout

```
Fatal error in MSS 8.3.4 on Linux-6.5.0-10027-tuxedo-x86_64-with-glibc2.35

  File "/home/user/PycharmProjects/MSS/mslib/msui/topview.py", line 365, in <lambda>
    self.mainwindow_signal_logout_mscolab.connect(lambda: self.signal_logout_mscolab.emit())
                                                          ^^^^^^^^^^^^^^^^^^^^^^^^^^
RuntimeError: wrapped C/C++ object of type MSUITopViewWindow has been deleted
```

close_topview_then_logout

ReimarBauer:
toggling flightrack, and operations, activating them and logout

**Purpose of PR?**:

Fixes #2309
or at least is a test for that.
Seems not easily reproducable

### DIFF
--- a/mslib/msui/topview.py
+++ b/mslib/msui/topview.py
@@ -359,7 +359,7 @@ class MSUITopViewWindow(MSUIMplViewWindow, ui.Ui_TopViewWindow):
                                                             mscolab_server_url=self.mscolab_server_url,
                                                             token=self.token)
 
-                self.mainwindow_signal_logout_mscolab.connect(lambda: self.signal_logout_mscolab.emit())
+                self.mainwindow_signal_logout_mscolab.connect(self.signal_logout_mscolab.emit)
                 self.mainwindow_signal_listFlighttrack_doubleClicked.connect(
                     lambda: self.signal_listFlighttrack_doubleClicked.emit())
                 self.mainwindow_signal_permission_revoked.connect(
--- a/mslib/msui/topview.py
+++ b/mslib/msui/topview.py
@@ -201,9 +201,6 @@ class MSUITopViewWindow(MSUIMplViewWindow, ui.Ui_TopViewWindow):
         self.mainwindow_signal_permission_revoked = mainwindow.signal_permission_revoked
         self.mainwindow_signal_render_new_permission = mainwindow.signal_render_new_permission
         self.mainwindow_signal_activate_flighttrack = mainwindow.signal_activate_flighttrack
-        self.mainwindow_signal_activate_operation = mainwindow.signal_activate_operation
-        self.mainwindow_signal_login_mscolab = mainwindow.signal_login_mscolab
-        self.mainwindow_signal_logout_mscolab = mainwindow.signal_logout_mscolab
         self.mainwindow_listFlightTracks = mainwindow.listFlightTracks
         self.mainwindow_filterCategoryCb = mainwindow.filterCategoryCb
         self.mainwindow_listOperationsMSC = mainwindow.listOperationsMSC
--- a/mslib/msui/viewwindows.py
+++ b/mslib/msui/viewwindows.py
@@ -76,13 +76,6 @@ class MSUIViewWindow(QtWidgets.QMainWindow):
         Overloads QtGui.QMainWindow.closeEvent(). This method is called if
         Qt receives a window close request for our application window.
         """
-        if self.docks[5] is not None and self.mainwindow_signal_logout_mscolab is not None:
-            # ToDo instead of numbers [5] use names
-            # we need to disconnect the mainwindow_signal_logout when the topview is closed
-            # the next topview will connect one again
-            self.mainwindow_signal_logout_mscolab.disconnect()
-            self.docks[5] = None
-
         if self.force_close:
             ret = QtWidgets.QMessageBox.Yes
         else:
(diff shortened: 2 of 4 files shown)


tests/_test_msui/test_mscolab.py
```diff
@@ -411,6 +411,64 @@ def assert_logout_text():
...
+
+        # activate all operation, this enables them in the docking widget too
+        self._activate_operation_at_index(1)
+        self._activate_operation_at_index(2)
+        self._activate_operation_at_index(0)
+        # ToDo refactor to be able to activate/deactivate by the docking widget and that it can be checked
+
+        self._activate_flight_track_at_index(0)
+        with mock.patch("PyQt5.QtWidgets.QMessageBox.warning", return_value=QtWidgets.QMessageBox.Yes):
+            topview_0.window.close()
```
matrss, line 459: Just closing the window does not automatically delete it. This might be the difference that makes the test not reproduce the issue (since the issue is a use-after-free, after Qt has deleted the widget). You can call `topview_0.window.deleteLater()` to explicitly mark it for deletion (the next time the Qt event loop does its thing) or you can set `Qt::WA_DeleteOnClose` for the widget.

matrss: And there is `sip.delete` as well.

ReimarBauer: and we have to wait until topview_0.window is a weak reference.

mslib/msui/topview.py
```diff
@@ -201,9 +201,6 @@ def __init__(self, parent=None, mainwindow=None, model=None, _id=None,
         self.mainwindow_signal_permission_revoked = mainwindow.signal_permission_revoked
         self.mainwindow_signal_render_new_permission = mainwindow.signal_render_new_permission
         self.mainwindow_signal_activate_flighttrack = mainwindow.signal_activate_flighttrack
-        self.mainwindow_signal_activate_operation = mainwindow.signal_activate_operation
-        self.mainwindow_signal_login_mscolab = mainwindow.signal_login_mscolab
-        self.mainwindow_signal_logout_mscolab = mainwindow.signal_logout_mscolab
         self.mainwindow_listFlightTracks = mainwindow.listFlightTracks
```
ReimarBauer, line 204: lines were doubled

mslib/msui/viewwindows.py
```diff
@@ -76,6 +76,13 @@ def closeEvent(self, event):
         Overloads QtGui.QMainWindow.closeEvent(). This method is called if
         Qt receives a window close request for our application window.
         """
+        if self.docks[5] is not None and self.mainwindow_signal_logout_mscolab is not None:
+            # ToDo instead of numbers [5] use names
+            # we need to disconnect the mainwindow_signal_logout when the topview is closed
+            # the next topview will connect one again
+            self.mainwindow_signal_logout_mscolab.disconnect()
```
matrss: AFAIU this would disconnect _all_ slots from the `mainwindow.signal_logout_mscolab` signal, no matter where they were connected from. This is wrong, unless there is only ever one topview and nothing else connects to the signal. The widget should only disconnect the slots that it itself connected.

ReimarBauer: I know, but not seen where this connect else is used.  And somehow the whole cleanup can be improved.

Manually this seem to work also with many topviews. 
weakref check may need a gc., currently I get many time a timeout and <[RuntimeError('wrapped C/C++ object of type MSUITopViewWindow has been deleted') raised in repr()] that needs something better too.

matrss: > I know, but not seen where this connect else is used.

Both the MSUITopViewWindow and the MultipleFlightpathControlWidget have a connect call on this signal AFAICT.

Even if it isn't a problem _right now_, it can become one at any time when a different widget decides to listen on that signal as well. It will be very hard to debug then. Also, it is just a violation of the separation of the widgets that is avoidable, i.e. it is easy enough to track the result of the connect call and specifically just disconnect that one (or maybe that is not even necessary if the slot wasn't a lambda but a method of the object, since Qt does a check if the slot still exists before calling the slot, AFAIU).

ReimarBauer: `Both the MSUITopViewWindow and the MultipleFlightpathControlWidget have a connect call on this signal AFAICT.`

The solution we need is on disconnect of the mscolab server on the main view then disconnect all connections used by the docking widget in all topviews. There should be no data kept based on an operation when we have no connected mscolab server.  

On all topviews which show the multiple dockingwidget the mscolab part needs to be emptied because of no connection and on a new mscolab server connection the data gets shown again.

ReimarBauer: the lambda was the culprit. I am a bit surprised about the QT magic, but seems to work at least manually.

setup.cfg
```diff
@@ -37,6 +37,6 @@ omit =
 norecursedirs = .git .idea .cache
 
 [flake8]
-ignore = E124,E125,E402,W504
+ignore = E124,E125,E402,W504,A005
```
ReimarBauer, line 40: that should not be in my PR, that is already in develop, how can I fix that?

ReimarBauer: @joernu76 @matrss do you have an idea, what I have to do?

matrss: How did you add the commit here? I am also surprised it shows up in the diff for all commits...

You can drop the commit from your branch and either merge in develop, or rebase your branch on top of develop. Those are the two ways I would generally update a PR branch.

ReimarBauer:
replaced by https://github.com/Open-MSS/MSS/pull/2319

PR #2313, opened 2024-04-09
Fix: Multiple flightpath dockingwidget updates only local new flighpathes

### LINKED ISSUE #2269: develop: multiple flightpath dockingwidget updates only local new flighpathes
When starting with local fligtpathes and the multiple flightpath dockingwidget a traceback occures when you then login.
Also the new operation is not shown.

problem

rohit2p:
**Purpose of PR?**:
I am able to reproduce the error (e.g., KeyError: 10, AttributeError: 'NoneType' object has no attribute 'all_waypoint_data') by creating or deleting operations, followed by attempting to switch operations. Sometimes the error does occurs but sometimes if i do sometime differently it doesn't occurs. In this PR i am trying to figure out the bug and fix it.

Fixes #2269 

**Checklist:**
- [X] Bug fix. Fixes #2269

### DIFF
--- a/tests/_test_msui/test_mscolab.py
+++ b/tests/_test_msui/test_mscolab.py
@@ -33,6 +33,7 @@ import fs.opener.errors
 import requests.exceptions
 import mock
 import pytest
+import random
 
 import mslib.utils.auth
 from mslib.mscolab.conf import mscolab_settings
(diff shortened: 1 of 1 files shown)


rohit2p:
@ReimarBauer @matrss Can you plz take a look, In case of tests I am not able to reproduce the key error? what am i doing wrong?

ReimarBauer:
@rohit2p I merged the other one,  so we have now conflicts here.

rohit2p:
@ReimarBauer when i debug `test_operation_while_switching_error` while switching between operations i should be able to get the error, then why i am not getting it ? can you help me a bit please

rohit2p:
@ReimarBauer when i debug `test_operation_while_switching_error` while switching between operations i should be able to get the error, then why i am not getting it ? can you help me a bit please

ReimarBauer:
when I look on test_operation_while_switching_error this does compared to the description the actions differently.

In the description the topview with the dockingwidget is opened and then the login is done. In the test you start with a login.

rohit2p:
@ReimarBauer see even with that the Key error is not happening during testing

rohit2p:
I think if i create a new operation or delete an old operations after login on top of the old operations and then switch between operations. In this case I'll be able to catch the error but when i try creating an operation after login. It doesn't reflect in the UI of MsColab operations section

joernu76:
The code is clean and it works nicely!

Close to being mergeable.

1) Wrt to the Options: You added a group for "flightpath" options in the TV/SV Options. I would say that both flightpath colour should be moved there as well. Maybe the tickbox "draw flightpath" in there and move the colour options there as well.

2) I also suggest to have a fully identical layout of this box in TV and SV.

3) Also, on my screen, the SV flightpath option box is not fully visible:
Screenshot at 2024-07-02 11-02-04
I think, this should be solvable using some of the QT options. There is still some space left on my screen for this purpose.

4) The TV box has the "Flight path option" label. The "Plot options" label is above the box. I suggest to handle this in a consistent fashion by having all these labels outside above the box:
Screenshot at 2024-07-02 11-03-54

5) Not necessarily for this PR. I find it confusing that when I can select the "active" flighttrack that the controls vanish and I cannot disable the tickbox. Ideally, this element should be "disabled" in QT such that it cannot be clicked and is visually marked as inactive.

6) Not necessarily for this project: When opening the box, all flightpaths are seemingly loaded from MSCOlab. I would suggest to do this in a lazy fashion only for selected flightpaths. Opening it takes an astonishing long time on our PHILEAS server; the user might think that the app has crashed as there is no loading splash screen or similar.

nice!

Issue #2261, opened 2024-03-02
Unexpected Application Termination When Accessing Top View/side view in MSS in Windows OS


Preetam-Das26:
I am encountering an issue while running the MSS application on Windows OS. I installed MSS using the conda-based installation with miniforge. The problem occurs specifically when attempting to use the top view or side view. After selecting any the view tab, clicking on the top view/side view option results in an abrupt exit due to disconnection.

Temporary fix :-
Running msui command in Miniforge prompt in the mssenv environment to use the top view/side view in MSUI.

ReimarBauer:
this problem may be related to a windows version? I can't reproduce it in my developer setup. In this example I do use a Windows Server 2016 Standard

![winreimar](https://github.com/Open-MSS/MSS/assets/278728/3772c4ff-a5fb-4f58-b4b9-c88a343792f4)


@Preetam-Das26 which windows version do you use?

Preetam-Das26:
Microsoft Windows 10

ReimarBauer:
we have to try if menuinst  #  <2.0.0 solves this

ReimarBauer:
@Preetam-Das26 please show a `mamba list ` output.

I got a windows 11 system for tests, but on a fresh mss installation I can't reproduce.

Preetam-Das26:
I am using conda based installation.
conda list output:
```
#
# Name                    Version                   Build  Channel
alabaster                 0.7.16             pyhd8ed1ab_0    conda-forge
alembic                   1.13.1             pyhd8ed1ab_1    conda-forge
appdirs                   1.4.4              pyh9f0ad1d_0    conda-forge
argon2-cffi               23.1.0             pyhd8ed1ab_0    conda-forge
argon2-cffi-bindings      21.2.0          py310h8d17308_4    conda-forge
attrs                     23.2.0             pyh71513ae_0    conda-forge
babel                     2.14.0             pyhd8ed1ab_0    conda-forge
basemap                   1.3.6           py310h791411c_2    conda-forge
basemap-data              1.3.2              pyhd8ed1ab_3    conda-forge
bcrypt                    4.1.2           py310h87d50f1_0    conda-forge
beautifulsoup4            4.12.3             pyha770c72_0    conda-forge
bidict                    0.23.1             pyhd8ed1ab_0    conda-forge
blinker                   1.7.0              pyhd8ed1ab_0    conda-forge
boa                       0.16.0             pyhd8ed1ab_1    conda-forge
[...]

PR #2314, opened 2024-04-10
Move the entry_points definitions into setup.py

### LINKED ISSUE #2294: Define entrypoints in setup.py instead of meta.yaml
Right now all entrypoints except for msui are declared in meta.yaml only. This has the disadvantage that they are bound to a single package definition for a single package manager. If we moved the entrypoints to setup.py any somewhat decent package manager would pick them up automatically (e.g. nix or pip). They could be dropped from the mss-feedstock as well then, I'd assume.

matrss:
Fixes #2294.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -52,6 +52,13 @@ setup(
     zip_safe=False,
     install_requires=[],  # we use conda build recipe
     entry_points=dict(
-        console_scripts=['msui = mslib.msui.msui:main'],
+        console_scripts=[
+            "mscolab = mslib.mscolab.mscolab:main",
+            "mss = mslib.msui.mss:main",
+            "mssautoplot = mslib.utils.mssautoplot:main",
+            "msui = mslib.msui.msui:main",
+            "mswms = mslib.mswms.mswms:main",
+            "mswms_demodata = mslib.mswms.demodata:main",
+        ],
     ),
 )
--- a/localbuild/meta.yaml
+++ b/localbuild/meta.yaml
@@ -12,13 +12,6 @@ build:
   skip: true  # [py<38]
   number: 1000
   script: "{{ PYTHON }} -m pip install . --no-deps -vv"  # [not win]
-  entry_points:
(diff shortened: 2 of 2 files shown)


setup.py
```diff
@@ -52,6 +52,15 @@
     zip_safe=False,
     install_requires=[],  # we use conda build recipe
     entry_points=dict(
-        console_scripts=['msui = mslib.msui.msui:main'],
+        console_scripts=[
+            "mscolab = mslib.mscolab.mscolab:main",
+            "mswms = mslib.mswms.mswms:main",
+            "mswms_demodata = mslib.mswms.demodata:main",
+        ],
+        gui_scripts=[
```
ReimarBauer: https://packaging.python.org/en/latest/specifications/entry-points/

The difference between console_scripts and gui_scripts only affects Windows systems. console_scripts are wrapped in a console executable, so they are attached to a console and can use [sys.stdin](https://docs.python.org/3/library/sys.html#sys.stdin), [sys.stdout](https://docs.python.org/3/library/sys.html#sys.stdout) and [sys.stderr](https://docs.python.org/3/library/sys.html#sys.stderr) for input and output. gui_scripts are wrapped in a GUI executable, so they can be started without a console, but cannot use standard streams unless application code redirects them. Other platforms do not have the same distinction.

We maybe should not do that extra for windows. Also we have some argparse arguments for msui  and click on mssautoplot.

There is also a windows problem which needs on some configurations a terminal start from msui.

ReimarBauer: https://github.com/Open-MSS/MSS/issues/2261

matrss: > We maybe should not do that extra for windows. Also we have some argparse arguments for msui and click on mssautoplot.

Commandline arguments are unrelated to stdin/stdout/stderr so AFAIU they should work perfectly fine with gui_scripts as well.

But I can move them all into console_scripts as well. EDIT: done.

ReimarBauer (approved):
Thx

PR #2316, opened 2024-04-10
renamed time to time_handling, don't shadow time

### LINKED ISSUE #2312: builtin checker findings for mslib/plugins/io/csv.py and mslib/utils/time.py
@matrss seems the builtin checker was updated?

I rather don't want to rename the csv io plugin, the time utils module can maybe refactored renamed to times (just plural?)

@joernu76 What do you think? 

I assume we need to solve this in stable.

```
Successfully installed flake8-7.0.0 flake8-builtins-2.4.0 mccabe-0.7.0 pycodestyle-2.11.1 pyflakes-3.2.0
mslib/plugins/io/csv.py:0:1: A005 the module is shadowing a Python builtin module "csv"
mslib/utils/time.py:0:1: A005 the module is shadowing a Python builtin module "time"
2     A005 the module is shadowing a Python builtin module "csv"
2
```

ReimarBauer:
**Purpose of PR?**:

Fixes 2312

renames module time to time_handling

skip the io plugins until refactoring for the flake8 builtin test


joernu76 (commented):
I disagree with the intent of the PR and the change itself.
The new name is not more expressive but simply longer.

ReimarBauer:
> I disagree with the intent of the PR and the change itself. The new name is not more expressive but simply longer.

@joernu76 what is your suggestion related to the flake8 checker? exclude it from the list?

ReimarBauer:
differently solved

PR #2317, opened 2024-04-11
fix: Remove A005 from flake8 checks

### LINKED ISSUE #2312: builtin checker findings for mslib/plugins/io/csv.py and mslib/utils/time.py

joernu76:
This test does not provide sensible warnings, currently.

**Purpose of PR?**:

Fixes #2312 

no
-

**Checklist:**
- [ ] Commit has unit tests

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -37,6 +37,6 @@ omit =
 norecursedirs = .git .idea .cache
 
 [flake8]
-ignore = E124,E125,E402,W504
+ignore = E124,E125,E402,W504,A005
 max-line-length = 120
 exclude = mslib/msui/qt5/*.py, mslib/mscolab/migrations/*.py


matrss:
I think this should go to develop, since we don't even have the flake8-builtins checks on stable. If it is really wanted on stable then the change would also need to go into the tool:pytest section of setup.cfg. On develop this duplication is gone already.

joernu76:
So to develop it goes.

In the flake8 test, I do not see either of the second or third command being output by github. i.e. the pip --upgrade pip command is shown, but neither the installation nor execution of flake8.

I presume flake8 is executed with the "0" as its output?

matrss:
> In the flake8 test, I do not see either of the second or third command being output by github. i.e. the pip --upgrade pip command is shown, but neither the installation nor execution of flake8.

pip's output from installing flake8 and flake8-builtins is shown (from this line onwards: https://github.com/Open-MSS/MSS/actions/runs/8650673932/job/23719710583?pr=2317#step:4:25).

> I presume flake8 is executed with the "0" as its output?

Yes, usually flake8 wouldn't print anything when there is no warning, but due to either the --count or the --statistics option it prints the total number of warnings (which is 0 in this case).

matrss (approved):

joernu76:
> > In the flake8 test, I do not see either of the second or third command being output by github. i.e. the pip --upgrade pip command is shown, but neither the installation nor execution of flake8.
> 
> pip's output from installing flake8 and flake8-builtins is shown (from this line onwards: https://github.com/Open-MSS/MSS/actions/runs/8650673932/job/23719710583?pr=2317#step:4:25).
> 
> > I presume flake8 is executed with the "0" as its output?
> 
> Yes, usually flake8 wouldn't print anything when there is no warning, but due to either the --count or the --statistics option it prints the total number of warnings (which is 0 in this case).

Ah, stupid me hasn't seen the ">" button, which hides all but the first command. Thanks.

Issue #2302, opened 2024-03-26
define a range of a "green" coverage


ReimarBauer:
Hey, I want to reach at least 70 % by coverage again,

we maybe should adjust a bit that not only a few lines below this goal the coverage test shows the CI has a failure.

What would be a good value for https://coverage.readthedocs.io/en/7.4.4/config.html#report-fail-under

matrss:
I don't think the setting you have referenced applies for coveralls, right? I.e. it is for coverage.py and would make the pytest run fail if coverage is below this value, but it would not affect what the status check from coveralls shows (which is green if coverage stayed the same or increased, and is red if coverage decreased, compared to a previous run), right?

I would prefer to stay with what coveralls already does, but fix our test suite to actually have a deterministic coverage (unlike what we have right now, where there is a handful of lines that may or may not be covered at random, for some reason).

ReimarBauer:
oh, right, we moved from there. It only illustrates what a solution can be.

I dislike that it currently looks like the tests have failed but it is only less line covered. 

But at all for much new lines I think it's a good idea to remember that with a merge we end up there again.

matrss:
The proper fix is to get the test suites coverage to be deterministic. Do you have any idea why it is not? I can only guess that it depends on what Qt has done already at the end of a test, so maybe a qWait at the end of every test would pave this over, _for now_. I would have to test that out.

ReimarBauer:
@matrss I think we can close this, a follow up may also get a different description, as my initial one.

Issue #2193, opened 2024-02-09
Fail tests if there are any Qt widgets left open


matrss:
Right now tests may leave widgets open. This can affect other tests. Not closing all widgets should instead be a test failure and any instance where this happens needs to be fixed.

~This is probably a blocker for #2192.~ It wasn't.

ReimarBauer:
I move this to the next major

PR #2318, opened 2024-04-12
Partially fix non-deterministic coverage

matrss:
This is a band-aid fix that mostly resolves the issue with non-deterministic coverage. Without the wait, tests abruptly end and coverage can fluctuate. With the wait, Qt has time to handle its stuff for a while longer, which stabilizes the coverage. The proper fix would be to clean up everything Qt-related that is created in a test after each test, but I couldn't get that to work yet.

There are also at least two more cases of non-deterministic coverage that seem to be unrelated to Qt and which aren't fixed with this.

Related to #2302.

### DIFF
--- a/tests/_test_msui/test_msui.py
+++ b/tests/_test_msui/test_msui.py
@@ -67,7 +67,7 @@ def test_main():
 
 class Test_MSS_TutorialMode:
     @pytest.fixture(autouse=True)
-    def setup(self, qapp):
+    def setup(self, qtbot, qapp):
         qapp.setApplicationDisplayName("MSUI")
         self.main_window = msui_mw.MSUIMainWindow(tutorial_mode=True)
         self.main_window.create_new_flight_track()
@@ -95,7 +95,7 @@ class Test_MSS_TutorialMode:
 
 class Test_MSS_AboutDialog:
     @pytest.fixture(autouse=True)
-    def setup(self, qapp):
+    def setup(self, qtbot):
         self.window = msui_mw.MSUI_AboutDialog()
         yield
         self.window.hide()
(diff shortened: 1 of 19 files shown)


matrss:
This PR should at least improve the situation with regard to the fluctuating coverage reports, but unfortunately it slows down the test suite a bit. A proper fix would probably include solving #2193, which lead to segfaults in my attempts so far.

matrss:
I've ran the test suite 10 times and let it report coverage to coveralls. As you can see here: https://coveralls.io/builds/66882529 between the 10 reported jobs 8 had the same coverage while 2 had something different. This is an improvement over what we have on develop, where basically every run reports something different.

tests/_test_msui/test_msui.py
```diff
@@ -67,7 +67,7 @@ def test_main():
 
 class Test_MSS_TutorialMode:
     @pytest.fixture(autouse=True)
-    def setup(self, qapp):
+    def setup(self, qtbot, qapp):
```
joernu76, line 70: there is no qapp fixture any more...?

joernu76:
we seem to redefine some qapp/qtbot fixtures defined somewhere else and replace the qapp with the qtbot one? Could we add a sentence to the fixture declaration on what is going on here?
It is really not apparent to...

tests/_test_msui/test_msui.py
matrss: The qapp and qtbot fixtures come from pytest-qt. I've previously overwritten the qapp fixture to add the additional checks for lingering message boxes and the likes, but needed to switch to overwriting qtbot to get access to the wait method.

The line here still needs qapp for `setApplicationDisplayName`, and has to request qtbot as well now, to also get the additional checks. In the future all tests should use qtbot's addWidget to do their cleanup properly, so all Qt tests will need to request qtbot anyway.

matrss:
Yes, it is not so obvious... I'll add a comment that the qtbot fixture just re-defines the qtbot fixture from pytest-qt with additional checks done after each test.

ReimarBauer (approved):
Let's do that, it's an intermediate step 
that makes it easier for us to evaluate other PRs

PR #2326, opened 2024-04-17
active_op_id needs to be shared

### LINKED ISSUE #2322: active_op_id needs a better implementation in topview (mscolab)
When you start with a topview and using mscolab you can set self.active_op_id by clicking on an operation but a second topview do start with None and not that value. You have to "activate" again.

The first activate seems also not to change that value to the operation id and keeps it None, it needs then a change to another operation.

ReimarBauer:
store widget data corresponding to docks

**Purpose of PR?**:

Fixes #2322
including tests

### DIFF
--- a/mslib/msui/topview.py
+++ b/mslib/msui/topview.py
@@ -211,9 +211,6 @@ class MSUITopViewWindow(MSUIMplViewWindow, ui.Ui_TopViewWindow):
         # Dock windows [WMS, Satellite, Trajectories, Remote Sensing, KML Overlay, Multiple Flightpath]:
         self.docks = [None, None, None, None, None, None]
 
-        # Corresponding widget objects
-        self.widgets = [None, None, None, None, None, None]
-
         # Initialise the GUI elements (map view, items of combo boxes etc.).
         self.setup_top_view()
 
--- a/mslib/msui/topview.py
+++ b/mslib/msui/topview.py
@@ -211,6 +211,9 @@ class MSUITopViewWindow(MSUIMplViewWindow, ui.Ui_TopViewWindow):
         # Dock windows [WMS, Satellite, Trajectories, Remote Sensing, KML Overlay, Multiple Flightpath]:
         self.docks = [None, None, None, None, None, None]
 
+        # Corresponding widget objects
+        self.widgets = [None, None, None, None, None, None]
+
         # Initialise the GUI elements (map view, items of combo boxes etc.).
         self.setup_top_view()
 
@@ -221,7 +224,7 @@ class MSUITopViewWindow(MSUIMplViewWindow, ui.Ui_TopViewWindow):
         self.active_flighttrack = active_flighttrack
 
         # Stores active mscolab operation id
-        self.active_op_id = None
+        self.active_op_id = mainwindow.mscolab.active_op_id
 
         # Mscolab Server Url and token
         self.mscolab_server_url = mscolab_server_url
@@ -356,6 +359,7 @@ class MSUITopViewWindow(MSUIMplViewWindow, ui.Ui_TopViewWindow):
                                                             listOperationsMSC=self.mainwindow_listOperationsMSC,
                                                             category=self.mainwindow_filterCategoryCb,
                                                             activeFlightTrack=self.active_flighttrack,
+                                                            active_op_id=self.active_op_id,
                                                             mscolab_server_url=self.mscolab_server_url,
                                                             token=self.token)
 
--- a/mslib/msui/viewwindows.py
+++ b/mslib/msui/viewwindows.py
@@ -56,8 +56,6 @@ class MSUIViewWindow(QtWidgets.QMainWindow):
         # in proper size in derived classes!
         self.docks = []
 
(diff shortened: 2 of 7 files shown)


mslib/msui/viewwindows.py
```diff
@@ -140,6 +142,8 @@ def createDockWidget(self, index, title, widget):
         # setWidget transfers the widget's ownership to Qt -- no setParent()
         # call is necessary:
         self.docks[index].setWidget(widget)
+        # store data of widget
+        self.widgets[index] = widget
```
ReimarBauer: I've not seen that this can be accessed anywhere else. This enables now also other tests for the docking widgets.

We have currently not a case where we need to access widget data from a different widget, maybe that is the reason it was missing?

mslib/msui/topview.py
```diff
@@ -211,6 +211,9 @@ def __init__(self, parent=None, mainwindow=None, model=None, _id=None,
         # Dock windows [WMS, Satellite, Trajectories, Remote Sensing, KML Overlay, Multiple Flightpath]:
         self.docks = [None, None, None, None, None, None]
 
+        # Corresponding widget objects
+        self.widgets = [None, None, None, None, None, None]
+
```
matrss: If I understand this right there is always a maximum of (in this case) 6 different docks/widgets that can either not exist (None) or exist (some object). Each index corresponds to a specific type of dock/widget. Is that correct?

If yes, why aren't these just 6 different named variables instead?

mslib/msui/topview.py
```diff
@@ -221,7 +224,7 @@ def __init__(self, parent=None, mainwindow=None, model=None, _id=None,
         self.active_flighttrack = active_flighttrack
 
         # Stores active mscolab operation id
-        self.active_op_id = None
+        self.active_op_id = mainwindow.mscolab.active_op_id
```
matrss, line 224: In which situation should the topview's active_op_id be something different than the mainwindow.mscolab.active_op_id? I.e. why does the topview need a copy of this attribute, instead of always using the one from the mainwindow?

mslib/msui/topview.py
ReimarBauer: This static implementation should become removed when these widgets gets plugable. At least I want to have the option that we not have to add any docking widget with all its dependencies to the core code.
It could become a list where we could append, maybe the docks and the widgets can become a tuple or a dict.

mslib/msui/topview.py
ReimarBauer: topview won't need it. The current implementation of the multiple_flighpath_dockingwidget has no access to the mainwindow, e.g.
https://github.com/Open-MSS/MSS/blob/develop/mslib/msui/multiple_flightpath_dockwidget.py#L507
There were more vars introduced which are already known in the mainwindow  https://github.com/Open-MSS/MSS/blob/develop/mslib/msui/topview.py#L204 


Currently I do prefer first a kind of logic test coverage before we refactor the inner.  It is somehow improving the Proof-Of-Concept before we can do the cleanup.

One of the bigger refactoring steps is to remove all the dups and focus on a better understandable API.

matrss: Couldn't the topview just hold a reference to the mainwindow then? Holding another reference to all of the mainwindow's signals is also something that feels wrong to me, and could be eliminated if the mainwindow was just accessible through the topview.

ReimarBauer: I think that would be a better solution and has to be done in a refactoring. @Jatin2020-24 did you have a reason not to do it that way?

This PR is related  to fix a bug in the order for opening the dockingwidget when a mscolab operation was selected.  We need a few more tests to figure that any combination for the docking widget works. 

I guess there is at least one more bug which needs to be test covered and fixed. 

When this part is  refactored we also have to refactor tests.

mslib/msui/viewwindows.py
matrss: Wouldn't `self.docks[n].widget() == self.widgets[n]`? (https://doc.qt.io/qt-6/qdockwidget.html#widget)

I.e. can't you just get the widget out of the dock?

ReimarBauer: Good idea, also in QT-5 available.  That simplifies it a lot.

ReimarBauer:
@matrss something more missing or should be improved?

matrss:
I still think we shouldn't duplicate this attribute everywhere. What is wrong with exposing a reference to the mainwindow on the topview? Then it would be possible for the MultipleFlightpathControlWidget to just access `self.ui.mainwindow.active_op_id`, or something like that.

matrss:
Alternatively, make the topview's active_op_id a getter/`@property` for the one from the mainwindow so that it can never be out of sync. Depends on how much of the mainwindow we want to have accessible everywhere.

matrss:
@ReimarBauer and I talked about this a bit this morning and we came to the conclusion that the PR should be merged as-is, but we need to take a look at things like this in the future and make an effort to refactor these things and clean up some of it.

matrss (approved):

PR #2334, opened 2024-04-22
Normalize files

### LINKED ISSUE #2330: Add lint rules to prevent CRLF and whitespace errors
Do we want to introduce a linter for these kinds of things? We could setup a https://editorconfig.org/ configuration for the repository and use https://github.com/editorconfig-checker/editorconfig-checker to enforce it.

_Originally posted by @matrss in https://github.com/Open-MSS/MSS/issues/2328#issuecomment-2063820151_
            
With such a setup we could enforce unix-style linebreaks, linebreaks at EOF, and other stuff.

matrss:
Add lint rules to avoid having CRLF endings in the repository and to prevent trailing whitespace at EOL and duplicate newlines at EOF.

Fixes #2330.


matrss:
For a diff of what meaningfully changed you can do `git diff stable --ignore-all-space --ignore-blank-lines` in a local checkout of this branch.

.github/workflows/lint.yml
```diff
@@ -0,0 +1,57 @@
+name: lint
+
+on:
+  push:
```
ReimarBauer, line 4: should we add here also new GSOC branches?

.github/workflows/lint.yml
```diff
@@ -0,0 +1,57 @@
...
+jobs:
+  flake8:
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    steps:
+    - uses: actions/checkout@v3
+    - name: Set up Python 3.8
+      uses: actions/setup-python@v3
+      with:
+        python-version: 3.8
```
ReimarBauer, line 22: pypi shows a requiremwent
 Requires: Python >=3.8.1

maybe we should increase this here to 3.9 ?

.github/workflows/lint.yml
```diff
@@ -0,0 +1,57 @@
...
+    branches:
+    - develop
+    - stable
+
+jobs:
+  flake8:
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    steps:
+    - uses: actions/checkout@v3
```
ReimarBauer, line 18: v4?

.github/workflows/lint.yml
```diff
@@ -0,0 +1,57 @@
...
+    - name: Set up Python 3.8
+      uses: actions/setup-python@v3
+      with:
```
ReimarBauer, line 21: v5?

.gitattributes
```diff
@@ -0,0 +1 @@
+* text=auto
```
ReimarBauer, line 1: We maybe should exclude png and nc files and other binaries from text 

https://github.com/desktop/desktop/issues/5323#issuecomment-410220449

We don't have  text eol=lf but it maybe useful it did not lookup binaries.

ReimarBauer:
sorry for the delay, had also an idea for another check, see https://github.com/Open-MSS/MSS/issues/2338

I added few remarks and questions.

.github/workflows/lint.yml
matrss: Yes. This is already the case for the python-flake8 workflow on develop, but since this is on stable I missed it. Again, the issue of having two development branches. ~I'll fix this up.~ EDIT: actually, this is something that needs to be fixed when merging stable into develop. flake8-builtins is also missing here and needs to be considered when merging.

.github/workflows/lint.yml
matrss: Depends on what minimum version of python we want to support for MSS.

.github/workflows/lint.yml
matrss: Right, this is only updated on develop so far. I don't want to manually update all of these, but we could configure dependabot to also send updates for the stable branch, if you want.

.github/workflows/lint.yml
matrss: See: https://github.com/Open-MSS/MSS/pull/2334#discussion_r1582626611

.gitattributes
matrss: `* text=auto` tells git to auto-detect if a file is text or binary, for all files in the repository. Does it wrongly detect a png or nc as text somewhere?

The issue you linked is solely caused by `* text`, which tells git that it should consider all files to be text files.

.github/workflows/lint.yml
ReimarBauer: ok on the next merge of develop to stable we have a new major release and also both branches have a similiar test infrastructure.

And yes stable is the branch we do release from. We have a major develop branch and sometimes based on that other feature branches. That is no issue.

.github/workflows/lint.yml
ReimarBauer: until 3.13  gets on conda it is 3.8 usually we have 5 versions

.github/workflows/lint.yml
ReimarBauer: I mentioned it because of the new lines.
Yes adding the bot for stable is a good thing. A stable release seems is renewed less frequently than the bot architecture.

.gitattributes
ReimarBauer: I tried in the normalize-files branch

```
files_with_crlf="$(git ls-files --eol)"
echo "$files_with_crlf"
```
current .gitattributes
```
i/-text w/-text attr/text=auto          tutorials/pictures/wms/win/selecttoopencontrol.png
i/-text w/-text attr/text=auto          tutorials/pictures/wms/win/star_filter.png
```

with *.png -text
```
i/-text w/-text attr/-text              tutorials/pictures/wms/win/selecttoopencontrol.png
i/-text w/-text attr/-text              tutorials/pictures/wms/win/star_filter.png
```
This changes the eloattr

.github/workflows/lint.yml
matrss: > ok on the next merge of develop to stable we have a new major release and also both branches have a similiar test infrastructure.

I am not sure if you misunderstood me. I was talking about the next merge from stable to develop, i.e. what should happen right after this PR has been merged, not on the next merge from develop to stable (a major release). These differences between the python-flake8.yml workflow on stable and the python-flake8.yml workflow on develop need to be accounted for in the merge of stable into develop.

.github/workflows/lint.yml
matrss: Then it should remain 3.8 to lint with the lowest common denominator of python, I think.

.github/workflows/lint.yml
matrss: I filed #2339.

.github/workflows/lint.yml
matrss: See #2339.

.gitattributes
matrss: Yes, since that is what .gitattributes does, changing the files attributes. `* text=auto` tells git to auto-detect if a file is text or binary and as seen in `i/-text w/-text` it successfully did so. With `text=auto` it detected those png files as "not text" (i.e. binary) in the index and in the working tree.

ReimarBauer (approved):
LGTM

ReimarBauer:
need to change rules. The flake8 mandatory check job was named lint only.

PR #2335, opened 2024-04-22
Normalize line endings

matrss:
Introduce a lint step that makes sure no CRLF ends up in the git repository.

Fixes part of #2330.


matrss:
Actually, I'll close this again, in favor of #2334, which should address the entire issue in one go.

PR #2341, opened 2024-05-06
fix: exception improved

### LINKED ISSUE #2274: Out of Scope Variable in Exception Block in constants.py
The variable `_fs` is out of scope in the except block as it was defined in the try block.
https://github.com/Open-MSS/MSS/blob/1d0a5ad604170f3a618c54b69b1ad949bba2ea91/mslib/msui/constants.py#L37-L43

ReimarBauer:
**Purpose of PR?**:

Fixes #2274

behaviour can be checked by 

```
import logging
import fs

def test_fs_path_not_existing():
    test_path = ("ftp://ftp.de.debian.org/debian/.config/msui", "foo://ftp.de.debian.org/debian/.config/msui")
    for fsurl in test_path:
        try:
            _fs = fs.open_fs(fsurl)
            _fs.makedirs(fsurl)
        except fs.errors.CreateFailed:
            logging.error('FS url "%s" create dir nor supported', fsurl)
            assert "_fs" not in locals()

        except fs.opener.errors.UnsupportedProtocol:
            logging.error('FS url "%s" not supported', fsurl)
            assert "_fs" not in locals()
```

**Checklist:**
- [X] Bug fix. Fixes #2274
- [ ] Commit has unit tests

### DIFF
--- a/mslib/msui/constants.py
+++ b/mslib/msui/constants.py
@@ -37,8 +37,9 @@ MSUI_CONFIG_PATH = os.getenv("MSUI_CONFIG_PATH", os.path.join(HOME, ".config", "
 if '://' in MSUI_CONFIG_PATH:
     try:
         _fs = fs.open_fs(MSUI_CONFIG_PATH)
-    except fs.errors.CreateFailed:
         _fs.makedirs(MSUI_CONFIG_PATH)
+    except fs.errors.CreateFailed:
+        logging.error('FS url "%s" create dir nor supported', MSUI_CONFIG_PATH)
     except fs.opener.errors.UnsupportedProtocol:
         logging.error('FS url "%s" not supported', MSUI_CONFIG_PATH)
 else:


mslib/msui/constants.py
```diff
@@ -37,8 +37,9 @@
 if '://' in MSUI_CONFIG_PATH:
     try:
         _fs = fs.open_fs(MSUI_CONFIG_PATH)
-    except fs.errors.CreateFailed:
         _fs.makedirs(MSUI_CONFIG_PATH)
+    except fs.errors.CreateFailed:
+        logging.error('FS url "%s" create dir nor supported', MSUI_CONFIG_PATH)
```
matrss, line 42: The logging message does not describe the same error case as `fs.errors.CreateFailed`. The error is described here: https://docs.pyfilesystem.org/en/latest/reference/errors.html#fs.errors.CreateFailed AFAICT it can be emitted when `fs.open_fs` fails for some reason and is not necessarily related to if the FS supports creating directories.

mslib/msui/constants.py
```diff
@@ -37,8 +37,9 @@
 if '://' in MSUI_CONFIG_PATH:
     try:
         _fs = fs.open_fs(MSUI_CONFIG_PATH)
-    except fs.errors.CreateFailed:
         _fs.makedirs(MSUI_CONFIG_PATH)
```
matrss, line 40: This would create a directory tree like MSUI_CONFIG_PATH inside of MSUI_CONFIG_PATH, i.e.:
```
$ tree test-fs
test-fs

0 directories, 0 files
$ python3
Python 3.11.7 | packaged by conda-forge | (main, Dec 23 2023, 14:43:09) [GCC 12.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import fs
>>> a = fs.open_fs("osfs://~/Playground/test-fs")
>>> b = a.makedirs("osfs://~/Playground/test-fs")
>>> 
$ tree test-fs
test-fs
`-- osfs:
    `-- ~
        `-- Playground
            `-- test-fs

5 directories, 0 files
```

I don't think that is what is intended to happen here. Instead, I think the wanted result is that of `fs.open_fs(MSUI_CONFIG_PATH, create=True)`.

And I think ~the entire `try:` block~ the entire `if:` block can be replaced with just that (i.e. if there is an exception it should crash MSS and not just be logged as an error, since I'd assume there is no reasonable path forward if the config directory cannot be used).

matrss (changes requested):

mslib/msui/constants.py
matrss: I think this can be radically simplified, I'll open a PR to demonstrate.

PR #2345, opened 2024-05-08
Use isoformat for mscolab timestamps

### LINKED ISSUE #2321: review and refactor usage of strptime to an iso time format (mscolab)
This is an unify task to change in several mscolab related methods somthing similar of `datetime.datetime.strptime(timestamp, "%Y-%m-%d, %H:%M:%S.%f %z") ` to an isoformat string

- https://docs.python.org/3/library/datetime.html#datetime.datetime.isoformat, - https://docs.python.org/3/library/datetime.html#datetime.datetime.fromisoformat.

if there are changes in the wms part possible, this should get a `# ToDo ...` comment

ReimarBauer:
**Purpose of PR?**:

Fixes #2321 

time format change for mscolab

### DIFF
--- a/mslib/mscolab/file_manager.py
+++ b/mslib/mscolab/file_manager.py
@@ -31,7 +31,6 @@ import logging
 import git
 import threading
 from sqlalchemy.exc import IntegrityError
-from mslib.mscolab.utils import created_at_isoformat
 from mslib.mscolab.models import db, Operation, Permission, User, Change, Message
 from mslib.mscolab.conf import mscolab_settings
 
@@ -400,7 +399,7 @@ class FileManager:
             'comment': change.comment,
             'version_name': change.version_name,
             'username': change.user.username,
-            'created_at': created_at_isoformat(change.created_at)
+            'created_at': change.created_at.isoformat()
         }, changes))
 
     def get_change_content(self, ch_id, user):
--- a/mslib/mscolab/file_manager.py
+++ b/mslib/mscolab/file_manager.py
@@ -31,6 +31,7 @@ import logging
 import git
 import threading
 from sqlalchemy.exc import IntegrityError
+from mslib.mscolab.utils import created_at_isoformat
 from mslib.mscolab.models import db, Operation, Permission, User, Change, Message
 from mslib.mscolab.conf import mscolab_settings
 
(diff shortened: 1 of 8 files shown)


matrss (changes requested):
I think this one needs to be changed as well: https://github.com/Open-MSS/MSS/blob/3ed347b43ed304da287e556b61c85de37a25a94d/mslib/mscolab/utils.py#L58

mslib/mscolab/file_manager.py
```diff
@@ -399,7 +400,7 @@ def get_all_changes(self, op_id, user, named_version=False):
             'comment': change.comment,
             'version_name': change.version_name,
             'username': change.user.username,
-            'created_at': change.created_at.strftime("%Y-%m-%d, %H:%M:%S.%f %z")
+            'created_at': created_at_isoformat(change.created_at)
```
matrss: What is wrong with `change.created_at.isoformat()`?

ReimarBauer: seems I looked only on that with the None state.

matrss: What None state?

ReimarBauer: on a test: test_get_message_dict

The function strftime returns None when created_at is initialized. But isoformat crashes.
But in that case isoformat gives an AttributeError

and that is because I need to change the test too.

tests/_test_mscolab/test_utils.py
```diff
@@ -31,7 +31,9 @@
 from mslib.mscolab.conf import mscolab_settings
 from mslib.mscolab.models import Operation, MessageType
 from mslib.mscolab.seed import add_user, get_user
-from mslib.mscolab.utils import get_recent_op_id, get_session_id, get_message_dict, create_files, os_fs_create_dir
+from mslib.mscolab.utils import (get_recent_op_id, get_session_id,
+                                 get_message_dict, create_files,
+                                 os_fs_create_dir)
 
 
 class Message:
```
matrss, line 39: I would like to get rid of this entire Message class. It is a very weird usage of a class (to kind of emulate an object of the actual Message class) and is not needed. The one instance where it is used could be replaced by an actual object of the actual class.

matrss: This is unrelated to the topic of this PR, I opened #2353 for this.

matrss (approved):

PR #2346, opened 2024-05-08
ARM-OSX installation documented

### LINKED ISSUE #2130: state of  osx-arm64
We got a request by a participant of the CAESAR campaign, Feb 20-April 7, if MSS can be installed on an M2 apple.

Currently this is not possible. I suggested to use our docker image, https://mss.readthedocs.io/en/stable/installation.html#docker-instance

Recently basemap  with 1.3.9 was added for osx-arm64. 

We have to review what blocks us to use a ARM OSX Migrator for our package, similiar to https://github.com/conda-forge/basemap-feedstock/pull/123

ReimarBauer:
**Purpose of PR?**:

Fixes #2130

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -51,6 +51,8 @@ As **Beginner** start with an installation of Miniforge3
 
 If you use already Mambaforge please read the `FAQ <https://github.com/conda-forge/miniforge#faq>`__
 
+
+
 Install MSS
 ~~~~~~~~~~~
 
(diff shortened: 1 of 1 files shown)


joernu76 (approved):

docs/installation.rst
```diff
@@ -63,6 +65,16 @@ versions for dependencies. ::
     (mssenv) $ msui
 
 
+.. hint::
+   For installing MSS on the ARM-MAC Apple devices because not all dependencies
+   are crosscompiled you need to set an additional subdir. ::
```
matrss, line 70: This is not required for develop (at least in CI we don't have to set this). Is is it required for stable? What dependency is not available on ARM?

ReimarBauer: because we use a more modern pyqt version in develop

matrss: Ah, OK, got it, `pyqt==5.12` doesn't exist for osx-arm64.

Issue #2148, opened 2024-01-11
The test suite leaves a lot of trash in /tmp and CWD


matrss:
The test suite is creating temporary directories that it does not cleanup afterwards. I am not yet sure where they are coming from or if this is only a side-effect of when the test suite crashes.

matrss:
After cleaning up, a single successful test suite run on the latest develop branch left this in /tmp:
```
$ ls -la /tmp/tmp*
/tmp/tmp16_y1fyq:
total 144
drwxrwxr-x  3 icg149 icg149   4096 Feb 16 12:30 .
drwxrwxrwt 36 root   root   135168 Feb 16 12:31 ..
drwxrwxr-x  2 icg149 icg149   4096 Feb 16 12:30 msui_wms_cache

/tmp/tmpiy183g69:
total 144
drwxrwxr-x  3 icg149 icg149   4096 Feb 16 12:30 .
drwxrwxrwt 36 root   root   135168 Feb 16 12:31 ..
drwxrwxr-x  2 icg149 icg149   4096 Feb 16 12:30 msui_wms_cache

/tmp/tmplv6_vzxc:
total 144
drwxrwxr-x  3 icg149 icg149   4096 Feb 16 12:29 .
drwxrwxrwt 36 root   root   135168 Feb 16 12:31 ..
drwxrwxr-x  2 icg149 icg149   4096 Feb 16 12:30 msui_wms_cache

/tmp/tmpxi8oehy0:
total 144
drwxrwxr-x  3 icg149 icg149   4096 Feb 16 12:31 .
drwxrwxrwt 36 root   root   135168 Feb 16 12:31 ..
drwxrwxr-x  2 icg149 icg149   4096 Feb 16 12:31 msui_wms_cache
```

It also created an osfs: directory in the CWD, for some reason:
```
$ ls -R osfs\:/
'osfs:/':
tmp

'osfs:/tmp':
tmpkb38g7u6msui167af85de1  tmpzsgyrmi7msui167af85de1

'osfs:/tmp/tmpkb38g7u6msui167af85de1':
msui-core.conf

'osfs:/tmp/tmpzsgyrmi7msui167af85de1':
msui-core.conf
```

ReimarBauer:
This happens when os.path is used but not getsyspath from the fs object https://docs.pyfilesystem.org/en/latest/concepts.html#system-paths

ReimarBauer:
> It also created an osfs: directory in the CWD, for some reason:

[...]

Issue #2103, opened 2023-11-19
Pyfilesystem Discussion: solve fs.path / os.path mismatch keep or deprecate fs


ReimarBauer:
on server we should have os.path. 
on ui we should have fs.path

fs.path means we can use pyfilesystem urls e.g. webdavfs or sshfs urls mapped to a "fs"path. 
That means our msui_settings.json can come from a nextcloud user dir and we can save images to an fs url too.

os.path is only capable to store on local or mounted filesystems.

if ui needs an os path then it gets converted by .getsyspath()

ReimarBauer:
Before working on this we should talk about if we want to keep Pyfilesystem2 in our lib or replace it completly  by os / pathlib.

When we introduced FS we had anything using it. Over the years it gots a mixup. At the time we introduced it,  there was no pathlib and we also had problems with tests on windows because of path. The idea that we use it with FS pathes seems not become the standard. We can use it with our nextcloud but we don't. etc.
Also with mscolab we can do things differently, e.g. exchange images, or config over chat. 

Let us collect pro's and con's

ReimarBauer:
related
https://github.com/Open-MSS/MSS/issues/2347

ReimarBauer:
https://github.com/Open-MSS/MSS/pull/2405/files#r1665854420

when we want to keep fs in the future we can't use send_from_directory. We have to replace it by send_file because this can also send a BytesIO object 

But this also needs an intermediate step, reading the fs data into a BytesIO object.

matrss:
If we want to keep using fs and use it properly we can't use _any_ function that expects a path-like object. `send_from_directory` is just one of them. This stops us from e.g. using any path sanitization features that flask does automatically, among other things.

What is the use case for fs, at least on the server side, anyway?

> But this also needs an intermediate step, reading the fs data into a BytesIO object.

Not really, what `fs.open` returns is a file-like object which should already be useable as a BytesIO-like object (if opened in bytes mode). So it would just need a `with:` statement opening the file, before sending it. Again, path sanitization would be on us, then.

ReimarBauer:
> What is the use case for fs, at least on the server side, anyway?

In the past there was no file storage on Google Cloud Platform.
It is an option to deploy on such platforms and to manage the file store over the network.

ReimarBauer:
When I look on the Release history on PyPi, https://pypi.org/project/fs/#history

More than 2 years ago 2.4.16 was released. 
[...]

PR #2348, opened 2024-05-11
osfs names name fix for better identifying what it is

### LINKED ISSUE #2347: better naming of some os/fs constants in tests/constants
Because MSUI_CONFIG_PATH can be an os string or a fs_url we can name others for their function

- ending with _FS means pyfilesystem, _URL means PyFilesystem
- _FILE_PATH, _DIR means os module

ReimarBauer:
**Purpose of PR?**:

Fixes #2347

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -32,7 +32,6 @@ import sys
 sys.dont_write_bytecode = True
 
 import pytest
-import fs
 import shutil
 import keyring
 from mslib.mswms.demodata import DataFiles
@@ -77,22 +76,23 @@ def keyring_reset():
 def generate_initial_config():
     """Generate an initial state for the configuration directory in tests.constants.ROOT_FS
     """
-    if not constants.ROOT_FS.exists("msui/testdata"):
-        constants.ROOT_FS.makedirs("msui/testdata")
+    _path = os.path.join(constants.ROOT_DIR, "msui", "testdata")
+    if not os.path.exists(_path):
+        os.makedirs(_path)
 
     # make a copy for mscolab test, so that we read different pathes during parallel tests.
     sample_path = os.path.join(os.path.dirname(__file__), "tests", "data")
     shutil.copy(os.path.join(sample_path, "example.ftml"), constants.ROOT_DIR)
 
-    if not constants.SERVER_CONFIG_FS.exists(constants.SERVER_CONFIG_FILE):
+    if not constants.MSWMS_SERVER_CONFIG_FS.exists(constants.MSWMS_SERVER_CONFIG_FILE):
         print('\n configure testdata')
         # ToDo check pytest tmpdir_factory
         examples = DataFiles(data_fs=constants.DATA_FS,
-                             server_config_fs=constants.SERVER_CONFIG_FS)
+                             server_config_fs=constants.MSWMS_SERVER_CONFIG_FS)
         examples.create_server_config(detailed_information=True)
         examples.create_data()
 
-    if not constants.SERVER_CONFIG_FS.exists(constants.MSCOLAB_CONFIG_FILE):
+    if not constants.MSWMS_SERVER_CONFIG_FS.exists(constants.MSCOLAB_CONFIG_FILE):
         config_string = f'''
 # SQLALCHEMY_DB_URI = 'mysql://user:pass@127.0.0.1/mscolab'
 import os
@@ -103,9 +103,9 @@ from urllib.parse import urljoin
 
 ROOT_DIR = '{constants.ROOT_DIR}'
 # directory where mss output files are stored
-root_fs = fs.open_fs(ROOT_DIR)
-if not root_fs.exists('colabTestData'):
-    root_fs.makedir('colabTestData')
+with fs.open_fs(ROOT_DIR) as root_fs:
+    if not root_fs.exists('colabTestData'):
+        root_fs.makedir('colabTestData')
 BASE_DIR = ROOT_DIR
 DATA_DIR = fs.path.join(ROOT_DIR, 'colabTestData')
 # mscolab data directory
@@ -176,16 +176,14 @@ enable_basic_http_authentication = False
 # enable login by identity provider
 USE_SAML2 = False
 '''
-        ROOT_FS = fs.open_fs(constants.ROOT_DIR)
-        if not ROOT_FS.exists('mscolab'):
-            ROOT_FS.makedir('mscolab')
-        with fs.open_fs(fs.path.join(constants.ROOT_DIR, "mscolab")) as mscolab_fs:
+        if not os.path.exists(constants.MSCOLAB_CONFIG_PATH):
+            os.makedirs(constants.MSCOLAB_CONFIG_PATH)
+        with open(constants.MSCOLAB_CONFIG_FILE_PATH, 'w') as _f:
             # windows needs \\ or / but mixed is terrible. *nix needs /
-            mscolab_fs.writetext(constants.MSCOLAB_CONFIG_FILE, config_string.replace('\\', '/'))
-        path = fs.path.join(constants.ROOT_DIR, 'mscolab', constants.MSCOLAB_CONFIG_FILE)
-        parent_path = fs.path.join(constants.ROOT_DIR, 'mscolab')
+            _f.write(config_string.replace('\\', '/'))
 
-    if not constants.SERVER_CONFIG_FS.exists(constants.MSCOLAB_AUTH_FILE):
+    mscolab_auth_path_file = os.path.join(constants.MSCOLAB_CONFIG_PATH, constants.MSCOLAB_AUTH_FILE)
+    if not os.path.exists(mscolab_auth_path_file):
         config_string = '''
 import hashlib
 
@@ -193,12 +191,9 @@ class mscolab_auth:
      password = "testvaluepassword"
      allowed_users = [("user", hashlib.md5(password.encode('utf-8')).hexdigest())]
 '''
-        ROOT_FS = fs.open_fs(constants.ROOT_DIR)
-        if not ROOT_FS.exists('mscolab'):
-            ROOT_FS.makedir('mscolab')
-        with fs.open_fs(fs.path.join(constants.ROOT_DIR, "mscolab")) as mscolab_fs:
+        with open(mscolab_auth_path_file, 'w') as _f:
             # windows needs \\ or / but mixed is terrible. *nix needs /
-            mscolab_fs.writetext(constants.MSCOLAB_AUTH_FILE, config_string.replace('\\', '/'))
+            _f.write(config_string.replace('\\', '/'))
 
     def _load_module(module_name, path):
         spec = importlib.util.spec_from_file_location(module_name, path)
@@ -207,8 +202,8 @@ class mscolab_auth:
         spec.loader.exec_module(module)
 
 
-    _load_module("mswms_settings", constants.SERVER_CONFIG_FILE_PATH)
-    _load_module("mscolab_settings", path)
+    _load_module("mswms_settings", constants.MSWMS_CONFIG_FILE_PATH)
+    _load_module("mscolab_settings", constants.MSCOLAB_CONFIG_FILE_PATH)
 
 
 generate_initial_config()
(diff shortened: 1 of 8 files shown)


conftest.py
```diff
@@ -220,14 +215,16 @@ def _load_module(module_name, path):
 
 @pytest.fixture(autouse=True)
 def reset_config():
-    """Reset the configuration directory used in the tests (tests.constants.ROOT_FS) after every test
+    """Reset the configuration directory used in the tests (tests.constants.ROOT_FS) before every test
     """
     # Ideally this would just be constants.ROOT_FS.removetree("/"), but SQLAlchemy complains if the SQLite file is deleted.
+    # We have between the workers a shared constants.ROOT_DIR
```
ReimarBauer, line 221: related to https://github.com/Open-MSS/MSS/issues/2148

mslib/utils/config.py
```diff
@@ -500,9 +498,7 @@ def load_settings_qsettings(tag, default_settings=None, ignore_test=False):
         return default_settings
 
     settings = {}
-
-    q_settings = QtCore.QSettings(os.path.join(constants.MSUI_CONFIG_PATH, "msui-core.conf"),
-                                  QtCore.QSettings.IniFormat)
+    q_settings = QtCore.QSettings(constants.MSUI_CORE_FILE_PATH, QtCore.QSettings.IniFormat)
```
ReimarBauer, line 501: https://github.com/Open-MSS/MSS/issues/2148

There was a mismatch for fs by os path joining. Because we check if pytest is in sys.modules I assume the finding I had were by setting the MSUI_SETTINGS env var myself.

conftest.py
```diff
@@ -77,22 +76,23 @@ def keyring_reset():
 def generate_initial_config():
     """Generate an initial state for the configuration directory in tests.constants.ROOT_FS
     """
-    if not constants.ROOT_FS.exists("msui/testdata"):
-        constants.ROOT_FS.makedirs("msui/testdata")
+    _path = os.path.join(constants.ROOT_DIR, "msui", "testdata")
+    if not os.path.exists(_path):
+        os.makedirs(_path)
```
matrss, line 81: Why does this revert to using os?

docs/conf.py
```diff
@@ -67,9 +67,11 @@ def get_tutorial_images():
...
     from mslib.mswms.demodata import DataFiles
 
-    root_fs = fs.open_fs("~/")
-    if not root_fs.exists("mss/testdata"):
-        root_fs.makedirs("mss/testdata")
+    with fs.open_fs("~/") as root_fs:
+        if not root_fs.exists("mss/testdata"):
+            root_fs.makedirs("mss/testdata")
+        if not root_fs.exists(".config/msui"):
+            root_fs.makedirs(".config/msui")
```
matrss, line 74: We should use `$XDG_CONFIG_HOME` instead of hardcoding `~/.config`.

EDIT: since we are already using pyfilesystem2 we could also make use of its ["App Filesystems"](https://docs.pyfilesystem.org/en/latest/reference/appfs.html), or utilize the underlying library, [appdirs](https://pypi.org/project/appdirs/), directly. This would make configuration and data locations automatically cross-platform compatible. The stuff that is currently dumped into ~/mss also belongs into one of these standardized locations, in my opinion.

EDIT again: appdirs is deprecated, [platformdirs](https://pypi.org/project/platformdirs/) exists.

mslib/msui/constants.py
```diff
@@ -30,10 +30,15 @@
 import fs
 import os
 import logging
+from fs import open_fs
```
matrss, line 33: Why that instead of `fs.open_fs`? Especially since `fs` is already imported and `fs.open_fs` used in other locations in this file. It is just three letters more and easier to read.

mslib/msui/constants.py
```diff
@@ -30,10 +30,15 @@
...
 
 # ToDo refactor to generic functions, keep only constants
 HOME = os.path.expanduser(f"~{os.path.sep}")
 MSUI_CONFIG_PATH = os.getenv("MSUI_CONFIG_PATH", os.path.join(HOME, ".config", "msui"))
+ROOT_FS = open_fs(MSUI_CONFIG_PATH)
+ROOT_DIR = ROOT_FS.getsyspath("")
```
matrss, line 39: Using `getsyspath` implicitly assumes that there is a local path available for the PyFilesystem2 fs. Apart from OSFS most don't support that and will throw an exception here.

Not that I see much of a use case putting `MSUI_CONFIG_PATH` on something other than the local filesystem, but I assume this was implemented for a reason?

tests/constants.py
```diff
@@ -42,27 +42,31 @@
         SHA = ""
 
 CACHED_CONFIG_FILE = None
-SERVER_CONFIG_FILE = "mswms_settings.py"
+MSWMS_SERVER_CONFIG_FILE = "mswms_settings.py"
```
matrss, line 45:
```suggestion
MSWMS_CONFIG_FILE = "mswms_settings.py"
```

To apply the same naming scheme that is also already used for `MSCOLAB_CONFIG_FILE` below.

tests/constants.py
```diff
@@ -42,27 +42,31 @@
...
-SERVER_CONFIG_FS = fs.open_fs(fs.path.join(ROOT_DIR, "msui"))
-DATA_FS = fs.open_fs(fs.path.join(ROOT_DIR, "msui/testdata"))
-
 MSUI_CONFIG_PATH = OSFS_URL
-# MSUI_CONFIG_PATH = SERVER_CONFIG_FS.getsyspath("") would use a none osfs path
 os.environ["MSUI_CONFIG_PATH"] = MSUI_CONFIG_PATH
-SERVER_CONFIG_FILE_PATH = fs.path.join(SERVER_CONFIG_FS.getsyspath(""), SERVER_CONFIG_FILE)
 
+ROOT_DIR = ROOT_FS.getsyspath("")
+# using pytest we skip using this conf store
```
matrss, line 54: I don't understand what this comment refers to.

conftest.py
matrss, line 221: This is not true. Each worker has their own ROOT_DIR. The ROOT_DIR is shared between consecutive tests on the same worker, which is why this reset fixture exists.

conftest.py
```diff
@@ -220,14 +215,16 @@ def _load_module(module_name, path):
...
 @pytest.fixture(autouse=True)
 def reset_config():
-    """Reset the configuration directory used in the tests (tests.constants.ROOT_FS) after every test
+    """Reset the configuration directory used in the tests (tests.constants.ROOT_FS) before every test
     """
     # Ideally this would just be constants.ROOT_FS.removetree("/"), but SQLAlchemy complains if the SQLite file is deleted.
+    # We have between the workers a shared constants.ROOT_DIR
     for e in constants.ROOT_FS.walk.files(exclude=["mscolab.db"]):
         constants.ROOT_FS.remove(e)
+    # mscolab_settings, mswms_settings, testdata is not changed so we can keep it
```
matrss, line 224: This is not true. We have a test that modifies the content of testdata here: https://github.com/Open-MSS/MSS/blob/037ba61c25a6cccdf0357779d800339a25857ae5/tests/_test_mswms/test_wms.py#L397-L436
It is currently disabled because it is flaky, but it exists.

Before exempting directories from the cleanup I would want a check that explicitly validates this assumption.

conftest.py
```diff
@@ -220,14 +215,16 @@ def _load_module(module_name, path):
...
     for e in constants.ROOT_FS.walk.dirs(search="depth"):
-        constants.ROOT_FS.removedir(e)
-
+        if not e in ("/mscolab", '/mswms', '/mswms/testdata'):
+            constants.ROOT_FS.removedir(e)
```
matrss, line 227: Apart from my concerns that the assumption that these paths are not modified doesn't hold true, I would also suggest splitting these paths out of the configuration directory (and by extensions this fixture). Test data is not part of the configuration, and if it really is static it could just be a separate session-scoped fixture. I am not sure what is stored in `/mscolab` and `/mswms` and if it qualifies as "configuration".

matrss: Also, as it is this change doesn't seem to speed up the tests, so I don't see what the motivation for it might be.

matrss (changes requested):

docs/conf.py
ReimarBauer: there should be already an issue for this by @joernu76  for $XDG_CONFIG_HOME.  On first attempt we had problems by that. But  maybe these are now solved.

App Filesystems can make a point for Pyfilesystem, sometimes I think if we should deprecate it.

tests/constants.py
ReimarBauer: qsettings store is not used on tests, see

https://github.com/Open-MSS/MSS/blob/develop/mslib/utils/config.py#L470

tests/constants.py
ReimarBauer: that was the idea, we know that it is a server.

docs/conf.py
matrss: I think you mean #908.

conftest.py
ReimarBauer: When I keep it, like:

```
diff --git a/conftest.py b/conftest.py
index 9bf23264..08bd8560 100644
--- a/conftest.py
+++ b/conftest.py
@@ -221,10 +221,8 @@ def reset_config():
     # We have between the workers a shared constants.ROOT_DIR
     for e in constants.ROOT_FS.walk.files(exclude=["mscolab.db"]):
         constants.ROOT_FS.remove(e)
-    # mscolab_settings, mswms_settings, testdata is not changed so we can keep it
     for e in constants.ROOT_FS.walk.dirs(search="depth"):
-        if not e in ("/mscolab", '/mswms', '/mswms/testdata'):
-            constants.ROOT_FS.removedir(e)
+        constants.ROOT_FS.removedir(e)
     generate_initial_config()
     create_msui_settings_file("{}")
     read_config_file()
```

this gives

```
ERROR tests/_test_utils/test_multidict.py::TestCIMultiDict::test_multidict - fs.errors.ResourceNotFound: resource 'mswms_settings.py' not found
=================================== 8 skipped, 7 warnings, 548 errors in 24.08s ====================================
```
all have the same failure, they can't load mswms_settings.py

EDIT: When we want to have that after the tests we should yield the testfunction. But tests without that PR are not crashing, so somehow I changed a bit more than wanted. Or we clean something different which did mostly not make problems.

mslib/msui/constants.py
ReimarBauer: I try to solve that differently

conftest.py
ReimarBauer: This fixture seems to run in front of the test now. It removes the created server settings and seems not to be able to reload the new existing ones.

I can't currently see why that behaviour is such different.

conftest.py
matrss: I see. That error is originating from `generate_initial_config()`, so either your changes there broke it or they surfaced a bug that was already there.

Not cleaning the environment between tests is not a fix though.

I think the issue is that the reset fixture fully cleans the environment, but due to the way the constants are loaded (once, at the beginning of the test execution, which is a pain already because it makes it impossible to just use a different directory for each test, which would be the best solution) the mswms/ directory is created only once, when tests.constants is imported, and not recreated for the next test. I think this is a major architectural problem that needs to be re-thought, while the simple fix is to re-create the mswms/ directory at the beginning of `generate_initial_config`.

docs/conf.py
matrss: > sometimes I think if we should deprecate it.

You mean remove fs entirely?

I think we should either fully remove it, or go all-in on fs. The current state of mixing everything everywhere is a mess and results in practically nothing but OSFS working, which makes fs somewhat pointless.

Not sure what would be better. I like python's builtin pathlib module, as an alternative to os, and haven't really seen a need for fs yet.

conftest.py
matrss: The behaviour is different because you have moved `MSWMS_SERVER_CONFIG_FS` and `DATA_FS` from msui/ to mswms/. `generate_initial_config` creates msui/ at the beginning, but doesn't create mswms/. The missing mswms/ directory is the issue.

docs/conf.py
ReimarBauer: I moved that to the next major, we should collect pro's and con's first. 

https://github.com/Open-MSS/MSS/issues/2103

When we introduced FS we had anything using it. Over the years it gots a mixup. At the time we introduced it there was no pathlib and we also had problems with tests on windows because of path. The idea that we use it with FS pathes seems not become the standard. We can use it with our nextcloud but we don't. etc.
Also with mscolab we can do things differently, e.g. exchange images, or config over chat.

conftest.py
ReimarBauer: constants.ROOT_DIR was for tests always an os path, see https://github.com/Open-MSS/MSS/blob/stable/tests/constants.py#L50

matrss: OK, at the time I wrote the comment I was off the impressions that we wanted to use fs everywhere.

This can be simplified to
```python
(pathlib.Path(constants.ROOT_DIR) / "msui" / "testdata").mkdir(parents=True, exist_ok=True)
```
(maybe we should make the constants `pathlib.Path` objects instead of just strings at some point?)

ReimarBauer: There are limitations by QT, qsettings needs an os.path. And for other parts we would need to change the filebrowser to the filepicker. 

After the release of 9.0.0 we can discuss and refactor that.  I think I close this PR and just fix the msui-core.conf mismatch.

matrss: > There are limitations by QT, qsettings needs an os.path.

There is no such thing as "a os.path". os.path is not a type. Historically, os and os.path represented paths simply as strings, but the python stdlib should accept so-called PathLike objects everywhere now. Qt might not, but pathlib.Path can trivially be cast to str to get a string reprensentation that would work.

ReimarBauer: It did not understand to handle an fs object.

ReimarBauer: That was too fast typed. The failure comes because qsettings can't handle fs_urls or strings for pyfilesystem2.
Also the Qt filebrowser can't. 

For me it is ok after the release of 9.0.0 to look how we can refactor this. Also pathlib is nice. But now I want to get a  new major before the next campaign starts released.

ReimarBauer:
This did show how complicated some last minutes changes can be. It did more harm as it  solves. 

We have to clarify that first
https://github.com/Open-MSS/MSS/issues/2103

PR #2355, opened 2024-05-14
Verify the dict representation of a message has an isoformatted timestamp

### LINKED ISSUE #2356: create a test to verify a socket created message has a propper time stamp
we have changed to use a tz info and require isoformatted time stamps. This is not reflected in the tests.

ReimarBauer:
**Purpose of PR?**:

Fixes #2356

### DIFF
--- a/tests/_test_msui/test_mscolab_operation.py
+++ b/tests/_test_msui/test_mscolab_operation.py
@@ -25,6 +25,7 @@
     limitations under the License.
 """
 import pytest
+import datetime
 
 from mslib.mscolab.conf import mscolab_settings
 from mslib.mscolab.models import Message, MessageType
@@ -83,6 +84,7 @@ class Test_MscolabOperation:
             message = Message.query.filter_by(text='**test message**').first()
             result = get_message_dict(message)
             assert result["message_type"] == MessageType.TEXT
+            assert datetime.datetime.fromisoformat(result["time"]).year == datetime.datetime.now().year
 
     def test_search_message(self, qtbot):
         self._send_message(qtbot, "**test message**")
--- a/tests/_test_msui/test_mscolab_operation.py
+++ b/tests/_test_msui/test_mscolab_operation.py
@@ -27,12 +27,13 @@
 import pytest
 
 from mslib.mscolab.conf import mscolab_settings
-from mslib.mscolab.models import Message
+from mslib.mscolab.models import Message, MessageType
 from PyQt5 import QtCore, QtTest, QtWidgets
 from mslib.msui import mscolab
 from mslib.msui import msui
 from mslib.mscolab.seed import add_user, get_user, add_operation, add_user_to_operation
 from mslib.utils.config import modify_config_file
+from mslib.mscolab.utils import get_message_dict
 
 
 class Actions:
@@ -79,6 +80,9 @@ class Test_MscolabOperation:
         self._send_message(qtbot, "**test message**")
         with self.app.app_context():
             assert Message.query.filter_by(text='**test message**').count() == 2
+            message = Message.query.filter_by(text='**test message**').first()
+            result = get_message_dict(message)
+            assert result["message_type"] == MessageType.TEXT
 
     def test_search_message(self, qtbot):
         self._send_message(qtbot, "**test message**")
--- a/tests/_test_msui/test_mscolab_operation.py
+++ b/tests/_test_msui/test_mscolab_operation.py
@@ -84,7 +84,7 @@ class Test_MscolabOperation:
             message = Message.query.filter_by(text='**test message**').first()
             result = get_message_dict(message)
             assert result["message_type"] == MessageType.TEXT
-            assert datetime.datetime.fromisoformat(result["time"]).year == datetime.datetime.now().year
+            assert datetime.datetime.fromisoformat(result["time"]) == message.created_at
 
     def test_search_message(self, qtbot):
         self._send_message(qtbot, "**test message**")
--- a/tests/_test_mscolab/test_utils.py
+++ b/tests/_test_mscolab/test_utils.py
@@ -29,28 +29,9 @@ import json
 
 from fs.tempfs import TempFS
 from mslib.mscolab.conf import mscolab_settings
-from mslib.mscolab.models import Operation, MessageType
+from mslib.mscolab.models import Operation
 from mslib.mscolab.seed import add_user, get_user
-from mslib.mscolab.utils import (get_recent_op_id, get_session_id,
-                                 get_message_dict, create_files,
-                                 os_fs_create_dir)
-
-
-class Message:
-    id = 1  # noqa: A003
-    u_id = 2
-
-    class user:
-        username = "name"
-    text = "Moin"
-    message_type = MessageType.TEXT
-    reply_id = 0
-    replies = []
-
-    class created_at:
-        @staticmethod
-        def isoformat():
-            pass
+from mslib.mscolab.utils import get_recent_op_id, get_session_id, create_files, os_fs_create_dir
 
 
 class Test_Utils:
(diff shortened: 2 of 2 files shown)


tests/_test_msui/test_mscolab_operation.py
```diff
@@ -79,6 +81,10 @@ def test_send_message(self, qtbot):
         self._send_message(qtbot, "**test message**")
         with self.app.app_context():
             assert Message.query.filter_by(text='**test message**').count() == 2
+            message = Message.query.filter_by(text='**test message**').first()
+            result = get_message_dict(message)
+            assert result["message_type"] == MessageType.TEXT
+            assert datetime.datetime.fromisoformat(result["time"]).year == datetime.datetime.now().year
```
ReimarBauer: that was missing for the isoformat change.

ReimarBauer:
I think we should have both tests. That isoformat test was missing for the last change.

matrss (changes requested):
The linked issue describes a different thing that has already been fixed. The title of the PR does not match the changes (move vs add).

In my opinion we don't need a test for this `get_message_dict` anyway, what it does assert is trivially true and the function is being exercised by broader tests already. If we do unit tests on a too fine-grained level (e.g. for private helper functions) we will get ourselves into a situation where any architectural change will break tests (e.g. because the tested function is removed/renamed/its interface changed).

In my opinion, the most valuable tests are those that test the public interface of the code (in our case that is the msui Qt app as well as the mscolab and mswms APIs, where the latter two are best tested through the UI) and we should thrive to get full coverage of the code base through those high level tests. Anything more fine-grained has to be considered individually for its value.

ReimarBauer:
> The linked issue describes a different thing that has already been fixed. The title of the PR does not match the changes (move vs add).
> 
changed issue and title

> In my opinion we don't need a test for this `get_message_dict` anyway, what it does assert is trivially true and the function is being exercised by broader tests already. If we do unit tests on a too fine-grained level (e.g. for private helper functions) we will get ourselves into a situation where any architectural change will break tests (e.g. because the tested function is removed/renamed/its interface changed).

Where do we have a test, that we now require isoformatted time strings with a tz information?

> 
> In my opinion, the most valuable tests are those that test the public interface of the code (in our case that is the msui Qt app as well as the mscolab and mswms APIs, where the latter two are best tested through the UI) and we should thrive to get full coverage of the code base through those high level tests. Anything more fine-grained has to be considered individually for its value.


This tests a socket created chat message if it is conform to an iso formatted time string. How is that different to your last paragraph?

matrss:
What I said was primarily directed at the MessageType check and the entirety of https://github.com/Open-MSS/MSS/blob/531cb75769e51ae7c1403dc202892873265ad66d/tests/_test_mscolab/test_utils.py#L66-L71. I would just remove them.

> Where do we have a test, that we now require isoformatted time strings with a tz information?

We don't, and I don't think we should have one unless using the iso format is part of our public interface. What we should have is a test that sends and receives messages from a client, exercising the serialization between client and server and making sure that messages reach each end as they should. We might need to have two or more instances of msui available in a test to properly do this. As it is, I consider the format of these timestamps as an implementation detail that is not important to check. Whenever we split MSS into multiple packages this might change, since then mscolab would have a public API and these timestamps would be part of it. In that case we should also think about versioning the API, though.

ReimarBauer:
> What I said was primarily directed at the MessageType check and the entirety of
> 
> https://github.com/Open-MSS/MSS/blob/531cb75769e51ae7c1403dc202892873265ad66d/tests/_test_mscolab/test_utils.py#L66-L71
> 
> . I would just remove them.

I moved it because I recognized that I can't use it in a common event based usecase. So it was only with that Message class a check for the definition of the model.  And by that a test for the get_message_dict.  I think we should not remove it yet but maybe at 80% coverage organize tests differently.


> > Where do we have a test, that we now require isoformatted time strings with a tz information?
> 
> We don't, and I don't think we should have one unless using the iso format is part of our public interface. What we should have is a test that sends and receives messages from a client, exercising the serialization between client and server and making sure that messages reach each end as they should. We might need to have two or more instances of msui available in a test to properly do this. As it is, I consider the format of these timestamps as an implementation detail that is not important to check. Whenever we split MSS into multiple packages this might change, since then mscolab would have a public API and these timestamps would be part of it. In that case we should also think about versioning the API, though.

I think this different, when we haven't a test we need to communicate that often, while a test says "no, that is wrong".
It is too easy to miss that tzinfo or isoformat is set. Also on the refactoring I wished I had an idea how to add that test, the string which I missed would may be crashed it.

matrss:
> I think this different, when we haven't a test we need to communicate that often, while a test says "no, that is wrong".

Yes, what I am saying is just that we should make a conscious decision on what we test, and why, because overdoing it on the details is a net-negative. If we want to declare that all timestamps must be encoded with isoformat, then that is fine to add as a test; it then becomes part of the expected interface instead of remaining an implementation detail.

My stab at `test_get_message_dict` is meant to say that this test encodes that we must have a helper function called `get_message_dict` and that it must return a dictionary with a `message_type` key, while the value of that key must be `MessageType.TEXT` for the specific message we gave to it. That is hardly a useful tests, and it would just be a hindrance when refactoring that part of the code. Instead, the important thing to test is that a message sent from one client to another arrives unaltered, and if it uses this helper function on its way then we get coverage for it for free.

matrss:
> If we want to declare that all timestamps must be encoded with isoformat, then that is fine to add as a test; it then becomes part of the expected interface instead of remaining an implementation detail.

I am still not sure how this could be tested though. Just making sure that this one helper function does it right doesn't say much.

ReimarBauer:
> > I think this different, when we haven't a test we need to communicate that often, while a test says "no, that is wrong".
> 
> Yes, what I am saying is just that we should make a conscious decision on what we test, and why, because overdoing it on the details is a net-negative. If we want to declare that all timestamps must be encoded with isoformat, then that is fine to add as a test; it then becomes part of the expected interface instead of remaining an implementation detail.
> 
> My stab at `test_get_message_dict` is meant to say that this test encodes that we must have a helper function called `get_message_dict` and that it must return a dictionary with a `message_type` key, while the value of that key must be `MessageType.TEXT` for the specific message we gave to it. That is hardly a useful tests, and it would just be a hindrance when refactoring that part of the code. Instead, the important thing to test is that a message sent from one client to another arrives unaltered, and if it uses this helper function on its way then we get coverage for it for free.

In principle it does pinning the definition of the get_message_dict, not detailed enough to use it for a TDD approach. Also not the origin version. When the lib would have been defined by TDD we likly would have there also very different quality of these tests now. That is maybe also an interesting aspect for a talk about the tests life cycle of a project by TDD. How it looks like for a young project and envolves on a mature project.
But back to this, when this would have been the TDD test for that function, we would maybe like it that we have it.
I don't know right now whether I want to reject tests that are now subsequently created in this form for existing code.


What we currently mostly do is to write tests afterwards for bugfixes or refactoring, for better understanding what it is doing and how and to get a coverage so it is harder for us to break something without realizing it. And important we continously improve.

tests/_test_msui/test_mscolab_operation.py
matrss:
```suggestion
            assert datetime.datetime.fromisoformat(result["time"]) == message.created_at
```

This will test that the full parsed datetime object matches the one recorded in the database, instead of just the year. This is required to actually check that the tzinfo is conveyed properly (just the year won't catch that).

ReimarBauer: will do that, I guess it throws an exception when the string does not match the format.

matrss: It doesn't necessarily throw an exception, since most parts of the iso8601 format are optional. E.g. a date alone is parsed as that date at 00:00:
```python
>>> datetime.datetime.fromisoformat("2022-01-01")
datetime.datetime(2022, 1, 1, 0, 0)
```
A timezone doesn't have to specified either as also seen in that example, it is parsed to a naive datetime then.

That's why the year comparison is not enough.

matrss (approved):
Seems OK to me. I think this ad-hoc Message-to-dict mapper should eventually be replaced by a real serialization/deserialization module that can be properly tested with e.g. property-based testing.

PR #2360, opened 2024-05-16
Fix leftover directories after tests

### LINKED ISSUE #2148: The test suite leaves a lot of trash in /tmp and CWD

matrss:
**Purpose of PR?**:

- Fix osfs:/ directory leftover in CWD after tests
- Fix leftover directories in /tmp after tests
    
    The way wms_cache was set previously made it so that after each tests
    run there would be one or more directories leftover in /tmp. This also
    happened for each start of MSUI. This change makes MSUI use the
    platform-specific cache directory of the user instead and sets
    XDG_CACHE_HOME to isolate those between test runs.

Fixes #2148.

**Checklist:**
- [ ] Commit has unit tests

### DIFF
--- a/mslib/msui/constants.py
+++ b/mslib/msui/constants.py
@@ -35,7 +35,9 @@ import logging
 HOME = os.path.expanduser(f"~{os.path.sep}")
 MSUI_CONFIG_PATH = os.getenv("MSUI_CONFIG_PATH", os.path.join(HOME, ".config", "msui"))
 # Make sure that MSUI_CONFIG_PATH exists
-_ = fs.open_fs(MSUI_CONFIG_PATH, create=True)
+_fs = fs.open_fs(MSUI_CONFIG_PATH, create=True)
+# MSUI does not actually support any PyFilesystem2 fs that is not available as a local path
+MSUI_CONFIG_SYSPATH = _fs.getsyspath("")
 
 GRAVATAR_DIR_PATH = fs.path.join(MSUI_CONFIG_PATH, "gravatars")
 
--- a/mslib/msui/constants.py
+++ b/mslib/msui/constants.py
@@ -30,6 +30,7 @@
 import fs
 import os
 import logging
+import platformdirs
 
 # ToDo refactor to generic functions, keep only constants
 HOME = os.path.expanduser(f"~{os.path.sep}")
@@ -38,6 +39,8 @@ MSUI_CONFIG_PATH = os.getenv("MSUI_CONFIG_PATH", os.path.join(HOME, ".config", "
 _fs = fs.open_fs(MSUI_CONFIG_PATH, create=True)
 # MSUI does not actually support any PyFilesystem2 fs that is not available as a local path
 MSUI_CONFIG_SYSPATH = _fs.getsyspath("")
+
+MSUI_CACHE_PATH = platformdirs.user_cache_path("msui", "mss")
 
 GRAVATAR_DIR_PATH = fs.path.join(MSUI_CONFIG_PATH, "gravatars")
 
--- a/mslib/utils/config.py
+++ b/mslib/utils/config.py
@@ -32,7 +32,6 @@ import json
 import logging
 import fs
 import os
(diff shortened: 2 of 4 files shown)


mslib/utils/config.py
```diff
@@ -151,8 +148,7 @@ class MSUIDefaultConfig:
     WMS_preload = []
 
     # WMS image cache settings:
-    # this changes on any start of msui, use ths msui_settings.json when you want a persistent path
-    wms_cache = os.path.join(tempfile.TemporaryDirectory().name, "msui_wms_cache")
+    wms_cache = str(constants.MSUI_CACHE_PATH / "wms_cache")
```
ReimarBauer, line 151: :)

mslib/msui/constants.py
```diff
@@ -30,12 +30,17 @@
...
 # ToDo refactor to generic functions, keep only constants
 HOME = os.path.expanduser(f"~{os.path.sep}")
 MSUI_CONFIG_PATH = os.getenv("MSUI_CONFIG_PATH", os.path.join(HOME, ".config", "msui"))
 # Make sure that MSUI_CONFIG_PATH exists
-_ = fs.open_fs(MSUI_CONFIG_PATH, create=True)
+_fs = fs.open_fs(MSUI_CONFIG_PATH, create=True)
+# MSUI does not actually support any PyFilesystem2 fs that is not available as a local path
+MSUI_CONFIG_SYSPATH = _fs.getsyspath("")
+
+MSUI_CACHE_PATH = platformdirs.user_cache_path("msui", "mss")
```
ReimarBauer, line 44: On the tutorials I have to remove these data too. There it would be easier to remove only the MSUI_CONFIG_PATH, which I can do by using a config, which I also clear.

Because it wasn't the same dir on a new tutorial I didn't care. 

It might easier to have a cache dir below the MSUI_CONFIG_PATH  so all is on one place.

ReimarBauer: Also before we run tests the platformdirs cache needs to be cleared, or stored randomized.

matrss: > It might easier to have a cache dir below the MSUI_CONFIG_PATH so all is on one place.

Cache files don't belong into the configuration directory. There is a reason those are separate XDG directories. As a user I should be able to remove `XDG_CACHE_HOME` without any issue and be confident that this has cleared all application caches.

matrss: > Also before we run tests the platformdirs cache needs to be cleared, or stored randomized.

That happens by setting `XDG_CACHE_HOME` to a temporary directory in `tests.constants`.

ReimarBauer: on my KDE that is empty. 

image

matrss: > On the tutorials I have to remove these data too. There it would be easier to remove only the MSUI_CONFIG_PATH, which I can do by using a config, which I also clear.

The cache should behave in a way that it doesn't matter if it exists already or not (just speeding things up if it does), so there should not necessarily be a need to remove it. Nonetheless, the tutorials could also just set XDG_CACHE_HOME (or a MSUI_CACHE_PATH, if we want to additionally introduce that).

> Because it wasn't the same dir on a new tutorial I didn't care.

It still left trash in /tmp though.

matrss: > on my KDE that is empty.
> 
> image

https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html:

>  $XDG_CACHE_HOME defines the base directory relative to which user-specific non-essential data files should be stored. If $XDG_CACHE_HOME is either not set or empty, a default equal to $HOME/.cache should be used.

ReimarBauer: Will see next days when I  recreate the tutorials if we have to adopt something.

ReimarBauer (approved):

PR #2382, opened 2024-05-21
Merge develop to stable  == 9.0.0

ReimarBauer:
**Purpose of PR?**:

this is the base of the next stable release 9.x 

385 files changed, 13871 insertions(+), 8039 deletions(-)

There was no merge conflict. 

This needs a **merge**.


matrss (dismissed):
`git diff develop` shows nothing, so no difference between this and develop. Therefore, LGTM.

.github/workflows/testing.yml
```diff
@@ -3,131 +3,84 @@ name: Pytest MSS
...
+        (echo Dependencies differ && echo "triggerdockerbuild=yes" >> $GITHUB_ENV )
 
     - name: Install pyvirtualdisplay if on stable
-      if: ${{ inputs.branch_name == 'stable' }}
+      if: ${{ inputs.image_suffix == 'stable' }}
       run: |
         source /opt/conda/etc/profile.d/conda.sh
         source /opt/conda/etc/profile.d/mamba.sh
-        mamba install -n mss-${{ inputs.branch_name }}-env pyvirtualdisplay
+        mamba install -n mss-${{ inputs.image_suffix }}-env pyvirtualdisplay
```
matrss: Spoke to soon, since CI treats stable and develop differently. This entire step has to be removed.

ReimarBauer: ok, we can do that for now, but then we should also drop pyvirtualdisplay in the new stable asap.

matrss: What you did isn't what I meant. The entire block I've commented on should be removed, because pyvirtualdisplay is no longer needed on stable after the merge.

`inputs.branch_name` doesn't exist, so your change has the same effect as removing the entire step (it gets skipped), but it should be fully removed.

ReimarBauer: done

ReimarBauer:
found a problem with demodata, checking and redo that later

PR #2391, opened 2024-05-24
Remove pytest-timeout

### LINKED ISSUE #2292: Remove the pytest-timeout usage
I'd argue that using pytest-timeout does not give us any benefit:
1. As discovered in https://github.com/Open-MSS/MSS/pull/2287#issuecomment-2013016954 it seems like its timeout is actively stopping test runs on macOS 13 that genuinely need more time, since the runners are just very slow.
2. Most of the time where it would be useful (e.g. when some remaining Qt window blocks further test execution) it actually fails to do its job, in my experience (i.e. the GitHub Actions timeout is reached instead).

Therefore I think we could get rid of it.

matrss:
**Purpose of PR?**:

Fixes #2292.

**Checklist:**
- [ ] Commit has unit tests

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -6,7 +6,6 @@ log_file = pytest.log
 log_file_level = DEBUG
 log_file_format = %(asctime)s %(levelname)s %(message)s
 log_file_date_format = %Y-%m-%d %H:%M:%S
-timeout = 60
 filterwarnings =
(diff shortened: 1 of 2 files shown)


ReimarBauer (approved):
good that this can be removed, that simplifies also debugging.

PR #2394, opened 2024-05-24
Run flake8 for all .py files tracked in git

### LINKED ISSUE #2374: flake8 in develop gets sloppy
I recognized that his was not show as failure by our flake8 checket.

https://github.com/Open-MSS/MSS/pull/2373#discussion_r1606099190

matrss:
**Purpose of PR?**:

Fixes #2374.

**Checklist:**
- [ ] Commit has unit tests

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -13,14 +13,15 @@
 # serve to show the default.
 import os
 import sys
-import logging
 import setuptools
 import subprocess
 import requests
 import zipfile
 import shutil
+import pathlib
 
 from string import Template
+
 
 def get_tutorial_images():
     TUTORIAL_URL = "https://fz-juelich.sciebo.de/s/KcF29hPNRzkxN6q/download"
@@ -97,24 +98,21 @@ if os.environ.get("GALLERY", "True") != "False":
     mslib.mswms.wms.server.generate_gallery(sphinx=True, generate_code=True, all_plots=True, levels="3,4,200,300",
                                             vtimes="2012-10-18T00:00:00,2012-10-19T00:00:00")
 
-# readthedocs has no past.builtins
-try:
-    from past.builtins import execfile
-except ImportError as ex:
-    logging.error("%s", ex)
-execfile('../mslib/version.py')
+version = {}
+exec(pathlib.Path("../mslib/version.py").read_text(), version)
+__version__ = version["__version__"]
 
 on_rtd = os.environ.get('READTHEDOCS') == 'True'
 
 # If extensions (or modules to document with autodoc) are in another directory,
 # add these directories to sys.path here. If the directory is relative to the
 # documentation root, use os.path.abspath to make it absolute, like shown here.
-#sys.path.insert(0, os.path.abspath('.'))
+# sys.path.insert(0, os.path.abspath('.'))
 
 # -- General configuration ------------------------------------------------
 
 # If your documentation needs a minimal Sphinx version, state it here.
-#needs_sphinx = '1.0'
+# needs_sphinx = '1.0'
 
 # Add any Sphinx extension module names here, as strings. They can be
 # extensions coming with Sphinx (named 'sphinx.ext.*') or your custom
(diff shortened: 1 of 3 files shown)


docs/conf.py
```diff
@@ -97,24 +97,19 @@ def get_tutorial_images():
     mslib.mswms.wms.server.generate_gallery(sphinx=True, generate_code=True, all_plots=True, levels="3,4,200,300",
                                             vtimes="2012-10-18T00:00:00,2012-10-19T00:00:00")
 
-# readthedocs has no past.builtins
-try:
-    from past.builtins import execfile
-except ImportError as ex:
-    logging.error("%s", ex)
-execfile('../mslib/version.py')
+from mslib.version import __version__
```
ReimarBauer: there is not an mss installation on readthedocs

So it can't become imported

additionally the dir should become added to the PYTHONPATH, e.g. sys.path.insert(0, '../')

docs/conf.py
```diff
@@ -133,14 +128,14 @@ def get_tutorial_images():
...
-#source_encoding = 'utf-8-sig'
+# source_encoding = 'utf-8-sig'
 
 # The master toctree document.
 master_doc = 'index'
 
 # General information about the project.
 project = 'MSS - Mission Support System'
-copyright = \
+project_copyright = \
```
ReimarBauer, line 141: this is a var which is used by the template.

and is used on bottom of the rtfd page

image

ReimarBauer: https://github.com/readthedocs/sphinx_rtd_theme/blob/master/sphinx_rtd_theme/footer.html#L21

ReimarBauer (changes requested):
needs few changes for making readthedocs happy.

docs/conf.py
matrss: Hopefully fixed by installing MSS into the readthedocs environment. Sidenote: is it possible to test this in any way, before it is merged?

docs/conf.py
matrss: `project_copyright` was added as an alias for `copyright` to fix this issue specifically: https://github.com/sphinx-doc/sphinx/issues/8132

docs/conf.py
ReimarBauer: I am not sure if we would exceed a limit on readthedocs with installing MSS. The env gets large, May has selected in the evironment.yml the needed modules for a gallery build. We could also use a copy of the _load_module of https://github.com/Open-MSS/MSS/blob/develop/conftest.py#L208

When it is merged and can't be builded the old version can't be updated and is kept.  

We maybe can setup a third version for testing, but have to figure how to do that.

docs/conf.py
ReimarBauer: the template in ReadTheDocs still uses that oldvalue. So you can add a ToDo at our file for renaming once it is fixed or send a PR to the template too. And when upstream it is changed we can use it.

docs/conf.py
matrss: It is installed without any dependencies (since the python package declares none), so it shouldn't be any larger than the repository that is already cloned.

If you really don't want it installed in that environment, and `__version__` is the only thing we need from the mslib module, then I would move the value of `mslib.version.__version__` into a separate `VERSION` file and just read it from there.

What's done in https://github.com/Open-MSS/MSS/blob/develop/conftest.py#L208 is IMO too complex for such a simple thing, and I don't even know what its purpose is in conftest.py and suspect that it could, and should, be removed there as well.

docs/conf.py
matrss: It is an alias, `copyright` uses the value of `project_copyright` as a default, so it is set to the same value: https://github.com/tk0miya/sphinx/blob/c5a9d04d453f457b087a03037ce49f88a4e439cb/sphinx/config.py#L101-L102. The template can continue to use `copyright`, that is not an issue.

I've also just validated that it does indeed work.

docs/conf.py
matrss: I've removed the mss installation in the rtd env and replaced execfile with an exec call derived from what is suggested here under 3.: https://packaging.python.org/en/latest/guides/single-sourcing-package-version/#single-sourcing-the-package-version

This makes the flake8 warning go away because it explicitly sets `__version__`.

ReimarBauer (approved):
LGTM

PR #2397, opened 2024-05-27
Fix #2396

### LINKED ISSUE #2396: Unhelpful error messages for missing data
Accessing data with a timestamp, for which no data raises misleading error messages in dataaccess class that are forwarded to the user.
The error message should contain a proper description of the state.

joernu76:
**Purpose of PR?**:

Fixes #2396

nope

### DIFF
--- a/mslib/mswms/dataaccess.py
+++ b/mslib/mswms/dataaccess.py
@@ -232,7 +232,8 @@ class DefaultDataAccess(NWPDataAccess):
             else:
                 logging.error("Could not identify filename. %s %s %s %s %s %s",
                               variable, vartype, init_time, valid_time, type(ex), ex)
-                raise ValueError(f"variable type {vartype} not available for variable {variable}")
+                raise ValueError(f"Variable '{variable}' not available for type '{vartype}', "
+                                 f"init_time '{init_time}', and valid_time '{valid_time}'")
 
     def is_reload_required(self, filenames):
         return False


jgrooss:
I am still working on how to run the mswms  develop version locally. 
- install developments in an env by [...] "mamba install mss=9.0.0 --only-deps"
- set python path by "export PYTHONPATH="`pwd`:$HOME/mss-dev"
- run server by "python mslib/mswms/mswms.py"
but I get some error message:
 ```
from mslib import __version__
ModuleNotFoundError: No module named 'mslib'

```

**Review by ReimarBauer — DISMISSED** [review]
thx

**Comment by ReimarBauer** [comment]
> I am still working on how to run the mswms develop version locally.
> 
> * install developments in an env by [...] "mamba install mss=9.0.0 --only-deps"
> * set python path by "export PYTHONPATH="`pwd`:$HOME/mss-dev"
> * run server by "python mslib/mswms/mswms.py"
>   but I get some error message:
> 
> ```
> from mslib import __version__
> ModuleNotFoundError: No module named 'mslib'
> ```

@jgrooss This likly should not get in the PR comment?

But to answer this:

```
$ ~/PycharmProjects/MSS$ mamba create -n meytenv mss==9.0.0 --only-deps
$ mamba activate meytenv
$ export PYTHONPATH=`pwd`:
$ python mslib/mswms/mswms.py
```

matrss:
Should this really go to develop, or rather to stable? It seems like a non-breaking, non-feature change to me, so I'd say stable.

joernu76:
We should discuss our strategy at some point in time.
I personally only merge hotfixes for (breaking) bugs to stable. Everything else goes to develop.

matrss:
I had the same view as well, but I see this as a similar situation to #2180. There (in the first 4 comments) Reimar told me that the change should target stable instead, even though it wasn't urgent or a "real" bug fix.

I think currently, as a rule of thumb, anything that requires a major release should go to develop, and everything else should go to stable.

Yes, we should discuss this process at some point. I personally dislike that we effectively have two development branches and I would rather want to treat stable as a maintenance branch that should only receive changes that already hit develop and were selected to be backported to the currently released version. But that is just my opinion.

@ReimarBauer can you clarify how to proceed here?

joernu76:
That is why we should have a meeting to come up with a common understanding and strategy. Having only small changes in stable certainly simplifies the there-and-back merging. This obviously only works for releasing reasonably often.

ReimarBauer:
@matrss is correct, I missed to look on the branch.

I add some few more points to the view. A voice meet for this is better to get all the facettes.

We are based on https://semver.org/

We anyway have often to merge from stable to develop. Because there we do our releases. And any bug fix (or even small enhancement) which don't break the API can go in to stable. We may discuss if we want only patch level releases, which is the last of the tripple or can do also use the minor number (second of the  tripple)  

Usually we go first for a patch level bug fix release, but also in  the 8.x we did minor releases. That means that when we for example have planned a 9.0.1 and we do add enhancements which don't need an API change the release can become a 9.1.0. (of course the issue lables have to be changed then, which I also did before) 
Doing it that way seems to me a very simple approach and gives us a bit flexebilty.

ReimarBauer:
I switched this to stable. We need to fix a bunch of problems, so it fits to a 9.0.1

ReimarBauer (approved):
fix for stable

matrss (approved):
Technically this PR now contains an unrelated commit that shouldn't be in here, because of changing the base branch without a proper rebase. Practically it shouldn't matter _in this case_, since that commit doesn't add any code and will be gone with the squash.

PR #2405, opened 2024-06-14
feat: Allow users to set custom profile pictures

workaryangupta:
This pull request aims to introduce a new feature to mscolab, enabling users to upload and set personalized profile pictures.

Screenshot (292)


matrss:
Please take a look at the CI test failures. You have introduced a new required positional argument to a constructor that is not supplied everywhere it is used, so it crashes a lot of tests.

I think the profile image should be optional and the default behavior should match what we had before this PR.

workaryangupta:
I made the profile image optional but still, the tests are failing. There is something more to it that I am not able to grasp.

matrss:
This is now a new test failure because some data files in skyfield-data have expired yesterday. This is unrelated to your code. Coincidentally, I've already noticed this last week and opened an issue: https://github.com/brunobord/skyfield-data/issues/36.

@ReimarBauer it looks like you fixed this the last time it came up, judging by the commit history in skyfield-data. Do you mind taking a look again? I am not quite sure what the process is, over there.

ReimarBauer:
You could xfail the test with a good hint to our issue, so it can be later enabled again.
https://github.com/Open-MSS/MSS/issues/2406

mslib/mscolab/server.py
```diff
@@ -351,6 +351,20 @@ def get_user():
     return json.dumps({'user': {'id': g.user.id, 'username': g.user.username}})
 
 
+@APP.route('/upload_profile_image', methods=["POST"])
+def upload_profile_image():
```
ReimarBauer, line 358: there should be check that the user is logged in, we have therefore a fixture.

without this unknown people can spam the server, because they can use prepared requests.

mslib/mscolab/server.py
```diff
@@ -351,6 +351,20 @@ def get_user():
...
+    user_id = request.form['user_id']
+    file = request.files['image']
+    if file:
```
ReimarBauer: and here should be configurable limit over the servers settings config.

I am not sure if I would give a user more than 1MB max size of a file.

mslib/msui/mscolab.py
```diff
@@ -51,9 +52,12 @@
 from mslib.msui import mscolab_admin_window as maw
 from mslib.msui import mscolab_version_history as mvh
 from mslib.msui import socket_control as sc
+from mslib.mscolab.file_manager import FileManager
```
ReimarBauer: keep in mind the UI should not do anything with the servers code

On a future release we also separate the server resources from the client resources.

mslib/msui/mscolab.py
```diff
@@ -694,6 +698,26 @@ def after_login(self, emailid, url, r):
             self.signal_login_mscolab.emit(self.mscolab_server_url, self.token)
 
     def fetch_gravatar(self, refresh=False):
+        # Display custom profile picture if exists
+        from mslib.mscolab.server import APP
+        with APP.app_context():
+            fm = FileManager(APP.config["MSCOLAB_DATA_DIR"])
```
ReimarBauer: this is broken by design. 

A user should be able to access a server which is not on his machine. 

Many users can share one server located at somewhere, you can not expect a local MSCOLAB_DATA_DIR has the same data for any user.

ReimarBauer: you need a request over the server to fetch the image from a database on a remote mscolab server.

for the uploading you do use the server already now you need it to get the image.

mslib/mscolab/server.py
ReimarBauer: file.read may need also some security checks.

e.g. is it an image mimtype

mslib/mscolab/file_manager.py
```diff
@@ -250,6 +250,31 @@ def modify_user(self, user, attribute=None, value=None, action=None):
...
         return True
 
+    def save_user_profile_image(self, user_id, image_data):
+        """
+        Save the user's profile image to the database.
+        """
+        user = User.query.get(user_id)
+        if user:
+            try:
+                user.profile_image = image_data
```
ReimarBauer: here we should somehow check what we are saving into the db.

ReimarBauer: it maybe also be better to store the file on the disk and the reference of the path in the database
see the discussion on

https://stackoverflow.com/questions/3748/storing-images-in-db-yea-or-nay

ReimarBauer (changes requested):
see comments

mslib/mscolab/conf.py
```diff
@@ -76,6 +76,12 @@ class default_mscolab_settings:
     # used to generate the password token
     SECURITY_PASSWORD_SALT = secrets.token_urlsafe(16)
 
+    # Max allowed file size for uploading user profile image
+    MAX_IMAGE_SIZE = 1 * 1024 * 1024  # 1MB
```
ReimarBauer: we have also attachments which can be images in the chat manager. The var maybe later used there too.

mslib/mscolab/conf.py
```diff
@@ -76,6 +76,12 @@ class default_mscolab_settings:
...
+
+    # Allowed file extensions for uploading user profile image
+    ALLOWED_EXTENSIONS = {'png', 'jpg', 'jpeg'}
```
ReimarBauer: that is confusing for the chat attachments, because we can share there also ftml files, csv files etc.
May be name this ALLOWED_IMAGE_EXTENSION, so it can be reused on other places.

ReimarBauer: I am also not sure if that is needed, when the server stores the image as file, it may be only one type, e.g. png for the users picture

mslib/mscolab/server.py
```diff
@@ -197,6 +197,11 @@ def wrapper(*args, **kwargs):
     return wrapper
 
 
+def allowed_file(filename):
+    """ Check if the uploaded file is in the allowed set of extensions """
```
ReimarBauer: when we store the file as image on the server by the server this is not needed, when we want to check incoming files of those types that is not enough, because when you add to a pdf, py, js a different extensions it wil gets over this.

mslib/msui/mscolab.py
```diff
@@ -699,25 +698,24 @@ def after_login(self, emailid, url, r):
...
-                self.ui.userOptionsTb.setIcon(icon)
-                return
+        url = urljoin(self.mscolab_server_url, 'fetch_profile_image')
+        data = {'user_id': str(self.user["id"])}
+        response = requests.get(url, data=data)
+        if response.status_code == 200:
+            img_data = response.content
+            img_byte_arr = io.BytesIO(img_data)
+            pixmap = QPixmap()
+            pixmap.loadFromData(img_byte_arr.getvalue())
```
ReimarBauer, line 708: when someone wants to feed in here some none image data, does this throw an exception?

mslib/mscolab/server.py
```diff
@@ -351,6 +356,40 @@ def get_user():
...
+def upload_profile_image():
+    user_id = request.form['user_id']
+    file = request.files['image']
+    if file and allowed_file(file.filename):
+        if file.mimetype.startswith('image/'):
+            try:
+                data = file.read()
+                if len(data) > mscolab_settings.MAX_IMAGE_SIZE:
+                    return jsonify({'message': 'File too large'}), 413
+                success, message = fm.save_user_profile_image(user_id, data)
```
ReimarBauer: instead of adding to the database you can use existing methods to store it as file on the server and save its name to the db, https://github.com/Open-MSS/MSS/blob/develop/mslib/mscolab/chat_manager.py#L100
e.g. move the function to the filemanager and refactor accordingly

mslib/mscolab/server.py
```diff
@@ -351,6 +356,40 @@ def get_user():
...
+                return jsonify({'message': str(e)}), 500
+        else:
+            return jsonify({'message': 'Invalid file type'}), 400
+    else:
+        return jsonify({'message': 'No file provided or invalid file type'}), 400
+
+
+@APP.route('/fetch_profile_image', methods=["GET"])
+def fetch_profile_image():
+    user_id = request.form['user_id']
```
ReimarBauer, line 378: There is not a protection for anyone using curl to fetch all images from the database.

When you use the method we defined for the uploaded chat files one has to guess more than a random string too. 

By this a DOS is possible, besides that we have a data leak.

mslib/mscolab/conf.py
matrss: The comment is incorrect, this is not 1 MB, it is 1 MiB.

mslib/mscolab/server.py
ReimarBauer: The current code can be used to add malicious content into the database.

mslib/mscolab/file_manager.py
ReimarBauer: over the server a user can add malicious content into the database.

mslib/mscolab/file_manager.py
```diff
@@ -250,6 +254,42 @@ def modify_user(self, user, attribute=None, value=None, action=None):
             db.session.commit()
         return True
 
+    def save_user_profile_image(self, user_id, image_file):
```
ReimarBauer: this is now a dup of existing code in the chat_manager 
https://github.com/Open-MSS/MSS/blob/develop/mslib/mscolab/chat_manager.py#L100

can you move this to the file_manager, refactor it so it can be used for all cases?

ReimarBauer: e.g. def upload_file(....

mslib/mscolab/models.py
```diff
@@ -58,16 +58,19 @@ class User(db.Model):
...
     password = db.Column(db.String(255))
+    profile_image_path = db.Column(db.String(255), nullable=True)
     registered_on = db.Column(AwareDateTime, nullable=False)
     confirmed = db.Column(db.Boolean, nullable=False, default=False)
     confirmed_on = db.Column(AwareDateTime, nullable=True)
     permissions = db.relationship('Permission', cascade='all,delete,delete-orphan', backref='user')
     authentication_backend = db.Column(db.String(255), nullable=False, default='local')
 
-    def __init__(self, emailid, username, password, confirmed=False, confirmed_on=None, authentication_backend='local'):
+    def __init__(self, emailid, username, password, profile_image_path=None, confirmed=False,
```
ReimarBauer, line 68: is this an absolut path or only a name?

mslib/mscolab/conf.py
```diff
@@ -68,7 +68,10 @@ class default_mscolab_settings:
 
     # mscolab file upload settings
     UPLOAD_FOLDER = os.path.join(DATA_DIR, 'uploads')
-    MAX_UPLOAD_SIZE = 2 * 1024 * 1024  # 2MB
+    MAX_UPLOAD_SIZE = 2 * 1024 * 1024  # 2MiB
+
```
ReimarBauer, line 72: I assume this shoul be used in https://github.com/Open-MSS/MSS/blob/develop/mslib/mscolab/app/__init__.py

mslib/mscolab/models.py
ReimarBauer: it better should be only the name because then easily it can be used with a path and a static url.

An url does not know about pathes. And a web server can also use an alias for an url.

mslib/mscolab/conf.py
```diff
@@ -68,7 +68,10 @@ class default_mscolab_settings:
...
+    # mscolab user profile image and image chat attachment limit
+    MAX_IMAGE_SIZE = 1 * 1024 * 1024  # 1MiB
```
ReimarBauer: with MAX_UPLOAD_SIZE this is not needed. Also the check for this isn't

mslib/mscolab/file_manager.py
```diff
@@ -250,6 +254,42 @@ def modify_user(self, user, attribute=None, value=None, action=None):
...
+        """
+        Save the user's profile image path to the database.
+        """
+        upload_folder = mscolab_settings.UPLOAD_FOLDER
+        file_token = secrets.token_urlsafe(20)
+
+        with fs.open_fs('/') as home_fs:
```
ReimarBauer: use create_files to ensure the needed pathes are available.

mslib/mscolab/file_manager.py
```diff
@@ -250,6 +254,42 @@ def modify_user(self, user, attribute=None, value=None, action=None):
             db.session.commit()
         return True
 
+    def save_user_profile_image(self, user_id, image_file):
+        """
+        Save the user's profile image path to the database.
+        """
+        upload_folder = mscolab_settings.UPLOAD_FOLDER
+        file_token = secrets.token_urlsafe(20)
```
matrss: Is there a reason to specify _exactly_ 20 bytes? If left unset `token_urlsafe` would use a "reasonable default", which currently amounts to 32 bytes and would, I'd imagine, grow if the stdlib people deem it necessary (e.g. from a cryptography stand point).

mslib/mscolab/file_manager.py
matrss: Opening "/" seems like a bad idea, potentially being vulnerable to path traversals (we already had something like that this year). Can this be a more specific directory from the get-go?

mslib/mscolab/file_manager.py
```diff
@@ -250,6 +254,42 @@ def modify_user(self, user, attribute=None, value=None, action=None):
...
+
+        with fs.open_fs('/') as home_fs:
+            file_dir = fs.path.join(upload_folder, 'profile')
+            # Checking if directory exists. Create the directory if it does not exist
+            if '\\' not in file_dir:
+                if not home_fs.exists(file_dir):
+                    home_fs.makedirs(file_dir)
+            else:
+                file_dir = file_dir.replace('\\', '/')
+                if not os.path.exists(file_dir):
```
matrss: This shouldn't mix `fs` and `os`/`os.path` stuff.

mslib/mscolab/file_manager.py
```diff
@@ -250,6 +254,42 @@ def modify_user(self, user, attribute=None, value=None, action=None):
...
+        """
+        Save the user's profile image path to the database.
+        """
+        upload_folder = mscolab_settings.UPLOAD_FOLDER
+        file_token = secrets.token_urlsafe(20)
+
+        with fs.open_fs('/') as home_fs:
+            file_dir = fs.path.join(upload_folder, 'profile')
+            # Checking if directory exists. Create the directory if it does not exist
+            if '\\' not in file_dir:
```
matrss: I don't quite understand the necessity for this distinction. "\\" is a perfectly valid character in a path on Linux and shouldn't silently be replaced with something else. If this is something for windows compatibility it should have a conditional based on if it is running on windows.

mslib/mscolab/file_manager.py
```diff
@@ -250,6 +254,42 @@ def modify_user(self, user, attribute=None, value=None, action=None):
...
+            file_name = secure_filename(file_name)
+
+            # Saving the file to the filepath
+            file_path = fs.path.join(file_dir, file_name)
+            image_file.save(file_path)
+
+            # Storing the file path to database
+            user = User.query.get(user_id)
+            if user:
+                user.profile_image_path = file_path
```
matrss: I don't think we should store the absolute path in the database. Instead, we should only put the path relative to the configured profile image folder in there, to support easy relocation in case of configuration changes. Also, it's less chars to store.

mslib/mscolab/server.py
```diff
@@ -351,6 +352,39 @@ def get_user():
...
+                return jsonify({'message': 'File too large'}), 413
+            success, message = fm.save_user_profile_image(user_id, file)
+            if success:
+                return jsonify({'message': message}), 200
+            else:
+                return jsonify({'message': message}), 400
+        else:
+            return jsonify({'message': 'Invalid file type'}), 400
+    else:
+        return jsonify({'message': 'No file provided or invalid file type'}), 400
```
matrss: A lot of the nesting in this function could be avoided by inverting the conditions and returning early. See https://softwareengineering.stackexchange.com/questions/18454/should-i-return-from-a-function-early-or-use-an-if-statement for an example of what I mean (specifically the second answer). This would make the function much more readable.

mslib/msui/mscolab.py
```diff
@@ -694,6 +697,30 @@ def after_login(self, emailid, url, r):
             self.signal_login_mscolab.emit(self.mscolab_server_url, self.token)
 
     def fetch_gravatar(self, refresh=False):
```
matrss: This function should probably be renamed, gravatar is a specific service and not a general name for profile images.

mslib/mscolab/file_manager.py
```diff
@@ -250,6 +254,42 @@ def modify_user(self, user, attribute=None, value=None, action=None):
...
+            if '\\' not in file_dir:
+                if not home_fs.exists(file_dir):
+                    home_fs.makedirs(file_dir)
+            else:
+                file_dir = file_dir.replace('\\', '/')
+                if not os.path.exists(file_dir):
+                    os.makedirs(file_dir)
+
+            # Creating unique and secure filename
+            file_name, file_ext = image_file.filename.rsplit('.', 1)
```
matrss: `file_ext` should better be derived from the mime type and the original filename not used at all, to avoid accidentally introducing vulnerabilities like path traversals (ideally I wouldn't want a filename to reach the flask endpoint at all, but I think it is part of flasks design that it is there).

mslib/mscolab/file_manager.py
ReimarBauer: I suggested that, then we can destinguish between chat images by the length.

mslib/mscolab/file_manager.py
ReimarBauer: windows

mslib/mscolab/file_manager.py
matrss: I'd suggest separate directories or a common prefix for each, e.g. `chat_attachment_` and `profile_image_`. Different lengths for a random part of the file name is very non-obvious.

mslib/mscolab/file_manager.py
matrss: So then this condition should be either `platform.system() != "Windows"` or `os.sep != "\\"`. Or we could define that the upload folder should always be specified with forward slashes, in which case I don't think any backslash could ever reach this point.

mslib/mscolab/file_manager.py
workaryangupta: Can you please explain this again after reviewing the recent changes.

ReimarBauer:
there are failing tests because of  the changes

- https://github.com/Open-MSS/MSS/actions/runs/9763173464/job/26948403618?pr=2405#step:10:2726

FAILED tests/_test_mscolab/test_sockets_manager.py::Test_Socket_Manager::test_upload_file - AssertionError: assert 'png' in 'mss-logo-20240702T145239-gg_WLzWnSYww0LGwa8mIRQ.unknown'

There the extension was removed. We want to know the extension.


- https://github.com/Open-MSS/MSS/actions/runs/9763173464/job/26948403618?pr=2405#step:10:2727

FAILED tests/_test_mscolab/test_chat_manager.py::Test_Chat_Manager::test_add_attachment - AttributeError: 'ChatManager' object has no attribute 'add_attachment'

There the method was changed?

workaryangupta:
> FAILED tests/_test_mscolab/test_chat_manager.py::Test_Chat_Manager::test_add_attachment - AttributeError: 'ChatManager' object has no attribute 'add_attachment'
> 
> There the method was changed?

add_attachment method was not needed, so I just removed it.

mslib/mscolab/file_manager.py
```diff
@@ -250,6 +256,59 @@ def modify_user(self, user, attribute=None, value=None, action=None):
...
+        """
+        Generic function to save files securely in specified directory with unique filename
+        and return the relative file path.
+        """
+        with fs.open_fs(upload_folder):
+            file_dir = fs.path.join(upload_folder, str(subfolder) if subfolder else "")
+            if sys.platform.startswith('win'):
+                file_dir = file_dir.replace('\\', '/')
+            if not os.path.exists(file_dir):
+                os.makedirs(file_dir)
```
matrss: This should fail on windows or at least create a wrong directory, because `os.makedirs` expects the OS-native separator, which was just replaced by "/".

Three options:
1. Fully use the `fs` module:
    ```python
    with fs.open_fs(upload_folder) as _fs:
        if sys.platform.startswith('win'):
            file_dir = file_dir.replace('\\', '/')
        _fs.makedirs(str(subfolder) if subfolder else "", recreate=True)
    ```
2. Fully use the `os` module:
    ```python
    file_dir = os.path.join(upload_folder, str(subfolder) if subfolder else "")
    os.makedirs(file_dir, exist_ok=True)
    ```
3. Use the more modern `pathlib` module:
    ```python
    file_dir = pathlib.Path(upload_folder) / (subfolder if subfolder is not None else "")
    file_dir.mkdir(parents=True, exist_ok=True)
    ```

Mixing `fs` and `os`/`pathlib` always leads to issues.

Which option is appropriate depends on if we need the fs-related advantages here. I am not sure if we do, I don't think so really; my preference would be `pathlib`. @ReimarBauer what is your opinion?

mslib/mscolab/file_manager.py
```diff
@@ -250,6 +256,59 @@ def modify_user(self, user, attribute=None, value=None, action=None):
...
+
+            # Creating unique and secure filename
+            file_name, _ = file.filename.rsplit('.', 1)
+            file_ext = guess_extension(file.mimetype) or '.unknown'
+
+            token = file_token if file_token else secrets.token_urlsafe(32)
```
matrss: Due to the reason described here: https://docs.python.org/3/library/secrets.html#how-many-bytes-should-tokens-use I don't think there should be an explicit number of bytes set for `token_urlsafe`. Instead, without an argument it will choose an "appropriate" default, which will be increased by the python stdlib maintainers as deemed necessary.

mslib/mscolab/file_manager.py
```diff
@@ -250,6 +256,59 @@ def modify_user(self, user, attribute=None, value=None, action=None):
             db.session.commit()
         return True
 
+    def upload_file(self, file, upload_folder, file_token=None, subfolder=None, identifier=None):
```
matrss: The signature can be slimmed down: `upload_folder` and `subfolder` can be one argument, `file_token` is not necessary as it can (and should) be created inside the method.

mslib/mscolab/server.py
```diff
@@ -351,6 +352,40 @@ def get_user():
...
+
+
+@APP.route('/fetch_profile_image', methods=["GET"])
+@verify_user
+def fetch_profile_image():
+    user_id = request.form['user_id']
+    user = User.query.get(user_id)
+    if user and user.profile_image_path:
+        base_directory = mscolab_settings.UPLOAD_FOLDER
+        base_path = os.path.join(base_directory, 'profile')
```
matrss: This hard-codes the profile path instead of using the config value. This is also the case in other places.

mslib/mscolab/file_manager.py
matrss: Maybe the upload_folder shouldn't be passed in at all and just taken from `mscolab_settings.UPLOAD_FOLDER`, and only a subfolder passed in then. That would prevent anyone from abusing this method to upload to a different location.

workaryangupta:
@matrss could you also guide me about resolving these test failures?
https://github.com/Open-MSS/MSS/actions/runs/9763173464/job/26948403618?pr=2405#step:10:2726
https://github.com/Open-MSS/MSS/actions/runs/9763173464/job/26948403618?pr=2405#step:10:2727

matrss:
`test_add_attachment` obviously fails because there is no `add_attachment` method anymore. This test could/should be moved to the other tests for the FileManager, where the replacement of `add_attachment` now resides, renamed and adapted to test `upload_file` instead.

`test_upload_file` fails because the `upload_file` tries to take the mimetype from `file`, but that mimetype is empty for some reason. I think this is a bug in the test because it doesn't specify the content type of the uploaded file in its post request.

matrss:
According to this: https://stackoverflow.com/a/51384608 flask doesn't do the right thing anyway to determine the mime type (it also just guesses from the file name and trusts the client to provide something correct). I don't want to introduce an additional dependency on python-magic though.

You could just leave out the file extension entirely, after all they aren't required on any sane system. Again, this would require a fix of `test_upload_file` though, since it currently enforces an extension for some reason.

workaryangupta:
> You could just leave out the file extension entirely, after all they aren't required on any sane system. Again, this would require a fix of `test_upload_file` though, since it currently enforces an extension for some reason.

So, should I just delete the line `assert 'png' in file`. Is that fine?

ReimarBauer:
> You could just leave out the file extension entirely, after all they aren't required on any sane system. Again, this would require a fix of `test_upload_file` though, since it currently enforces an extension for some reason.

When the file is downloaded from the chat again to disk it needs the correct extension. You see this in the context menu on right click on an  image in the context menu - Download.


https://github.com/Open-MSS/MSS/assets/278728/00c6fc29-8e66-4dfb-bd05-a2e8c94775a4

ReimarBauer:
There might be also a comatibility feature or migration script for existing servers needed when the mimetypes are not stored in the database. There should be no dataloss.

mslib/mscolab/file_manager.py
ReimarBauer: I've not looked up more on this yet:
https://github.com/Open-MSS/MSS/issues/2103

When we do this we will do a complete refactoring in hopefully one shot including docs and examples.

At current state because this is also not in the students proposal work we should keep fs.

workaryangupta:
> When the file is downloaded from the chat again to disk it needs the correct extension. You see this in the context menu >on right click on an image in the context menu - Download.


So, regarding the extension, should I just roll back to how things were earlier?

ReimarBauer:
> > When the file is downloaded from the chat again to disk it needs the correct extension. You see this in the context menu >on right click on an image in the context menu - Download.
> 
> So, regarding the extension, should I just roll back to how things were earlier?

That is on your choice, you can also improve the missing parts. When you rollback add a todo or an issue so that someone knows that it should be changed but then it should give a working version, maybe also more tests.

matrss:
> When the file is downloaded from the chat again to disk it needs the correct extension.

Oh, my bad, I only considered the profile images, which aren't downloaded to disk. In that case with the chat attachments I think the most reasonable option is to take it from the supplied filename in the request (mostly what you had, but using `mimetypes.guess_extension(mimetypes.guess_type(<filename from request>))` instead of just `guess_extension`).

mslib/mscolab/file_manager.py
```diff
@@ -250,6 +256,58 @@ def modify_user(self, user, attribute=None, value=None, action=None):
...
+    def upload_file(self, file, subfolder, identifier=None):
+        """
+        Generic function to save files securely in specified directory with unique filename
+        and return the relative file path.
+        """
+        upload_folder = mscolab_settings.UPLOAD_FOLDER
+        with fs.open_fs(upload_folder) as _fs:
+            file_dir = fs.path.join(upload_folder, str(subfolder) if subfolder else "")
+            if sys.platform.startswith('win'):
+                file_dir = file_dir.replace('\\', '/')
```
matrss: This needs to happen for `upload_folder`, before the `fs.open_fs` call.

mslib/mscolab/file_manager.py
```diff
@@ -250,6 +256,58 @@ def modify_user(self, user, attribute=None, value=None, action=None):
             db.session.commit()
         return True
 
+    def upload_file(self, file, subfolder, identifier=None):
+        """
+        Generic function to save files securely in specified directory with unique filename
+        and return the relative file path.
+        """
+        upload_folder = mscolab_settings.UPLOAD_FOLDER
+        with fs.open_fs(upload_folder) as _fs:
```
matrss: This could directly open `fs.path.join(upload_folder, str(subfolder) if subfolder else "")`, since anything outside of that doesn't need to be accessed. If `fs.open_fs` is called with `create=True` this would make the later `_fs.makedirs` call unnecessary.

mslib/mscolab/file_manager.py
```diff
@@ -250,6 +256,58 @@ def modify_user(self, user, attribute=None, value=None, action=None):
...
+            file_dir = fs.path.join(upload_folder, str(subfolder) if subfolder else "")
```
matrss: Joining the absolute path of `upload_folder` with anything else isn't necessary when using `_fs` for all filesystem operations, since those always operate within the opened directory.

mslib/mscolab/file_manager.py
```diff
@@ -250,6 +256,58 @@ def modify_user(self, user, attribute=None, value=None, action=None):
...
+
+            if identifier:
+                file_name = f'{identifier}-{timestamp}-{token}{file_ext}'
+            else:
+                file_name = f'{file_name}-{timestamp}-{token}{file_ext}'
+            file_name = secure_filename(file_name)
+
+            # Saving the file
+            file_path = fs.path.join(file_dir, file_name)
+            file.save(file_path)
```
matrss: This supplies a PyFilesystem2 path to something that expects `os.path` semantics. Instead, `file.save` can receive a file-like object:
```python
with _fs.open(fs.path.join(str(subfolder) if subfolder else "", file_name), mode="wb") as f:
    file.save(f)
```
or, if already specifying the subfolder in the `open_fs` call:
```python
with _fs.open(file_name, mode="wb") as f:
    file.save(f)
```

mslib/mscolab/file_manager.py
```diff
@@ -250,6 +256,58 @@ def modify_user(self, user, attribute=None, value=None, action=None):
...
+
+            # Relative File path
+            upload_folder = upload_folder.replace('\\', '/')
+            static_dir = fs.path.basename(upload_folder)
+            static_file_path = os.path.join(static_dir, str(subfolder), file_name)
+            static_file_path = static_file_path.replace('\\', '/')
+            logging.debug(f'Relative Path: {static_file_path}')
```
matrss: This again mixes `fs` and `os`. This could be simplified to
```python
fs.path.relativefrom(upload_folder, fs.path.join(<what was supplied to open_fs>, file_name))
```
when the subfolder is opened as `_fs`.

mslib/mscolab/file_manager.py
```diff
@@ -250,6 +256,58 @@ def modify_user(self, user, attribute=None, value=None, action=None):
...
+
+            return static_file_path
+
+    def save_user_profile_image(self, user_id, image_file):
+        """
+        Save the user's profile image path to the database.
+        """
+        relative_file_path = self.upload_file(image_file, subfolder='profile', identifier=user_id)
```
matrss, line 316: This is hardcoding the "profile" directory as the subfolder under the `UPLOAD_FOLDER`. I think this would be fine, but then we can not also have `PROFILE_IMG_FOLDER` as a config option. It must be consistent.

mslib/mscolab/server.py
```diff
@@ -351,6 +352,40 @@ def get_user():
...
+@APP.route('/fetch_profile_image', methods=["GET"])
+@verify_user
+def fetch_profile_image():
+    user_id = request.form['user_id']
+    user = User.query.get(user_id)
+    if user and user.profile_image_path:
+        # base_directory = mscolab_settings.UPLOAD_FOLDER
+        # base_path = os.path.join(base_directory, 'profile')
+        base_path = mscolab_settings.PROFILE_IMG_FOLDER
+        filename = os.path.basename(user.profile_image_path)
```
matrss: The image path should be saved in a way in which it is relative to some configurable path (either `UPLOAD_FOLDER` or `PROFILE_IMG_FOLDER`). Using basename suggests that the saved path currently contains more information than it should (and using basename might throw away more than it should).

With my suggestions the saved path would be relative to `mscolab_settings.UPLOAD_FOLDER`. Relative to `PROFILE_IMG_FOLDER` would also be possible, that depends on if we want to make that configurable or not (again, see the other remark that it is currently configurable _and_ hardcoded in different places).

mslib/mscolab/server.py
```diff
@@ -351,6 +352,40 @@ def get_user():
...
+        return send_from_directory(base_path, filename)
```
matrss: At this point the whole `fs` abstraction breaks down, because `send_from_directory` expects `os.path` semantics and we shouldn't really use anything else. Until here, everything should be treated as `fs`, and then can be somewhat cleanly converted to something `send_from_directory` understands with `send_from_directory(_fs.getsyspath(""), user.profile_image_path)` (`_fs` would need to be opened first).

ReimarBauer:
trying manually:

starting server
```
rm -rf ~/colabdata
python mslib/mscolab/mscolab.py db --init
 python mslib/mscolab/mscolab.py start
```

- open msui
- python mslib/msui/msui.py 
- connect
- add user

The user is logged in, that works. 

But the server outputs

```
ERROR: Exception on /fetch_profile_image [GET]
  File "/home/user/Miniforge/envs/mssdev/lib/python3.11/site-packages/flask/app.py", line 2190, in wsgi_app
    response = self.full_dispatch_request()
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/user/Miniforge/envs/mssdev/lib/python3.11/site-packages/flask/app.py", line 1487, in full_dispatch_request
    return self.finalize_request(rv)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/user/Miniforge/envs/mssdev/lib/python3.11/site-packages/flask/app.py", line 1506, in finalize_request
    response = self.make_response(rv)
               ^^^^^^^^^^^^^^^^^^^^^^
  File "/home/user/Miniforge/envs/mssdev/lib/python3.11/site-packages/flask/app.py", line 1837, in make_response
    raise TypeError(
TypeError: The view function did not return a valid response. The return type must be a string, dict, list, tuple with headers or status, Response instance, or WSGI callable, but it was a int.
```

You need to check for a new user, or also existing users which don't have an image set.

mslib/mscolab/server.py
```diff
@@ -351,6 +352,40 @@ def get_user():
...
+    else:
+        return 404
```
ReimarBauer: this must be done differently, usually we use as return by the server json. For the case no file is there it may be enough to do nothing.

mslib/msui/mscolab.py
```diff
@@ -798,16 +825,51 @@ def on_context_menu(point):
...
+            # Resize the image and load it into QPixmap
+            image = Image.open(file_name)
+            image = image.resize((64, 64), Image.ANTIALIAS)
+            img_byte_arr = io.BytesIO()
+            image.save(img_byte_arr, format='PNG')
+            img_byte_arr.seek(0)
+
+            pixmap = QPixmap()
+            pixmap.loadFromData(img_byte_arr.getvalue())
+            self.profile_dialog.gravatarLabel.setPixmap(pixmap.scaled(64, 64))
```
matrss: Semantically the same thing is also done in `fetch_profile_image`. Give it a name and make it its own method.

matrss (changes requested):

ReimarBauer:
When a profile image is uploaded the small icon did update on the next request of the profile image
It also not changes immeaditly when the image gets updated

e.g.

https://github.com/Open-MSS/MSS/assets/278728/f97c5a57-89ae-476b-a537-0490c518ec5d


With an ok, and a changed image I would expect to see it.

mslib/mscolab/server.py
matrss: A 404 is a good choice if there is no profile image set for the user. That's what the "Not Found" status expresses in this context. But with flask this would need to be `abort(404)` instead of a return.

mslib/msui/mscolab.py
```diff
@@ -798,16 +825,51 @@ def on_context_menu(point):
...
+
+            # Prepare the file data for upload
+            try:
+                img_byte_arr.seek(0)  # Reset buffer position
+                files = {'image': (os.path.basename(file_name), img_byte_arr, 'image/png')}
+                data = {
```
ReimarBauer: an uploaded jpg keeps its extension but is saved as png internally

```
~/colabdata/uploads/profile$ file *
1-20240704T173912-84N0itmzs-5tFn66Kmoxze8e6P1JQqZYohJMZo45qcs.jpg: PNG image data, 64 x 64, 8-bit/color RGB, non-interlaced
1-20240704T174129-nT7DKeSC1xttahv5281m7_eFfEOIhiSN1cSCIWFGvg0.png: PNG image data, 64 x 64, 8-bit/color RGBA, non-interlaced
1-20240704T174326-85GpnpCRFZhhW7BqVq3-LJGtK6IPVM8HycdUBjNFejw.png: PNG image data, 64 x 64, 8-bit/color RGBA, non-interlaced
```

mslib/mscolab/server.py
workaryangupta: @matrss The image path is actually saved relative to the `UPLOAD_FOLDER`
For example : `uploads/profile/1-20240627T210302-qqJvzG6QD3OrzBPAZKMkYbXArgI.jpg`
`os.path.basename(user.profile_image_path)` in this case is just used to get the file name `1-20240627T210302-qqJvzG6QD3OrzBPAZKMkYbXArgI.jpg`. I don't get whats wrong with this?

matrss: That's not relative to UPLOAD_FOLDER, that is relative to the parent directory of UPLOAD_FOLDER. Taking the basename would fail if there was another directory in the path. If `user.profile_image_path` didn't start with uploads/ they could just be joined, without first using basename.

mslib/mscolab/file_manager.py
workaryangupta: @matrss I refactored the code according to this, now the static path is returned is relative to the uploads directory, which works fine for profile images,
BUT, as this function also handles chat attachment functionality, the chat attachments now return a path like 
`6/{filename}`, which breaks the code for adding the messages to the chat as it expects a path like `uploads/6/{filename}`
Please suggest me what I must do.

matrss: Hmm, without seeing the exact error I can only guess, but if the chat attachment stuff breaks then that code has the same bug and also needs to be fixed.

But that is a bit more work: it would require a database migration to fix the wrong paths that are already saved and the migration infrastructure to do that properly doesn't exist yet. Also, that would be well out of scope for your project.

I think you could just introduce a keyword argument to specify if the method returns the path with or without the prefix to allow the chat attachment code to get the old wrongly-prefixed path, while using the proper path for profile images. The default should be to not add the prefix.

That can then be fixed and refactored later.

@ReimarBauer does that sound reasonable or do you have a different idea?

workaryangupta:
> When a profile image is uploaded the small icon did update on the next request of the profile image It also not changes immeaditly when the image gets updated
> With an ok, and a changed image I would expect to see it.

Fixed it, now when you upload an image, it immediately updates on both the small pic and the gravatar label.

mslib/msui/mscolab.py
```diff
@@ -798,16 +827,53 @@ def on_context_menu(point):
...
+
+    def upload_image(self):
+        file_name, _ = QFileDialog.getOpenFileName(self.prof_diag, "Open Image", "", "Image Files (*.png *.jpg *.jpeg)")
+        if file_name:
+            # Determine the image format
+            mime_type, _ = mimetypes.guess_type(file_name)
+            file_format = mime_type.split('/')[1].upper()
+
+            # Resize the image and set profile image pixmap
+            image = Image.open(file_name)
```
ReimarBauer: this needs an try: except PIL.UnidentifiedImageError: because one can have wrong extensions used, e.g. renamed a mp4 to a jpg

currently this gives a traceback:



Fatal error in MSS 9.0.0 on Linux-6.5.0-10043-tuxedo-x86_64-with-glibc2.35

```

  File "/home/user/PycharmProjects/2024/workaryangupta/MSS/mslib/msui/mscolab.py", line 850, in upload_image
    image = Image.open(file_name)
            ^^^^^^^^^^^^^^^^^^^^^
  File "/home/user/Miniforge/envs/mssdev/lib/python3.11/site-packages/PIL/Image.py", line 3283, in open
    raise UnidentifiedImageError(msg)
PIL.UnidentifiedImageError: cannot identify image file '/home/user/tmp/1.jpg'

```

ReimarBauer: renaming a gif to a png and uploading works, so gif can maybe added to the list of extensions

ReimarBauer: renaming a gif to a jpg and uploading gives

```

  File "/home/user/Miniforge/envs/mssdev/lib/python3.11/site-packages/PIL/JpegImagePlugin.py", line 643, in _save
    rawmode = RAWMODE[im.mode]
              ~~~~~~~^^^^^^^^^
KeyError: 'P'

The above exception was the direct cause of the following exception:

  File "/home/user/PycharmProjects/2024/workaryangupta/MSS/mslib/msui/mscolab.py", line 853, in upload_image
    image.save(img_byte_arr, format=file_format)
  File "/home/user/Miniforge/envs/mssdev/lib/python3.11/site-packages/PIL/Image.py", line 2431, in save
    save_handler(self, fp, filename)
  File "/home/user/Miniforge/envs/mssdev/lib/python3.11/site-packages/PIL/JpegImagePlugin.py", line 646, in _save
    raise OSError(msg) from e
OSError: cannot write mode P as JPEG
```

So also catch OSError

 Becasue this is not server code it is not critical, we help here the user and maybe get no issues when someone has files like this.

mslib/mscolab/server.py
```diff
@@ -351,6 +351,38 @@ def get_user():
...
+        base_path = mscolab_settings.UPLOAD_FOLDER
+        filename = user.profile_image_path
+        return send_from_directory(base_path, filename)
+    else:
+        abort(404)
+
+
 @APP.route("/delete_own_account", methods=["POST"])
 @verify_user
 def delete_own_account():
```
ReimarBauer, line 393: when a user deletes his own account, we also have to delete the profile images, this is not much effort from a GDPR point of view. GDPR is also the reason why we have this button.

ReimarBauer: Because all images are kept on uploading a new one, we need to remove all profile images of that user not only the most recent.

mslib/mscolab/server.py
```diff
@@ -351,6 +351,38 @@ def get_user():
     return json.dumps({'user': {'id': g.user.id, 'username': g.user.username}})
 
 
+@APP.route('/upload_profile_image', methods=["POST"])
+@verify_user
+def upload_profile_image():
+    user_id = request.form['user_id']
```
ReimarBauer, line 359: telegram has an option to slide through old profile images. Currently we save all versions of a ever used profile image. 

What's your plan with this?

mslib/mscolab/server.py
ReimarBauer: added this to https://github.com/Open-MSS/MSS/issues/2103

ReimarBauer: @workaryangupta 
Here a ToDo can be added, e.g. see open Discussion about PyFilesystem2 on ticket
or it can be solved as @matrss wrote.

ReimarBauer (changes requested):
see comments

mslib/mscolab/server.py
workaryangupta: Not anymore, we delete the old image when saving a new one!

mslib/mscolab/server.py
workaryangupta: done

ReimarBauer (changes requested):
When I have a new user created add an image and logout and create the next user and logout and then login thee first user I get
```
  File "/home/user/PycharmProjects/2024/workaryangupta/MSS/mslib/msui/mscolab.py", line 353, in login_handler
    self.mscolab.after_login(data["email"], self.mscolab_server_url, r)
  File "/home/user/PycharmProjects/2024/workaryangupta/MSS/mslib/msui/mscolab.py", line 682, in after_login
    self.fetch_profile_image()
  File "/home/user/PycharmProjects/2024/workaryangupta/MSS/mslib/msui/mscolab.py", line 720, in fetch_profile_image
    self.set_profile_pixmap(response.content)
  File "/home/user/PycharmProjects/2024/workaryangupta/MSS/mslib/msui/mscolab.py", line 705, in set_profile_pixmap
    self.profile_dialog.gravatarLabel.setPixmap(resized_pixmap)
RuntimeError: wrapped C/C++ object of type QLabel has been deleted
CRITICAL: MSS Version: 9.0.0
CRITICAL: Python Version: 3.11.6 | packaged by conda-forge | (main, Oct  3 2023, 10:40:35) [GCC 12.3.0]
CRITICAL: Platform: Linux-6.5.0-10043-tuxedo-x86_64-with-glibc2.35 (('64bit', 'ELF'))
  File "/home/user/PycharmProjects/2024/workaryangupta/MSS/mslib/msui/mscolab.py", line 353, in login_handler
    self.mscolab.after_login(data["email"], self.mscolab_server_url, r)
  File "/home/user/PycharmProjects/2024/workaryangupta/MSS/mslib/msui/mscolab.py", line 682, in after_login
    self.fetch_profile_image()
  File "/home/user/PycharmProjects/2024/workaryangupta/MSS/mslib/msui/mscolab.py", line 720, in fetch_profile_image
    self.set_profile_pixmap(response.content)
  File "/home/user/PycharmProjects/2024/workaryangupta/MSS/mslib/msui/mscolab.py", line 705, in set_profile_pixmap
    self.profile_dialog.gravatarLabel.setPixmap(resized_pixmap)
RuntimeError: wrapped C/C++ object of type QLabel has been deleted
```


https://github.com/Open-MSS/MSS/assets/278728/431abc53-b225-4211-9259-b5e0c2bfc708


you can also when you want to populate the db with some test users by


`mscolab.py db --seed`

The data is described:
https://github.com/Open-MSS/MSS/blob/develop/mslib/mscolab/seed.py#L228

only users which don't have a profile image can login without this failure

ReimarBauer:
on the first two users from the seed example I added the profile image, same problem.



https://github.com/Open-MSS/MSS/assets/278728/7aa2dae8-ce6e-420d-88b6-f0b9460b72a1

workaryangupta:
I am looking into why this is happening. Could be somewhat related to how UI elements are set after the logout.

mslib/mscolab/file_manager.py
```diff
@@ -250,6 +258,71 @@ def modify_user(self, user, attribute=None, value=None, action=None):
             db.session.commit()
         return True
 
+    def delete_user_profile_images(self, image_to_be_deleted):
```
ReimarBauer: singular, it is only deleting one image

mslib/mscolab/server.py
```diff
@@ -351,6 +351,38 @@ def get_user():
...
+
+@APP.route('/fetch_profile_image', methods=["GET"])
+@verify_user
+def fetch_profile_image():
+    user_id = request.form['user_id']
+    user = User.query.get(user_id)
+    if user and user.profile_image_path:
+        base_path = mscolab_settings.UPLOAD_FOLDER
+        filename = user.profile_image_path
+        return send_from_directory(base_path, filename)  # todo : Supply os path semantics in this and def uploads()
```
ReimarBauer: your todo is what it is currently. 

A ToDo means there can be done something I ack this but we do it later.

so you better write

`#` ToDo needs to use fs for the path handling and send_directory needs os.path see discussion on #2103

mslib/msui/mscolab.py
```diff
@@ -798,16 +830,60 @@ def on_context_menu(point):
...
         self.profile_dialog.gravatarLabel.setContextMenuPolicy(QtCore.Qt.CustomContextMenu)
         self.profile_dialog.gravatarLabel.customContextMenuRequested.connect(on_context_menu)
 
         self.prof_diag.show()
-        self.fetch_gravatar()
+        self.fetch_profile_image()
+
+    def upload_image(self):
+        file_name, _ = QFileDialog.getOpenFileName(self.prof_diag, "Open Image", "", "Image Files (*.png *.jpg *.jpeg)")
+        if file_name:
```
ReimarBauer, line 848: is there a reason not to use the same image definition as in the chat?
image

workaryangupta: No reason, I'll change it to this

mslib/msui/mscolab.py
```diff
@@ -693,7 +697,35 @@ def after_login(self, emailid, url, r):
 
             self.signal_login_mscolab.emit(self.mscolab_server_url, self.token)
 
-    def fetch_gravatar(self, refresh=False):
+    def set_profile_pixmap(self, img_data):
+        pixmap = QPixmap()
+        pixmap.loadFromData(img_data)
+        resized_pixmap = pixmap.scaled(64, 64)
+
+        if (hasattr(self, 'profile_dialog') and self.profile_dialog is not None and
```
ReimarBauer, line 706: here you can add also a Todo
e.g. verify by a test if the condition can be simplified, by if self.profile_dialog is not None

workaryangupta: well, I tried it and it did not work
still I might add a todo to check if the condition could be simplified. Maybe there could be a scope

mslib/msui/mscolab.py
```diff
@@ -693,7 +697,35 @@ def after_login(self, emailid, url, r):
...
+                hasattr(self.profile_dialog, 'gravatarLabel') and self.profile_dialog.gravatarLabel is not None):
+            self.profile_dialog.gravatarLabel.setPixmap(resized_pixmap)
+
+        if hasattr(self, 'ui'):
+            icon = QtGui.QIcon()
```
ReimarBauer: likly this hasattr has also some yet hidden problems.

mslib/msui/mscolab.py
ReimarBauer: that's not for now, this needs a test coverage

mslib/msui/mscolab.py
ReimarBauer: when is it not True?

workaryangupta: my bad, it is truthy

workaryangupta: I think I did it as a preventive measure, but it is not needed

mslib/mscolab/file_manager.py
ReimarBauer: I'm fine with the suggested solution.

for the chat window this is the operation id, there can be many images. So that needs better a subdir.

ReimarBauer: @workaryangupta please add also a ToDo here

ToDo: add a namespace for the chat, similar as for profile

https://github.com/workaryangupta/MSS/blob/feat/upload-profile-img/mslib/mscolab/file_manager.py#L300

ReimarBauer: with a manually migration the existing chat message attachment can be seen. so that needs no additional migration, but when we implement the namespace it definitly does.

ReimarBauer (changes requested):
Almost all done, a last ToDo should be added to the file_manager that we ack to add a namespace for the chat, see comments

ReimarBauer (approved):
Thx, for the pretty nice feature

mslib/mscolab/server.py
matrss: I disagree, this shouldn't be a todo, this is a bug that has already bitten us in other places. The fix is trivial, so it can be done here:
```python
        base_path = mscolab_settings.UPLOAD_FOLDER
        if sys.platform.startswith('win'):
            base_path = base_path.replace('\\', '/')
        filename = user.profile_image_path
        with fs.open_fs(base_path) as _fs:
            return send_from_directory(_fs.getsyspath(""), filename)
```

mslib/mscolab/file_manager.py
```diff
@@ -250,6 +258,71 @@ def modify_user(self, user, attribute=None, value=None, action=None):
             db.session.commit()
         return True
 
+    def delete_user_profile_image(self, image_to_be_deleted):
+        '''
+        This function is called when deleting account or updating the profile picture
+        '''
+        upload_folder = mscolab_settings.UPLOAD_FOLDER
+        with fs.open_fs(upload_folder) as profile_fs:
```
matrss, line 269: This uses UPLOAD_FOLDER without first replacing backslashes with slashes on Windows.

matrss (changes requested):

matrss:
I'll do a thorough review tomorrow, but I think this is looking good now.

ReimarBauer:
To be noted I looked that also on the send_from_directory before. Itselfs it has a builtin secureness by safe_join. But I also prefer the way it is done now.  

https://github.com/pallets/werkzeug/blob/main/src/werkzeug/utils.py#L538

https://github.com/pallets/werkzeug/blob/main/src/werkzeug/security.py#L131

Edit: not readed correctly it is that windows backslash. Thx.

matrss:
@ReimarBauer I think you are replying to https://github.com/Open-MSS/MSS/pull/2405#discussion_r1667654129, right? If that is the case: the change was neither for path sanitization nor for the backslash/slash conversion. It was to properly interpret UPLOAD_FOLDER as an fs URL and then use the only supported option to convert those to a path-like value (getsyspath). It would have errored previously as soon as someone actually tried to use something other than a plain path (e.g. an `osfs://` would have failed due to the mis-interpretation of that config option). Now the only requirement is that it is convertible to a local path (which still mostly defeats the purpose for fs, but at least it won't try to erroneously open a directory named after the fs scheme).

mslib/mscolab/file_manager.py
```diff
@@ -250,6 +258,74 @@ def modify_user(self, user, attribute=None, value=None, action=None):
             db.session.commit()
         return True
 
+    def delete_user_profile_image(self, image_to_be_deleted):
```
matrss, line 261: Just a sidenote, nothing to do now: since this method is named `delete_user_profile_image` I think it would semantically make more sense to pass a user to this method, instead of a path.

It works as-is though, so this is something that could be refactored in the future.

matrss (approved):
LGTM, Thanks!

AFAICT there is currently no way to unset the profile image after one was set (it can just be replaced with a new one, not brought back into a state in which there is none set for the account). That could be a future improvement.

mslib/mscolab/file_manager.py
workaryangupta: Okay

PR #2419, opened 2024-06-28
add JuRSE badge

### LINKED ISSUE #2415: congrats and new badge from JuRSE
**Congrats all** 

we were selected for https://www.fz-juelich.de/en/rse/jurse-community/jurse-code-of-the-month/july-2024

https://img.shields.io/badge/JuRSE_Code_Pick-July_2024-blue?link=https://www.fz-juelich.de/en/rse/jurse-community/jurse-code-of-the-month/july-2024

We should add this to
- RTFD
- README
- pages
- Helmholtz Software
- blackduck

ReimarBauer:
**Purpose of PR?**:

Fixes #2415

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,11 +72,15 @@ Current release info
 ====================
 [![Conda Version](https://img.shields.io/conda/vn/conda-forge/mss.svg)](https://anaconda.org/conda-forge/mss)
 [![DOI](https://zenodo.org/badge/DOI/10.5281/zenodo.6572620.svg)](https://doi.org/10.5281/zenodo.6572620)
+[![JuRSE Code Pick](https://img.shields.io/badge/JuRSE_Code_Pick-July_2024-blue)](https://www.fz-juelich.de/en/rse/jurse-community/jurse-code-of-the-month/july-2024)
 [![Conda Platforms](https://img.shields.io/conda/pn/conda-forge/mss.svg)](https://anaconda.org/conda-forge/mss)
 [![DOCS](https://img.shields.io/badge/%F0%9F%95%AE-docs-green.svg)](http://mss.rtd.io)
 [![Conda Recipe](https://img.shields.io/badge/recipe-mss-green.svg)](https://anaconda.org/conda-forge/mss)
 [![Conda Downloads](https://img.shields.io/conda/dn/conda-forge/mss.svg)](https://anaconda.org/conda-forge/mss)
 [![Coverage Status](https://coveralls.io/repos/github/Open-MSS/MSS/badge.svg?branch=develop)](https://coveralls.io/github/Open-MSS/MSS?branch=develop)
+
+
+
 
 
 Publications
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -3,7 +3,7 @@
 
 Mission Support System's Documentation
 ======================================
-|Conda Version| |GitHub Issues| |Coverage Status| |DOI| |Conda Platforms| |Conda Recipe| |Conda Downloads|
+|Conda Version| |DOI| |JuRSE Code Pick| |GitHub Issues| |Coverage Status| |Conda Platforms| |Conda Recipe| |Conda Downloads|
 
 .. |Conda Version| image:: https://img.shields.io/conda/vn/conda-forge/mss.svg
    :target: https://anaconda.org/conda-forge/mss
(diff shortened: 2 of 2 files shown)


gisi90 (approved):

README.md
```diff
@@ -72,13 +72,17 @@ Current release info
...
 [![Conda Platforms](https://img.shields.io/conda/pn/conda-forge/mss.svg)](https://anaconda.org/conda-forge/mss)
 [![DOCS](https://img.shields.io/badge/%F0%9F%95%AE-docs-green.svg)](http://mss.rtd.io)
 [![Conda Recipe](https://img.shields.io/badge/recipe-mss-green.svg)](https://anaconda.org/conda-forge/mss)
 [![Conda Downloads](https://img.shields.io/conda/dn/conda-forge/mss.svg)](https://anaconda.org/conda-forge/mss)
 [![Coverage Status](https://coveralls.io/repos/github/Open-MSS/MSS/badge.svg?branch=develop)](https://coveralls.io/github/Open-MSS/MSS?branch=develop)
 
 
+
+
+
```
matrss, line 85: Unnecessary whitespace change

ReimarBauer: That will be fixed on next time changing that file.

matrss: Yea, I'm pretty sure it won't, unless we add a linter for that. Insane amounts of blank lines are also everywhere in the docs. Not that this is really important or that I care too much, otherwise I would have made this a blocking review.

ReimarBauer: It happened because of arranging the badges. We will have more badges. I got some ideas by libertem @uellue and @sk1p :)

sk1p: > Yea, I'm pretty sure it won't, unless we add a linter for that. Insane amounts of blank lines are also everywhere in the docs.

Just because I was mentioned, have you considered adding [pre-commit](https://pre-commit.com/) hooks for this? It's mostly a few lines of configuration and it'll automatically fix files with issues like this for you before they even end up in the commit/PR/repo (and you can use [pre-commit-ci](https://pre-commit.ci/) to very quickly run these checks as part of your CI, together with other linters and formatters like blank, flake8, ruff, or whatever you are using, in case a contributor doesn't use the local git hooks).

A `.pre-commit-config.yaml` like this should do the job:

```yaml
repos:
- repo: https://github.com/pre-commit/pre-commit-hooks
  rev: v4.6.0
  hooks:
  - id: trailing-whitespace
    files: ^(mslib|docs)/  # <- just an example, in reality this can be a bit more complicated, with excludes etc.
```

\</ad\> :smile: 

Reimar, let me know if this is interesting to you, and if I should show you anything in-person!

matrss: I've used pre-commit in different projects in the past, but I remember it having issues on NixOS because it made invalid FHS related assumptions (at least with the official usage method). Maybe that has changed in the meantime. The "shift-left" idea in this case is good though, as long as it doesn't introduce noticeable slowdowns.

That said, I don't think the trailing-whitespace hook (or any hook provided by pre-commit, really) would help in this particular case. This isn't about trailing whitespace at EOL or even multiple blank lines at EOF, this is about multiple blank lines (5 in this case) where 1, or arguably maybe 2, would have the same meaning and 5 are just excessive. We also have that a lot in the rst docs.

sk1p: > That said, I don't think the trailing-whitespace hook (or any hook provided by pre-commit, really) would help in this particular case. This isn't about trailing whitespace at EOL or even multiple blank lines at EOF, this is about multiple blank lines (5 in this case) where 1, or arguably maybe 2, would have the same meaning and 5 are just excessive. We also have that a lot in the rst docs.

Ah, gotcha, then it's probably more in the realm of something like [docstrfmt](https://github.com/LilSpazJoekp/docstrfmt), which I haven't use yet.

ReimarBauer: > Just because I was mentioned, have you considered adding [pre-commit](https://pre-commit.com/) hooks for this?

I've seen the maintainabilty code analysis badge and the need of a JOSS publication for us too. :)
Have you used codeclimate now longer and can exchange some of its capabilities?
I know something like this from the past, but is was not further developed, clonedigger (https://sourceforge.net/projects/clonedigger/) and quantifiedcode (https://github.com/quantifiedcode/quantifiedcode). quantifiedcode was pretty good for detecting antipatterns.  In the first years of MSS it was helpful to find what else needs refactoring.
Added a ticket for that:
https://github.com/Open-MSS/MSS/issues/2424

PR #2425, opened 2024-07-03
fix: interpolation to flight track altitude

### LINKED ISSUE #2409: Linear view: interpolation to flight track altitude?
Does linear view interpolate to the flight altitude of the flight track? 
I just realized this is not the case for our WMS. 
grafik

joernu76:
**Purpose of PR?**:

Fixes #2409

No

Checked on PHILEAS data in PL.


gisi90 (approved):
Works! (see comments #2409 )

PR #2332, opened 2024-04-20
Merge stable to develop

ReimarBauer:
**Purpose of PR?**:

This needs a merge no squash

PR #2427, opened 2024-07-04
Merge stable to develop

ReimarBauer:
**Purpose of PR?**:

merge stable to develop


matrss (commented):
This PR mixes a bunch of unrelated changes, which makes it basically impossible to review for correctness of the intention of each change when applied to develop.

It was stated in #2332 that these PRs shouldn't be reviewed for the correctness of the intention of each commit, but instead just for if they are technically applicable to the target branch. But since I only find the former worthwhile for a review (after all, the latter is already done automatically by git through merge conflicts) I won't review this, as also stated in #2332.

gisi90 (approved):
fixes mostly typos

PR #2431, opened 2024-07-08
feat: Allow users to set custom profile pictures (#2405)

### LINKED ISSUE #1287: upload of avatar image for MSC

ReimarBauer:
Enhancing the functionality of the MSColab server to allow users to upload small images which would replace generic initials as identifiers during login. This feature personalizes user accounts and significantly enhances the visual aspect of user identification and interaction within the MSColab platform.

**Purpose of PR?**:
Part of the GSoC 2024 improve MSUI project:  Introduce a new feature to mscolab, enabling users to upload and set personalized profile pictures.

Fixes #1287


mslib/mscolab/models.py
```diff
@@ -58,16 +58,19 @@ class User(db.Model):
     username = db.Column(db.String(255))
     emailid = db.Column(db.String(255), unique=True)
     password = db.Column(db.String(255))
+    profile_image_path = db.Column(db.String(255), nullable=True)  # relative path
```
matrss, line 61: We definitely need that proper migration machinery and an automated check for model changes, then we would have noticed that this requires a migration... I am working on that already, so I think we can add that later, but it would mean that develop would temporarily not be in a releasable state.

ReimarBauer: We recognized this also during the work on that feature, it was first a kind of datastore. We discussed there also what a developer should do and at that time "create a new database because its not finished".

And it is not more a problem than already we have since some version. The old option is deprecated for the new major version. Then when your work is finished we have a better solution. Great that you work on that. :)

Because we announced that change we can make a major release only after the correct migration solution is implemented. That is a burden that we impose on ourself with deprecations. Somehow the release is on hold ...

Without a change there would also no migration needed. That is now the first difference to 9.x

At the moment we principle can release we can lookup what from the others GSOC project also can be merged and do a new release. Then we can do the next deprecation etc.  

Currently we reset the database for tests. That's why the current setup did not fail.

matrss (approved):

ReimarBauer:
reminder: never delete a GSOC branch

PR #2436, opened 2024-07-10
tests: Tests for feature - Custom Profile Images

workaryangupta:
This PR contains tests for testing the new feature - upload custom profile images

### DIFF
--- a/mslib/mscolab/server.py
+++ b/mslib/mscolab/server.py
@@ -356,7 +356,7 @@ def get_user():
 @APP.route('/upload_profile_image', methods=["POST"])
 @verify_user
 def upload_profile_image():
-    user_id = request.form['user_id']
+    user_id = g.user.id
     file = request.files['image']
     if not file:
         return jsonify({'message': 'No file provided or invalid file type'}), 400
@@ -375,7 +375,7 @@ def upload_profile_image():
 @APP.route('/fetch_profile_image', methods=["GET"])
 @verify_user
 def fetch_profile_image():
-    user_id = request.form['user_id']
+    user_id = g.user.id
     user = User.query.get(user_id)
     if user and user.profile_image_path:
         base_path = mscolab_settings.UPLOAD_FOLDER
--- a/mslib/mscolab/server.py
+++ b/mslib/mscolab/server.py
@@ -375,7 +375,7 @@ def upload_profile_image():
 @APP.route('/fetch_profile_image', methods=["GET"])
 @verify_user
 def fetch_profile_image():
-    user_id = g.user.id
+    user_id = request.form['user_id']
     user = User.query.get(user_id)
     if user and user.profile_image_path:
         base_path = mscolab_settings.UPLOAD_FOLDER
--- a/tests/_test_mscolab/test_file_manager.py
+++ b/tests/_test_mscolab/test_file_manager.py
@@ -32,6 +32,7 @@ from werkzeug.datastructures import FileStorage
 
 from mslib.mscolab.models import Operation, User
 from mslib.mscolab.seed import add_user, get_user, add_operation
+from mslib.mscolab.conf import mscolab_settings
 
 
 class Test_FileManager:
@@ -267,6 +268,24 @@ class Test_FileManager:
             static_path = self.fm.upload_file(file, subfolder=str(operation.id), identifier=None)
             assert name in static_path
             assert static_path.endswith(ext)
+
+    def test_upload_file(self):
+        sample_file_path = os.path.join(os.path.dirname(__file__), "..", "data")
+        filename = "example.txt"
+        _, ext = filename.split('.')
+
+        open_txt = os.path.join(sample_file_path, "example.txt")
+        with open(open_txt, 'rb') as fp:
+            file = FileStorage(fp, filename=filename, content_type="text/plain")
+
+            subfolder = 'test_subfolder'
+            identifier = 'unique_identifier'
+            relative_path = self.fm.upload_file(file, subfolder=subfolder, identifier=identifier)
+
+            assert os.path.isfile(os.path.join(mscolab_settings.UPLOAD_FOLDER, relative_path))
+            assert identifier in relative_path
+            assert subfolder in relative_path
+            assert relative_path.endswith(ext)
 
     def test_get_file(self):
         with self.app.test_client():
--- a/tests/_test_mscolab/test_file_manager.py
+++ b/tests/_test_mscolab/test_file_manager.py
@@ -276,16 +276,22 @@ class Test_FileManager:
 
         open_txt = os.path.join(sample_file_path, "example.txt")
         with open(open_txt, 'rb') as fp:
+            file_content = fp.read()
+            fp.seek(0)                  # reset file pointer
             file = FileStorage(fp, filename=filename, content_type="text/plain")
-
             subfolder = 'test_subfolder'
             identifier = 'unique_identifier'
             relative_path = self.fm.upload_file(file, subfolder=subfolder, identifier=identifier)
+            full_path = os.path.join(mscolab_settings.UPLOAD_FOLDER, relative_path)
 
-            assert os.path.isfile(os.path.join(mscolab_settings.UPLOAD_FOLDER, relative_path))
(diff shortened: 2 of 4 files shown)


tests/_test_msui/test_mscolab.py
```diff
@@ -901,6 +904,39 @@ def test_profile_dialog(self, qtbot):
...
+        initial_pixmap = self.window.mscolab.profile_dialog.gravatarLabel.pixmap().toImage()
+
+        # Creating a new image
+        new_image = Image.new('RGB', (64, 64), color='cyan')
+        img_byte_arr = io.BytesIO()
+        new_image.save(img_byte_arr, format='JPEG')
+        img_byte_arr.seek(0)
+
+        # Path where the image will be temporarily stored
+        temp_image_path = os.path.join(os.getcwd(), 'new_profile_image.jpg')
```
matrss: Instead of writing to CWD this should use a temporary path. You can use the `tmp_path` pytest fixture to get one: https://docs.pytest.org/en/latest/reference/reference.html#tmp-path.

This would also make the explicit cleanup at the end unnecessary.

tests/_test_msui/test_mscolab.py
```diff
@@ -901,6 +904,39 @@ def test_profile_dialog(self, qtbot):
...
+
+        # Saving the imaeg on path
+        with open(temp_image_path, 'wb') as f:
+            f.write(img_byte_arr.getvalue())
+
+        # Mocking the QFileDialog to select the image
+        with mock.patch('PyQt5.QtWidgets.QFileDialog.getOpenFileName', return_value=(temp_image_path, 'Image (*.jpg)')):
+            self.window.mscolab.upload_image()
+
+        qtbot.wait(1000)
```
matrss: `qtbot.wait` should only be used as a last resort, since using it doesn't solve the underlying race condition, it just hides it (most of the time). It is better to use `qtbot.wait_until` and an assertion that ensures whatever is waited for has happened. `wait_until` would also only wait for as long as necessary, while a static duration with `wait` will be too long most of the time and slow the test suite down.

tests/_test_msui/test_mscolab.py
```diff
@@ -901,6 +904,39 @@ def test_profile_dialog(self, qtbot):
...
+        self.window.mscolab.fetch_profile_image(refresh=True)
```
matrss: I think there should be some assert for the updated pixmap before the `fetch_profile_image` call, because it is expected that `upload_image` updates the local images already, right?

This is related to not using `qtbot.wait`.

matrss (changes requested):

tests/_test_msui/test_mscolab.py
workaryangupta: Yes you are right, it would definitely make this test more robust

tests/_test_mscolab/test_server.py
```diff
@@ -139,6 +142,19 @@ def test_delete_user(self):
...
+            # create user again, this time the user has a profile image set
+            assert add_user(self.userdata[0], self.userdata[1], self.userdata[2])
+            token = self._get_token(test_client, self.userdata)
+            response = self._upload_profile_image(test_client, token, self.userdata[0])
+            assert response.status_code == 200  # this will ensure image was uploaded
+            response = test_client.post('/delete_own_account', data={"token": token})
+            assert response.status_code == 200
+            data = json.loads(response.data.decode('utf-8'))
+            assert data["success"] is True
+            response = test_client.post('/delete_own_account', data={"token": "dsdsds"})
```
ReimarBauer: what should this verify regarding to delete the own user account? The account was deleted. The token used is wrong.

tests/_test_mscolab/test_server.py
```diff
@@ -139,6 +142,19 @@ def test_delete_user(self):
...
+            assert response.status_code == 200
+            assert response.data.decode('utf-8') == "False"
+
```
ReimarBauer, line 180: is this utf-8 decode necessary?

ReimarBauer (commented):
There should be a test for an oversized image
MAX_UPLOAD_SIZE

It looks like that on the chat there is no feedback to the user and nothing is uploaded.

tests/_test_mscolab/test_file_manager.py
```diff
@@ -268,6 +269,24 @@ def test_upload_chat_attachment(self):
...
+            file = FileStorage(fp, filename=filename, content_type="text/plain")
+
+            subfolder = 'test_subfolder'
+            identifier = 'unique_identifier'
+            relative_path = self.fm.upload_file(file, subfolder=subfolder, identifier=identifier)
+
+            assert os.path.isfile(os.path.join(mscolab_settings.UPLOAD_FOLDER, relative_path))
+            assert identifier in relative_path
+            assert subfolder in relative_path
+            assert relative_path.endswith(ext)
```
matrss, line 290: I think there should also be an assert that compares the file content of the uploaded file for equality with the sample file that was uploaded.

tests/_test_mscolab/test_server.py
matrss, line 1: It would be great to have another test that ensures a user cannot upload a profile image for a different user.

matrss: And that an unauthenticated client cannot upload anything.

matrss (changes requested):

ReimarBauer (changes requested):
take care to use in the server code  g.user.id and not something submitted

and the test can verify the need for this

tests/_test_mscolab/test_server.py
```diff
@@ -130,14 +132,45 @@ def test_get_user(self):
...
+    def test_unauthorized_profile_image_upload(self):
+        other_user_data = 'other@ex.com', 'other', 'other'
+        assert add_user(self.userdata[0], self.userdata[1], self.userdata[2])
+        assert add_user(other_user_data[0], other_user_data[1], other_user_data[2])
+        with self.app.test_client() as test_client:
+            # Case 1: Unauthenticated upload attempt
+            user = get_user(self.userdata[0])
+            assert user.profile_image_path is None
+            self._upload_profile_image(test_client, token="random-string", email=self.userdata[0])
```
ReimarBauer, line 171: This random-string does not exist and does not depend on the SECRET_KEY, so it can't get further.

workaryangupta: isn't that what we want?

ReimarBauer: this now tests verify_user used, so yes we can keep it. This was requested.

mslib/mscolab/server.py
```diff
@@ -375,7 +375,7 @@ def upload_profile_image():
 @APP.route('/fetch_profile_image', methods=["GET"])
 @verify_user
 def fetch_profile_image():
-    user_id = request.form['user_id']
+    user_id = g.user.id
```
matrss: This doesn't seem right. On upload it makes sense, to ensure that a user can only upload a profile image for themselves. But on fetch this would lead to every user being shown with the same profile image - the one belonging to the user who fetched the images.

tests/_test_mscolab/test_server.py
```diff
@@ -130,14 +132,45 @@ def test_get_user(self):
...
             response = test_client.post('/delete_own_account', data={"token": token})
             assert response.status_code == 200
-            data = json.loads(response.data.decode('utf-8'))
-            assert data["success"] is True
-            response = test_client.post('/delete_own_account', data={"token": "dsdsds"})
+            assert response.get_json()["success"] is True
+            # ToDo: Check if user token was cleared after deleting account as assert returns True instead of False
+            # assert verify_user_token(config_loader(dataset="mscolab_server_url"), token) is False
+
+            # Case 2 : The user has a custom profile image set
```
matrss, line 144: Shouldn't this test case also check that the deleted users profile image is removed?

mslib/mscolab/server.py
workaryangupta: Ahh okay

ReimarBauer (approved):
LGTM

matrss (approved):

PR #2437, opened 2024-07-16
feat: Real-Time tracking of active users in an operation

workaryangupta:
This pull request aims to introduce a new feature to mscolab that allows a user to see the current number of active users in an operation.

sss


ReimarBauer:
I think the output should be near by the operation name: Description Active Users: 2

currently it is on the title line describing operations at all.

mslib/mscolab/sockets_manager.py
```diff
@@ -132,10 +154,31 @@ def handle_start_event(self, json_config):
         self.sockets.append(socket_storage)
 
     def handle_disconnect(self):
-        logging.info("disconnected")
-        logging.info(request.sid)
+        logging.info("Handling disconnect.")
```
ReimarBauer: this should become debug

all new .info should become debug.

They are logged into ~/.config/msui/msui.log

while info goes always on screen

mslib/mscolab/sockets_manager.py
```diff
@@ -283,6 +326,7 @@ def setup_managers(app):
     socketio.on_event('file-save', sm.handle_file_save)
     socketio.on_event('add-user-to-operation', sm.join_creator_to_operation)
     socketio.on_event('update-operation-list', sm.update_operation_list)
+    socketio.on_event('operation-selected', sm.handle_operation_selected)  # Register new event
```
ReimarBauer: the comment is no new informations, thats what happens by the line.  you may want to comment for what it is used. We later can do with the others too.

ReimarBauer:
When a user gets removed from the activated OP this needs an update of the counter

https://github.com/user-attachments/assets/a769f321-7ed3-4623-856a-4409db577f5f

ReimarBauer (changes requested):
see comments

mslib/mscolab/sockets_manager.py
```diff
@@ -51,12 +51,33 @@ def __init__(self, chat_manager, file_manager):
...
 
     def handle_connect(self):
         logging.debug(request.sid)
 
+    def handle_operation_selected(self, json_config):
+        logging.debug("Operation selected: {}".format(json_config))
+        token = json_config['token']
+        op_id = json_config['op_id']
+        user = User.verify_auth_token(token)
+        if not user:
```
ReimarBauer: if user is None:
   return None ?

ReimarBauer (commented):
currently the size of the description changes the space of the label for the active users.



image

The activity counter has a precendence over a long description which is good.  The border could have a bit more space. 

image

I am also not sure if the word Users: would be enough (without the word Active)

mslib/mscolab/utils.py
```diff
@@ -46,6 +46,14 @@ def get_session_id(sockets, u_id):
     return s_id
 
 
+def get_user_id(sockets, s_id):
```
ReimarBauer, line 49: could you add for this also a test, or have you?

mslib/mscolab/sockets_manager.py
workaryangupta: In other functions, it was written like this only so I just wrote this condition like those. 
Should I return None?

ReimarBauer: None is the default for return.

I meant the condition check before. It is also on other places wrong when it is done similiar.

The C based python allows currently this but from the definition it better should compare by "is" instead of abusing boolean.

Another python Implementation may use the specification different.

workaryangupta: I see, I will correct this in other places too

workaryangupta:
> currently the size of the description changes the space of the label for the active users.
 Fixed it

ReimarBauer:
Please update your branch with develop

ReimarBauer (approved):
LGTM

Issue #2442, opened 2024-07-23
Refactor MSColab away from a bunch of global variables


matrss:
Currently the MSColab Flask app is defined by a bunch of global variables (the Flask app instance that is imported and reused everywhere, and mscolab_settings mainly). The often seen alternative approach is to have a factory function constructing and returning a Flask instance (like described here: https://flask.palletsprojects.com/en/3.0.x/patterns/appfactories/), which could then be fed with the configuration to build a matching app instance.

This would make it easier to get independent Flask app instances in tests. I suspect that the current construction is one of the reasons why I still see inter-dependencies between tests on the database level.

It would also make it possible to construct differently configured Flask app instances for different tests. This would mainly come in handy to make tests/_test_mscolab/test_server_auth_required.py run as part of an ordinary run of the test suite.

This came up while I am currently struggling with building a test for migrating data from one database to a new one, as part of #2432. Since (almost?) all of the database related helper functions import and use the global app instance it is basically impossible to make them act on a different database.

matrss:
No idea what to tag this as. It isn't a bug per-se, but also not an enhancement. This is an architectural change that should be invisible to a user, and as such doesn't require a major release. I've picked bug for now. The most fitting label would be "technical debt" IMO.

matrss:
In the context of #2437 it was found that different locations in code had different instances of SocketManager, for some reason. I think this is a bug resulting from the very convoluted application setup. Some tests also instantiate their own objects of the manager types and then mix interactions with them and the global instances, which again is inconsistent usage of the objects and seems to make some issues in #2437's tests.

matrss:
@ReimarBauer this is not fixed, our Flask app instance is still a global object.

ReimarBauer:
@matrss ok, the further work on that is a deeper API change. 

I don't remove the bug label, because that API change is more than only an improvement. We can mention this later in CHANGES.

JanEisermann:
Hi, I would like to work on it.

PR #2444, opened 2024-07-26
Fix inconsistent SocketsManager objects

matrss:
On import of mslib.mscolab.server initialize_managers was called once, hooking events up to handlers in a SocketsManager instance and initializing the Flask app with the Flask-SocketIO extension. Then, in handle_start, initialize_managers was called a second time, creating another SocketsManager object, overwriting the instance saved with the SocketIO object, and trying to initialize the Flask app with it again.

The last step does not work though, a Flask app can only be initialized once with each extension (i.e. the sockio.init_app(app) call can only be done once per Flask app object).

What this led to is that the Flask-SocketIO extension was using the first SocketsManager object to handle incoming events, while the server side was using the second SocketsManager object when it called methods on sockio.sm.

This changes the startup process to do the initialization just once.

This is more of a minimally-invasive band-aid to get it working for now; a proper fix would entail doing #2442 and getting the ability to build multiple app objects, IMO.

### DIFF
--- a/mslib/mscolab/server.py
+++ b/mslib/mscolab/server.py
@@ -46,7 +46,7 @@ from flask.wrappers import Response
 
 from mslib.mscolab.conf import mscolab_settings, setup_saml2_backend
 from mslib.mscolab.models import Change, MessageType, User
-from mslib.mscolab.sockets_manager import setup_managers
+from mslib.mscolab.sockets_manager import _setup_managers
 from mslib.mscolab.utils import create_files, get_message_dict
 from mslib.utils import conditional_decorator
 from mslib.index import create_app
@@ -124,8 +124,8 @@ def confirm_token(token, expiration=3600):
     return email
 
 
-def initialize_managers(app):
-    sockio, cm, fm = setup_managers(app)
+def _initialize_managers(app):
+    sockio, cm, fm = _setup_managers(app)
     # initializing socketio and db
     app.wsgi_app = socketio.Middleware(socketio.server, app.wsgi_app)
     sockio.init_app(app)
@@ -133,7 +133,7 @@ def initialize_managers(app):
     return app, sockio, cm, fm
 
 
-_app, sockio, cm, fm = initialize_managers(APP)
+_app, sockio, cm, fm = _initialize_managers(APP)
 
 
 def check_login(emailid, password):
--- a/mslib/mscolab/mscolab.py
+++ b/mslib/mscolab/mscolab.py
@@ -45,15 +45,13 @@ from mslib.utils.qt import Worker, Updater
 
 
 def handle_start(args):
-    from mslib.mscolab.server import APP, initialize_managers, start_server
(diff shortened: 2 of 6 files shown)


mslib/mscolab/server.py
```diff
@@ -124,7 +124,7 @@ def confirm_token(token, expiration=3600):
     return email
 
 
-def initialize_managers(app):
+def _initialize_managers(app):
```
ReimarBauer, line 127: The rename could be done also on the different refactoring in develop later?

ReimarBauer (commented):
Oh, good catch.

mslib/mscolab/server.py
matrss: If anything imports `initialize_managers` and calls it, that would break MSColab in the same fashion. Calling `initialize_managers` more than once is always a bug. If anyone is doing that I would rather break their usage explicitly then silently run a wrongly initialized MSColab instance.

I forgot to do the same to `setup_managers` though. I fixed that now.

matrss: In develop this function should just be removed entirely and the complete initialization process should be replaced with Flasks app factory approach. That would bring us closer to a global-variable-free and consistent application setup.

matrss: Or actually, that _could_ happen on stable as well, if we had a defined public API. Right now basically the entirety of mslib has to be considered as the public API for SemVer changes, which turns even simple architectural changes into breaking changes. We should consider making private parts explicitly private (starting with an underscore, as is convention).

matrss: I could keep the old names and make it so that calling the function twice would `sys.exit` with an appropriate error message. That would still technically be a breaking change though, but anything else doesn't fix the bug.

ReimarBauer: I am still not sure if renaming is the best option when we plan to remove them. We should deprecate them with a warning too.

That is how we designed the lib currently an edge case for SemVer.  
As a human I can live with an exception but in general I think we should not start with exceptions.

matrss: Some more options:
- emit a warning when `setup_managers` is called a second time, with the risk that people don't read warnings and run a broken MSColab
- turn the `ChatManager`, `FileManager` and `SocketsManager` objects created in `setup_managers` into global objects instead of creating new ones on each call; making calling `setup_managers` twice safe, but pointless

ReimarBauer: not global by global that makes much problems.  Sometimes I wished python did not know "global" . That would make it better understandable in talking about states of objects.

ReimarBauer (approved):
thx

mslib/mscolab/server.py
matrss: Well, the "global" keyword isn't really the issue, global variables can be set without it. But yes, mutable global variables make problems, as is evident from this very issue: the Flask app instance is such a global object, and the mismatch between that and the not-global SocketsManager broke stuff.

PR #2447, opened 2024-08-01
Fix tests failing due to socketio connection issues

matrss:
**Purpose of PR?**:

The tests started to fail more often after the changes to test_sockets_manager.py made it no longer request the mscolab_server fixture.

~Presumably this delayed the startup of said fixture, leading to a race condition in other tests where the running server instance is not yet ready to accept SocketIO connections. This was hidden previously because the fixture simply had more time to get started.~

Presumably this delayed the startup of said fixture to a point in time at which the pytest processes MSColab server objects were no longer in a state in which they could be properly started in a forked process. Making the fixture be autouse'd to set it up early fixes this issue for now.

**Checklist:**
- [ ] Commit has unit tests

### DIFF
--- a/tests/fixtures.py
+++ b/tests/fixtures.py
@@ -28,6 +28,7 @@ import mock
 import multiprocessing
 import time
 import urllib
+import socketio
 import mslib.mswms.mswms
 import eventlet
 import eventlet.wsgi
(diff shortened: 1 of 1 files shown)


matrss:
The socket connection errors seem to be gone, but now there are some tests that hang until a timeout occurs...

tests/fixtures.py
```diff
@@ -115,6 +116,14 @@ def mscolab_session_server(mscolab_session_app, mscolab_session_managers):
     handles per-test cleanup as well.
     """
     with _running_eventlet_server(mscolab_session_app) as url:
+        # Wait until the Flask-SocketIO server is ready for connections
+        while True:
+            try:
+                with socketio.SimpleClient() as sio:
+                    sio.connect(url)
```
ReimarBauer: does it make a difference with sio.connect(url, transports='polling')

matrss: No, it doesn't. Somehow the tests that request this fixture are permanently stuck in this loop now, as if the server never gets ready.

matrss:
Tests seem to be as reliable as before now, just occasionally failing on macOS.

ReimarBauer (approved):
thx

PR #2457, opened 2024-08-12
feat: new regex filter feature

rohit2p:
**Purpose of PR?**:

introduce a new filter feature using re


matrss:
I think the Apply and Cancel buttons aren't really needed and just clutter the interface, instead any fetch or similar could just take the current content of the filter into account and hitting enter in the regex input should just do an update.

Also, the current changes somehow break the operations list when resizing, it no longer changes sizes with the window.

rohit2p:
> I think the Apply and Cancel buttons aren't really needed and just clutter the interface, instead any fetch or similar could just take the current content of the filter into account and hitting enter in the regex input should just do an update.
> 
> Also, the current changes somehow break the operations list when resizing, it no longer changes sizes with the window.

yes great unless i hit enter the filtered operations will only appear in the mainwindow as i hit enter the docking widget will refresh and show the filtered operations only. Thx

joernu76:
Nice!
1) The regex filter overrides the category filter. I think we want to have bot in an *AND* fashion. 
2) I do not understand the "ignore archived" tickbox. In this view "archived" operations shall always be ignored. There is a view opened by the "Operation Archive" button. "Someone" broke the archiving fully and completely, but my intent is to remedy this soon. Ideally the Operation archive view gets its own regex filter.

joernu76:
Further, we need to think about the interaction between the multiple flightpath widget and this filter. Shall there be any?
Currently, the widget does not change its view when the category in the mainwindow is changed, which I deem a bug (not necessarily yours to fix, but would be nice and should be included in the fix that only "attaches" checked operations).

rohit2p:
> Further, we need to think about the interaction between the multiple flightpath widget and this filter. Shall there be any? Currently, the widget does not change its view when the category in the mainwindow is changed, which I deem a bug (not necessarily yours to fix, but would be nice and should be included in the fix that only "attaches" checked operations).

yes you have to reopen the topview for the reflection to happen definitely need a fix and I'll do my best and i also want the filtered operations to reflect in docking widget but i having trouble to find how this category filtered operations are reflecting in the docking widget

PR #2467, opened 2024-08-16
Fixes for using PostgreSQL

matrss:
Created enum types should also be dropped on downgrade. Dropping tables must be committed to become visible.

**Purpose of PR?**:

Fixes #

**Checklist:**
- [ ] Commit has unit tests


ReimarBauer (approved):
I expect that his works. I can verify this again later when we have merged both to develop.

mslib/mscolab/migrations/versions/92eaba86a92e_to_version_8_3_5_initial_migration.py
```diff
@@ -81,7 +81,9 @@ def upgrade():
 def downgrade():
     # ### commands auto generated by Alembic - please adjust! ###
     op.drop_table('permissions')
+    sa.Enum(name='access_level').drop(op.get_bind(), checkfirst=False)
     op.drop_table('messages')
+    sa.Enum(name='messagetype').drop(op.get_bind(), checkfirst=False)
```
ReimarBauer, line 86: I just looked into the model definition: "message_type"

ReimarBauer (changes requested):
I think it is  "message_type" please double check

mslib/mscolab/migrations/versions/92eaba86a92e_to_version_8_3_5_initial_migration.py
matrss: The column that uses the type is called `message_type`, but the enum type in the database is created with the name `messagetype`. You can see that here: https://github.com/Open-MSS/MSS/blob/c60241503e9e576d95b01c713dc55e11326d450b/mslib/mscolab/migrations/versions/92eaba86a92e_to_version_8_3_5_initial_migration.py#L61, the relevant part being `name='messagetype'`.

matrss: I think it derives that name from simply lowercasing the class name of the enum defined here: https://github.com/Open-MSS/MSS/blob/c60241503e9e576d95b01c713dc55e11326d450b/mslib/mscolab/message_type.py#L28

ReimarBauer (approved):
ok, thx.

PR #2468, opened 2024-08-16
bug fix for incomplete xml imports

### LINKED ISSUE #2456: problem in gcpoints_path
With some of our data I get

```
Fatal error in MSS 9.0.0 on Linux-6.5.0-10043-tuxedo-x86_64-with-glibc2.35

  File "/home/reimar/PycharmProjects/2024/rohit2p/MSS/mslib/msui/multiple_flightpath_dockwidget.py", line 764, in set_flag
    self.draw_inactive_operations()
  File "/home/reimar/PycharmProjects/2024/rohit2p/MSS/mslib/msui/multiple_flightpath_dockwidget.py", line 898, in draw_inactive_operations
    patch = MultipleFlightpath(self.view.map,
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/reimar/PycharmProjects/2024/rohit2p/MSS/mslib/msui/multiple_flightpath_dockwidget.py", line 71, in __init__
    self.draw()
  File "/home/reimar/PycharmProjects/2024/rohit2p/MSS/mslib/msui/multiple_flightpath_dockwidget.py", line 110, in draw
    x, y = self.compute_xy(lon, lat)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/reimar/PycharmProjects/2024/rohit2p/MSS/mslib/msui/multiple_flightpath_dockwidget.py", line 78, in compute_xy
    x, y = self.map.gcpoints_path(lon, lat)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/reimar/PycharmProjects/2024/rohit2p/MSS/mslib/msui/mpl_map.py", line 723, in gcpoints_path
    assert len(lons) > 1
           ^^^^^^^^
```

ReimarBauer:
**Purpose of PR?**:

Fixes #2456

**Checklist:**
- [X] Commit has unit tests


mslib/mscolab/file_manager.py
```diff
@@ -83,11 +84,11 @@ def create_operation(self, path, description, user, last_used=None, content=None
...
-            operation_file = data.open(fs.path.combine(operation.path, 'main.ftml'), 'w')
-            if content is not None:
-                operation_file.write(content)
-            else:
-                operation_file.write(mscolab_settings.STUB_CODE)
+            with data.open(fs.path.combine(operation.path, 'main.ftml'), 'w') as operation_file:
+                if content is not None and verify_waypoint_data(content):
+                    operation_file.write(content)
+                else:
+                    operation_file.write(mscolab_settings.STUB_CODE)
```
matrss: So, this would write STUB_CODE if the content supplied to create_operation does not verify, creating an operation with different content then supplied. I think create_operation should instead behave just like save_file, i.e. error out on invalid content and not creating the operation.

mslib/utils/verify_waypoint_data.py
```diff
@@ -0,0 +1,53 @@
...
+    See the License for the specific language governing permissions and
+    limitations under the License.
+"""
+
+
+import xml.dom.minidom
+import xml.parsers.expat
+
+
+def verify_waypoint_data(xml_content):
```
matrss, line 32: This could be an ideal use-case for [XML Schemas](https://en.wikipedia.org/wiki/XML_Schema_(W3C)). That would require more than the stdlib xml support though.

mslib/utils/verify_waypoint_data.py
```diff
@@ -0,0 +1,53 @@
...
+
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+"""
+
+
+import xml.dom.minidom
```
matrss: This might be a more fundamental issue: the xml processing modules from python's stdlib are _not_ suitable to parse untrusted xml documents. See the first warning here: https://docs.python.org/3/library/xml.html.

My understanding is that we allow users to upload those xml documents directly, so a malicious user might be able to craft the xml in such a way that it e.g. crashes the MSColab server.

tests/_test_mscolab/test_server.py
```diff
@@ -233,7 +234,7 @@ def test_get_all_changes(self):
         with self.app.test_client() as test_client:
             operation, token = self._create_operation(test_client, self.userdata)
             fm, user = self._save_content(operation, self.userdata)
-            fm.save_file(operation.id, "content2", user)
+            fm.save_file(operation.id, XML_CONTENT2, user)
```
matrss: My understanding is that the tests in this file should exercise the server's API and so should _not_ interact with MSColab internals like a FileManager instance (I subscribe to the idea that the _only_ thing worth testing is the external interface of the code, so tests should _never_ interact with internals at all, so as to not hinder future refactors and the like). Instead, this should use the same code path that is also used by the MSUI client (so, socketio, I think).

tests/_test_utils/test_verify_waypoint_data.py
```diff
@@ -0,0 +1,98 @@
...
+    See the License for the specific language governing permissions and
+    limitations under the License.
+"""
+
+from mslib.utils.verify_waypoint_data import verify_waypoint_data
+
+
+def test_verify_xml_waypoint():
+    """
+    tests various
```
matrss: That sentence seems incomplete.

tests/_test_utils/test_verify_waypoint_data.py
```diff
@@ -0,0 +1,98 @@
...
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+"""
+
+from mslib.utils.verify_waypoint_data import verify_waypoint_data
+
+
+def test_verify_xml_waypoint():
```
matrss: Do we need to test this specifically or can we cover the entirety of that function by only interacting with MSColab's public interface?

mslib/mscolab/file_manager.py
ReimarBauer: this is a different problem and should discussed/done afterwards. This may have impact to server configuration defaults for operations, which we also have for flightpathes. 

I've just seen that we not care on closing the file and solved that too.

mslib/utils/verify_waypoint_data.py
ReimarBauer: this is currently only a simple solution. A refactoring needs to be done too.

tests/_test_mscolab/test_server.py
ReimarBauer: If we have other tests that give the same result, with the same coverage, we can reduce the internal tests you mentioned

tests/_test_utils/test_verify_waypoint_data.py
ReimarBauer: We have more than one interface. I want to be sure that this is covered when a refactoring is done and whatever sends us data.

mslib/mscolab/file_manager.py
matrss: It is new behavior that is introduced in this PR, so I think it should be discussed here.

Previously, the STUB_CODE was used when no content was specified. Now, the STUB_CODE is used when either no content is specified, or the specified content did not validate. The latter is unintuitive behavior to me, and I think create_operation should just error out in this case.

tests/_test_utils/test_verify_waypoint_data.py
matrss: > We have more than one interface.

And all of them that interact with the real world have to be tested. The important part is that the servers interface to the outside world does indeed do the validation, not that the validation in itself exists and responds as expected.

I just want to emphasize that a test is meant to codify an expected behavior of the code, and the existence of an internal function called `verify_xml_waypoint` is not a crucial part to ensure for the future, in my opinion.

Ensuring that coverage doesn't decrease through a refactoring could be accomplished by checking that coverage does not decrease, instead.

> whatever sends us data

That's the point, this test does not concern itself with if data send from whereever is actually validated.

tests/_test_mscolab/test_server.py
matrss: This case here is different. This is not an internal test. This is a test that exercises the external interface to interact with an operation, but in doing so erroneously uses internal structures.

Instead of calling fm.save_file this should go through a socketio test_client and emit the file-save event, so that it actually tests the server.

mslib/mscolab/file_manager.py
ReimarBauer: Oh,  then I remove it I didn't want to do that

tests/_test_mscolab/test_server.py
ReimarBauer: not thought on that, indeed that should now with the improved test setup doable.

I thought on just fixing the wrong string...

tests/_test_utils/test_verify_waypoint_data.py
ReimarBauer: I disagree, I first want to know what kind of failures we are aware off. I also want to know that a new function has no failure when it is designed. For me testdriven means to define what the function should solve.  The test describes what I had in mind as I wrote the function.

tests/_test_mscolab/test_files.py
```diff
@@ -179,8 +179,7 @@ def test_delete_operation(self):
             assert len(messages) == 0
 
     def _example_data(self):
-        self.content1 = """\
-    <?xml version="1.0" encoding="utf-8"?>
+        self.content1 = """<?xml version="1.0" encoding="utf-8"?>
```
ReimarBauer, line 189: this was also giving an xml.parsers.expat.ExpatError

tests/_test_utils/test_verify_waypoint_data.py
joernu76: I am not sure whether this should be discussed on an open channel. However, the XML is never parsed by the server, so the server should be "safe".
The clients are a different matter though. If we think that XML cannot be trusted, simply switching to defusedxml on the client side should solve this...?

matrss: @joernu76 I think you wanted to respond to https://github.com/Open-MSS/MSS/pull/2468#discussion_r1721359848. Yes, you are right, we should take that to a closed channel.

matrss: > For me testdriven means to define what the function should solve.

I see that differently. Test-driven means writing tests for the correctness of the change that is to be implemented, in this case for the change that the server and client should verify the xml data. That in the end this function implements the change is an irrelevant implementation detail, as far as testing is concerned IMO. It could be a thousand different functions, and only a single test, as long as the test exercises the change they implement from the codes exposed surface. Testing all of the thousand functions would even be detrimental, because it locks in a particular architectural design, with little to no benefit.

Anyway, no matter if this test is present or not, the client-side validation isn't tested for yet. I.e. I can remove the client-side validation without a test failing.

mslib/utils/verify_waypoint_data.py
ReimarBauer: better check :)

mslib/mscolab/file_manager.py
```diff
@@ -57,43 +58,50 @@ def create_operation(self, path, description, user, last_used=None, content=None
         description: description of the operation
         """
         # set codes on these later
-        if path.find("/") != -1 or path.find("\\") != -1 or (" " in path):
-            logging.debug("malicious request: %s", user)
-            return False
-        proj_available = Operation.query.filter_by(path=path).first()
-        if proj_available is not None:
+        if content is None or verify_waypoint_data(content):
```
matrss: I think this condition should be inverted into an early return. I.e.:
```python
if content is not None and not verify_waypoint_data(content):
    return False
```

The rest of the method could then stay as it was and assume that content is either None or verified, making the diff much smaller and the method more readable.

mslib/msui/flighttrack.py
```diff
@@ -644,13 +645,19 @@ def load_from_ftml(self, filename):
         _dirname, _name = os.path.split(filename)
         _fs = fs.open_fs(_dirname)
         xml_content = _fs.readtext(_name)
-        name = os.path.basename(filename.replace(".ftml", "").strip())
-        self.load_from_xml_data(xml_content, name)
+        if verify_waypoint_data(xml_content):
+            name = os.path.basename(filename.replace(".ftml", "").strip())
+            self.load_from_xml_data(xml_content, name)
```
matrss: This does the verification twice, once here and once in `self.load_from_xml_data`.

matrss: I'd argue that the verification only belongs into the top-level `load_from_xml_data` function here: https://github.com/Open-MSS/MSS/blob/ff744f5f4dadf09407e81688a0da18b1cd771d3e/mslib/msui/flighttrack.py#L98

mslib/msui/mscolab.py
```diff
@@ -1977,6 +1978,9 @@ def handle_import_msc(self, file_path, extension, function, pickertype):
                 model = ft.WaypointsTableModel(waypoints=new_waypoints)
                 xml_doc = self.waypoints_model.get_xml_doc()
                 xml_content = xml_doc.toprettyxml(indent="  ", newl="\n")
+            if not verify_waypoint_data(xml_content):
+                show_popup(self.ui, "Import Failed", f"The file - {file_name}, was not imported!", 0)
+                return
```
matrss, line 1985: If the intention is to protect against e.g. maliciously crafted XML or the like, then verification must be the first thing that is done, before using the XML for anything else. In this case though, the XML might be used already here: https://github.com/Open-MSS/MSS/blob/ff744f5f4dadf09407e81688a0da18b1cd771d3e/mslib/msui/mscolab.py#L1971

matrss: Also, if `function` is not None, then that supplied function would need to do the verification as well. But this is a theoretical concern, it seems like for XML files it is always None, and it is only used to provide a way to read CSV files instead.

tests/_test_mscolab/test_server.py
```diff
@@ -33,13 +33,19 @@
...
 class Test_Server:
     @pytest.fixture(autouse=True)
-    def setup(self, mscolab_app):
+    def setup(self, mscolab_app, mscolab_managers):
         self.app = mscolab_app
+        self.sockio, self.cm, self.fm = mscolab_managers
+        self.sm = self.sockio.sm
+        self.sockets = []
         self.userdata = 'UV10@uv10', 'UV10', 'uv10'
+        self.sio = self.sockio.test_client(self.app)
```
matrss: Please use this as a contextmanager, so it is disconnected properly afterwards.

tests/_test_mscolab/test_server.py
```diff
@@ -33,13 +33,19 @@
...
 
 
 class Test_Server:
     @pytest.fixture(autouse=True)
-    def setup(self, mscolab_app):
+    def setup(self, mscolab_app, mscolab_managers):
         self.app = mscolab_app
+        self.sockio, self.cm, self.fm = mscolab_managers
+        self.sm = self.sockio.sm
+        self.sockets = []
```
matrss: This list is unused and unneeded.

tests/_test_mscolab/test_server.py
```diff
@@ -33,13 +33,19 @@
...
+from tests.utils import XML_CONTENT1, XML_CONTENT2
 
 
 class Test_Server:
     @pytest.fixture(autouse=True)
-    def setup(self, mscolab_app):
+    def setup(self, mscolab_app, mscolab_managers):
         self.app = mscolab_app
+        self.sockio, self.cm, self.fm = mscolab_managers
+        self.sm = self.sockio.sm
```
matrss: There should be no need for `cm`, `fm` and `sm` in this class, to test what it is intended to test.

tests/_test_mscolab/test_server.py
```diff
@@ -410,5 +443,5 @@ def _save_content(self, operation, userdata=None):
             userdata = self.userdata
         user = get_user(userdata[0])
         fm = FileManager(self.app.config["MSCOLAB_DATA_DIR"])
```
matrss: The `SocketsManager`, `FileManager` and `ChatManager` instances are only really designed to be singleton objects at the moment; creating more instances of them does not reflect reality, so isn't useful for tests, and might even break stuff as it did with multiple `SocketsManager`s.

tests/_test_utils/test_verify_waypoint_data.py
```diff
@@ -0,0 +1,98 @@
...
+            </Waypoint>
+          </ListOfWaypoints>
+        </FlightTrack>"""
+
+    xml_content = (
+        (flight_track_with_waypoints, True),
+        (flight_track_empty, False),
+        (flight_track_incomplete, False),
+        (flight_track_with_typo, False),
+    )
```
matrss: The pytest-idiomatic way for this would be to parametrize the test over these cases: https://docs.pytest.org/en/7.1.x/how-to/parametrize.html

mslib/msui/mscolab.py
ReimarBauer: a users functions has to return name, waypoints e.g. https://github.com/Open-MSS/MSS/blob/develop/mslib/plugins/io/flitestar.py

How a user can test his functions is a different scope.

tests/_test_utils/test_verify_waypoint_data.py
ReimarBauer: I know, but then the cases outside the function and I don't see that as better readable.

tests/_test_mscolab/test_server.py
matrss: Oh, actually, I don't think it is useable as a contextmanager, unlike the flask test_client. Still, the disconnect is missing.

tests/_test_utils/test_verify_waypoint_data.py
matrss: Yes, they would be outside the function because each case would then actually be a separate test, as far as pytest is concerned. I find it more understandable when it is clear from the failing test itself which case fails, instead of having to look at the test function and figuring out in which loop iteration the assert fails.

tests/_test_mscolab/test_server.py
ReimarBauer: unfortunately context manager protocol is not implemented into the SocketIOTestClient for flask-socketio, but hey I can just use it there where I need it.

tests/_test_utils/test_verify_waypoint_data.py
ReimarBauer: separating the tests by different names I can agree but for xml the fixture gets unreadable.

ReimarBauer: I added one of the linebreak patterns too, which fails too.

matrss: Fixture? What fixture?

Sorry, I don't see why something like this would be less readable:
```python
flight_track_with_waypoints = """<?xml version="1.0" encoding="utf-8"?>
<FlightTrack version="9.0.0">
  <ListOfWaypoints>
    <Waypoint flightlevel="233.0" lat="41.601070573320186" location="" lon="41.355120439498535">
      <Comments></Comments>
    </Waypoint>
    <Waypoint flightlevel="374.0" lat="48.19354838709677" location="" lon="33.74841526975632">
      <Comments></Comments>
    </Waypoint>
    <Waypoint flightlevel="372.0" lat="51.23623045499366" location="" lon="-34.20481757994082">
      <Comments></Comments>
    </Waypoint>
    <Waypoint flightlevel="20.0" lat="37.037047471474835" location="" lon="-40.797295393717434">
      <Comments></Comments>
    </Waypoint>
  </ListOfWaypoints>
</FlightTrack>"""

flight_track_empty = """<?xml version="1.0" encoding="utf-8"?>
<FlightTrack version="9.0.0">
  <ListOfWaypoints>
  </ListOfWaypoints>
</FlightTrack>"""

flight_track_incomplete = """<?xml version="1.0" encoding="utf-8"?>
<FlightTrack version="9.0.0">
  <ListOfWaypoints>
    <Waypoint flightlevel="233.0" lat="41.601070573320186">
      <Comments></Comments>
    </Waypoint>
    <Waypoint flightlevel="374.0">
      <Comments></Comments>
    </Waypoint>
    <Waypoint flightlevel="372.0" lat="51.23623045499366" location="" lon="-34.20481757994082">
      <Comments></Comments>
    </Waypoint>
    <Waypoint flightlevel="20.0" lat="37.037047471474835" location="">
      <Comments></Comments>
    </Waypoint>
  </ListOfWaypoints>
</FlightTrack>"""

flight_track_with_typo = """<?xml version="1.0" encoding="utf-8"?>
<FlightTrack version="9.0.0">
  <ListOfWaypoints>
    <Waypoint flightlevel="233.0"" lat="41.601070573320186" location="" lon="41.355120439498535">
      <Comments></Comments>
    </Waypoint>
    <Waypoint flightlevel="374.0" lat="48.19354838709677" location="" lon="33.74841526975632">
      <Comments></Comments>
    </Waypoint>
      <Comments></Comments>
    </Waypoint>
  </ListOfWaypoints>
</FlightTrack>"""

cases = [
    (flight_track_with_waypoints, True),
    (flight_track_empty, False),
    (flight_track_incomplete, False),
    (flight_track_with_typo, False),
]

@pytest.mark.parametrize("xml_content,verification_result", cases)
def test_verify_xml_waypoint(xml_content, verification_result):
    """Test xml verification."""
    assert verify_waypoint_data(xml_content) is verification_result
```

ReimarBauer: I misunderstood you because you where pointing on the example.

ReimarBauer: next update on the way

matrss: Having it split into individual functions with descriptive names like you did looks good too though, pick what you prefer.

ReimarBauer: compared to individual functions in a failure case that would look like

image

How would I see now that flight_track_with_typo was the reason?

ReimarBauer: I think we keep it, it is good to have such an examples with more content.

matrss (approved):
LGTM. The broken encapsulation in the test suite, e.g. by using `fm` in test_server.py, is a running theme in the entirety of the test suite from what I have seen and I think a refactor to address that is a bit out of scope for this PR.

tests/_test_utils/test_verify_waypoint_data.py
matrss: Ahh right, it shows the content of the variable as the test name by default. That can be changed with the `ids=` option to `pytest.mark.parametrize`.

PR #2471, opened 2024-08-18
test for the new features in the MFDW, topview & sideview

rohit2p:
**Purpose of PR?**:

This PR introduce testing code for the new features in the docking widget, topview & sideview

### DIFF
--- a/tests/_test_msui/test_multiple_flightpath_dockwidget.py
+++ b/tests/_test_msui/test_multiple_flightpath_dockwidget.py
@@ -123,12 +123,12 @@ def test_set_linewidth(main_window):
     with mock.patch.object(multiple_flightpath_widget, "update_flighttrack_patch") as mock_update_patch:
         # s.et the line width
         multiple_flightpath_widget.dsbx_linewidth.setValue(3.0)
+        mock_update_patch.assert_called_once_with(item.flighttrack_model)
 
-        # Verify the line width has been updated in dict_flighttrack and the patch was updated
-        wp_model = item.flighttrack_model
-        assert multiple_flightpath_widget.dict_flighttrack[wp_model]["linewidth"] == 3.0
-        assert multiple_flightpath_widget.change_linewidth is True
-        mock_update_patch.assert_called_once_with(wp_model)
+    # Verify the line width has been updated in dict_flighttrack
+    wp_model = item.flighttrack_model
+    assert multiple_flightpath_widget.dict_flighttrack[wp_model]["linewidth"] == 3.0
+    assert multiple_flightpath_widget.change_linewidth is True
 
 
 def test_set_transparency(main_window):
@@ -147,12 +147,12 @@ def test_set_transparency(main_window):
     with mock.patch.object(multiple_flightpath_widget, "update_flighttrack_patch") as mock_update_patch:
 
         multiple_flightpath_widget.hsTransparencyControl.setValue(50)
+        mock_update_patch.assert_called_once_with(item.flighttrack_model)
 
-        # Verify the transparency has been updated in dict_flighttrack and update_flighttrack_patch was called
-        wp_model = item.flighttrack_model
-        assert multiple_flightpath_widget.dict_flighttrack[wp_model]["line_transparency"] == (50 / 100)
-        assert multiple_flightpath_widget.change_line_transparency is True
-        mock_update_patch.assert_called_once_with(wp_model)
+    # Verify the transparency has been updated in dict_flighttrack
+    wp_model = item.flighttrack_model
+    assert multiple_flightpath_widget.dict_flighttrack[wp_model]["line_transparency"] == (50 / 100)
+    assert multiple_flightpath_widget.change_line_transparency is True
 
 
 def test_set_linestyle(main_window):
@@ -171,14 +171,14 @@ def test_set_linestyle(main_window):
     with mock.patch.object(multiple_flightpath_widget, "update_flighttrack_patch") as mock_update_patch:
 
         multiple_flightpath_widget.cbLineStyle.setCurrentText('Dashed')
+        mock_update_patch.assert_called_once_with(item.flighttrack_model)
 
-        # Verify the line style has been updated in dict_flighttrack and update_flighttrack_patch was called
-        wp_model = item.flighttrack_model
-
-        expected_style = '--'
-        assert multiple_flightpath_widget.dict_flighttrack[wp_model]["line_style"] == expected_style
-        assert multiple_flightpath_widget.change_line_style is True
-        mock_update_patch.assert_called_once_with(wp_model)
+    # Verify the line style has been updated in dict_flighttrack
+    wp_model = item.flighttrack_model
+    expected_style = '--'
+    assert multiple_flightpath_widget.dict_flighttrack[wp_model]["line_style"] == expected_style
+    assert multiple_flightpath_widget.change_line_style is True
+    mock_update_patch.assert_called_once_with(wp_model)
 
 
 def test_selectAll(main_window):
@@ -211,23 +211,17 @@ def test_random_custom_color_selection(main_window):
     """
     _, multiple_flightpath_widget = main_window
 
-    # Mock the get_random_color method to test random selection
-    with mock.patch.object(multiple_flightpath_widget, "get_random_color",
-                           wraps=multiple_flightpath_widget.get_random_color) as mock_get_random_color:
-        # Select random colors multiple times
-        selected_colors = set()
-        for _ in range(10):  # Test with 10 random selections
-            color = multiple_flightpath_widget.get_random_color()
-            normalized_color = tuple(int(c * 255) for c in color)
-            assert normalized_color not in selected_colors, "Duplicate color selected!"
-            selected_colors.add(normalized_color)
+    # Select random colors multiple times
+    selected_colors = set()
+    for _ in range(10):  # Test with 10 random selections
+        color = multiple_flightpath_widget.get_random_color()
+        normalized_color = tuple(int(c * 255) for c in color)
+        assert normalized_color not in selected_colors, "Duplicate color selected!"
+        selected_colors.add(normalized_color)
 
-        # Ensure the mock method was called 10 times
-        assert mock_get_random_color.call_count == 10
-
-        # Check that all selected colors are from the custom_colors list
-        for color in selected_colors:
-            assert color in multiple_flightpath_widget.custom_colors
+    # Check that all selected colors are from the custom_colors list
+    for color in selected_colors:
+        assert color in multiple_flightpath_widget.custom_colors
 
 
 def activate_flight_track_at_index(main_window, index):
--- a/tests/_test_msui/test_multiple_flightpath_dockwidget.py
+++ b/tests/_test_msui/test_multiple_flightpath_dockwidget.py
@@ -53,6 +53,8 @@ def main_window(qtbot):
     # Switch to the Multiple Flightpath Widget
     topview_window.cbTools.setCurrentIndex(6)
 
+    QtTest.QTest.qWait(200)
+
     # Get a reference to the created MultipleFlightpathControlWidget
     multiple_flightpath_widget = topview_window.docks[tv.MULTIPLEFLIGHTPATH].widget()
 
@@ -76,16 +78,14 @@ def test_setColour(main_window):
     """
     test for the filghttrack colour
     """
-    _, multiple_flightpath_widget = main_window
+    main_window, multiple_flightpath_widget = main_window
     color_button = multiple_flightpath_widget.pushButton_color
 
-    # Mock the exec_ method and color_selected signal of the CustomColorDialog
+    # Mock the exec_ method to accept the dialog otherwise test get stuck
     with mock.patch("mslib.utils.colordialog.CustomColorDialog.exec_",
-                    return_value=QtWidgets.QDialog.Accepted) as mock_exec, \
-            mock.patch("mslib.utils.colordialog.CustomColorDialog.color_selected",
-                       new_callable=mock.Mock) as mock_color_selected:
+                    return_value=QtWidgets.QDialog.Accepted) as mock_exec:
         # Activate the first flight track
-        activate_flight_track_at_index(main_window[0], 0)
+        activate_flight_track_at_index(main_window, 0)
 
         # Click on the second flight track in the docking widget
         click_on_flight_track_in_docking_widget_at_index(multiple_flightpath_widget, 1)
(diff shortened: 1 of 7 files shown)


rohit2p:
@matrss why `"feat: New UI for option in topview & sideview" ` this commit also got in here...is this supposed to happen or is it a mistake. How do i fix it?

rohit2p:
Also my PR is not merged yet then how can i write testing code on them. It should merge tho the works is done in that PR

matrss:
> @matrss why `"feat: New UI for option in topview & sideview" ` this commit also got in here...is this supposed to happen or is it a mistake. How do i fix it?

This commit is shown here because this PR targets develop instead of your GSoC branch. develop doesn't have the commit, so it is shown as one that would be merged with this PR.

matrss:
> Also my PR is not merged yet then how can i write testing code on them. It should merge tho the works is done in that PR

I consider tests a necessary condition for that PR to be merged, so IMO the tests belong into that PR. Testing a new feature is part of implementing said feature, and the tests could surface some possibly remaining issues with it. I don't know if @ReimarBauer or @joernu76 see that differently though.

ReimarBauer:
> > Also my PR is not merged yet then how can i write testing code on them. It should merge tho the works is done in that PR
> 
> I consider tests a necessary condition for that PR to be merged, so IMO the tests belong into that PR. Testing a new feature is part of implementing said feature, and the tests could surface some possibly remaining issues with it. I don't know if @ReimarBauer or @joernu76 see that differently though.

It is ideal to write the tests with or before the development or even before a new development. Without a prototype, the tests could be seen as such.
Even if you have some experience, it is not always easy to write tests that test what you want or what is important. 
However, tests have a big influence on how you write software or sometimes you realize that you can't test it that way or something is broken and you have to write the software differently. 

For someone who is still learning this, it can be an insurmountable challenge, so we are a little lenient with students

tests/_test_msui/test_multiple_flightpath_dockwidget.py
```diff
@@ -25,33 +25,78 @@
...
+
+        # Open a Top View window
+        self.main_window.actionTopView.trigger()
+        self.topview_window = self.main_window.listViews.currentItem().window
+
+        # Switch to the Multiple Flightpath Widget
+        self.topview_window.cbTools.setCurrentIndex(6)
+
+        # Get a reference to the created MultipleFlightpathControlWidget
+        self.multiple_flightpath_widget = self.topview_window.docks[tv.MULTIPLEFLIGHTPATH].widget()
```
matrss, line 57: There is no need to re-use this class. The class is a remnant from before MSS used pytest and I would prefer not to use this style of test definition. Test can just be simple functions. If you want to extract the setup part into something separate to use in multiple test functions, then just define a pytest fixture for that.

tests/_test_msui/test_multiple_flightpath_dockwidget.py
```diff
@@ -25,33 +25,78 @@
...
+
 from mslib.msui import msui
-from mslib.msui.multiple_flightpath_dockwidget import MultipleFlightpathControlWidget
-from mslib.msui import flighttrack as ft
+# from mslib.msui.multiple_flightpath_dockwidget import MultipleFlightpathControlWidget
+# from mslib.msui import flighttrack as ft
 import mslib.msui.topview as tv
 
 
 class Test_MultipleFlightpathControlWidget:
```
matrss: There is no need to re-use this class. The class is a remnant from before MSS used pytest and I would prefer not to use this style of test definition. Tests can just be simple functions. If you want to extract the setup part into something separate to use in multiple test functions, then just define a pytest fixture for that or make it a contextmanager.

tests/_test_msui/test_multiple_flightpath_dockwidget.py
```diff
@@ -25,33 +25,78 @@
...
+    @mock.patch("mslib.utils.colordialog.CustomColorDialog.exec_", return_value=QtWidgets.QDialog.Accepted)
+    @mock.patch("mslib.utils.colordialog.CustomColorDialog.color_selected", new_callable=mock.Mock)
+    def test_setColour(self, mock_color_selected, mockdlg):
+        color_button = self.multiple_flightpath_widget.pushButton_color
+
+        self._activate_flight_track_at_index(0)
+        self.click_on_flight_track_in_docking_widget_at_index(1)
+
+        # Simulate clicking the button to open the color dialog
+        QtTest.QTest.mouseClick(color_button, QtCore.Qt.LeftButton)
```
matrss: In my experience, simulating mouse clicks should be a last resort, since they often lead to situations in which Qt will behave asynchronously and stuff like `qtbot.wait_until` or similar will have to be used to avoid race conditions making the tests flaky. With buttons it should be possible to just do `button.click()` instead.

tests/_test_msui/test_multiple_flightpath_dockwidget.py
rohit2p: Line this you mean https://github.com/Open-MSS/MSS/pull/2471/commits/1320879e6ca44e55b028a33ed70352e9c1d37d0f

tests/_test_msui/test_multiple_flightpath_dockwidget.py
```diff
@@ -25,33 +25,244 @@
...
+    _, multiple_flightpath_widget = main_window
+    color_button = multiple_flightpath_widget.pushButton_color
+
+    # Mock the exec_ method and color_selected signal of the CustomColorDialog
+    with mock.patch("mslib.utils.colordialog.CustomColorDialog.exec_",
+                    return_value=QtWidgets.QDialog.Accepted) as mock_exec, \
+            mock.patch("mslib.utils.colordialog.CustomColorDialog.color_selected",
+                       new_callable=mock.Mock) as mock_color_selected:
+        # Activate the first flight track
+        activate_flight_track_at_index(main_window[0], 0)
```
matrss: If you assign the first element from the main_window tuple on line 79 as well then you don't have to index into it here. I think that would be cleaner.

tests/_test_msui/test_multiple_flightpath_dockwidget.py
```diff
@@ -25,33 +25,244 @@
...
+        # Simulate clicking the button to open the color dialog
+        color_button.click()
+        assert mock_exec.call_count == 1
+
+        # Simulate a color being selected
+        color = QtGui.QColor("#0000ff")  # Example color
+        mock_color_selected.emit(color)
+
+        # Ensure the color_selected signal was emitted
+        mock_color_selected.emit.assert_called_with(color)
```
matrss: This is really a tautology, you call the function and then assert that it is called. Instead, it would make more sense to check that `multiple_flightpath_widget.apply_color` is called with the right color, or even better would be to ensure that the drawn flighttrack has the right color afterwards. The latter might be a bit more complicated though.

tests/_test_msui/test_multiple_flightpath_dockwidget.py
```diff
@@ -25,33 +25,244 @@
...
+    click_on_flight_track_in_docking_widget_at_index(multiple_flightpath_widget, 1)
+
+    # Ensure the current item is checked
+    item = multiple_flightpath_widget.list_flighttrack.currentItem()
+    item.setCheckState(QtCore.Qt.Checked)
+
+    # Mock the dsbx_linewidth setValue and the update_flighttrack_patch method
+    with mock.patch.object(multiple_flightpath_widget.dsbx_linewidth, "setValue", return_value=3.0) as mock_set_value, \
+            mock.patch.object(multiple_flightpath_widget, "update_flighttrack_patch") as mock_update_patch, \
+            mock.patch.object(multiple_flightpath_widget.dsbx_linewidth, "value", return_value=3.0) as mock_value:
```
matrss: So, as a rule of thumb, mocking should be done sparingly as it always means that the mocked out parts do no longer reflect the reality of the application code. It is sometimes necessary as a last resort with regard to PyQt5, but in this case I think it can be reduced a bit.

In this case, I think the `multiple_flightpath_widget.dsbx_linewidth.setValue` and `multiple_flightpath_widget.dsbx_linewidth.value` mocks are not needed, as just calling `multiple_flightpath_widget.dsbx_linewidth.setValue(3.0)` should have the same effect as changing its value in the UI. This in turn should cause `multiple_flightpath_widget.set_linewidth` to be called, the way it is hooked up via signals. Then the test should assert that the result of that is an updated linewidth.

tests/_test_msui/test_multiple_flightpath_dockwidget.py
```diff
@@ -25,33 +25,244 @@
...
+    test for the selectALL method
+    """
+    _, multiple_flightpath_widget = main_window
+
+    # Mock the selectAll method to check if it gets called
+    with mock.patch.object(multiple_flightpath_widget, "selectAll",
+                           wraps=multiple_flightpath_widget.selectAll) as mock_selectAll:
+
+        multiple_flightpath_widget.selectAll(QtCore.Qt.Checked)
+        mock_selectAll.assert_called_once_with(QtCore.Qt.Checked)
```
matrss: Again, this kind of thing is a tautology and without it the mock becomes unnecessary.

matrss (changes requested):
I think this is looking pretty good already. I have some comments with regards to the test cases themselves and mocking, but structurally this is what I had in mind :slightly_smiling_face:

tests/_test_msui/test_multiple_flightpath_dockwidget.py
```diff
@@ -25,33 +25,244 @@
...
+    # Create two flight tracks
+    window.actionNewFlightTrack.trigger()
+    window.actionNewFlightTrack.trigger()
+
+    # Open a Top View window
+    window.actionTopView.trigger()
+    topview_window = window.listViews.currentItem().window
+
+    # Switch to the Multiple Flightpath Widget
+    topview_window.cbTools.setCurrentIndex(6)
```
joernu76, line 54: I am by no means an expert on QT timing, but unless you call the openTool method directly, I would assume, that a QT wait statement would belong after setting the Index to ensure that the reaction of instantiating the widget has triggered before getting the reference.

tests/_test_msui/test_multiple_flightpath_dockwidget.py
```diff
@@ -25,33 +25,244 @@
...
+
+def test_setColour(main_window):
+    """
+    test for the filghttrack colour
+    """
+    _, multiple_flightpath_widget = main_window
+    color_button = multiple_flightpath_widget.pushButton_color
+
+    # Mock the exec_ method and color_selected signal of the CustomColorDialog
+    with mock.patch("mslib.utils.colordialog.CustomColorDialog.exec_",
```
joernu76: Why not do the mock-patching as fixture? I just double checked the code and "your" syntax is commonly used by other peoples tests. See
 _tests/_test_msui/test_satellite_dockwidget.py
as example.
I think this is more elegant (similar to your rewriting of setup as fixture; nice!)

tests/_test_msui/test_multiple_flightpath_dockwidget.py
joernu76: agreed this does not test anything. In fact, I think you do not need mocking here. You can trigger the dialogue, select some colours and accept it via code? And then check that the flightpath data structure has the new color?

joernu76 (commented):
I think you have not really understood the purpose of mocking and testing. You mock only things, which are "external" to the tested function, to provide inputs that'd otherwise not be there. 

E.g. you can mock a colour selection dialogue of QT away to directly provide a colour.
Then you trigger the dialogue and test that the colour was properly applied to the plot and/or internal data structure.
You do not want to test the QT dialogue, so you can mock it away to avoid the hassle of programmatically manipulating the dialogue.
Important is however the proper test of the effect!

tests/_test_msui/test_multiple_flightpath_dockwidget.py
rohit2p: I have to use mock for accepting the dialog otherwise it won't manually idk why, I have tried different things but you are right i have used mock unnecessary in many places

tests/_test_msui/test_multiple_flightpath_dockwidget.py
matrss: I am not quite sure if I understand you correctly, are you suggesting to do the mock as a decorator on the test function? I.e. with
```python
@mock.patch(...)
def test_...():
```
, as is used in some other places?

If yes, then I'd argue against that. The mock should only be in effect for the action that needs something to be mocked. This is easy to achieve using a contextmanager like here, while using it as a decorator will apply the mock broadly for the entirety of the test function.

I think it is most readable to use it as a contextmanager and explain right there in a comment why the mock is necessary and no alternative was usable.

tests/_test_msui/test_multiple_flightpath_dockwidget.py
```diff
@@ -25,33 +25,230 @@
...
+
+def test_setColour(main_window):
+    """
+    test for the filghttrack colour
+    """
+    main_window, multiple_flightpath_widget = main_window
+    color_button = multiple_flightpath_widget.pushButton_color
+
+    # Mock the exec_ method to accept the dialog otherwise test get stuck
+    with mock.patch("mslib.utils.colordialog.CustomColorDialog.exec_",
```
matrss: Even this mock shouldn't be necessary. I think it should be possible to programmatically follow the entire flow of selecting a color doing this:
```python
# Activate the first flight track
activate_flight_track_at_index(main_window, 0)

# Click on the second flight track in the docking widget
click_on_flight_track_in_docking_widget_at_index(multiple_flightpath_widget, 1)

# Simulate clicking the button to open the color dialog
color_button.click()

# Get a reference to the custom color dialog
color_dialog = multiple_flightpath_widget.findChild(QtWidgets.QDialog)
assert color_dialog is not None

# Select the first color
color = QtGui.QColor(color_dialog.colors[0])
color_dialog.color_buttons[0].click()

# Verify that the flight track data structure was updated with the new color
wp_model = multiple_flightpath_widget.list_flighttrack.currentItem().flighttrack_model
applied_color_rgba = multiple_flightpath_widget.get_color(wp_model)

# Convert the applied_color_rgba to a QColor object
applied_color = QtGui.QColor.fromRgbF(*applied_color_rgba)

assert applied_color == color
```

This just requires some minor adjustments in other parts:
```patch
diff --git a/mslib/msui/multiple_flightpath_dockwidget.py b/mslib/msui/multiple_flightpath_dockwidget.py
index 380fb6bc..bdaf9cfa 100644
--- a/mslib/msui/multiple_flightpath_dockwidget.py
+++ b/mslib/msui/multiple_flightpath_dockwidget.py
@@ -422,7 +422,7 @@ class MultipleFlightpathControlWidget(QtWidgets.QWidget, ui.Ui_MultipleViewWidge
                 else:
                     color_dialog = CustomColorDialog(self)
                     color_dialog.color_selected.connect(lambda color: self.apply_color(wp_model, color))
-                    color_dialog.exec_()
+                    color_dialog.show()
             else:
                 self.labelStatus.setText("Status: Check Mark the flighttrack to change its color.")
         elif self.list_operation_track.currentItem() is not None:
@@ -983,7 +983,7 @@ class MultipleFlightpathOperations:
                 else:
                     color_dialog = CustomColorDialog(self.parent)
                     color_dialog.color_selected.connect(lambda color: self.apply_color(op_id, color))
-                    color_dialog.exec_()
+                    color_dialog.show()
             else:
                 self.parent.labelStatus.setText("Status: Check Mark the flighttrack to change its color.")
 
diff --git a/mslib/utils/colordialog.py b/mslib/utils/colordialog.py
index 53ac19f2..6a2a0f0f 100644
--- a/mslib/utils/colordialog.py
+++ b/mslib/utils/colordialog.py
@@ -60,6 +60,7 @@ class CustomColorDialog(QtWidgets.QDialog):
             row = i // 5
             col = i % 5
             swatch_layout.addWidget(button, row, col)
+            self.color_buttons.append(button)
 
         # Add "Pick Custom Color" button
         self.custom_color_button = QtWidgets.QPushButton("Pick Custom Color")
```

Adding the buttons to self.color_buttons is something that I presume was just overlooked anyway, since the list exists but is unused so far. The change from exec_ to show is a more appropriate use of the widget IMO, since exec_ blocks further execution and show doesn't. The former is only really the right choice when the return value of exec_ is then used, but in this case the dialog communicates via signals, which is an asynchronous style compatible with show.

tests/_test_msui/test_multiple_flightpath_dockwidget.py
```diff
@@ -25,33 +25,230 @@
...
+        # Simulate clicking the button to open the color dialog
+        color_button.click()
+        assert mock_exec.call_count == 1
+
+        color_dialog = multiple_flightpath_widget.findChild(QtWidgets.QDialog)
+        assert color_dialog is not None
+
+        # Simulate a color being selected
+        color = QtGui.QColor("#0000ff")  # Example color
+        color_dialog.color_selected.emit(color)
```
matrss: My above suggestion would eliminate this signal emission, which I think doesn't belong into a test anyway. Completely mocking the exec_ method and directly emitting this signal breaks encapsulation and entirely skips a significant path through the color dialog and how it is hooked up with the MFDW, which I think is undesirable in a test.

tests/_test_msui/test_multiple_flightpath_dockwidget.py
```diff
@@ -25,33 +25,227 @@
...
+        # Select random colors multiple times
+        selected_colors = set()
+        for _ in range(10):  # Test with 10 random selections
+            color = multiple_flightpath_widget.get_random_color()
+            normalized_color = tuple(int(c * 255) for c in color)
+            assert normalized_color not in selected_colors, "Duplicate color selected!"
+            selected_colors.add(normalized_color)
+
+        # Ensure the mock method was called 10 times
+        assert mock_get_random_color.call_count == 10
```
matrss: The mock and this assert aren't needed: obviously the assert is true, calling that function 10 times is what the test does.

tests/_test_msui/test_multiple_flightpath_dockwidget.py
```diff
@@ -25,33 +25,227 @@
...
+    # mock update_flighttrack_patch method to check if it gets called
+    with mock.patch.object(multiple_flightpath_widget, "update_flighttrack_patch") as mock_update_patch:
+        # s.et the line width
+        multiple_flightpath_widget.dsbx_linewidth.setValue(3.0)
+
+        # Verify the line width has been updated in dict_flighttrack and the patch was updated
+        wp_model = item.flighttrack_model
+        assert multiple_flightpath_widget.dict_flighttrack[wp_model]["linewidth"] == 3.0
+        assert multiple_flightpath_widget.change_linewidth is True
+        mock_update_patch.assert_called_once_with(wp_model)
```
matrss: This assert could be moved up to right after the `multiple_flightpath_widget.dsbx_linewidth.setValue` call. Then, the remaining asserts don't actually need the mock anymore, so can be dedented out of the with block. The mock should only be held as long as it is needed, and not any longer.

matrss (changes requested):
Some more minor comments, otherwise this is looking good!

matrss (approved):
LGTM, Thanks!

PR #2477, opened 2024-08-22
Fix #2476

### LINKED ISSUE #2476: archive operation triggers exception in mscolab
due to a too large value. The days needs to be significantly smaller to not trigger an error in a) conversion to timedelta and b) addition to "now".

```
File "/home/mss-mscolab/Miniforge/envs/mssenv/lib/python3.10/site-packages/mslib/mscolab/server.py", line 551, in set_last_used
    datetime.datetime.now(tz=datetime.timezone.utc) - datetime.timedelta(days=days_ago),
OverflowError: Python int too large to convert to C int
```
and
```
  File "/home/mss-mscolab/Miniforge/envs/mssenv/lib/python3.10/site-packages/mslib/mscolab/server.py", line 551, in set_last_used
    datetime.datetime.now(tz=datetime.timezone.utc) - datetime.timedelta(days=days_ago),
OverflowError: date value out of range
```

joernu76:
**Purpose of PR?**:

Fixes #2476 

no

### DIFF
--- a/mslib/msui/mscolab.py
+++ b/mslib/msui/mscolab.py
@@ -1788,7 +1788,7 @@ class MSUIMscolab(QtCore.QObject):
                     "token": self.token,
                     "op_id": self.active_op_id,
                     # when a user archives an operation we set the max “natural” integer in days
-                    "days": sys.maxsize,
+                    "days": 99999,  # larger values run into a lot of issues
                 }
                 url = urljoin(self.mscolab_server_url, 'set_last_used')
                 try:
--- a/mslib/mscolab/server.py
+++ b/mslib/mscolab/server.py
@@ -676,6 +676,10 @@ def set_last_used():
     op_id = request.form.get('op_id', None)
     user = g.user
     days_ago = int(request.form.get('days', 0))
+    if days_ago > 99999:
+        days_ago = 99999
+    elif days_ago < -99999:
+        days_ago = -99999
     fm.update_operation(int(op_id), 'last_used',
                         datetime.datetime.now(tz=datetime.timezone.utc) - datetime.timedelta(days=days_ago),
                         user)


joernu76:
This goes to develop as it is part of a larger set of hopefully smaller fixes to make archiving work again.

ReimarBauer:
it can't be based on stable without merge conflicts

mslib/msui/mscolab.py
```diff
@@ -1788,7 +1788,7 @@ def archive_operation(self):
                     "token": self.token,
                     "op_id": self.active_op_id,
                     # when a user archives an operation we set the max “natural” integer in days
-                    "days": sys.maxsize,
+                    "days": 99999,  # larger values run into a lot of issues
```
ReimarBauer, line 1791: Unfortunately someone with an old version will set the value where you see a lot of issues. This can be solved when we change the server code,  the server has to do where we used sys.maxsize always your new value. 

We likly have to deprecate the API that the client sends the days at all. 

Because the user can't alter the value we should force the value on the server, and remove the form data of days

please fix it in the server code
https://github.com/Open-MSS/MSS/blob/develop/mslib/mscolab/server.py#L678

And on a migration script for the database we can adjust that number too.
Please add a ToDo or issue as follow up.

ReimarBauer (changes requested):
we have to fix this in the server or some clients will send the old constant

ReimarBauer:
For the time based there is also an ARCHIVE_THRESHOLD defined, which can be altered on the server config.  

https://github.com/Open-MSS/MSS/blob/develop/mslib/mscolab/conf.py#L44

mslib/msui/mscolab.py
joernu76: I do not understand the migration scripts in mss/mslib/mscolab/migrations/versions in particular the hashes? What do they relate to. I do not find them in git log.
I am also not sure that a fix is necessary as the server will crash *before* setting an illegal date, so the dates in the database should be fine.
So this can probably be merged as is.

joernu76:
> For the time based there is also an ARCHIVE_THRESHOLD defined, which can be altered on the server config.
> 
> https://github.com/Open-MSS/MSS/blob/develop/mslib/mscolab/conf.py#L44

? Yes, this MR fix does not change that and should use this variable.

mslib/msui/mscolab.py
joernu76: The server checks the value supplied by client and truncates the value. So old clients should work as well. "Normal" legit use for days>99999 are not possible.

ReimarBauer: > I do not understand the migration scripts in mss/mslib/mscolab/migrations/versions in particular the hashes? What do they relate to. I do not find them in git log. I am also not sure that a fix is necessary as the server will crash _before_ setting an illegal date, so the dates in the database should be fine. So this can probably be merged as is.

In the database the hashes are set and connected by that to the scripts. 


https://www.digitalocean.com/community/tutorials/how-to-perform-flask-sqlalchemy-migrations-using-flask-migrate

joernu76: OK, the link gives some context, but I still have issues with that. However, this PR does not change the schema, thus flask db migrate does not generate a script *and* this script does not even affect values currently in the database, where datetimes are stored not the values being communicated via API.
So, I will look into migration for the other draft PR and this can be merged as it is.

ReimarBauer (approved):
Thx

PR #2479, opened 2024-08-22
Fix #2478

### LINKED ISSUE #2478: Operation Archive always empty for certain configuration options
When the configuration option of skipping archived operations is selected, the Operation Archive will always be empty.
Due to the default settings, this will confuse the user who likely assumes that the feature is broken or that the archived operation has been lost.

joernu76:
**Purpose of PR?**:

Fixes #2478

no


ReimarBauer (approved):
ok

PR #2480, opened 2024-08-22
rework active+archiving

### LINKED ISSUE #2497: Rework archiving
Archiving is currently both automatic and manual, which sometimes interacts in a surprising fashion.
In a first step, a deterministic, stable behaviour should be created by making archiving fully manual such that no desired operations vanish from sight, neither archived operations become activated unitentionally.

joernu76:
Fix #2497

### DIFF
--- a/mslib/msui/mscolab.py
+++ b/mslib/msui/mscolab.py
@@ -87,7 +87,7 @@ class MSColab_OperationArchiveBrowser(QtWidgets.QDialog, ui_opar.Ui_OperationArc
 
     def select_archived_operation(self, item):
         logging.debug('select_inactive_operation')
-        if item.access_level == "creator":
+        if item.access_level in ["creator", "admin"]:
             self.archived_op_id = item.op_id
             self.pbUnarchiveOperation.setEnabled(True)
         else:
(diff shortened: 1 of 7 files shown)


joernu76:
This is a first draft to simplify things. A cronjob or similar *could* do an automated archiving on server side, but I dislike this unless we can replace the active flag but a status enum which could give us a further state like "keep".

1) So, can we "simply" replace the "active" bool by a "status" int/enum?
2) I am unsure if/how a change to the status will cause an update of the operation lists of users. Would update_operation need to trigger something here, or is this the job of the function that called update_operation?

ReimarBauer:
@joernu76 have a look at https://github.com/kbr/autocron 
I helped packaging this for conda-forge

ReimarBauer:
Please add also a line to the docs about the archive feature.
https://github.com/Open-MSS/MSS/blob/develop/docs/mscolab.rst?plain=1#L200

ReimarBauer:
> 2\. I am unsure if/how a change to the status will cause an update of the operation lists of users. Would update_operation need to trigger something here, or is this the job of the function that called update_operation?

with reload_operations the client updates its list. 

The server can call update_operation_list. This emits 'operation-list-update'. The client receives this and makes an self.signal_operation_list_updated.emit() The signal is on the client connected to self.reload_operation_list which then triggers reload_operations

joernu76:
> Please add also a line to the docs about the archive feature. https://github.com/Open-MSS/MSS/blob/develop/docs/mscolab.rst?plain=1#L200

done

joernu76:
> > 2. I am unsure if/how a change to the status will cause an update of the operation lists of users. Would update_operation need to trigger something here, or is this the job of the function that called update_operation?
> 
> with reload_operations the client updates its list.
> 
> The server can call update_operation_list. This emits 'operation-list-update'. The client receives this and makes an self.signal_operation_list_updated.emit() The signal is on the client connected to self.reload_operation_list which then triggers reload_operations

I will have a closer look at a couple of interactions between "normal" and "mscolab" in the next weeks, and maybe integrate both into a single class because the split is causing a couple of issues, currently. I hope that this can also be done in smaller steps.

mslib/msui/mscolab.py
```diff
@@ -87,7 +87,7 @@ def __init__(self, parent=None, mscolab=None):
 
     def select_archived_operation(self, item):
         logging.debug('select_inactive_operation')
-        if item.access_level == "creator":
+        if item.access_level in ["creator", "admin"]:
```
ReimarBauer, line 90: this change must also be done in the UI too, or an admin cannot click the Operation - Maintenance - Archiv in the menu


need_changes_to_activate_button

ReimarBauer: There you need to enable the functions on the given role
https://github.com/Open-MSS/MSS/blob/develop/mslib/msui/mscolab.py#L1967

ReimarBauer (changes requested):
see comment

joernu76:
nice catch

mslib/msui/mscolab.py
joernu76: nice catch

ReimarBauer (approved):
I found a problem for the "other" users  not seeing on their interface that an operation they have open got archieved. As long they are connected they are doing changes.

That maybe worked on in a different PR

archive_switch_does_not_trigger_the_default_operation

The Active Users label sees 2 connected and also the chat. 

image

joernu76:
I would agree. Please open an issue for this. I would assume that a review for the general behaviour due to server side changes (e.g. change of the category of an operation) would be in order.

Issue #2206, opened 2024-02-15
Re-think the develop/stable workflow


matrss:
I don't like the way it is now.

- PRs go to two different branches depending on what they do. This means development is happening on two base branches which will always produce unnecessary conflicts.
- CI failures on stable are basically invisible, since the repository landing page does not show them prominently.
- PRs going to a non-default branch do not automatically close issues that are mentioned as "Fixes #" in the PR description. This means PRs to stable always have to remember to manually add this link in GitHub.

My intuition on how this should work and what I initially thought it was like is this:
1. all PRs go to develop
2. for each PR, one can decide to also backport this PR to stable (e.g. make a cherry-pick on stable to apply the change there too) (note: I think it would be better to just call this branch by its released version number, e.g. a v8.3 or v8 branch that gets backports for fixes that should go to that version as a minor/patch version)
3. whenever a new release is cut from develop stable is set to point at the same commit as develop

This would also completely eliminate any merge commits in the git history for the future. I don't like merge commits either.

matrss:
Even when explicitly linking a PR to an issue the issue is not automatically closed, for some reason. This happened in #2205: I've linked the PR to #2199 manually, then it was merged, and then the issue was still not automatically closed.

ReimarBauer:
github has some limitations compared to bitbucket where we started.

We have for a good reason a stable branch. This is used to destinguish what features and bug fixes are for.

Bug fixes have to go to the stable branch and a minor or patch level release is the follow up.  We have to regular merge from stable to develop. This is a pretty good constant where we also know what a release is based on and what only gets to a user. 

The main problem is, that we need long to make new major releases. But when we have constantly more people this can speedup.

matrss:
> We have for a good reason a stable branch. This is used to destinguish what features and bug fixes are for.

I am not suggesting to get rid of stable, and I am also not suggesting to no longer differentiate between bug fixes and features. I am just suggesting to re-think the workflow between the two, since right now we effectively have two development branches, which leads to unnecessary conflicts.

[...]

PR #2482, opened 2024-08-22
Merge stable to develop

ReimarBauer:
**Purpose of PR?**:

get recent fixes into develop

### DIFF
--- a/.github/workflows/testing-all-oses.yml
+++ b/.github/workflows/testing-all-oses.yml
@@ -48,6 +48,4 @@ jobs:
       # I have no idea yet on why this happens and how to fix it.
       # Even a module level skip is not enough, they need to be completely ignored.
       # TODO: fix those tests and drop the ignores
-      run: micromamba run -n ci env QT_QPA_PLATFORM=offscreen pytest -v -n logical --durations=20 --cov=mslib
-        --ignore=tests/_test_msui/test_sideview.py --ignore=tests/_test_msui/test_topview.py --ignore=tests/_test_msui/test_wms_control.py
-        tests
+      run: micromamba run -n ci env QT_QPA_PLATFORM=offscreen pytest -v -n logical --durations=20 --cov=mslib tests
--- a/.github/workflows/testing-scheduled.yml
+++ b/.github/workflows/testing-scheduled.yml
@@ -10,7 +10,7 @@ jobs:
     permissions:
       actions: write
     steps:
-      - uses: benc-uk/workflow-dispatch@v1.2.3
+      - uses: benc-uk/workflow-dispatch@v1.2.4
         with:
           workflow: testing-stable.yml
           ref: stable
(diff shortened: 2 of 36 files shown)


joernu76 (approved):

matrss:
My understanding was that these kinds of PRs should be merged using a real merge, not by squashing them.

joernu76:
dammit, my bad. shall I revert this?

joernu76:
We can revert by MR or by force push. probably should use the revert feature?

matrss:
I think force pushing would be a bad idea since it is the public develop branch, not some feature branch. At least one of the GSoC students has already pulled the squash commit, so force pushing would necessitate some cleanup. Revert seems better, but ultimately I would defer that decision to @ReimarBauer.

ReimarBauer:
I think also revert is better. Please prepare that. We learn also.

joernu76:
Yes, we should rediscuss this. This squashing/non-squashing is always an issue. gitlab handles this much better IMO.

matrss:
FWIW, if we used a backports-based workflow as suggested in #2206 then this kind of accident would become impossible. Every PR merge would always be a squash. Also, there would be no confusion as to which branch to target with PRs, as (almost) all changes would go to develop first.

I don't see how GitLab would handle this differently, as long as there are different kinds of PRs that need to be treated differently there will be a potential for these mistakes.

ReimarBauer:
The downside of the approach is that I think it fits good to EffVer and I have doubts that we will do the right changes for a released version having SemVer.
We need to have a different discussion than over all kinds of PRs to get to a decission.

joernu76:
Gitlab allows to determine the kind of merge upon creating the Merge Request via checkboxes. I.e. the creator can set them properly and the reviewer can check. 4-eye-principle. Whereas github forces the sole responsibility on the button-presser.
Another way to reduce the problem would be to immediately merge stable into develop after each-and-every fix, which would probably be a good idea anyway.
We really should schedule a day to discuss things. I doubt that there is an obvious solution to our problems, but would like to be positively surprised.

matrss:
> Gitlab allows to determine the kind of merge upon creating the Merge Request via checkboxes.

OK, I didn't know that. That indeed would make the situation somewhat better, but the result would be the equivalent of what we have with targeting stable vs develop for each PR: the creator of the PR must know what to choose, and the reviewer must check it. We have seen that this review is also easy to forget with the target branch.

A backports-based workflow would completely eliminate these potential pitfalls: essentially every new change goes to develop, every PR is squashed. Every merged or un-merged PR can be labeled as to-be-backported at any point in time, even long after it was merged, and as long as there are no conflicts an automated process can create a backport PR targeting stable. If there are conflicts the backport must be done manually. But then again, this would be a situation in which, right now, we would get semantic conflicts in the "merge stable to develop" PRs, which can be hard to spot when these PRs combine a bunch of unrelated changes into one and aren't even reviewed for semantic correctness of the changes.

I think there is a reason that many major open source projects supporting more than their develop version use this kind of workflow.

> We really should schedule a day to discuss things.

Yes please.

PR #2488, opened 2024-08-28
replace ANTIALIAS instances with LANCZOS

### LINKED ISSUE #2487: replace ANTIALIAS instances with LANCZOS
ANTIALIAS was removed in Pillow 10.0.0 (after being deprecated through many previous versions). Now we need to use PIL.Image.LANCZOS instead of PIL.Image.ANTIALIAS or else the tests suit will fail.

workaryangupta:
**Purpose of PR?**:

Fixes #2487


matrss:
This PR should target the stable branch, as it is also affected by the issue.

matrss (approved):

PR #2494, opened 2024-08-29
Merge develop into GSOC2024-AryanGupta

workaryangupta:
**Purpose of PR?**:

Merge recent changes from develop branch to my gsoc branch so as to keep it updated and avoid merge conflicts in the future.


ReimarBauer (commented):
did a first pass through.

tests/_test_mscolab/test_server.py
```diff
@@ -448,6 +483,8 @@ def _save_content(self, operation, userdata=None):
         if userdata is None:
             userdata = self.userdata
         user = get_user(userdata[0])
+        self.fm.save_file(operation.id, XML_CONTENT1, user)
+        return user
         fm = FileManager(self.app.config["MSCOLAB_DATA_DIR"])
```
ReimarBauer: Please check this. This looks not correct, the lines after the return can't be reached

tests/_test_mscolab/test_server.py
```diff
@@ -448,6 +483,8 @@ def _save_content(self, operation, userdata=None):
...
         fm.save_file(operation.id, "content1", user)
         return fm, user
```
ReimarBauer: when this the _save_content is, this should only return user

ReimarBauer (changes requested):
I assume this was the merge conflict.  When you look where _save_content is used, there it needs only user, not a tuple.




```

    def _save_content(self, operation, userdata=None):
        if userdata is None:
            userdata = self.userdata
        user = get_user(userdata[0])
<<<<<<< HEAD
        fm = FileManager(self.app.config["MSCOLAB_DATA_DIR"])
        fm.save_file(operation.id, "content1", user)
        return fm, user

    def _upload_profile_image(self, test_client, token, email):
        # Creating a dummy image
        img = Image.new('RGB', (64, 64), color='yellow')
        img_byte_arr = io.BytesIO()
        img.save(img_byte_arr, format='JPEG')
        img_byte_arr.seek(0)
        filename = "test.jpeg"

        # Post request for uploading the image
        user = get_user(email)
        data = {
            "user_id": str(user.id),
            "token": token,
            'image': (img_byte_arr, filename, 'image/jpeg')
        }
        response = test_client.post('/upload_profile_image', data=data)
        return response
=======
        self.fm.save_file(operation.id, XML_CONTENT1, user)
        return user
>>>>>>> develop
```


When that becomes merged the result would be

```
    def _save_content(self, operation, userdata=None):
        if userdata is None:
            userdata = self.userdata
        user = get_user(userdata[0])
        self.fm.save_file(operation.id, XML_CONTENT1, user)
        return user
        fm = FileManager(self.app.config["MSCOLAB_DATA_DIR"])
        fm.save_file(operation.id, "content1", user)
        return fm, user
```

The tests don't crash because the other lines not reached. And we don't have a test which checks our code for that problem.

ReimarBauer (approved):
Thx

PR #2495, opened 2024-08-30
remove the conda/mamba based updater.

### LINKED ISSUE #2256: Do not assume that MSS is installed with conda/mamba
Some code assumes that MSS is installed with conda or mamba. This happens e.g. here: https://github.com/Open-MSS/MSS/blob/f5034357cef3602afdb59ece9be683a990896c76/mslib/msidp/idp_conf.py#L40

Also, based on a rough skim of the code, the updater seems to assume that if conda is available in the PATH then MSS must have been installed with it. This is not guaranteed either.

ReimarBauer:
Show in About a new version available.

**Purpose of PR?**:

Partial fix of #2256

removes the updater, because it is not sufficient  when we release a pinning or when we install by pixi or a different package manager

removes a feature


joernu76 (approved):

mslib/utils/release_info.py
```diff
@@ -0,0 +1,75 @@
...
+        latest_release = {
+            'tag_name': release_data['tag_name'],
+            'release_name': release_data['name'],
+            'published_at': release_data['published_at'],
+            'url': release_data['html_url']
+        }
+        return latest_release
+
+    except requests.exceptions.RequestException as e:
+        print(f"Error fetching release data: {e}")
```
matrss: This should be a log message.

mslib/utils/release_info.py
```diff
@@ -0,0 +1,75 @@
...
+    # Compare the latest release with the previous known version
+    if latest_release['tag_name'] != installed_version:
+        github_url = f'<a href="{latest_release["url"]}#target">{latest_release["url"]}</a>'
+        result.append(f"New release found: {latest_release['release_name']} ({latest_release['tag_name']})")
+        result.append(f"Published at: {latest_release['published_at']}")
+        result.append(f"Release URL: {github_url}")
+        result = ' | '.join(result)
+    else:
+        result = no_new_release_found
+    return result
```
matrss: This could be simplified quite a bit:

```suggestion
    if latest_release is None or latest_release['tag_name'] == installed_version:
        return no_new_release_found

    github_url = f'<a href="{latest_release["url"]}#target">{latest_release["url"]}</a>'
    return ' | '.join([
        f"New release found: {latest_release['release_name']} ({latest_release['tag_name']})",
        f"Published at: {latest_release['published_at']}",
        f"Release URL: {github_url}",
    ])
```

mslib/mscolab/app/__init__.py
```diff
@@ -34,7 +35,10 @@
 from flask import Flask, url_for
 from mslib.mscolab.conf import mscolab_settings
 from flask_sqlalchemy import SQLAlchemy
-from mslib.utils import prefix_route
+from mslib.utils import prefix_route, release_info
+
+
+logging.debug(release_info.check_for_new_release())
```
matrss: A debug message seems to be the wrong log level to me. I think it should be at least info, maybe even a warning, or not a log message at all and instead a print.

matrss (changes requested):
Great to see this! Just some quick comments at first glance. Also, it doesn't actually address the issue described in #2256, the explicit reference to CONDA_PREFIX is still there.

mslib/utils/release_info.py
ReimarBauer: Yes, we should avoid print, because that didn't make it to the logging file and in worst case it can crash the server.

mslib/mscolab/app/__init__.py
ReimarBauer: The server output of a productive system usually is not seen on a terminal. 
No output and only a warning in the log when an update is needed is the better solution for the servers.

ReimarBauer:
> Great to see this! Just some quick comments at first glance. Also, it doesn't actually address the issue described in #2256, the explicit reference to CONDA_PREFIX is still there.

I know, it is a first step, this part needs further changes. There needs more based on the mscolab_settings.

ReimarBauer:
There will be a seperate PR applying changes to the gallery workflow. This needs currently an update of the docker image.
I prefer to use the way how RTFD did it.

docs/installation.rst
```diff
@@ -124,21 +124,7 @@ versions for dependencies. ::
...
-After you started the MSS UI it informs you after a while if there is a new update available.
-From the command line you can trigger this update feature by ::
-
-    (mssenv) $ msui --update
-
-
-
-Other Methods
+Update Methods
 .............
```
matrss: According to the rst docs I could find, a section underline must be at least as wide as the title.

mslib/utils/qt.py
```diff
@@ -385,9 +384,6 @@ def __init__(self, function):
             self.finished = NonQtCallback()
             self.failed = NonQtCallback()
 
-        self.failed.connect(lambda e: self._update_gui())
-        self.finished.connect(lambda x: self._update_gui())
```
matrss: What's the reasoning for removing these? I wouldn't be surprised if this breaks some of the other uses of the Worker class, but I also have no idea why this was here anyway.

mslib/utils/qt.py
```diff
@@ -420,96 +416,6 @@ def create(function, on_success=None, on_failure=None, start=True):
...
-        c_list = self._execute_command(f"{self.command} list -f mss")
-        self.old_version = c_list.split("\n")[-2].split()[1]
-        if any(c.isdigit() for c in self.new_version):
-            if self.new_version > self.old_version:
-                self.on_status_update.emit("Your version of MSS is outdated!")
-                self.on_update_available.emit(self.old_version, self.new_version)
-            else:
-                self.on_status_update.emit("Your MSS is up to date.")
-
     def _restart_msui(self):
```
matrss: Shouldn't this method be removed as well?

mslib/utils/release_info.py
```diff
@@ -0,0 +1,80 @@
...
+    except requests.exceptions.RequestException as e:
+        logging.debug(f"Error fetching release data: {e}")
+        return None
+
+
+def check_for_new_release():
+    no_new_release_found = f"{datetime.date.today()}: No new release found."
+    try:
+        # we use the @concurrent.process(timeout=1) on the function to stop requests trying to establish a connection
+        latest_release = get_latest_release().result()
```
matrss: Immediately calling `.result()` IMO defeats the purpose of using pebble at all. What does this do that isn't already accomplished by just the timeout of `requests.get`?

matrss (changes requested):

matrss:
> > Great to see this! Just some quick comments at first glance. Also, it doesn't actually address the issue described in #2256, the explicit reference to CONDA_PREFIX is still there.
> 
> I know, it is a first step, this part needs further changes. There needs more based on the mscolab_settings.

That's fine, but then the PR description can't mention that it fixes that issue, when it really doesn't.

ReimarBauer:
It doesn't fix the name but in the discussion in the issue the problem of what needs to be changed has more facettes.

mslib/utils/release_info.py
ReimarBauer: requests.get timeout is used when the connection is established.

But it retrys 10 seconds until it gives up when there is no connection possible.

This is far to long for an user clicking on about, and on the server the startup or restart can be blocked too.

matrss: Do you have an example to reproduce the issue you see with it? The `requests.get` timeout specifies both the time to wait for a connection to be established and the time to wait for responses from the server. With `timeout=(1, 1)` it should also fail after at most 1 second if there is no connection possible (well, unless the target domain has multiple IP addresses, but that doesn't seem to be the case here).

The only other situation that I am aware of where the request might take longer is when the server is responding very slowly, waiting for almost the timeout before sending another few bytes. That seems to be an unlikely edge case though, especially when accessing GitHub's API?

matrss:
> It doesn't fix the name but in the discussion in the issue the problem of what needs to be changed has more facettes.

Yes, but merging this PR would auto-close that issue, while the issue isn't solved. And that is just plain wrong. If you want an issue attached to this PR, please make removing the updater a separate issue, since that was only brought up in #2256, but not the main point of that issue, or make removing the usage of CONDA_PREFIX part of this PR.

docs/installation.rst
ReimarBauer: at least same length, longer was in the past also correct.

ReimarBauer:
> > It doesn't fix the name but in the discussion in the issue the problem of what needs to be changed has more facettes.
> 
> Yes, but merging this PR would auto-close that issue, while the issue isn't solved. And that is just plain wrong. If you want an issue attached to this PR, please make removing the updater a separate issue, since that was only brought up in #2256, but not the main point of that issue, or make removing the usage of CONDA_PREFIX part of this PR.

It is only a matter of the merge order. When it is already closed, it is connected which is also valueable.

from the issue "Some code assumes that MSS is installed with conda or mamba. This happens e.g. here:"

for me e.g. means lookup also others.

matrss:
OK, makes sense, then both this PR and #2499 must be merged at the same time, as only the combination of both addresses #2256.

mslib/utils/release_info.py
ReimarBauer: Just load the requests module and disconnect (or have no) internet before you execute

requests.get('https://duckduckgo.com/', timeout=(1,1))

with a connection available this shows a 200, without you wait until you get a requests.exceptions.ConnectionError

We can use MSS offline, we should not add something which can block this.


This behaves a bit different, it is faster than the about dialog

ReimarBauer: Let's look on the gifcycle how it can  block my system by commenting of the  decorator and starting the local server
problem

ReimarBauer:
> OK, makes sense, then both this PR and #2499 must be merged at the same time, as only the combination of both addresses #2256.

It is also common in the "unperfect" world of github to add comments to already closed issues. I splitted the PR from an undo or reading perspective.

mslib/utils/release_info.py
ReimarBauer: there is also a longish discussion for this problem on https://stackoverflow.com/questions/21965484/timeout-for-python-requests-get-entire-response and other places.

The decorator solves the  problem of requests waiting long when it can't establish a connection.

mslib/utils/qt.py
ReimarBauer: I had the same question. The tests didn't fail, but actually I am not sure besides creating tutorials if I would recognize a problem too.  I add it back.

mslib/utils/release_info.py
matrss: > Just load the requests module and disconnect (or have no) internet before you execute
> 
> requests.get('https://duckduckgo.com/', timeout=(1,1))
> 
> with a connection available this shows a 200, without you wait until you get a requests.exceptions.ConnectionError

```
$ time python3 -c "import requests; requests.get('https://duckduckgo.com/', timeout=(1,1))"
[...]
requests.exceptions.ConnectionError: HTTPSConnectionPool(host='duckduckgo.com', port=443): Max retries exceeded with url: / (Caused by NameResolutionError("<urllib3.connection.HTTPSConnection object at 0x77467b735540>: Failed to resolve 'duckduckgo.com' ([Errno -2] Name or service not known)"))

real	0m0,166s
user	0m0,121s
sys	0m0,024s
[ble: exit 1]
```

I don't see the problem. Same when I remove the decorator from that function, it is actually faster without an internet connection than with one because the name resolution fails immediately. Both with and without an internet connection it doesn't get close to the 1s timeout on my system.

> there is also a longish discussion for this problem on https://stackoverflow.com/questions/21965484/timeout-for-python-requests-get-entire-response and other places.

The issue described there is what I mentioned here: 

> The only other situation that I am aware of where the request might take longer is when the server is responding very slowly, waiting for almost the timeout before sending another few bytes. That seems to be an unlikely edge case though, especially when accessing GitHub's API?

And I am fairly certain that this situation is rather unlikely to actually occur.

Most notably, that link _does not_ discuss problems with timeouts for establishing a connection.

> The decorator solves the problem of requests waiting long when it can't establish a connection.

The `timeout` argument already solves that problem.

ReimarBauer:
```
url = "https://api.github.com/repos/Open-MSS/MSS/releases/latest"
requests.get(url, timeout=(1,1))
```
takes 10 sec. without network

matrss: > Let's look on the gifcycle how it can block my system by commenting of the decorator and starting the local server

It's impossible to see from that gif where that time is actually spent. Can you please measure the execution duration of `get_latest_release()`? Something like this should work:
```diff
diff --git a/mslib/utils/release_info.py b/mslib/utils/release_info.py
index e03ca686..c6809362 100644
--- a/mslib/utils/release_info.py
+++ b/mslib/utils/release_info.py
@@ -33,7 +33,6 @@ from pebble import concurrent
 from mslib.version import __version__ as installed_version
 
 
-@concurrent.process(timeout=1)
 def get_latest_release():
     # GitHub API URL for the MSS Release
     url = "https://api.github.com/repos/Open-MSS/MSS/releases/latest"
@@ -62,12 +61,18 @@ def get_latest_release():
 
 def check_for_new_release():
     no_new_release_found = f"{datetime.date.today()}: No new release found."
+    import time
+    start = time.perf_counter()
     try:
         # we use the @concurrent.process(timeout=1) on the function to stop requests trying to establish a connection
-        latest_release = get_latest_release().result()
+        latest_release = get_latest_release()
     except TimeoutError as e:
         logging.debug(f"Error fetching release data: {e}")
         latest_release = None
+    finally:
+        elapsed = time.perf_counter() - start
+        print("Elapsed time:", elapsed)
+
 
     if latest_release is None or latest_release['tag_name'] == installed_version:
         return no_new_release_found, False
```

ReimarBauer:
```
(mssdev) reimarb@infinity:~/PycharmProjects/MSS$ python mslib/mswms/mswms.py 
WARNING:root:Couldn't import mswms_settings (ImportError:'No module named 'mswms_settings''), Using dummy config.
Elapsed time: 10.012056518999998
WARNING:root:Couldn't import mswms_auth (ImportError:'{No module named 'mswms_auth'), creating dummy config.
INFO: Configuration File: 'None'
 * Serving Flask app 'mslib.mswms.app'
 * Debug mode: off
INFO: WARNING: This is a development server. Do not use it in a production deployment. Use a production WSGI server instead.
 * Running on http://127.0.0.1:8081
INFO: Press CTRL+C to quit


```

ReimarBauer:
```
url = "https://api.github.com/repos/Open-MSS/MSS/releases/latest"
requests.get(url, timeout=(1,1))
```

my click time is slow ;) but it is now several times repeated and it is ~10secs,

github_api

ReimarBauer: I also recognized, when I have tried the url with network, I need 2-3 times calling it after the network is off to get to the problem.

ReimarBauer: When we agree to a solution like this, we could also do similiar with the airdata etc. download

matrss: Hmm, you are getting a different error than I do when disabling the network connection, for some reason. Any idea why? I would consider this a bug in requests, as it clearly does not obey the timeout setting. Could you please report that?

I am still not convinced that pulling in an entire library for threading is the right move to hide a bug behind the heavy weight of setting up an entire thread for a single http request. Where is the thread pool set up for this? How much does setting it up cost for this single usage? Will it be reused if other places also use pebble, or will it need a new one? Wouldn't the python stdlib suffice (there is similar stuff already in `concurrent.futures`)?

ReimarBauer: Which network error do you have? and which requests version, I've just seen I hadn't the  most recent.

Well we change our stack for our needs. When we can replace the other worker usages with it after the GSOC merges it would be usefull too.


I look if that concurrent.futures also helps.

ReimarBauer: With requests                  2.32.3             pyhd8ed1ab_0    conda-forge

seems we need to uplift the pinning

it seems to work 
```

  File "<stdin>", line 1, in <module>
  File "/home/reimarb/Miniforge/envs/mssdev/lib/python3.11/site-packages/requests/api.py", line 73, in get
    return request("get", url, params=params, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/reimarb/Miniforge/envs/mssdev/lib/python3.11/site-packages/requests/api.py", line 59, in request
    return session.request(method=method, url=url, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/reimarb/Miniforge/envs/mssdev/lib/python3.11/site-packages/requests/sessions.py", line 589, in request
    resp = self.send(prep, **send_kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/reimarb/Miniforge/envs/mssdev/lib/python3.11/site-packages/requests/sessions.py", line 703, in send
    r = adapter.send(request, **kwargs)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/reimarb/Miniforge/envs/mssdev/lib/python3.11/site-packages/requests/adapters.py", line 700, in send
    raise ConnectionError(e, request=request)
requests.exceptions.ConnectionError: HTTPSConnectionPool(host='api.github.com', port=443): Max retries exceeded with url: /repos/Open-MSS/MSS/releases/latest (Caused by NameResolutionError("<urllib3.connection.HTTPSConnection object at 0x7862fb9f88d0>: Failed to resolve 'api.github.com' ([Errno -3] Temporary failure in name resolution)"))
```

ReimarBauer: mamba list -r
```
2024-09-02 18:09:17  (rev 6)
     certifi  {2024.7.4 (conda-forge/noarch) -> 2024.8.30 (conda-forge/noarch)}
     requests  {2.31.0 (conda-forge/noarch) -> 2.32.3 (conda-forge/noarch)}

2024-09-02 18:16:53  (rev 7)
     requests  {2.32.3 (conda-forge/noarch) -> 2.32.1 (conda-forge/noarch)}

2024-09-02 18:18:25  (rev 8)
     requests  {2.32.1 (conda-forge/noarch) -> 2.31.0 (conda-forge/noarch)}

```
Maybe the problem was only on my system?  It is also gone currently with the old version.

matrss: > Which network error do you have?

As mentioned above:
```
requests.exceptions.ConnectionError: HTTPSConnectionPool(host='duckduckgo.com', port=443): Max retries exceeded with url: / (Caused by NameResolutionError("<urllib3.connection.HTTPSConnection object at 0x77467b735540>: Failed to resolve 'duckduckgo.com' ([Errno -2] Name or service not known)"))
```

So I am getting a "Errno -2 Name or service not known" while you are getting an "Errno -3 Temporary failure in name resolution".

ReimarBauer: I still can't reproduce the problem I had before.  But I always now have  Errno -3 independent from cable or wireless disconnection

matrss (approved):
I've updated the PR description to reflect that this is a partial fix of said issue. LGTM now.

Issue #1632, opened 2022-12-06
drag and drop for operations and flight tracks


ReimarBauer:
On UIs usually one could move items from one section to another one.

We can't do this for a flight track to become a new operation and also not for an operation to become a new flight track

To exchange between both we have to use a local file export. That could be improved.

SimranPatil26:
**Purpose of PR -**

This PR introduces drag and drop functionality to allow seamless movement of flight tracks and operations within the UI. Previously, users could not convert a flight track into a new operation or vice versa without exporting and re-importing a local file. This PR resolves that limitation.

**Fixes #1632**

_**No**_, this PR enhances functionality without affecting existing workflows.

**Manual Verification and Scenarios Covered -**

**_Before Fix:_**
Drag-and-drop was not supported between flight tracks and operations.
Users had to manually export/import local files to exchange data.

**_After Fix:_**
Drag-and-drop enabled for seamless conversion between flight tracks and operations.
New UI elements allow users to drop items into the desired section.
Backend updates ensure proper data handling after drag-and-drop actions.

**Additional Information for Reviewer -**

This PR introduces event-based drag-and-drop handling in the UI.
Updates existing data handling methods to ensure smooth data exchange between flight tracks and operations.

**_Yes_**, the user guide needs an update to document the new drag-and-drop feature.

**Code Changes Implemented -**

    import tkinter as tk
    from tkinter import messagebox

     class DragDropManager:
    def _init_(self, root):
        self.root = root
        self.source = None
        self.create_ui()

    def create_ui(self):
        self.root.title("Drag and Drop for Flight Tracks and Operations")
        
        # Flight Track Section
        self.flight_track_listbox = tk.Listbox(self.root, height=5, selectmode=tk.SINGLE)
        self.flight_track_listbox.insert(tk.END, "Flight Track 1", "Flight Track 2")
        self.flight_track_listbox.bind("<ButtonPress-1>", self.start_drag)
        self.flight_track_listbox.pack(pady=10)

        # Operations Section
        self.operation_listbox = tk.Listbox(self.root, height=5, selectmode=tk.SINGLE)
        self.operation_listbox.insert(tk.END, "Operation A", "Operation B")
        self.operation_listbox.bind("<ButtonRelease-1>", self.drop_item)
        self.operation_listbox.pack(pady=10)

        # Instruction Label
        tk.Label(self.root, text="Drag Flight Tracks and drop them into Operations").pack(pady=5)

    def start_drag(self, event):
[...]

PR #2499, opened 2024-09-01
renamed and refactored names for mscolab 

### LINKED ISSUE #2498: review / rename / refactor MSCOLAB config vars
Some names have a prefix of MSCOLAB which is superflous.

Some are hardcoded

ReimarBauer:
**Purpose of PR?**:

Fixes #2498, #2256

This needs a HINT on the next CHANGES because setting vars are renamed

**Checklist:**
- [ ] Bug fix. Fixes #2498, #2256
- [X] Commit has unit tests


mslib/mscolab/conf.py
```diff
@@ -57,8 +57,20 @@ class default_mscolab_settings:
...
+
+    # dir where mscolab single sign-on process files are stored
+    SSO_DIR = os.path.join(DATA_DIR, 'datasso')
+
+    # Enable SSL certificate verification during SSO between MSColab and IdP
+    SSO_SSL_CERT_VERIFICATION = True
+
+    # path to xmlsec1 (sign, verify, encrypt and decrypt XML documents)
+    # this is only needed for a testing IDP setup
+    # SSO_XMLSEC_PATH = os.path.join(os.environ["CONDA_PREFIX"], "bin", "xmlsec1")
```
matrss: I don't think making this configurable is a good idea either. There is a standard environment variable to communicate the search path for executables, it's called `PATH`. Why not just use that?

matrss (changes requested):

mslib/mscolab/conf.py
ReimarBauer: Because of the assumptation https://pysaml2.readthedocs.io/en/latest/howto/config.html#xmlsec-binary you don't have a version you want to use not in the correct "system" path.

ReimarBauer: looks like it is fixed
```
>>> from saml2.sigver import get_xmlsec_binary
>>> get_xmlsec_binary()
'/home/reimar/Miniforge/envs/mssdev/bin/xmlsec1'
```

conftest.py
```diff
@@ -112,8 +112,8 @@ def generate_initial_config():
 BASE_DIR = ROOT_DIR
 DATA_DIR = fs.path.join(ROOT_DIR, 'colabTestData')
 # mscolab data directory
-MSCOLAB_DATA_DIR = fs.path.join(DATA_DIR, 'filedata')
-MSCOLAB_SSO_DIR = fs.path.join(DATA_DIR, 'datasso')
+FILE_DATA = fs.path.join(DATA_DIR, 'filedata')
```
matrss: Is there some kind of naming convention to follow? Other directories have the _DIR suffix for example. Also, from the name it is unclear what this even contains. Maybe `OPERATIONS_DATA` might make more sense?

mslib/mscolab/app/__init__.py
```diff
@@ -47,7 +47,7 @@
 APP.config.from_object(__name__)
 APP.route = prefix_route(APP.route, SCRIPT_NAME)
 
-APP.config['MSCOLAB_DATA_DIR'] = mscolab_settings.MSCOLAB_DATA_DIR
+APP.config['MSCOLAB_DATA_DIR'] = mscolab_settings.FILE_DATA
```
matrss: Shouldn't the key used with APP.config be updated as well?

matrss: Actually, why do we even have two global configuration places instead of just one?

ReimarBauer: Do you mean moving the APP.config[*] to the users scope?

matrss: Question 1 is: should it be `APP.config['FILE_DATA']` instead?

Question 2 is: why do we have both `APP.config` and `mscolab_settings`?

ReimarBauer: 1. yes :)
2. the idea was to have a different layer for users than we have in the installed system.  
In my other projects we used this approach and I copied it to MSS too. It is a python script because it is also used to add functionality by the user. For mscolab we could check if a yaml is an option. For mswms it is needed to add different dataaccess or plots. At the time we created the configuration option we wanted to have it similar to mswms.

matrss: That doesn't really explain why there need to be two global key-value stores for configuration, does it? The configuration could just as well be loaded right into `APP.config`.

Anyway, that isn't really relevant for this PR and would rather be a separate refactor. Keeping names consistent is relevant here though.

ReimarBauer: We can discuss that here, but it is then not the best place for refactoring that. An issue is better.

mslib/msidp/idp_conf.py
```diff
@@ -36,28 +35,21 @@
 from saml2.saml import NAME_FORMAT_URI
 from saml2.saml import NAMEID_FORMAT_PERSISTENT
 from saml2.saml import NAMEID_FORMAT_TRANSIENT
+from saml2.sigver import get_xmlsec_binary
+
+from mslib.mscolab.conf import mscolab_settings
```
matrss: msidp is a separate module from mscolab. With msui we deliberately don't use anything from mscolab. Should msidp likewise not depend on mscolab internals?

mslib/msidp/idp_conf.py
```diff
@@ -36,28 +35,21 @@
...
-def full_path(local_file):
-    """Return the full path by joining the BASEDIR and local_file."""
-    return os.path.join(BASEDIR, local_file)
+BASE_DIR = mscolab_settings.BASE_DIR
+DATA_DIR = mscolab_settings.DATA_DIR
+SSO_DIR = mscolab_settings.SSO_DIR
 
 
 def sso_dir_path(local_file):
     """Return the full path by joining the MSCOLAB_SSO_DIR and local_file."""
```
matrss: This comment still refers to `MSCOLAB_SSO_DIR`.

tests/_test_msui/test_mscolab.py
```diff
@@ -56,7 +56,7 @@ def setup(self, qtbot, mscolab_server):
         assert add_user_to_operation(path=self.operation_name, emailid=self.userdata[0])
         self.user = get_user(self.userdata[0])
 
-        self.main_window = msui.MSUIMainWindow(mscolab_data_dir=mscolab_settings.MSCOLAB_DATA_DIR)
+        self.main_window = msui.MSUIMainWindow(operations_data=mscolab_settings.OPERATIONS_DATA)
```
matrss: Why does msui reuse a path from mscolab?

matrss (changes requested):

tests/_test_msui/test_mscolab.py
ReimarBauer: using the same name is also not that good.

ReimarBauer: The dir there is used when we switch "Work Asynchronously" on.

docs/samples/config/msidp/msidp_settings.py.sample
```diff
@@ -0,0 +1,36 @@
...
+
+    mslib.msidp.conf.py.example
+    ~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+    config for msidp
+
+    This file is part of mss.
+
+    :copyright: Copyright 2019 Shivashis Padhi
+    :copyright: Copyright 2019-2024 by the MSS team, see AUTHORS.
```
matrss: Author and year should be updated.

mslib/msidp/conf.py
```diff
@@ -0,0 +1,48 @@
+# -*- coding: utf-8 -*-
+"""
+
+    mslib.msidp.conf
+    ~~~~~~~~~~~~~~~f
```
matrss:
```suggestion
    ~~~~~~~~~~~~~~~~
```

mslib/msidp/conf.py
```diff
@@ -0,0 +1,48 @@
...
+
+    config for msidp.
+
+    This file is part of MSS.
+
+    :copyright: Copyright 2019 Shivashis Padhi
+    :copyright: Copyright 2019-2024 by the MSS team, see AUTHORS.
```
matrss: Also needs to be updated.

tests/_test_msui/test_mscolab.py
```diff
@@ -59,7 +59,7 @@ def setup(self, qtbot, mscolab_server):
         assert add_user_to_operation(path=self.operation_name, emailid=self.userdata[0])
         self.user = get_user(self.userdata[0])
 
-        self.main_window = msui.MSUIMainWindow(mscolab_data_dir=mscolab_settings.MSCOLAB_DATA_DIR)
+        self.main_window = msui.MSUIMainWindow(local_operations_data=ROOT_DIR)
```
matrss, line 62: Is reusing ROOT_DIR a good idea, or should it better be a sub-directory (or even an entirely different temporary directory)?

matrss (changes requested):

docs/samples/config/msidp/msidp_settings.py.sample
ReimarBauer: Don't understand the comment with author and year. 

Can you give an example?

tests/_test_msui/test_mscolab.py
ReimarBauer: ROOT_DIR is for now ok. On tests that is a tmp dir.

and the local asynchronous operations have a different dir than the server below that dir.

docs/samples/config/msidp/msidp_settings.py.sample
ReimarBauer: comment removed, correct, I was blind

tests/_test_msui/test_mscolab.py
ReimarBauer: We have currently not an own name for this async data dir (local_colabdata). That is hardcoded. With the changes now all data is below the ROOT_DIR in this case ~/mss. 
The current setup of the dirs for the asynchronous data (local_colabdata) take care for the tests or developers using different accounts.
The setup on a developers machine of an mscolab server uses colabdata. There each dir is a git repo, while in local_colabdata we have only 1 file in recent version


```
~/mss$ tree
.
├── colabdata
│   ├── datasso
│   ├── filedata
│   │   ├── Admin_Test
│   │   │   └── main.ftml
│   │   ├── four
│   │   │   └── main.ftml
│   │   ├── one
│   │   │   └── main.ftml
│   │   ├── test_mscolab
│   │   │   └── main.ftml
│   │   ├── three
│   │   │   └── main.ftml
│   │   └── two
│   │       └── main.ftml
│   ├── mscolab.db
│   └── uploads
└── local_colabdata
    └── a
        └── one
            └── mscolab_operation.ftml
```

For the user local_colabdata files could be principle readed by the flightrack browser, but likly this will need a change, e.g. pre or postfix of the common name. 

When we add a feature like
https://github.com/Open-MSS/MSS/issues/1632
Then the async operations can be also added to the list of possible flight tracks. 

But if you continue the idea, one possibility for MSUI is that we end up with only 1 list instead of 2 and distinguish the items by their attributes

For now a hardcoded  local_colabdata for the MSUI user is sufficient.

matrss: So essentially, MSUI only ever writes to `$ROOT_DIR/local_colabdata`? If that is the case, then there shouldn't be any conflicts by using it, but I still think that it is passed the wrong directory then if it needs to suffix it with another directory unconditionally.

I.e. I think it should be passed `local_operations_data=(ROOT_DIR / "local_colabdata")` (here in pathlib notation, don't know what it would really need to look like right now) since that is where local operations data is stored, and don't append another directory itself.

This could be a future refactor though, as not reusing mscolab_settings was already tangential to this PR. I just want to find out if my understanding of what it does is somewhat correct.

ReimarBauer: additional the word "root dir" does come from the pyfilesystem2 idea. e.g. https://docs.pyfilesystem.org/en/latest/concepts.html?highlight=root#sandboxing

That wording needs then also refactored when we do https://github.com/Open-MSS/MSS/issues/2103

matrss (approved):

tests/_test_msui/test_mscolab.py
ReimarBauer: > So essentially, MSUI only ever writes to `$ROOT_DIR/local_colabdata`? If that is the case, then there shouldn't be any conflicts by using it, but I still think that it is passed the wrong directory then if it needs to suffix it with another directory unconditionally.
> 
> I.e. I think it should be passed `local_operations_data=(ROOT_DIR / "local_colabdata")` (here in pathlib notation, don't know what it would really need to look like right now) since that is where local operations data is stored, and don't append another directory itself.
> 
> This could be a future refactor though, as not reusing mscolab_settings was already tangential to this PR. I just want to find out if my understanding of what it does is somewhat correct.

MSUI writes also ftml, csv, etc. files when mscolab is not used. They have a different dir.  We should do a different refactoring (not here) to use the same base dir for both. This would be the first step to reuse the mscolab async files in a diffferent way. 

I've not used now the data_dir from the settings, because I want to have that refactoring in one step, It is easier to understand what only for that is changed

ReimarBauer: I am currently also not sure if we have to keep all levels. user-operation-file(with same name: mscolab_operation.ftml)

a follow up: https://github.com/Open-MSS/MSS/issues/2508

ReimarBauer: https://github.com/Open-MSS/MSS/issues/2508

PR #2485, opened 2024-08-26
Feat: Legend integration

rohit2p:
**Purpose of PR?**:

integration of legend

PR #2513, opened 2024-09-07
Feat: clickable legend and annotation on the plot

rohit2p:
**Purpose of PR?**:

This PR is a follow up on the legend feature PR #2485. in this the legend is clickable in order to store a custom label insted of only name and that custom label will be used as annotation to plot on the map beside the line of (flight track/operations)


ReimarBauer:
on toggling the active flighpath the color gets wrong

image

ReimarBauer:
renaming "by save as" a fligh track will add it twice to the legend but keeps only 2 in the selected list

image

PR #2555, opened 2024-10-06
reorder imports in autoplot_dockwidget

Preetam-Das26:
**Purpose of PR?**:
reorder imports in autoplot_dockwidget
Fixes #

**Checklist:**
- [ ] Commit has unit tests

### DIFF
--- a/mslib/msui/autoplot_dockwidget.py
+++ b/mslib/msui/autoplot_dockwidget.py
@@ -31,11 +31,11 @@ import os
 import json
 from datetime import datetime
 
+import click
 from PyQt5.QtWidgets import QWidget, QFileDialog, QTreeWidgetItem, QMessageBox
 from PyQt5 import QtCore
-import click
 
-from mslib.utils.mssautoplot import cli_tool
+from mslib.utils.mssautoplot import main as cli_tool
 from mslib.msui.qt5.ui_mss_autoplot import Ui_AutoplotDockWidget
 from mslib.msui import constants as const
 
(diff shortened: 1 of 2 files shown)


mslib/msui/autoplot_dockwidget.py
```diff
@@ -27,15 +27,17 @@
...
-import json
 import os
-import click
-from mslib.utils.mssautoplot import main as cli_tool
+import json
+from datetime import datetime
+
 from PyQt5.QtWidgets import QWidget, QFileDialog, QTreeWidgetItem, QMessageBox
 from PyQt5 import QtCore
+import click
```
ReimarBauer: move this before the first from PyQt5 import then we have all from after each other. both are also not in the standardlib

mslib/msui/autoplot_dockwidget.py
```diff
@@ -27,15 +27,17 @@
...
+
+from mslib.utils.mssautoplot import cli_tool
```
ReimarBauer: this needs to become changed it is 

from mslib.utils.mssautoplot import main as cli_tool

not only the mac os tests fail because of that

ReimarBauer (changes requested):
one import has a name problem, because the cli_tool was renamed back to main.

ReimarBauer:
maybe do this also for mslib.utils.mssautoplot, there e.g. requests needs to become separately installed is not in stdlib.

```
from datetime import datetime, timedelta
import io
import logging
import re
import os
import sys
import requests
from urllib.parse import urljoin
import json
from PyQt5.QtCore import Qt
```

ReimarBauer (approved):

PR #2556, opened 2024-10-06
Autoplot has to use http auth for mscolab and login

ReimarBauer:
**Purpose of PR?**:

Used a similiar method as in mscolab to login to protected mscolab servers
- signature of objects changed
- names changed to common names
- some docstrings added


ReimarBauer:
After this is merged, a followup is to add mscolab support for SideView and LinearView.

Preetam-Das26 (approved):

PR #2557, opened 2024-10-07
Autoplot show message when not used correctly

ReimarBauer:
**Purpose of PR?**:

Avoid a crash when a mssautoplot.json file is loaded before the corresponding operations or fligt tracks are added to MSUI. 

You can try this on a fresh start of msui, loading topview, autoplot, selecting mssautoplot.json, clicking on an entry.

The data in the tree widget gets now  reduced to those matching in the interface. The crash of requesting WMS data before get_capabilties is catched too. In a future vesion it should be possible to call get_capabilties on loading the mssautoplot.json.


ReimarBauer:
Some of the other changes changed the behaviour, it did not crash

ReimarBauer:
I assume thats the better solution
https://github.com/Open-MSS/MSS/pull/2561

PR #2570, opened 2024-10-19
fixed DeprecationWarning in requests use in auth.py

### LINKED ISSUE #1787: Fix DeprecationWarning in requests use
/opt/conda/envs/mss-develop-env/lib/python3.10/site-packages/requests/auth.py:46: DeprecationWarning: Non-string passwords will no longer be supported in Requests 3.0.0. Please convert the object you've passed in (<class 'NoneType'>) to a string or bytes object in the near future to avoid problems.
    warnings.warn(

YogeshRathee512:
Purpose of PR?:

Fixes #1787
Resolves the DeprecationWarning for non-string passwords in requests when using keyring.

No, it ensures future compatibility with requests 3.0.0 by handling password strings correctly.

Verified that credentials are correctly retrieved from keyring.
Checked that the system no longer triggers the DeprecationWarning.
Tested login flow for HTTP authentication using the stored credentials.
Ensured that errors are handled properly when keyring is unavailable or credentials do not exist.
Additional information for reviewer?:
This PR is part of a follow-up on improving error handling and avoiding deprecated practices with third-party libraries such as requests. It is a continuation of previous work on authentication refactoring.

Does this PR result in some Documentation changes?
No.

Code Changes:
Initial Code:
image

image

Updated Code:
image

image

Checklist:

 Bug fix. Fixes #1787
 New feature (Non-API breaking changes that add functionality)
 PR Title follows the convention of <type>: <subject>
 Commit has unit tests

### DIFF
--- a/mslib/utils/auth.py
+++ b/mslib/utils/auth.py
@@ -4,10 +4,10 @@
     mslib.utils.auth
     ~~~~~~~~~~~~~~~~
 
-    handles passwords from the keyring for login and http_auuth
+    handles passwords from the keyring for login and http_auth
 
 
-    To better understand of the code, look at the 'ships' example from
+    To better understand the code, look at the 'ships' example from
     chapter 14/16 of 'Rapid GUI Programming with Python and Qt: The
     Definitive Guide to PyQt Programming' (Mark Summerfield).
 
@@ -39,7 +39,7 @@ try:
 except (ImportError, ModuleNotFoundError):
     class DBusErrorResponse(Exception):
         """
-        Fallback definition on not DBus systems
+        Fallback definition on non-DBus systems
         """
         def __init__(self, message):
             super().__init__(message)
@@ -64,20 +64,18 @@ def del_password_from_keyring(service_name=NAME, username=""):
 def get_password_from_keyring(service_name=NAME, username=""):
     """
     When we request a username we use this function to fill in a form field with a password
-    In this case by none existing credentials in the keyring we have to return an empty string
+    In this case by non-existing credentials in the keyring we have to return an empty string
     """
     if username.strip() != "":
         try:
             cred = keyring.get_credential(service_name=service_name, username=username)
-            if username is not None and cred is None:
-                return ""
-            elif cred is None:
-                return None
+            if cred is None:
+                return ""  # Return empty string instead of None
             else:
                 return cred.password
         except (keyring.errors.KeyringLocked, keyring.errors.InitError, DBusErrorResponse) as ex:
             logging.warning(ex)
-            return None
+            return ""  # Return empty string if an error occurs
 
 
 def save_password_to_keyring(service_name=NAME, username="", password=""):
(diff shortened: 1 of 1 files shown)


mslib/utils/auth.py
```diff
@@ -100,14 +98,21 @@ def get_auth_from_url_and_name(server_url, http_auth, overwrite_login_cache=True
         if server_url == url:
             try:
                 password = get_password_from_keyring(service_name=url, username=auth_name)
+                if password is None:
```
ReimarBauer, line 101: We don't want to save empty passwords to the keyring

ReimarBauer (commented):
How is the change related to the requests module? Can you provide a test which shows what you want to fix?

The warning in the issue is not created in our code. How is this connected?

mslib/utils/auth.py
```diff
@@ -4,10 +4,10 @@
     mslib.utils.auth
     ~~~~~~~~~~~~~~~~
 
-    handles passwords from the keyring for login and http_auuth
+    handles passwords from the keyring for login and http_auth
```
ReimarBauer, line 7: good, our spellchecker had not found that :)

mslib/utils/auth.py
ReimarBauer, line 101: We don't allow to save empty strings as password, that is not a good idea, to remove None here.

mslib/utils/auth.py
```diff
@@ -100,14 +98,21 @@ def get_auth_from_url_and_name(server_url, http_auth, overwrite_login_cache=True
...
+
             name = auth_name
             break
+
     if name == "":
         name = None
-    auth = constants.AUTH_LOGIN_CACHE.get(server_url, (name, None))
+
+    # Always return a valid string for the password
+    auth = constants.AUTH_LOGIN_CACHE.get(server_url, (name, ""))
```
ReimarBauer, line 117: We don't accept empty passwords. And should not start using them in the AUTH_LOGIN_CACHE.

ReimarBauer:
When we have an example how we submit None over requests, then we should catch that where we did that. Because there will be no login possible with an empty or None string. It is not needed to submit such values at all. 
Can you look into this. But from the bug description and discussion I am currently unsure if this is still a problem. 
For this it is better to provide a test first.

ReimarBauer (changes requested):
see comments

YogeshRathee512:
Respected ReimarBauer , I saw your message late at night and wanted to keep you updated. As I’m still a beginner with open source, I’ll need a bit more time to review the code and understand my mistakes. I’ll be working on it tomorrow and will reach out as soon as possible with any new solutions or updates.

ReimarBauer:
Welcome @YogeshRathee512 

there are two options currently where we use the current functionality. 
 - https://github.com/Open-MSS/MSS/blob/develop/mslib/msui/mscolab.py#L64
This is not yet merged to develop. 
 - https://github.com/Open-MSS/MSS/blob/GSOC2024-rebased-autoplot/mslib/utils/mssautoplot.py#L49

see if it is necessary to intercept the sending of None here. You can look at it via debugging, or via a test.

ReimarBauer:
the proposed PR does not take care, where we could send None. The issue meanwhile was closed.

PR #2574, opened 2024-11-13
enabled mscolab for linearview and linearview

ReimarBauer:
**Purpose of PR?**:

enabled Linearview with mscolab and linearview in autoplot


ReimarBauer:
next step is cleanup

PR #2470, opened 2024-08-17
fixed the model.Message description and tests now using op_id

ReimarBauer:
**Purpose of PR?**:

Fixes #2456

**Does this PR introduce a breaking change?**
no

**If the changes in this PR are manually verified, list down the scenarios covered:**:

**Additional information for reviewer?** :
_Mention if this PR is part of any design or a continuation of previous PRs_

**Does this PR results in some Documentation changes?**
_If yes, include the list of Documentation changes_

**Checklist:**
- [X] Bug fix. Fixes #<issue number>
- [X] New feature (Non-API breaking changes that adds functionality)
- [] PR Title follows the convention of  `<type>: <subject>`
- [X] Commit has unit tests

PR #2588, opened 2024-12-10
undo part of 02999a0

### LINKED ISSUE #2586: type casting of None in models.Changes to 'None' is wrong
By fixing #1860 with https://github.com/Open-MSS/MSS/pull/2585 I have seen that the https://github.com/Open-MSS/MSS/blob/develop/mslib/mscolab/models.py#L210 version_name default is None. But the comparison to None https://github.com/Open-MSS/MSS/blob/develop/mslib/mscolab/file_manager.py#L466 gives not the expected result. It looks like adding waypoints stores it as "None".

In the PR I do use now .filter(str(Change.version_name) != "None")\

I don't know currently where the Type Cast happens or what's the reason.

### LINKED ISSUE #2686: log the file location of msui
For debugging path related issues of msui starts it would be helpful if we can see in the log 
$HOME/.config/msui/msui.log

from which file location msui was started.

ReimarBauer:
**Purpose of PR?**:

None should not become str 'None'
Fixes #2586

### DIFF
--- a/mslib/mscolab/models.py
+++ b/mslib/mscolab/models.py
@@ -213,5 +213,7 @@ class Change(db.Model):
         self.op_id = int(op_id)
         self.u_id = int(u_id)
         self.commit_hash = str(commit_hash)
-        self.version_name = str(version_name)
-        self.comment = str(comment)
+        if version_name is not None:
+            self.version_name = str(version_name)
+        if comment is not None:
+            self.comment = str(comment)


mslib/mscolab/models.py
```diff
@@ -216,5 +216,5 @@ def __init__(self, op_id, u_id, commit_hash, version_name=None, comment=None):
         self.op_id = int(op_id)
         self.u_id = int(u_id)
         self.commit_hash = str(commit_hash)
-        self.version_name = str(version_name)
-        self.comment = str(comment)
+        self.version_name = version_name
+        self.comment = comment
```
matrss: The casts were introduced with the reasoning here: https://github.com/Open-MSS/MSS/pull/2470#discussion_r1721425806

We could keep them while still handling None properly, like so:
```suggestion
        if version_name is not None:
            self.version_name = str(version_name)
        if comment is not None:
            self.comment = str(comment)
```

This would also make more sense semantically, as it would use optional arguments only when they are supplied. Otherwise having the defaults on the columns is kind of useless.

What do you think?

matrss (changes requested):

matrss (commented):
LGTM. But shouldn't this target stable, as the issue exists there too? Also, should a migration script be included that fixes wrongly created database entries?

ReimarBauer:
Absolutly. My fault. I thought I introduced that shortly.

ReimarBauer:
@matrss The migration will be adressed later. 
https://github.com/Open-MSS/MSS/issues/2590

With the old solution a migration is rather ugly and I don't want to add a new function which is in the next major pointless.

Also that it is now handled "None" here as string did not crash our usecases.

matrss (approved):

PR #2589, opened 2024-12-13
some fixes for  et

### LINKED ISSUE #542: replace xml.etree.ElementTree
**[Original report](https://bitbucket.org/wxmetvis/mss/issue/542) by Reimar Bauer (Bitbucket: [ReimarBauer](https://bitbucket.org/ReimarBauer), GitHub: [ReimarBauer](https://github.com/ReimarBauer)).**

at least on server site we should use defusedxml.ElementTree

ReimarBauer:
**Purpose of PR?**:

Fixes #542

some more to replace

### DIFF
--- a/mslib/msui/flighttrack.py
+++ b/mslib/msui/flighttrack.py
@@ -39,7 +39,7 @@ import logging
 import os
 
 import fs
-import xml.dom.minidom  # nosec
+import xml.dom.minidom
 import defusedxml.minidom
 from defusedxml import DefusedXmlException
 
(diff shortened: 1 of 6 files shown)


mslib/msui/flighttrack.py
```diff
@@ -39,8 +39,9 @@
 import os
 
 import fs
-import xml.dom.minidom
-import xml.parsers.expat
+import xml.dom.minidom  # nosec
```
matrss: I don't like that the nosec marker is on the import, instead of on the actual usage that was deemed unproblematic. This would disable warnings for new places using the module, if any were to be added in the future. Is it possible to instead mark every line using `xml.dom.minidom` as nosec instead, with an explanation why they are safe?

ReimarBauer: explanation moved. At the moment we add a sec scanner we maybe have to configure that differently.

joernu76 (approved):

PR #2604, opened 2025-01-24
a tmp dir does not need a 2nd random var

### LINKED ISSUE #2536: review / refactor tests/constants.py
This looks to me currently a bit overengeenered.  On github also each worker creates anyway a new tmp dir. 
Is there any case on local development where we need a second information about the state of the git repository?

```
try:
    import git
except ImportError:
    SHA = ""
else:
    path = os.path.dirname(os.path.realpath(__file__))
    repo = git.Repo(path, search_parent_directories=True)
    logging.debug(path)
    try:
        # this is for local development important to get a fresh named tmpdir
        SHA = repo.head.object.hexsha[0:10]
    except (ValueError, BrokenPipeError) as ex:
        logging.debug("Unknown Problem: %s", ex)
        # mounted dir in docker container and owner is not root
        SHA = ""

ROOT_FS = TempFS(identifier=f"msui{SHA}")
```

ReimarBauer:
**Purpose of PR?**:

Fixes #2536

### DIFF
--- a/docs/development.rst
+++ b/docs/development.rst
@@ -226,7 +226,7 @@ This can be used to run tests on an invisible virtual display by prepending the 
   $ xvfb-run pytest ...
 
 We have implemented demodata as data base for testing. On first call of pytest a set of demodata becomes stored
-in a /tmp/mss* folder. If you have installed gitpython a postfix of the revision head is added.
+in a /tmp/mss* folder.
 
 
 Setup MSWMS server
--- a/requirements.d/development.txt
+++ b/requirements.d/development.txt
@@ -12,7 +12,6 @@ pytest-cov
 sphinx
 sphinx_rtd_theme
 sphinxcontrib-video
-gitpython
 pynco
 conda-verify
 pytest-reverse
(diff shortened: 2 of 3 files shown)


matrss (changes requested):
This actually blocked running the test suite in a non-git checkout of the source code (e.g. extracted source archive) as it wouldn't find a commit sha then. So it is good that it gets removed.

But some changes are missing:
1. This was the only usage of the git module in tests/, so the development dependency on gitpython can be removed: <https://github.com/Open-MSS/MSS/blob/a3a98d88481dd013cef66e1b52ea6051b26e1626/requirements.d/development.txt#L15>
2. The docs are outdated now: <https://github.com/Open-MSS/MSS/blob/a3a98d88481dd013cef66e1b52ea6051b26e1626/docs/development.rst?plain=1#L228-L229>

ReimarBauer:
> 1. This was the only usage of the git module in tests/, so the development dependency on gitpython can be removed:

good catch, this should have been removed there already some time longer. The move to the meta.yaml was a copy.

matrss:
~~All test jobs are failing now... I suspect that to only be indirectly triggered by the changes though, because the dependency change causes a new environment resolution. I have seen the same test failures in https://github.com/Open-MSS/MSS/pull/2298, but thought it was something specific to that PR until now.~~ I misremembered, I think.

ReimarBauer:
> ~All test jobs are failing now... I suspect that to only be indirectly triggered by the changes though, because the dependency change causes a new environment resolution. I have seen the same test failures in #2298, but thought it was something specific to that PR until now.~ I misremembered, I think.

yeah, one timouts, 4 numpy related, https://github.com/Open-MSS/MSS/pull/2616

matrss (approved):

PR #2606, opened 2025-01-24
autoplot plots use layout definitions

### LINKED ISSUE #1898: Simplified export of png overlays compatible to the PLANET system on HALO
png overlays without axis, titles etc. could be used in systems like PLANET as map overlays that are also accessible for operators on the aeroplane. I am looking for a simple workflow/option to export png overlays from the MSS UI to png overlays. An option "export as overlay" would be the easiest way in my opinion. The overlays should have a high enough resolution to allow zooming into them within PLANET a bit, but not too high because of performance. Furthermore, they need to be in cylindrical coordinates, without borders, axis and so on.

ReimarBauer:
**Purpose of PR?**:

Fixes #1898

this uses now the layout definition in mssautoplot.json

```
"layout": {
        "topview": [
            916,
            702
        ],
        "sideview": [
            496,
            370
        ],
        "linearview": [
            913,
            557
        ],
        "tableview": [
            1236,
            424
        ],

```
currently verified with a topview example

### DIFF
--- a/docs/mssautoplot.rst
+++ b/docs/mssautoplot.rst
@@ -25,6 +25,9 @@ The CLI tool has the following parameters:
 +--------------+-------+----------------------------------------------------------------------+
 | ``--etime``  | TEXT  |   Ending time for downloading multiple plots with a fixed interval.  |
 +--------------+-------+----------------------------------------------------------------------+
+| ``--raw``    | BOOL  |   Saves the raw image with its projection in topview.                |
++--------------+-------+----------------------------------------------------------------------+
+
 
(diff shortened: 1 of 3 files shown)


ReimarBauer:
the figure has now that size but the plot doesn't scale.

ReimarBauer:
The --raw=True option creates images like

ecmwf-eur-ll015-pltemp01-08-global-cyl-20121017t120000z-1-30-0

This is not the best projection in the example

swsrkty (approved):

PR #2608, opened 2025-01-25
Add Fullname Support to User Profiles

### LINKED ISSUE #1282: user profile options

annapurna-gupta:
Fixes #1282


mslib/mscolab/file_manager.py
```diff
@@ -236,7 +236,6 @@ def modify_user(self, user, attribute=None, value=None, action=None):
                 db.session.delete(user)
                 db.session.commit()
             user_query = User.query.filter_by(id=user.id).first()
-            # on delete we return successful deleted
```
ReimarBauer: why have you removed this comment?

mslib/mscolab/migrations/versions/922e4d9c94e2_to_version_10_0_0.py
```diff
@@ -29,5 +30,6 @@ def downgrade():
     # ### commands auto generated by Alembic - please adjust! ###
     with op.batch_alter_table('users', schema=None) as batch_op:
         batch_op.drop_column('profile_image_path')
+        batch_op.drop_column('fullname')
 
-    # ### end Alembic commands ###
+     # ### end Alembic commands ###
```
ReimarBauer: please check, is there a leading blank too much?

mslib/msui/qt5/ui_add_user_dialog.py
```diff
@@ -56,18 +70,20 @@ def setupUi(self, addUserDialog):
         self.gridLayout.addLayout(self.verticalLayout, 0, 0, 1, 1)
 
         self.retranslateUi(addUserDialog)
-        self.buttonBox.accepted.connect(addUserDialog.accept)
-        self.buttonBox.rejected.connect(addUserDialog.reject)
+        self.buttonBox.accepted.connect(addUserDialog.accept) # type: ignore
```
ReimarBauer, line 73: why is there a  # type: ignore now?

usually this is not in a pyuic5 autogenerated file

ReimarBauer:
try to alter a test that now also the fullname is checked if one was added

annapurna-gupta:
> try to alter a test that now also the fullname is checked if one was added

i am sorry, i didn't understand...can you please explain

mslib/msui/qt5/ui_add_user_dialog.py
annapurna-gupta: i will check this and revert it.

mslib/mscolab/migrations/versions/922e4d9c94e2_to_version_10_0_0.py
annapurna-gupta: okkk

mslib/mscolab/file_manager.py
annapurna-gupta: I didn't do it intentionally; I'll be careful next time.

ReimarBauer:
image

Where we currenly have Change Avatar, there should be an "Edit"
The Edit should be used to change the Avatar and the Full Name.  Changing the e-mail is currently not allowed by user action.

mslib/mscolab/file_manager.py
annapurna-gupta: i am putting it back.

annapurna-gupta:
> image
> 
> Where we currenly have Change Avatar, there should be an "Edit" The Edit should be used to change the Avatar and the Full Name. Changing the e-mail is currently not allowed by user action.

Screenshot 2025-01-26 172121
Whenever I try to add a user, an error is displayed.

annapurna-gupta:
> image
> 
> Where we currenly have Change Avatar, there should be an "Edit" The Edit should be used to change the Avatar and the Full Name. Changing the e-mail is currently not allowed by user action.

Is it possible to create a new issue for this change? I would be happy to work on it.

ReimarBauer:
> > image
> > Where we currenly have Change Avatar, there should be an "Edit" The Edit should be used to change the Avatar and the Full Name. Changing the e-mail is currently not allowed by user action.
> 
> Is it possible to create a new issue for this change? I would be happy to work on it.
yes, after the half feature is added, currently someone will ask where is the fullname...

ReimarBauer:
> > try to alter a test that now also the fullname is checked if one was added
> 
> i am sorry, i didn't understand...can you please explain

We have now changed the API and the model. The CI runs should cover the changes. The tests are all in your fork too 
https://github.com/annapurna-gupta/MSS/tree/feature/add-fullname-nickname/tests

We should have at least a test which covers adding a fullname on user creation
e.g. https://github.com/annapurna-gupta/MSS/blob/feature/add-fullname-nickname/tests/_test_msui/test_mscolab.py#L178

mslib/mscolab/server.py
```diff
@@ -616,6 +617,15 @@ def set_version_name():
     return jsonify({"success": True, "message": "Successfully set version name"})
 
 
+@APP.route("/edit_user_info", methods=["POST"])
+@verify_user
+def edit_user_info():
+    user = g.user
```
ReimarBauer: This code is currently unused, when an edit feature gets implemented this maybe a start. 

Maybe save it as a patch and remove it for now.

annapurna-gupta: done

ReimarBauer:
@annapurna-gupta Eric has a good documentation about pytest https://pytest-with-eric.com/
There some more sources on the net https://realpython.com/pytest-python-testing/ 
on pyvideo you do find many https://pyvideo.org/tag/pytest/

annapurna-gupta:
> @annapurna-gupta Eric has a good documentation about pytest https://pytest-with-eric.com/
> There some more sources on the net https://realpython.com/pytest-python-testing/
> on pyvideo you do find many https://pyvideo.org/tag/pytest/


Thanks for the pytest resources! I couldn’t work on the issue as my college practicals were going on.

annapurna-gupta:
Hi @ReimarBauer , I’ve pushed PR, please review it...meanwhile, I’ll fix the tests.

tests/_test_mscolab/test_seed.py
```diff
@@ -43,7 +43,6 @@ def setup(self, mscolab_app, mscolab_managers):
         self.userdata_2 = "UV2@v2", "V2", "v2"
 
         assert add_user(self.userdata_0[0], self.userdata_0[1], self.userdata_0[2])
-        assert add_operation(self.operation_name, self.description)
```
ReimarBauer, line 46: why is this removed?

maybe look on the signature of add_user and improve where that is used in the test

ReimarBauer (changes requested):
have a look on the signature of add_user and improve it and add tests which use fullname
```
export QT_QPA_PLATFORM=offscreen
pytest -k test_url_combo
ERROR tests/_test_msui/test_mscolab.py::Test_Mscolab_connect_window::test_url_combo - TypeError: add_user() missing 1 required positional argument: 'fullname'
```

tests/_test_mscolab/test_seed.py
annapurna-gupta: it was a mistake .. i will put it back.

annapurna-gupta:
@ReimarBauer can you please review the changes.

mslib/mscolab/migrations/versions/89cbaa26ae78_to_version_10_0_0.py
```diff
@@ -0,0 +1,35 @@
+"""To version_10_0_0
+
```
ReimarBauer: this is wrong, we can have only 1 migration script to version 10. You need to add your needs to the existing one.

compared to the other your file does nothing and can become removed

mslib/mscolab/server.py
```diff
@@ -248,9 +248,27 @@ def check_login(emailid, password):
...
+
+    return {"success": True}  # If everything is valid
+
+
+def register_user(email, password, username, fullname):
+    if len(str(email.strip())) == 0 or len(str(username.strip())) == 0 or len(str(fullname.strip())) == 0:
+        return {"success": False, "message": "Your username or email or fullname cannot be empty"}
+    fullname_validation = is_valid_fullname(fullname)
+    if not fullname_validation["success"]:
+        return fullname_validation
```
ReimarBauer: we need a json string and not only False, or?

{"success":  False}

mslib/mscolab/server.py
```diff
@@ -248,9 +248,27 @@ def check_login(emailid, password):
...
+    # Check each character in fullname
+    for char in fullname:
+        if not (char.isalpha() or char in [" ", "-", "'"]):  # Allow letters, space, hyphen, and apostrophe
+            return {"success": False, "message": "Full name must contain only letters (no numbers or symbols)."}
+
+    return {"success": True}  # If everything is valid
+
+
+def register_user(email, password, username, fullname):
+    if len(str(email.strip())) == 0 or len(str(username.strip())) == 0 or len(str(fullname.strip())) == 0:
```
ReimarBauer: In the first step of implementing it is not needed to have the full name mandatory.

It can become mandatory when we use it on some other places

mslib/mscolab/server.py
```diff
@@ -370,14 +388,14 @@ def get_auth_token():
                 token = user.generate_auth_token()
                 return json.dumps({
                     'token': token,
-                    'user': {'username': user.username, 'id': user.id}})
+                    'user': {'username': user.username, 'id': user.id}, 'fullname': user.fulllname})
             else:
```
ReimarBauer, line 376: the linter should have found that typo `user.fulllname`

mslib/msui/qt5/ui_add_user_dialog.py
```diff
@@ -56,18 +70,30 @@ def setupUi(self, addUserDialog):
...
+        self.emailIDLabel.setText(_translate("addUserDialog", "Email:"))
         self.emailid.setPlaceholderText(_translate("addUserDialog", "johndoe@gmail.com"))
         self.passwordLabel.setText(_translate("addUserDialog", "Password:"))
         self.password.setPlaceholderText(_translate("addUserDialog", "Your password"))
         self.confirmPasswordLabel.setText(_translate("addUserDialog", "Confirm Password:"))
         self.rePassword.setPlaceholderText(_translate("addUserDialog", "Confirm your password"))
-        self.emailIDLabel.setText(_translate("addUserDialog", "Email:"))
-        self.usernameLabel.setText(_translate("addUserDialog", "Username:"))
+
+
```
ReimarBauer: it is not wanted to have the `__name__ == "__main__":` in the conversion from a ui file. Likly you used an option to create that.

ReimarBauer (changes requested):
see comments and also look into the tests output of the CI

mslib/mscolab/server.py
annapurna-gupta: okk

mslib/mscolab/server.py
annapurna-gupta: okk..got you

mslib/mscolab/server.py
annapurna-gupta: strange it didn't.

mslib/msui/qt5/ui_add_user_dialog.py
annapurna-gupta: okk..i will remove it

mslib/mscolab/migrations/versions/89cbaa26ae78_to_version_10_0_0.py
annapurna-gupta: Screenshot 2025-02-07 171917
After deleting that migration script, the mscolab server is no longer running

annapurna-gupta: Screenshot 2025-02-07 172350

annapurna-gupta:
> see comments and also look into the tests output of the CI

Could you please guide me on how to fix the migration script issue? Once that is resolved, I will fix the CI results.

ReimarBauer:
Look into the content of the added file. We we will have only 1 version 10.0.0 release. 
So we want only 1 migration file.

Your new file is not needed.

Move the changes to the existing mig script for version 10.0.0.

annapurna-gupta:
> Look into the content of the added file. We we will have only 1 version 10.0.0 release. So we want only 1 migration file.
> 
> Your new file is not needed.
> 
> Move the changes to the existing mig script for version 10.0.0.

okk

annapurna-gupta:
@ReimarBauer could you please look at the changes and the workflow error.

annapurna-gupta:
@ReimarBauer could you please take a look in this pr.

tests/_test_msui/test_mscolab.py
```diff
@@ -194,7 +194,7 @@ def test_add_users_without_updating_credentials_in_config_file(self, mockmessage
         assert config_loader(dataset="MSS_auth").get(self.url) == "something@something.org"
         self._connect_to_mscolab(qtbot)
         assert self.window.mscolab_server_url is not None
-        self._create_user("anand", "anand@something.org", "anand_pass")
+        self._create_user("anand", "anand@something.org", "anand_pass", fullname="Anand User")
```
ReimarBauer: the signature of _create_user has no keyword, it has only parameters,

check always that you use the correct signature

tests/_test_msui/test_mscolab.py
```diff
@@ -1007,18 +1008,20 @@ def assert_():
...
         self.connect_window.newPasswordLe.setText(str(password))
         self.connect_window.newConfirmPasswordLe.setText(str(password))
+        self.connect_window.newFullnameLe.setText(str(fullname))
         okWidget = self.connect_window.newUserBb.button(self.connect_window.newUserBb.Ok)
         QtTest.QTest.mouseClick(okWidget, QtCore.Qt.LeftButton)
 
         def assert_user_created():
             assert self.window.usernameLabel.text() == username
             assert self.window.connectBtn.isVisible() is False
+            assert self.window.fullnameLabel.text() == fullname
```
ReimarBauer: where on the screen is after a successfull login this self.window.fullnameLabel ?
Can it be used for comparison?
As I know we have it only in the profile, not anywhere else. or?

ReimarBauer (changes requested):
see comments

mslib/mscolab/server.py
```diff
@@ -248,7 +248,17 @@ def check_login(emailid, password):
     return False
 
 
-def register_user(email, password, username):
+def is_valid_fullname(fullname):
+    # Check each character in fullname
+    for char in fullname:
```
ReimarBauer: This would exclude users from Greece, China, etc. We should rework the inputs using slugify.

So this check is not a good idea.

mslib/mscolab/server.py
```diff
@@ -263,7 +273,14 @@ def register_user(email, password, username):
     user_exists = User.query.filter_by(username=str(username)).first()
     if user_exists:
         return {"success": False, "message": "This username is already registered"}
-    user = User(email, username, password)
+    if fullname.strip():
+        if not fullname[0].isupper():
```
ReimarBauer: This would exclude users from Greece, China, etc. We should rework the inputs using slugify.

split the names by " " or "-" and use for each part slugify and concatinate them by the separator again.

see https://pypi.org/project/python-slugify/

we have slugify alread on some other places

It is always success also when someone does not enter something.

tests/_test_msui/test_mscolab.py
annapurna-gupta: got you.

tests/_test_msui/test_mscolab.py
annapurna-gupta: you are right , the fullname should only be shown in the user profile after adding a user or logging in.

mslib/mscolab/server.py
annapurna-gupta: You're right; I hadn't thought about it that way

mslib/mscolab/server.py
annapurna-gupta: i have updated the test , please review.

mslib/mscolab/server.py
```diff
@@ -248,7 +249,32 @@ def check_login(emailid, password):
...
+        if "-" in part:
+            subparts = part.split("-")
+            processed_subparts = [slugify(subpart) for subpart in subparts]
+            processed_parts.append("-".join(processed_subparts))
+        else:
+            processed_parts.append(slugify(part))
+
+    final_name = " ".join(processed_parts)
+
+    if not any(processed_parts):
```
ReimarBauer: Have you a test for this?

Isn't that check a bit late?

mslib/mscolab/migrations/versions/922e4d9c94e2_to_version_10_0_0.py
```diff
@@ -1,8 +1,8 @@
-"""To version 10.0.0
+"""To version <10_0_0>
```
ReimarBauer: The version should be 10.0.0

mslib/mscolab/server.py
```diff
@@ -248,7 +249,32 @@ def check_login(emailid, password):
     return False
 
 
-def register_user(email, password, username):
+def is_valid_fullname(fullname):
```
ReimarBauer: That name is misleading to what the function does

Rename it to what it is doing

mslib/mscolab/server.py
```diff
@@ -351,6 +382,8 @@ def hello():
             'direct_login': mscolab_settings.DIRECT_LOGIN
         })
     else:
+        if mscolab_settings.__dict__.get('enable_basic_http_authentication', False):
```
ReimarBauer: What has this change to do with the issue?

ReimarBauer (changes requested):

mslib/mscolab/server.py
annapurna-gupta: i will revert it, i am sorry.

mslib/mscolab/server.py
annapurna-gupta: okk

mslib/mscolab/server.py
annapurna-gupta: yes i have written a test for this function, and i will check in the meantime.

mslib/mscolab/migrations/versions/922e4d9c94e2_to_version_10_0_0.py
annapurna-gupta: okkk

ReimarBauer:
@annapurna-gupta please update by e.g. github your develop branch and afterwards rebase this branch to develop.
Or rebase it to the origin develop.

mslib/mscolab/server.py
```diff
@@ -248,7 +249,39 @@ def check_login(emailid, password):
     return False
 
 
-def register_user(email, password, username):
+def process_fullname(fullname):
+    print(f"Original fullname: '{fullname}'")
```
ReimarBauer: a print crashes on some setups, please remove it.

If you need output use logging

mslib/mscolab/server.py
```diff
@@ -248,7 +249,39 @@ def check_login(emailid, password):
...
+    fullname = fullname.strip()
+    print(f"Stripped fullname: '{fullname}'")
+
```
ReimarBauer: logging

mslib/mscolab/server.py
```diff
@@ -248,7 +249,39 @@ def check_login(emailid, password):
...
+    if not fullname:
+        return {"success": True, "processed_name": ""}
+
+    parts = fullname.split()
+    print(f"Name parts: {parts}")
+
```
ReimarBauer: logging

mslib/mscolab/server.py
```diff
@@ -248,7 +249,39 @@ def check_login(emailid, password):
...
+            processed_parts.append("-".join(processed_subparts))
+        else:
+            processed_part = slugify(part)
+            print(f"Processed part: '{processed_part}'")
+            if not processed_part:
+                return {"success": False, "message": "Fullname contains invalid part after processing."}
+            processed_parts.append(processed_part)
+
+    final_name = " ".join(processed_parts)
+    print(f"Final processed name: '{final_name}'")
```
ReimarBauer: logging

mslib/mscolab/server.py
```diff
@@ -248,7 +249,39 @@ def check_login(emailid, password):
...
+        if "-" in part:
+            subparts = part.split("-")
+            processed_subparts = [slugify(subpart) for subpart in subparts]
+            print(f"Subparts: {subparts}, Processed subparts: {processed_subparts}")
+            if not all(processed_subparts):
+                return {"success": False, "message": "Fullname contains invalid part after processing."}
+            processed_parts.append("-".join(processed_subparts))
+        else:
+            processed_part = slugify(part)
+            print(f"Processed part: '{processed_part}'")
```
ReimarBauer: logging

ReimarBauer (changes requested):
see comments

annapurna-gupta:
> @annapurna-gupta please update by e.g. github your develop branch and afterwards rebase this branch to develop. Or rebase it to the origin develop.

I just updated it.

mslib/mscolab/server.py
annapurna-gupta: okk

mslib/mscolab/server.py
annapurna-gupta: okk, got you

mslib/mscolab/server.py
annapurna-gupta: okk

mslib/mscolab/server.py
annapurna-gupta: okk

mslib/mscolab/server.py
annapurna-gupta: okk

mslib/mscolab/migrations/versions/922e4d9c94e2_to_version_10_0_0.py
```diff
@@ -1,8 +1,8 @@
 """To version 10.0.0
 
-Revision ID: 922e4d9c94e2
-Revises: c171019fe3ee
-Create Date: 2024-07-24 15:28:42.009581
+Revision ID: 3b4e20f78d34
+Revises: 89cbaa26ae78
+Create Date: 2025-02-06 17:37:12.666326
```
matrss: Please don't change these, the revision ID should match up with the file name and the "revises:" ID must be the previous migrations ID (this right now references a non-existent migration, which can not work).

mslib/mscolab/migrations/versions/922e4d9c94e2_to_version_10_0_0.py
```diff
@@ -21,15 +21,16 @@ def upgrade():
     # ### commands auto generated by Alembic - please adjust! ###
     with op.batch_alter_table('users', schema=None) as batch_op:
         batch_op.add_column(sa.Column('profile_image_path', sa.String(length=255), nullable=True))
-    op.execute("UPDATE changes SET version_name = NULL WHERE version_name = 'None';")
-    op.execute("UPDATE changes SET comment = NULL WHERE comment = 'None';")
```
matrss: These should not be removed, they are here from a previous PR.

matrss (changes requested):

mslib/mscolab/migrations/versions/922e4d9c94e2_to_version_10_0_0.py
annapurna-gupta: okk, now ik why this was not working.

mslib/mscolab/migrations/versions/922e4d9c94e2_to_version_10_0_0.py
annapurna-gupta: okk i will put it back.

mslib/mscolab/migrations/versions/922e4d9c94e2_to_version_10_0_0.py
annapurna-gupta: It is still giving an error.

mslib/mscolab/migrations/versions/922e4d9c94e2_to_version_10_0_0.py
```diff
@@ -11,7 +11,7 @@
 
 
 # revision identifiers, used by Alembic.
-revision = '922e4d9c94e2'
+revision = '3b4e20f78d34'
```
matrss: This revision change is still incorrect.

mslib/mscolab/migrations/versions/922e4d9c94e2_to_version_10_0_0.py
```diff
@@ -21,15 +21,22 @@ def upgrade():
     # ### commands auto generated by Alembic - please adjust! ###
     with op.batch_alter_table('users', schema=None) as batch_op:
         batch_op.add_column(sa.Column('profile_image_path', sa.String(length=255), nullable=True))
+        batch_op.add_column(sa.Column('fullname',
+            sa.String(length=255), nullable=True))
```
matrss, line 25: You define the column as nullable here, but in the model it is not nullable.

mslib/mscolab/migrations/versions/922e4d9c94e2_to_version_10_0_0.py
annapurna-gupta: okkk

mslib/mscolab/migrations/versions/922e4d9c94e2_to_version_10_0_0.py
annapurna-gupta: ohhh, i will change it.

mslib/mscolab/migrations/versions/922e4d9c94e2_to_version_10_0_0.py
annapurna-gupta: Screenshot 2025-02-18 230905


Then what should i do with this error?

matrss: This error happens because you have previously applied the migration with a wrong revision identifier. Your local database now contains this identifier, but there no longer is a migration associated with it, so the database is in invalid state. The easiest would be to delete the database file and just create a new one. By default the database is in `~/colabdata/mscolab.db`.

annapurna-gupta: On my system, the colabdata folder is already empty, so there is no mscolab.db file to delete. and if there is no database file how i am getting error, do i need to create db file from scratch any suggestion how can i do it? or am i missing somthing?

matrss: Sorry, I was looking at stable instead of develop. The default location for the db file has been changed to `~/mss/colabdata/mscolab.db` at some point. The database is created and migrated automatically at server startup.

annapurna-gupta: okkk, i have just pushed a pr.. please review .

mslib/mscolab/server.py
```diff
@@ -248,7 +249,31 @@ def check_login(emailid, password):
...
+    fullname = fullname.strip()
+    if not fullname:
+        return {"success": True, "processed_name": ""}
+
+    parts = fullname.split()
+
+    processed_parts = []
+    for part in parts:
+        if "-" in part:
+            processed_part = part.lower()
```
ReimarBauer: why lowercase?

mslib/mscolab/server.py
```diff
@@ -248,7 +249,31 @@ def check_login(emailid, password):
     return False
 
 
-def register_user(email, password, username):
+def process_fullname(fullname):
+    fullname = fullname.strip()
```
ReimarBauer: I thought a bit on that. The reason for this is not to get unwanted characters stored in the database. 

But it means also we can store the origin string if it doesn´t have problems.

how is it with that example?
```
name = "John Smith-Doe"
result_with_regex = slugify(name, regex_pattern=r"[^a-zA-Z\s]")
if name.lower() == result_with_regex:
     return {"success": True, "processed_name": name}
else:
    return {"success": Falsle, "message": "Invalid chars detected"}
````

When there is a unit test we could verify if another solution gives an identical result.

File: tests/_test_msui/test_mscolab.py [hunk 30]
```diff
@@ -178,13 +178,13 @@ def test_logout(self, qtbot):
...
         modify_config_file({"MSS_auth": {self.url: "something@something.org"}})
-        self._create_user("something", "something@something.org", "something")
+        self._create_user("something", "something@something.org", "something", "Something")
         assert config_loader(dataset="MSS_auth").get(self.url) == "something@something.org"
         assert mslib.utils.auth.get_password_from_keyring("MSCOLAB",
                                                           "something@something.org") == "password from TestKeyring"
         # assert self.window.stackedWidget.currentWidget() == self.window.newuserPage
         assert self.main_window.usernameLabel.text() == 'something'
-        assert self.main_window.mscolab.connect_window is None
+        # assert self.main_window.mscolab.connect_window is None
```
ReimarBauer: why is this changed?

annapurna-gupta: i was trying to fix an error, then i forget to uncomment it, i will revert it now.

mslib/mscolab/server.py
annapurna-gupta: so that it could not break "Jean--Luc" this kind of input further and process it as whole.

mslib/mscolab/server.py
annapurna-gupta: Thank you for the explanation and example, i will update the function and the test_function as suggested.

mslib/mscolab/migrations/versions/922e4d9c94e2_to_version_10_0_0.py
```diff
@@ -21,15 +21,22 @@ def upgrade():
...
+
     op.execute("UPDATE changes SET version_name = NULL WHERE version_name = 'None';")
     op.execute("UPDATE changes SET comment = NULL WHERE comment = 'None';")
+
     # ### end Alembic commands ###
 
 
 def downgrade():
-    # ### commands auto generated by Alembic - please adjust! ###
+# ### commands auto generated by Alembic - please adjust! ###
```
ReimarBauer: This looks like an indenting failure 

Why not keeping the origin line?

mslib/mscolab/server.py
```diff
@@ -248,7 +249,35 @@ def check_login(emailid, password):
...
+def process_fullname(fullname):
+    fullname = " ".join(fullname.split())
+    if not fullname:
+        return {"success": True, "processed_name": ""}
+
+    result_with_regex = slugify(fullname, regex_pattern=r"[^a-zA-Z\s\-]")
+    if fullname.lower() != result_with_regex:
+        return {"success": False, "message": "Invalid characters detected in fullname."}
+
+    parts = fullname.split()
```
ReimarBauer: When is this reached?
Or why don't you return the name when it matches the regex rule?

Please write a unittest in pytest format for this function which has different input as examples and shows by asserts the expected result.

ReimarBauer (changes requested):
Please add a test for the new function test_process_name

mslib/mscolab/server.py
annapurna-gupta: I was thinking that if I applied the code you provided at the top of the function, everything after the else block would become unreachable, that is why did that. But now I realize that I don’t need to check the string again if it has no problems—it should simply be returned as it is. Sorry, I was confused earlier, but I understand what you wanted me to get now. I’ll work on it.

mslib/mscolab/migrations/versions/922e4d9c94e2_to_version_10_0_0.py
annapurna-gupta: sorry that is a mistake, i will revert it.

annapurna-gupta:
Hii @ReimarBauer, half of the test are still falling.

ReimarBauer:
> Hii @ReimarBauer, half of the test are still falling.

Hi

Debug this one first


https://github.com/Open-MSS/MSS/actions/runs/13435389961/job/37536411745#step:9:1819

FAILED tests/_test_mscolab/test_server.py::Test_Server::test_check_login - assert False is True

Those are failing because of timeout you can ignore for now.
Those related to your change, need changes or your Implantation does.

annapurna-gupta:
> > Hii @ReimarBauer, half of the test are still falling.
> 
> Hi
> 
> Debug this one first
> 
> https://github.com/Open-MSS/MSS/actions/runs/13435389961/job/37536411745#step:9:1819
> 
> FAILED tests/_test_mscolab/test_server.py::Test_Server::test_check_login - assert False is True

okk

mslib/mscolab/server.py
```diff
@@ -248,7 +249,16 @@ def check_login(emailid, password):
     return False
 
 
-def register_user(email, password, username):
+def process_fullname(fullname):
+    fullname = " ".join(fullname.split())
```
ReimarBauer: the reason for this is to have only 1 blank used in a name?
maybe add a comment if so.

mslib/mscolab/server.py
```diff
@@ -248,7 +249,16 @@ def check_login(emailid, password):
     return False
 
 
-def register_user(email, password, username):
+def process_fullname(fullname):
```
ReimarBauer: I've not seen a pytest for this new function. This can be done in a further step.

ReimarBauer: I found it :)

> I've not seen a pytest for this new function. This can be done in a further step.

tests/_test_mscolab/test_server_auth_required.py
```diff
@@ -60,10 +60,6 @@ def test_verify_pw(self):
         assert verify_pw("unknown", "unknown") is False
         assert verify_pw("user", "wrong") is False
 
-    def test_register_user(self):
```
ReimarBauer, line 63: why was this removed?

ReimarBauer (changes requested):
see my question about the removed test.

mslib/mscolab/server.py
annapurna-gupta: yes, the reason of this is to handle multiple spaces between fullname, i will add commant.

mslib/mscolab/server.py
annapurna-gupta: okk

tests/_test_mscolab/test_server_auth_required.py
annapurna-gupta: sorry, i was trying to fix an error that is why i removed it.. i will put it back.

annapurna-gupta:
Hi @ReimarBauer, i have made the requested changes.

ReimarBauer:
Thx I look later today.

annapurna-gupta:
> Thx I look later today.

btw what is your time zone.

ReimarBauer (approved):
Thx

ReimarBauer:
@matrss please do a final review :)

ReimarBauer:
> > Thx I look later today.
> 
> btw what is your time zone.

MEZ

ReimarBauer:
@annapurna-gupta please look at the merge conflict and solve it

annapurna-gupta:
> @annapurna-gupta please look at the merge conflict and solve it

done

mslib/mscolab/server.py
```diff
@@ -248,7 +249,16 @@ def check_login(emailid, password):
...
+    fullname = " ".join(fullname.split())    # Removes extra spaces from fullname
+    result_with_regex = slugify(fullname, regex_pattern=r"[^a-zA-Z\s\-]")
```
matrss: This seems rather restrictive, my own name would not be allowed according to these rules. Why is the name validated against these rules at all?

There are a lot of falsehoods that people believe about the structure of names: https://www.kalzumeus.com/2010/06/17/falsehoods-programmers-believe-about-names/. In the end those falsehoods boil down to "a software must allow an arbitrary UTF-8 string of arbitrary (or at least sufficient) length as a name to not inadvertently declare a real name as invalid.".

mslib/mscolab/server.py
```diff
@@ -372,14 +387,14 @@ def get_auth_token():
                 token = user.generate_auth_token()
                 return json.dumps({
                     'token': token,
-                    'user': {'username': user.username, 'id': user.id}})
+                    'user': {'username': user.username, 'id': user.id}, 'fullname': user.fullname})
```
matrss: The 'fullname' key should be part of the inner 'user' dict.

mslib/msui/msui_mainwindow.py
```diff
@@ -628,7 +628,7 @@ def add_import_plugins(self, picker_default):
             except Exception as ex:
                 logging.error("Error on import: %s: %s", type(ex), ex)
                 QtWidgets.QMessageBox.critical(
-                    self.tr(f"ERROR: Configuration\n\n{plugins}\n\nthrows {type(ex)} error:\n{ex}"))
+                    self.tr(f"ERROR: Configuration\n\n{plugins, }\n\nthrows {type(ex)} error:\n{ex}"))
```
matrss: Where are these commas coming from? This turns the interpolated value into a one-element tuple, but I don't see why.

mslib/msui/qt5/ui_add_user_dialog.py
matrss: Should this have been reverted now or was this created by the conversion process?

matrss (changes requested):

mslib/mscolab/server.py
annapurna-gupta: okkk

mslib/msui/msui_mainwindow.py
annapurna-gupta: i don't know why comma is their, i will remove it.

mslib/msui/qt5/ui_add_user_dialog.py
annapurna-gupta: i think this is created by the conversion process, because i don't remember doing it manually, should i remove this?

mslib/mscolab/server.py
annapurna-gupta: You're right, I think I shouldn't restrict the full name, I should pass it as it is.

ReimarBauer:
This is maybe a difference on the OS where it is created. At least I can reproduce it on my MBP too.

```
pyuic5 --version
Python User Interface Compiler 5.15.9 for Qt version 5.15.8
pyuic5 ui_add_user_dialog.ui > ../qt5/ui_add_user_dialog.py

reimarbauer@MBP-von-Reimar ui % grep -i type ../qt5/ui_add_user_dialog.py
        self.buttonBox.accepted.connect(addUserDialog.accept) # type: ignore
        self.buttonBox.rejected.connect(addUserDialog.reject) # type: ignore
```

mslib/msui/qt5/ui_add_user_dialog.py
ReimarBauer: it is maybe OS dependent, I can reproduce it.

```
 pyuic5 --version
Python User Interface Compiler 5.15.9 for Qt version 5.15.8

pyuic5 ui_add_user_dialog.ui > ../qt5/ui_add_user_dialog.py
reimarbauer@MBP-von-Reimar ui % grep -i type ../qt5/ui_add_user_dialog.py
        self.buttonBox.accepted.connect(addUserDialog.accept) # type: ignore
        self.buttonBox.rejected.connect(addUserDialog.reject) # type: ignore

```

annapurna-gupta:
@matrss could you please review the changes.

ReimarBauer:
Likly you want to ask @matrss  or?

mslib/mscolab/server.py
```diff
@@ -248,7 +248,16 @@ def check_login(emailid, password):
...
 
 
-def register_user(email, password, username):
+def process_fullname(fullname):
+    if fullname == "":
+        return {"success": True, "processed_name": fullname}
+    if any(char.isalpha() for char in fullname):
+        return {"success": True, "processed_name": fullname}
+    else:
+        return {"success": False, "message": "Full name must contain at least one letter"}
```
matrss: What is the rationale for the remaining restrictions?

annapurna-gupta:
> Likly you want to ask @matrss or?

yes..sorry for the confusion..

mslib/mscolab/server.py
annapurna-gupta: > What is the rationale for the remaining restrictions?


I thought a full name shouldn't contain only numbers or special characters..i will remove it.

mslib/mscolab/server.py
```diff
@@ -248,7 +248,11 @@ def check_login(emailid, password):
     return False
 
 
-def register_user(email, password, username):
+def process_fullname(fullname):
+    return fullname
```
matrss: This is an identity function now and can be removed.

annapurna-gupta: okkk

matrss (approved):
LGTM now, thank you!

A next step should be to make the fullname changeable after registration. Existing user accounts should be able to make use of this feature too.

annapurna-gupta:
> LGTM now, thank you!
> 
> A next step should be to make the fullname changeable after registration. Existing user accounts should be able to make use of this feature too.

I got you, and @ReimarBauer  has already suggested changing the avatar button to an edit button that will allow editing the full name and changing the avatar. I think it would be better to create a separate issue for this, and I will definitely work on it. Since updating the avatar button might take some time, and I need to start working on my proposal, I’ll work on this in the meantime or maybe after the proposal period.

PR #2612, opened 2025-01-26
docs:Added documentation of Keyring Feature

### LINKED ISSUE #2421: document an example for removing a password/key from the keyring
At least in the developer section we maybe should document an example for the usage of the keyring command

anj20:
**Purpose of PR?**:

Fixes #2421 

*No*

*Yes*
*Added relevant documentation updates to Test Keyring Features.*

**Checklist:**
- [x] Doc Changes. Fixes #2421 


docs/development.rst
```diff
@@ -398,7 +398,130 @@ Common resources that a test might need,
 like e.g. a running MSColab server or a QApplication instance for GUI tests,
 are collected in :mod:`tests.fixtures` in the form of pytest fixtures that can be requested as needed in tests.
 
+Testing Keyring Features
+========================
```
ReimarBauer: This will give a H1 section, on same level as the topic of the Page "Development"

The heading hierarchy is described at line 300 ... 

Please make it a H2 and change the others accordingly

ReimarBauer (changes requested):
see comment

docs/development.rst
ReimarBauer: When you render by `make html ` with recent sphinx and sphinx_rtd_theme the current result is and shows the problem

image

ReimarBauer: and when you click on that, the whole document gets a different meaning

ReimarBauer (changes requested):
the description is currently to generic, it should be for a dev looking into this document become faster than looking into the code. 

please look into the auth module  where it is used and use the default names of the configuration, e.g. 
for mscolab it is defined in MSCOLAB_auth_user_name, with "mscolab"

e.g. when auth on the mscolab server is enabled you want to ask. The URL is then the service url.

`keyring get MSCOLAB_AUTH_{URL} mscolab`


provide the examples based on what we do and need.

look on e.g. our usage of `get_password_from_keyring` e.g. https://github.com/Open-MSS/MSS/blob/develop/mslib/msui/mscolab.py#L205


all the descriptions done here can be later also retrieved by perplexity.ai or similiar.

ReimarBauer (changes requested):
just looked at the form of the new docs. 

Bildschirmfoto 2025-02-03 um 09 54 58


This needs to become a section of development and not on same level.

See for the section codes[building-the-docs-with-sphinx](https://mss.readthedocs.io/en/stable/development.html#building-the-docs-with-sphinx)

You need to start with a H2

looking on the content next.

docs/development.rst
```diff
@@ -398,6 +398,81 @@ Common resources that a test might need,
 like e.g. a running MSColab server or a QApplication instance for GUI tests,
 are collected in :mod:`tests.fixtures` in the form of pytest fixtures that can be requested as needed in tests.
 
+Testing Keyring Features
```
ReimarBauer: You can name it 
**Keyring Features**

testing has some different meaning

docs/development.rst
```diff
@@ -398,6 +398,81 @@ Common resources that a test might need,
...
+Command-Line Commands for Keyring
+-----------------------------------
+
+1. **Set a Password**
+
+   Store a password for a specific service and user:
+
+   .. code-block:: bash
+
+       keyring set SERVICE_NAME USERNAME
```
ReimarBauer, line 426: the examples maybe should destinguish the http_auth protection and a user password. 
When you use the` keyring set MSCOLAB_AUTH_{url} user@example.com` then it needs a change to `keyring set MSCOLAB_AUTH_{url} MSCOLAB_auth_user_name` the default in the configuration is mscolab.

ReimarBauer: for the WMS Servers this username can differ for each server, an example is given in [config.py ](https://github.com/Open-MSS/MSS/blob/develop/mslib/utils/config.py#L220)

ReimarBauer: mscolab has besides the common http_auth also user accounts and they are not described here.

These can become tested when you add a user with a password to your local running server.

docs/development.rst
```diff
@@ -398,6 +398,81 @@ Common resources that a test might need,
 like e.g. a running MSColab server or a QApplication instance for GUI tests,
 are collected in :mod:`tests.fixtures` in the form of pytest fixtures that can be requested as needed in tests.
 
+Testing Keyring Features
+=========================
+
+This document outlines step-by-step instructions for testing keyring features using the command line.
```
ReimarBauer: for using the keyring feature

docs/development.rst
ReimarBauer: an example here would be

keyring set http://localhost:8083 myname@mydomain

Then when you started mscolab and use the myname@mydomain for this url it fills in your saved credentials and you are logged in by login.

docs/development.rst
```diff
@@ -398,6 +398,81 @@ Common resources that a test might need,
...
+
+   .. code-block:: bash
+
+       keyring set SERVICE_NAME USERNAME
+
+   **Example:**
+
+   .. code-block:: bash
+
+       keyring set MSCOLAB_AUTH_{url} user@example.com
```
ReimarBauer: that is for an auth protection, when you setup an mscolab_auth.py and have this in your search path.

In that case it is as username not user@example.com. We use a name, in the default config it is just mscolab

using this as an example is then 
`keyring set MSCOLAB_AUTH_http://localhost:8083 mscolab`

docs/development.rst
```diff
@@ -398,6 +398,81 @@ Common resources that a test might need,
...
+   .. code-block:: bash
+
+       keyring get SERVICE_NAME USERNAME
+
+   **Example:**
+
+   .. code-block:: bash
+
+       keyring get MSCOLAB_AUTH_{url} user@example.com
+
```
ReimarBauer, line 461: this is again mixed two cases. here you would do an example 
`keyring get http://localhost:8083 myname@mydomain`

ReimarBauer: and for the case of an http_auth 

`keyring get MSCOLAB_AUTH_http://localhost:8083 mscolab
`

docs/development.rst
```diff
@@ -398,6 +398,81 @@ Common resources that a test might need,
...
+   .. code-block:: bash
+
+       keyring delete SERVICE_NAME USERNAME
+
+   **Example:**
+
+   .. code-block:: bash
+
+       keyring delete MSCOLAB_AUTH_{url} user@example.com
```
ReimarBauer: mixed again

docs/development.rst
```diff
@@ -398,6 +398,81 @@ Common resources that a test might need,
...
+
+   To confirm the deletion, attempt to retrieve the password:
+
+   .. code-block:: bash
+
+       keyring get MSCOLAB_AUTH_{url} user@example.com
+
```
ReimarBauer: mixed again

we should have in the docs something a developer can use for the application.

This is also usefull when you have a new computer and want to transfer account  data.

ReimarBauer (approved):
Thx

PR #2617, opened 2025-02-02
fix: set pinning of numpy<2.0

### LINKED ISSUE #2483: Review if we need to refactor for numpy >=2.1.0
currently we not have set a pinning for numpy but our dependencies have.

We need to check if we are compatible when they release their pinning.

divyaajainn:
**Purpose of PR?**: Pins NumPy to version below 2.0. Ensures compatibility.

Bug Fix:
Fixes #2483

### DIFF
--- a/localbuild/meta.yaml
+++ b/localbuild/meta.yaml
@@ -86,6 +86,7 @@ requirements:
     - flask-login
     - pysaml2
     - libxmlsec1  # [not win]
+    - numpy <2.0
   run_constrained:
     - menuinst >=2.0.2
 
--- a/mslib/msui/mpl_pathinteractor.py
+++ b/mslib/msui/mpl_pathinteractor.py
@@ -111,7 +111,7 @@ class WaypointsPath(mpath.Path):
            matplotlib.Path) at the given index.
         """
         self.vertices = np.insert(self.vertices, index,
-                                  np.asarray(vertex, np.float_), axis=0)
+                                  np.asarray(vertex, np.float64), axis=0)
         self.codes = np.insert(self.codes, index, code, axis=0)
 
     def index_of_closest_segment(self, x, y, eps=5):
--- a/requirements.d/development.txt
+++ b/requirements.d/development.txt
@@ -20,3 +20,4 @@ eventlet>0.30.2
 dnspython>=2.0.0, <2.3.0
 gsl==2.7.0
 boa
+numpy<2.0


ReimarBauer:
Please pin it in the meta.yaml. you maybe can catch it in the localbuild dir. 

development won't see a user.

localbuild/meta.yaml
```diff
@@ -86,6 +86,7 @@ requirements:
     - flask-login
     - pysaml2
     - libxmlsec1  # [not win]
+    - numpy <2.0
```
ReimarBauer, line 89: This will install numpy 1.x. 

The PR has to become sended to stable branch.

requirements.d/development.txt
```diff
@@ -20,3 +20,4 @@ eventlet>0.30.2
 dnspython>=2.0.0, <2.3.0
 gsl==2.7.0
 boa
+numpy<2.0
```
ReimarBauer, line 23: It is not wanted to be added for testing again.

ReimarBauer (changes requested):

ReimarBauer:
When this should go to stable the PR needs to be changed to stable and not develop.

mslib/msui/mpl_pathinteractor.py
```diff
@@ -111,7 +111,7 @@ def insert_vertex(self, index, vertex, code):
            matplotlib.Path) at the given index.
         """
         self.vertices = np.insert(self.vertices, index,
-                                  np.asarray(vertex, np.float_), axis=0)
+                                  np.asarray(vertex, np.float64), axis=0)
```
ReimarBauer, line 114: When we keep the old version, why is this change needed?

localbuild/meta.yaml
divyaajainn: do we want a specific version of numpy like numpy>=1.26, numpy<2.0?

requirements.d/development.txt
ReimarBauer: this line needs to become removed

ReimarBauer:
please do the fix in a PR to the stable branch, you likly need to base it on stable. We don't want to merge all changes of develop into stable for this fix.

There are many ways. 

One for a minor release is:

```
git checkout stable
git checkout -b "name_of_fix"
work on it, send a PR, link it to stable
```

divyaajainn:
yes i understand it now, did it by mistake. Closing this pr and making a new one linked to stable branch. Also, do we need a specific version of numpy like numpy>=1.26, numpy<2.0 which ensures latest version of numpy before 2.0?

ReimarBauer:
the other one is merged

PR #2618, opened 2025-02-03
fix: set pinning of numpy<2.0

### LINKED ISSUE #2483: Review if we need to refactor for numpy >=2.1.0

divyaajainn:
**Purpose of PR?**: Pinning of numpy < 2.0 as a temporary fix for compatibility issues in mpl_pathinteractor.py. 

Bug fix:
Fixes #2483


ReimarBauer (approved):
Thx

PR #2627, opened 2025-02-05
fix: remove user credential from keyring

### LINKED ISSUE #2624: on mscolab delete account credentials should be removed from keyring
When someone removes his account from a mscolab server also his credentials and the servers auth should become removed from the users keyring.

A test should become added as well or an existing extend.

divyaajainn:
**Purpose of PR?**: User password is removed from keyring when account is deleted from MSColab. 

Fixes #2624

### DIFF
--- a/mslib/mscolab/file_manager.py
+++ b/mslib/mscolab/file_manager.py
@@ -39,8 +39,6 @@ from sqlalchemy.exc import IntegrityError
 from mslib.utils.verify_waypoint_data import verify_waypoint_data
 from mslib.mscolab.models import db, Operation, Permission, User, Change, Message
 from mslib.mscolab.conf import mscolab_settings
-from mslib.utils.auth import del_password_from_keyring
-from mslib.utils.config import config_loader
 
 
 class FileManager:
@@ -235,8 +233,6 @@ class FileManager:
                 # Delete profile image if it exists
                 if user.profile_image_path:
                     self.delete_user_profile_image(user.profile_image_path)
-                mscolab_server_url = config_loader(dataset="default_MSCOLAB")
-                del_password_from_keyring(service_name=mscolab_server_url[0], username=user.emailid)
                 db.session.delete(user)
                 db.session.commit()
             user_query = User.query.filter_by(id=user.id).first()
--- a/mslib/mscolab/file_manager.py
+++ b/mslib/mscolab/file_manager.py
@@ -39,6 +39,8 @@ from sqlalchemy.exc import IntegrityError
 from mslib.utils.verify_waypoint_data import verify_waypoint_data
 from mslib.mscolab.models import db, Operation, Permission, User, Change, Message
 from mslib.mscolab.conf import mscolab_settings
+from mslib.utils.auth import del_password_from_keyring
+from mslib.utils.config import config_loader
 
 
 class FileManager:
@@ -233,6 +235,8 @@ class FileManager:
                 # Delete profile image if it exists
                 if user.profile_image_path:
                     self.delete_user_profile_image(user.profile_image_path)
+                mscolab_server_url = config_loader(dataset="default_MSCOLAB")
+                del_password_from_keyring(service_name=mscolab_server_url[0], username=user.emailid)
                 db.session.delete(user)
                 db.session.commit()
             user_query = User.query.filter_by(id=user.id).first()
--- a/mslib/msui/mscolab.py
+++ b/mslib/msui/mscolab.py
@@ -937,7 +937,11 @@ class MSUIMscolab(QtCore.QObject):
         if reply == QMessageBox.No:
             return
         try:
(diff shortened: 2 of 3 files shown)


mslib/mscolab/file_manager.py
```diff
@@ -39,6 +39,8 @@
 from mslib.utils.verify_waypoint_data import verify_waypoint_data
 from mslib.mscolab.models import db, Operation, Permission, User, Change, Message
 from mslib.mscolab.conf import mscolab_settings
+from mslib.utils.auth import del_password_from_keyring
+from mslib.utils.config import config_loader
```
ReimarBauer: this won't work,

the config_loader belongs to.the MSUI and this is the application which is installed on the client side.

The server MSCOLAB is on some other location and of course has no idea about the passwords the user has in his machines keyring.

You need to lookup where in the MSUI the Delete Account is used and implement it nearby.

delete_account

ReimarBauer (changes requested):

mslib/mscolab/file_manager.py
divyaajainn: Got it.

ReimarBauer (changes requested):
This is one of three from the issue.

You need to extend the existing test to verify that the keyring is cleared, see 

https://github.com/Open-MSS/MSS/blob/develop/tests/_test_msui/test_mscolab.py#L873

There can be another key/value pair in the keyring when the server has the login protected. 

https://mss.readthedocs.io/en/stable/mscolab.html#protecting-login

The auth_username is configured in the users configuration and can be retrieved from there https://github.com/Open-MSS/MSS/blob/develop/mslib/utils/config.py#L208

you need a second delete for this, e.g. keyring del f"MSCOLAB_AUTH_{url}" MSCOLAB_auth_user_name

ReimarBauer:
This looks good. we are discussing on slack adding a test fot this.

tests/_test_msui/test_mscolab.py
```diff
@@ -876,10 +876,14 @@ def test_user_delete(self, mockmessage, qtbot):
         self._create_user(qtbot, "something", "something@something.org", "something")
         u_id = self.window.mscolab.user['id']
         self.window.mscolab.open_profile_window()
+        assert self.url + "something@something.org" in mslib.utils.auth.keyring.get_keyring().passwords
```
ReimarBauer, line 880: it would be good to add a comment here. That the veryfications is done that way because of the TestKeyring in conftest.

For another newer developer this is not obvious. ;)

divyaajainn: Got it!

mslib/msui/mscolab.py
```diff
@@ -937,6 +937,11 @@ def delete_account(self, _=None):
         if reply == QMessageBox.No:
             return
         try:
+            auth_name = config_loader(dataset="MSCOLAB_auth_user_name")
+            del_password_from_keyring(self.mscolab_server_url, self.email)
```
ReimarBauer, line 941: the default name is "mscolab" and that should not become replaced by None.

mslib/msui/mscolab.py
```diff
@@ -937,6 +937,11 @@ def delete_account(self, _=None):
...
+            if auth_name:
+                del_password_from_keyring(f"MSCOLAB_AUTH_{self.mscolab_server_url}", auth_name)
```
ReimarBauer: can it be always called, does it make problems when it is not defined in the keyring? In that case use an exception handler.

mslib/msui/mscolab.py
```diff
@@ -937,6 +937,11 @@ def delete_account(self, _=None):
...
+            modify_config_file({"MSCOLAB_auth_user_name": None})
```
ReimarBauer: it is not needed to change config_file of the user.

or at least it should use the "defaults" then which is "mscolab"

mslib/msui/mscolab.py
ReimarBauer, line 941: @matrss please have a look here. For cleanup that would be good to do. Do we have to add message box too, informing the user that we want to do that?

matrss: I don't see a reason why an additional message box would be necessary. When the account is deleted the saved credentials no longer have any value, so why would someone want to retain them? IMO msui already has too many message boxes... Though the one asking for confirmation to actually delete the account makes sense.

mslib/msui/mscolab.py
divyaajainn: I don't think it should create a problem if the key is not defined in the keyring since exception handling is already implemented inside the del_password_from_keyring function. So, I'll just remove the "if auth_name" check as it makes more sense?

ReimarBauer (approved):
verified with a local setup

Thx

PR #2635, opened 2025-02-08
Fixes : MSCOLAB File Import Abort Error

### LINKED ISSUE #2629: MSCOLAB . error when aborting import file
When aborting file import in MSCOLAB, I get the following error message

```
Fatal error in MSS 9.3.0 on Linux-6.8.0-110052-tuxedo-x86_64-with-glibc2.35
Python 3.11.11 | packaged by conda-forge | (main, Dec  5 2024, 14:17:24) [GCC 13.3.0]

  File "/home/icg122/miniforge3/envs/mssenv/lib/python3.11/site-packages/mslib/msui/msui_mainwindow.py", line 683, in handle_import_local
    for name in filenames:
TypeError: 'NoneType' object is not iterable

```

annapurna-gupta:
Fixes #2629

### DIFF
--- a/mslib/msui/msui_mainwindow.py
+++ b/mslib/msui/msui_mainwindow.py
@@ -671,6 +671,8 @@ class MSUIMainWindow(QtWidgets.QMainWindow, ui.Ui_MSUIMainWindow):
             self.last_save_directory,
             f"Flight Track (*.{extension});;All files (*.*)",
             pickertype=pickertype)
+        if filenames is None:
+            raise TypeError("'NoneType' object is not iterable")
         if self.local_active:
             if filenames is not None:
                 activate = True
--- a/mslib/msui/msui_mainwindow.py
+++ b/mslib/msui/msui_mainwindow.py
@@ -671,19 +671,18 @@ class MSUIMainWindow(QtWidgets.QMainWindow, ui.Ui_MSUIMainWindow):
             self.last_save_directory,
             f"Flight Track (*.{extension});;All files (*.*)",
             pickertype=pickertype)
-        if filenames is None:
-            raise TypeError("'NoneType' object is not iterable")
-        if self.local_active:
-            if filenames is not None:
-                activate = True
-                if len(filenames) > 1:
-                    activate = False
+        if filenames is not None:
+            if self.local_active:
+                if filenames is not None:
+                    activate = True
+                    if len(filenames) > 1:
+                        activate = False
+                    for name in filenames:
+                        self.create_new_flight_track(filename=name, function=function, activate=activate)
+                    self.last_save_directory = fs.path.dirname(name)
+            else:
                 for name in filenames:
-                    self.create_new_flight_track(filename=name, function=function, activate=activate)
-                self.last_save_directory = fs.path.dirname(name)
-        else:
-            for name in filenames:
-                self.mscolab.handle_import_msc(name, extension, function, pickertype)
+                    self.mscolab.handle_import_msc(name, extension, function, pickertype)
 
     def handle_export_local(self, extension, function, pickertype):
         if self.local_active:
--- a/mslib/msui/msui_mainwindow.py
+++ b/mslib/msui/msui_mainwindow.py
@@ -671,14 +671,15 @@ class MSUIMainWindow(QtWidgets.QMainWindow, ui.Ui_MSUIMainWindow):
             self.last_save_directory,
             f"Flight Track (*.{extension});;All files (*.*)",
             pickertype=pickertype)
-        if filenames is not None:
-            if self.local_active:
-                activate = True
-                if len(filenames) > 1:
-                    activate = False
-                for name in filenames:
-                    self.create_new_flight_track(filename=name, function=function, activate=activate)
-                self.last_save_directory = fs.path.dirname(name)
+        if filenames is None:
+            return
+        if self.local_active:
+            activate = True
+            if len(filenames) > 1:
+                activate = False
+            for name in filenames:
+                self.create_new_flight_track(filename=name, function=function, activate=activate)
+            self.last_save_directory = fs.path.dirname(name)
         else:
             for name in filenames:
                 self.mscolab.handle_import_msc(name, extension, function, pickertype)
--- a/mslib/msui/msui_mainwindow.py
+++ b/mslib/msui/msui_mainwindow.py
@@ -673,16 +673,15 @@ class MSUIMainWindow(QtWidgets.QMainWindow, ui.Ui_MSUIMainWindow):
             pickertype=pickertype)
         if filenames is not None:
             if self.local_active:
-                if filenames is not None:
-                    activate = True
-                    if len(filenames) > 1:
-                        activate = False
-                    for name in filenames:
-                        self.create_new_flight_track(filename=name, function=function, activate=activate)
-                    self.last_save_directory = fs.path.dirname(name)
-            else:
+                activate = True
+                if len(filenames) > 1:
+                    activate = False
                 for name in filenames:
-                    self.mscolab.handle_import_msc(name, extension, function, pickertype)
+                    self.create_new_flight_track(filename=name, function=function, activate=activate)
+                self.last_save_directory = fs.path.dirname(name)
+        else:
+            for name in filenames:
+                self.mscolab.handle_import_msc(name, extension, function, pickertype)
 
     def handle_export_local(self, extension, function, pickertype):
         if self.local_active:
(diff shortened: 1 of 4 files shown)


annapurna-gupta:
@ReimarBauer could you please review this pr.

mslib/utils/qt.py
```diff
@@ -108,7 +108,7 @@ def get_open_filenames(parent, title, dirname, filt, pickertype=None):
         raise FatalUserError(f"Unknown file picker type '{pickertype}'.")
     logging.debug("Selected '%s'", filename)
     if filename == []:
-        filename = None
+        filename = []
```
ReimarBauer: this is a solution but not the preferred one. 

The origin failure is
```
  File "/Users/reimarbauer/PycharmProjects/2025/annapurna-gupta/MSS/mslib/msui/msui_mainwindow.py", line 706, in handle_import_local
    for name in filenames:
TypeError: 'NoneType' object is not iterable
```

solve the problem as early as possible. When you verify that it is not None before it gets into the For loop then it is solved.

A few lines below someone had done such a check for the export.

tests/_test_utils/test_qt.py
```diff
@@ -142,7 +142,7 @@ def test_get_open_filenames():
         assert type(exc_info.value.__cause__) is FatalUserError
     with mock.patch("mslib.utils.qt.get_open_filenames_qt", return_value=[]):
         filenames = mqt.get_open_filenames(None, "", "", "csv", pickertype="qt")
-        assert filenames is None
```
ReimarBauer: we have a test_import_file this maybe the right place to verify a situation where we have a None filename

ReimarBauer (changes requested):
see comments

tests/_test_utils/test_qt.py
ReimarBauer: take this as an example and write a specific new test `test_none_import_file` for the bug. This should not have a trouble as the disabled test you can use as example. 
When you get by the test the same error as the user has then you can lookup with your fix if you solved it.

start with

```
def test_none_import_file(self, qtbot)
    with mock.patch("mslib.msui.msui_mainwindow.get_open_filenames", return_value=None) as mockopen:
```

annapurna-gupta: sorry i was busy..i am working on it and thankyou so much .

tests/_test_msui/test_mscolab.py
```diff
@@ -598,7 +598,7 @@ def test_import_file(self, name, qtbot):
 
     def test_none_import_file(self, qtbot):
         with mock.patch("mslib.msui.msui_mainwindow.get_open_filenames", return_value=None):
-            with pytest.raises(TypeError, match="'NoneType' object is not iterable"):
+            with pytest.raises(TypeError, match="'NoneType' object is not iterable'"):
```
ReimarBauer: the test should look like similiar to

```
def test_none_import_file(self, qtbot):
        with mock.patch("mslib.msui.msui_mainwindow.get_open_filenames", return_value=None) as mockopen:
            self._connect_to_mscolab(qtbot)
            modify_config_file({"MSS_auth": {self.url: self.userdata[0]}})
            self._login(qtbot, emailid=self.userdata[0], password=self.userdata[2])
            self._activate_operation_at_index(0)
            wp = self.window.mscolab.waypoints_model
            assert len(wp.waypoints) == 2
            for action in self.window.menuImportFlightTrack.actions():
                if action.objectName() == "actionImportFlightTrackFTML":
                    action.trigger()
                    break
            assert mockopen.call_count == 1
            imported_wp = self.window.mscolab.waypoints_model
            assert len(imported_wp.waypoints) == 2
```

mslib/msui/msui_mainwindow.py
```diff
@@ -671,6 +671,8 @@ def handle_import_local(self, extension, function, pickertype):
             self.last_save_directory,
             f"Flight Track (*.{extension});;All files (*.*)",
             pickertype=pickertype)
+        if filenames is None:
+            raise TypeError("'NoneType' object is not iterable")
```
ReimarBauer: a raise is not needed, the user decided to cancel his request. The user knows already that it was canceled.

Try `if filenames is not None:`

tests/_test_msui/test_mscolab.py
ReimarBauer: Bildschirmfoto 2025-02-11 um 11 45 52

ReimarBauer (changes requested):
see comments

mslib/utils/qt.py
annapurna-gupta: okkk

tests/_test_msui/test_mscolab.py
annapurna-gupta: i got you..Thank you for the suggestion.

mslib/msui/msui_mainwindow.py
annapurna-gupta: okkkk

tests/_test_msui/test_mscolab.py
```diff
@@ -596,6 +596,11 @@ def test_import_file(self, name, qtbot):
             imported_wp = self.window.mscolab.waypoints_model
             assert len(imported_wp.waypoints) == name[2]
 
+    def test_none_import_file(self, qtbot):
+        with mock.patch("mslib.msui.msui_mainwindow.get_open_filenames", return_value=None):
```
ReimarBauer: The test shows the failure:

```
FAILED tests/_test_msui/test_mscolab.py::Test_Mscolab::test_none_import_file - AssertionError: Regex pattern did not match.
```

tests/_test_msui/test_mscolab.py
```diff
@@ -596,6 +596,11 @@ def test_import_file(self, name, qtbot):
...
+            with pytest.raises(TypeError, match="'NoneType' object is not iterable'"):
+                self.window.handle_import_local("txt", "load_from_txt", None)
+
```
ReimarBauer, line 614: the test should fail when a problem occures, so you have to write it to show the problem.

And when you have this you fix the problem. When you did the correct fix, then also the test succeeds.

tests/_test_msui/test_mscolab.py
annapurna-gupta: i will implement the test function you have given me.

tests/_test_msui/test_mscolab.py
ReimarBauer: There is am example how that works

https://www.youtube.com/watch?v=B1j6k2j2eJg

annapurna-gupta: i got you, thanks for the guidance.

annapurna-gupta: I have just made a PR. Please take a look and suggest any changes I need to make.

mslib/msui/msui_mainwindow.py
```diff
@@ -671,17 +671,18 @@ def handle_import_local(self, extension, function, pickertype):
...
             f"Flight Track (*.{extension});;All files (*.*)",
             pickertype=pickertype)
-        if self.local_active:
-            if filenames is not None:
-                activate = True
-                if len(filenames) > 1:
-                    activate = False
+        if filenames is not None:
+            if self.local_active:
+                if filenames is not None:
```
ReimarBauer: this line won't be reached when 2 lines before the same check is given

annapurna-gupta: i am so sorry that was a mistake

ReimarBauer:
Because this is a Bugfix we need a PR to stable. You can always lookup the issue description. When the milestone is for 9.x it is currently stable because develop will be the next major.

annapurna-gupta:
> Because this is a Bugfix we need a PR to stable. You can always lookup the issue description. When the milestone is for 9.x it is currently stable because develop will be the next major.

This PR is already targeting the stable branch

matrss:
> This PR is already targeting the stable branch

It is not: 
Screenshot from 2025-02-12 11-15-05

Your branch seems to contain your commits on top of the stable branch (a slightly out-of-date version of it, but that shouldn't matter much), but the PR itself targets the develop branch and merging it would change develop instead of stable. You should only need to change the base branch of the PR in this case.

annapurna-gupta:
> > This PR is already targeting the stable branch
> 
> It is not: Screenshot from 2025-02-12 11-15-05
> 
> Your branch seems to contain your commits on top of the stable branch (a slightly out-of-date version of it, but that shouldn't matter much), but the PR itself targets the develop branch and merging it would change develop instead of stable. You should only need to change the base branch of the PR in this case.

yes you are right, i am sorry for the confusion.

ReimarBauer (approved):
Thx

PR #2642, opened 2025-02-11
more colors added for the zenith angle 

### LINKED ISSUE #2638: add solar zenith angle to top view/Remote sensing
For optical observations like remote sensing  and also for observations of chemical compounds that change significantly 
near the sunset, it would be nice to add information about the solar zenith angle (SZA)

It could be done by adding a "zenith" option in  
top view/Remote sensing/show angle (degree)

the SZA would then be calculated using the time, latitude and longitude information similar to the azimuth angle

ReimarBauer:
**Purpose of PR?**:

Fixes #2638 

suggested by @joernu76

### DIFF
--- a/mslib/msui/remotesensing_dockwidget.py
+++ b/mslib/msui/remotesensing_dockwidget.py
@@ -259,7 +259,8 @@ class RemoteSensingControlWidget(QtWidgets.QWidget, ui.Ui_RemoteSensingDockWidge
             points[i][0][0], points[i][0][1] = bmap(points[i][0][0], points[i][0][1])
         points = np.concatenate([points[:-1], points[1:]], axis=1)
         # plot
-        solar_lines = LineCollection(points, cmap=self.solar_cmap, norm=self.solar_norm,
+        norm = BoundaryNorm(np.arange(0, 180,5), 256)
+        solar_lines = LineCollection(points, cmap="turbo", norm=norm,
                                      zorder=2, linewidths=3, animated=True)
         solar_lines.set_array(np.array(vals))
         return solar_lines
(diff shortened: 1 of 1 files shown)


ReimarBauer:
that´s an color example
Bildschirmfoto 2025-02-11 um 20 25 39

jgrooss:
I don't see the difference. 
topview_sza_new
topview_sza_old

I should be clearer in what was requested by the remote sensing (DOAS) group: Klaus writes:
"...wäre die Frage ob man auch mehr Bereiche definieren kann, also z.B. immer von erste Zahl bis zur zweiten Zahl von 80, 81, 82, 83, 84, 85, 86, 87.0, 87.5, 88.0, 88.5, 90.0, 90.5, 91.0, 91.5, 92.0, also in Summe 15 Bereiche, die man auf Regenbogenfarben abbilden müsste."

I would also propose to change the wording from "altitude" to "zenith"

ReimarBauer:
```
 should be clearer in what was requested by the remote sensing (DOAS) group: Klaus writes:
"...wäre die Frage ob man auch mehr Bereiche definieren kann, also z.B. immer von erste Zahl bis zur zweiten Zahl von 80, 81, 82, 83, 84, 85, 86, 87.0, 87.5, 88.0, 88.5, 90.0, 90.5, 91.0, 91.5, 92.0, also in Summe 15 Bereiche, die man auf Regenbogenfarben abbilden müsste."

I would also propose to change the wording from "altitude" to "zenith"
```

What is your takeoff time and the destination time of your 2nd point?
altitude ?

The range is not something I want to add in this minor change. This is then something Joern can lookup. Likly the UI for the docking widget needs more options.

For a takeoff at 10:34 to destination at 13:25 it looks for me like:
gradient

jgrooss:
The times were given in the table view. In my case it was 2023-08-06 09:00 and 12:31 UTC

ReimarBauer:
That can´t be easily done in a patch level or minor in the next hours, so we save it as as task for later

PR #2643, opened 2025-02-12
prepare v9.3.1

ReimarBauer:
**Purpose of PR?**:
preparing minor release

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,14 @@
 Changelog
 =========
 
+
+Version 9.3.1
+~~~~~~~~~~~~~
+Bug fix release:
+
+All changes:
+https://github.com/Open-MSS/MSS/milestone/110?closed=1
+
 Version 9.3.0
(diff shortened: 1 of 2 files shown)


jgrooss (approved):
I only see the version entry in CHANGES.rst and in version .py
This I can approve. 
I cannot judge the underlying individual changes in the devel branch.

matrss (approved):

ReimarBauer:
> I only see the version entry in CHANGES.rst and in version .py This I can approve. I cannot judge the underlying individual changes in the devel branch.

That`s not asked here ;)

I think that's a misunderstanding, I'm doing a patchlevel release here.

To judge whether it should be a patch level or a minor release, you can always look at the link in Changes. Then of course in the changes there. Usually bugfixes are on the patch level. When we can add an enhancement (feature) without an API change then we need to change the minor level.

And the major number to version 10 will happen when we merged develop to stable. Maybe next week.

So thx for the approve.

PR #2648, opened 2025-02-18
further conditional verifications

### LINKED ISSUE #2646: http_auth failures should become better visible
when an http_auth password is required the user currently can without entering use the UI  to create an user account (without success).
The user gets no feedback that he has no chance to succeed. At the end of the process he sees an "unexpected error" and of course can't login. 

That error message in that case should show that the http_auth had failed. This also helps to destinguish a known problem from other unexpected.

When the connection works we show green messages, but when it is not working, it makes no sense to continue.

ReimarBauer:
**Purpose of PR?**:

Fixes #2646

### DIFF
--- a/mslib/mscolab/server.py
+++ b/mslib/mscolab/server.py
@@ -236,11 +236,13 @@ def create_or_update_idp_user(email, username, token, authentication_backend):
 
 
 @APP.route('/')
+@conditional_decorator(auth.login_required, mscolab_settings.__dict__.get('enable_basic_http_authentication', False))
 def home():
     return render_template("/index.html")
 
 
 @APP.route("/status")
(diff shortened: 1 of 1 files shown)


ReimarBauer:
This shows on a missing auth entry:

Bildschirmfoto 2025-02-18 um 11 37 24

mslib/mscolab/server.py
```diff
@@ -236,11 +236,13 @@ def create_or_update_idp_user(email, username, token, authentication_backend):
 
 
 @APP.route('/')
+@conditional_decorator(auth.login_required, mscolab_settings.__dict__.get('enable_basic_http_authentication', False))
```
ReimarBauer, line 239: this was also missing, so I added it here too.

joernu76 (approved):

Issue #419, opened 2019-06-14
cleanup ToDos


ReimarBauer:
**[Original report](https://bitbucket.org/wxmetvis/mss/issue/419) by Reimar Bauer (Bitbucket: [ReimarBauer](https://bitbucket.org/ReimarBauer), GitHub: [ReimarBauer](https://github.com/ReimarBauer)).**

----------------------------------------

We have some outdated ToDOs left over.  Cleanup ToDOs

ReimarBauer:
**Original comment by Reimar Bauer (Bitbucket: [ReimarBauer](https://bitbucket.org/ReimarBauer), GitHub: [ReimarBauer](https://github.com/ReimarBauer)).**

----------------------------------------

I do after mscolab finished a pass through for the older ones

ReimarBauer:
**Original comment by Reimar Bauer (Bitbucket: [ReimarBauer](https://bitbucket.org/ReimarBauer), GitHub: [ReimarBauer](https://github.com/ReimarBauer)).**

----------------------------------------

gradually create issues from ToDo for improvement

PR #2650, opened 2025-02-20
few todos solved

ReimarBauer:
**Purpose of PR?**:

partial #419 solved

After the fs/os refactoring also some more ToDos are solved. For some we should do use one of our meets  to review them, kill them or write issues.

### DIFF
--- a/mslib/msui/mscolab.py
+++ b/mslib/msui/mscolab.py
@@ -929,7 +929,7 @@ class MSUIMscolab(QtCore.QObject):
 
     @verify_user_token
     def delete_own_account(self, _=None):
-         reply = QMessageBox.question(
+        reply = QMessageBox.question(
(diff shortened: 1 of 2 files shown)


joernu76 (approved):

mslib/msui/mscolab.py
```diff
@@ -928,9 +928,8 @@ def upload_image(self, _=None):
                                      f'Cannot identify image file. Please check the file format. Error: {e}')
 
     @verify_user_token
-    def delete_account(self, _=None):
-        # ToDo rename to delete_own_account
-        reply = QMessageBox.question(
+    def delete_own_account(self, _=None):
+         reply = QMessageBox.question(
```
matrss:
```suggestion
        reply = QMessageBox.question(
```

This space breaks linting and tests right now.

PR #2660, opened 2025-02-27
update lockfile to v10 of MSS

### LINKED ISSUE #2659: pixi.lock has wrong MSS version
Until https://github.com/Open-MSS/MSS/issues/2658 we solved new releases should be done on linux.

The workflow needs after the tag a pixi update to set up the new version into the lock.

We need to update the lockfile.

ReimarBauer:
**Purpose of PR?**:

Fixes #2659

### DIFF
--- a/pixi.lock
+++ b/pixi.lock
@@ -29,10 +29,10 @@ environments:
       - conda: https://conda.anaconda.org/conda-forge/linux-64/bzip2-1.0.8-h4bc722e_7.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-64/c-ares-1.34.4-hb9d3cd8_0.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-64/ca-certificates-2025.1.31-hbcca054_0.conda
-      - conda: https://conda.anaconda.org/conda-forge/noarch/cachetools-5.5.1-pyhd8ed1ab_0.conda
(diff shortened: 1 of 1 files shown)


matrss:
Does this actually break anything right now? I didn't see a CI test yet that failed due to a wrong open_mss version in the lock from the v10 release.

I'd rather get the automated update workflow to work than start doing it manually. Right now it fails because the configured PAT does not have permission to create a branch and PR in this repository.

I think we will have to either configure the PAT and user to be a developer of this repository, or make the workflow push to a fork: https://github.com/peter-evans/create-pull-request/blob/main/docs/concepts-guidelines.md#push-pull-request-branches-to-a-fork.

ReimarBauer:
> I'd rather get the automated update workflow to work than start doing it manually. Right now it fails because the configured PAT does not have permission to create a branch and PR in this repository.
> 
> I think we will have to either configure the PAT and user to be a developer of this repository, or make the workflow push to a fork:

The PAT comes from the open-mss-build user. That has owner role in Open-MSS. I look later in the configuration.

ReimarBauer:
> Does this actually break anything right now? I didn't see a CI test yet that failed due to a wrong open_mss version in the lock from the v10 release.

Not now, I am waiting for develop again, but I know now that I can't update currently on the MBP.

Issue #2233, opened 2024-02-23
Replace multiprocessing usage in test fixtures with subprocess


matrss:
Some advantages would be:
- no longer a necessity to use the "fork" method, which should allow the tests to run on windows as well
- no potential for an unclean server start due to the "fork" method, the server processes would start in their own new python interpreter
- due to the better separation it would become possible to e.g. start both an mscolab server with and without basic auth, enabling us to test different server configuration more effectively (right now the basic auth case can only be run separately)
- it would become possible to test with different WSGI servers, if we want that

Myrausman:
Could you assign this issue to me?

Myrausman:
are these only changes I must do, or are there any more requirements ?

```
def _running_eventlet_server(app):
    """Context manager that starts the app in an eventlet server and returns its URL."""
    scheme = "http"
    host = "127.0.0.1"
    socket = eventlet.listen((host, 0))
    port = socket.getsockname()[1]
    url = f"{scheme}://{host}:{port}"
    app.config['URL'] = url
    start_method = multiprocessing.get_start_method()
    if start_method != "fork":
        pytest.skip("requires the multiprocessing start_method 'fork', which is unavailable on this system")
    args = ["python", "-m", "eventlet.wsgi.server"]
    process = subprocess.Popen(args, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
    try:
        process.start()
        while not is_url_response_ok(urllib.parse.urljoin(url, "index")):
            time.sleep(0.5)
        yield url
    finally:
        process.kill()
        process.communicate()
        process.close()

```

matrss:
No, I think this should effectively start `python mslib/mscolab/server.py start` or its mswms counterpart using subprocess, depending on the fixture. This would also require a new approach to get the assigned port from the subprocess (or maybe a listening socket can be opened beforehand and its port can be re-used for the subprocess, not sure yet). This might also entail changes to tests (hopefully not many) and the configuration setup in the test suite might have to be adapted.

This is definitely a more complicated issue.

JanEisermann:
Hello, i would like to work on it.

ReimarBauer:
enabling basic auth for tests should be worked on after we can merge https://github.com/Open-MSS/MSS/pull/3033 
But we can lookup how windows behaves with this change and fix bugs which blocks windows.

PR #2671, opened 2025-03-06
feat(mscolab): Add config option to allow unsafe Werkzeug in development

### LINKED ISSUE #2656: tutorial modes findings for the mscolab example
That "The Werkzeug web server is not designed to run in production." is correct. We need a config option to run it only for development.

```
INFO: MSS Version: 9.3.1
INFO: Python Version: 3.11.6 | packaged by conda-forge | (main, Oct  3 2023, 10:40:35) [GCC 12.3.0]
INFO: Platform: Linux-5.15.0-125-generic-x86_64-with-glibc2.35 (('64bit', 'ELF'))
INFO: Launching MSColab Server
  File "/home/r.bauer/MAIN/MSS/.pixi/envs/tutorials/bin/mscolab", line 10, in <module>
    sys.exit(main())
             ^^^^^^
  File "/home/r.bauer/MAIN/MSS/mslib/mscolab/mscolab.py", line 410, in main
    handle_start(args)
  File "/home/r.bauer/MAIN/MSS/mslib/mscolab/mscolab.py", line 58, in handle_start
    start_server(APP, sockio, cm, fm)
  File "/home/r.bauer/MAIN/MSS/mslib/mscolab/server.py", line 1020, in start_server
    sockio.run(app, port=port, debug=mscolab_settings.DEBUG)
  File "/home/r.bauer/MAIN/MSS/.pixi/envs/tutorials/lib/python3.11/site-packages/flask_socketio/__init__.py", line 651, in run
    raise RuntimeError('The Werkzeug web server is not '
RuntimeError: The Werkzeug web server is not designed to run in production. Pass allow_unsafe_werkzeug=True to the run() method to disable this error.
Start and Seed MSColab server

```

anj20:
**Purpose of PR?**: This PR introduces a configuration option to allow running the Werkzeug development server in a non-secure mode for testing and development purposes. It introduces an environment variable (MSC_ALLOW_UNSAFE_WERKZEUG) that, when set to True, will pass allow_unsafe_werkzeug=True to the Flask-SocketIO server. This prevents the RuntimeError when running the MSColab server in a development environment.

Fixes #2656 

**Checklist:**
- [X] Bug fix. Fixes #2656 
- [ ] Commit has unit tests

### DIFF
--- a/mslib/mscolab/server.py
+++ b/mslib/mscolab/server.py
@@ -24,6 +24,7 @@
     See the License for the specific language governing permissions and
     limitations under the License.
 """
+import os
 import fs
 import sys
 import functools
@@ -1018,7 +1019,8 @@ if mscolab_settings.USE_SAML2:
 
 def start_server(app, sockio, cm, fm, port=8083):
     create_files()
-    sockio.run(app, port=port, debug=mscolab_settings.DEBUG)
+    allow_unsafe = os.getenv("MSC_ALLOW_UNSAFE_WERKZEUG", "False").lower() in ("1", "true", "yes")
+    sockio.run(app, port=port, debug=mscolab_settings.DEBUG, allow_unsafe_werkzeug=allow_unsafe)
 
 
 def main():


ReimarBauer:
The need for this is still unclear to me. Why don't we see this in all of our running tests? Don't we have tests where this is covered? 

When this is needed we need additional changes and documentation.

matrss:
> Why don't we see this in all of our running tests? Don't we have tests where this is covered?

The tests do not cover the server startup end-to-end. The servers are instead started differently here: https://github.com/Open-MSS/MSS/blob/e5bcba2b964050edf8163af3ee7c1486b97cf495/tests/fixtures.py#L189-L210

This is necessary because it is otherwise impossible to select a free port for the server without having a race condition.

Proper end-to-end testing of the server would be achieved if #2233 was fixed.

But I can't even reproduce the issue, I only get a (valid) warning for this...

matrss:
Aha, it only errors out if `allow_unsafe_werkzeug=False` _and_ stdin is closed or not a TTY: https://github.com/miguelgrinberg/Flask-SocketIO/blob/78dda9b64b275f9962d7a40de0db411b6aafac07/src/flask_socketio/__init__.py#L648-L659

So in "normal" development where you start the server in a terminal everything works fine, it just emits a warning. In the tutorials on the other hand it is started as a background process in a script, where stdin might be closed or at least not a TTY, so the error happens.

anj20:
> Aha, it only errors out if `allow_unsafe_werkzeug=False` _and_ stdin is closed or not a TTY: https://github.com/miguelgrinberg/Flask-SocketIO/blob/78dda9b64b275f9962d7a40de0db411b6aafac07/src/flask_socketio/__init__.py#L648-L659
> 
> So in "normal" development where you start the server in a terminal everything works fine, it just emits a warning. In the tutorials on the other hand it is started as a background process in a script, where stdin might be closed or at least not a TTY, so the error happens.

i think then the issue is solved whenever `allow_unsafe_werkzeug=True` then it won't give any error

ReimarBauer:
> i think then the issue is solved whenever `allow_unsafe_werkzeug=True` then it won't give any error

That is extremly dangerous because we would remove the important warning for users. A common user should never use werkzeug. The default should always be False. The tutorials maybe can also use waitress.

werkzeug is only a development server.

ReimarBauer:
Some issues should be better discussed in the issue itself. In this example, if we bypass the security warning, we would give every user a risk that is as incomprehensible to them as it was for me when I did the tutorials. They wouldn't even be warned.

In the tutorials, it needs to be handled differently; we should always set a good example.

PR #2672, opened 2025-03-07
docs:change to upgrade

### LINKED ISSUE #2663: version upgrade should be described in the docs
currently we describe update but we need upgrade

anj20:
**Purpose of PR?**: docs update

Fixes #2663 

**Checklist:**
- [X] Bug fix. Fixes #2663 
- [ ] Commit has unit tests

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -41,12 +41,12 @@ Usage
     mscolab -h
     mssautoplot -h
 
-Updating
+Upgrading
 ........
 
 ::
 
-    pixi global update mss
+    pixi global upgrade mss
 
 
 Project installation
@@ -82,14 +82,14 @@ On the very first start of **msui** it takes a bit longer because it setups font
     (MSS) mscolab -h
     (MSS) mssautoplot -h
 
-Updating
+Upgrading
 ........
 
 ::
 
     cd MSS
     pixi shell
-    (MSS) pixi update mss
+    (MSS) pixi upgrade mss
 
 
 Server based installation example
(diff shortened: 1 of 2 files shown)


anj20:
@ReimarBauer created a new pr plz review it.

docs/installation.rst
```diff
@@ -41,12 +41,12 @@ Usage
...
     mssautoplot -h
 
-Updating
+Upgrading
 ........
 
 ::
 
-    pixi global update mss
+    pixi global upgrade mss
```
ReimarBauer, line 49: please lookup the documtentation, what would be the equivalent for doing a version upgrade

```
pixi global upgrade mss
Error:   × `pixi global upgrade` has been removed, and will be re-added in future releases
  ╰─▶ You can call `pixi global update` for most use cases
```

ReimarBauer (changes requested):
see comments

docs/installation.rst
anj20: I think pixi global upgrade has been removed, and update is used more frequently.

anj20: Screenshot 2025-03-07 at 7 12 21 PM

there is no alternative global command to update now i guess

ReimarBauer: We should use the descriptive help text for the commands to explain the meanings. Add a line what the command does below the headline.

ReimarBauer:
@anj20 could you finish this in the next 2-3 hours?

ReimarBauer:
meanwhile I had to prepare a release and because users have to get that release I added it.

PR #2690, opened 2025-03-14
don't block pixi commands on none linux platforms because of tutorials

### LINKED ISSUE #2658: pixi update / pixi lock on OSX-64, ARM-OSX-64, WIN-64 fails
I can currently not update on the MBP the pixi.lock. 

```
(base) reimarbauer@pool-18-54 MSS % pixi update
Error:
  × Unable to solve pypi dependencies for the tutorials environment because no
  │ compatible python interpreter can be installed for the current platform
  help: Try converting your [pypi-dependencies] to conda [dependencies]
```

ReimarBauer:
**Purpose of PR?**:

Fixes #2658

when we only remove osx-arm64 from tutorials these machiches use automatically rosetta and available packages for osx.

This does not mean that the tutorials now can become created on osx-arm64 or windows but all other pixi commands like pixi lock, pixi update did not fail because of the tutorial definition.
Without that change the tutorials problems on osx or windows can't become debugged. 
When we want to limit the tutorials to linux we should in a further step add a check of the platform there.

### DIFF
--- a/pixi.lock
+++ b/pixi.lock
@@ -3822,6 +3822,631 @@ environments:
       - conda: https://conda.anaconda.org/conda-forge/linux-64/zstandard-0.23.0-py311hbc35293_1.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-64/zstd-1.5.6-ha6fb4c9_0.conda
       - pypi: .
+      osx-64:
+      - conda: https://conda.anaconda.org/conda-forge/noarch/alembic-1.15.1-pyhd8ed1ab_0.conda
+      - conda: https://conda.anaconda.org/conda-forge/osx-64/aom-3.9.1-hf036a51_0.conda
+      - conda: https://conda.anaconda.org/conda-forge/noarch/appdirs-1.4.4-pyhd8ed1ab_1.conda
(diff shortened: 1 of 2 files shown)


matrss:
Are we planning to support the tutorials on Windows and macOS? If yes this change makes sense, but is there an issue for it already?

Irrespective of that, the underlying issue seems to be a pixi bug and deserves a report upstream: it should be possible to run `pixi update` on macOS even if the specified platform is Linux only. Doing an update on Linux for a Windows only project works fine for example, and it should for any combination of platforms.

ReimarBauer:
> Irrespective of that, the underlying issue seems to be a pixi bug and deserves a report upstream: it should be possible to run `pixi update` on macOS even if the specified platform is Linux only. Doing an update on Linux for a Windows only project works fine for example, and it should for any combination of platforms.

ah ok, I was not aware on that.

ReimarBauer:
> Are we planning to support the tutorials on Windows and macOS? If yes this change makes sense, but is there an issue for it already?

initially it was partly, I focused for the refactoring on linux first. It was just difficult without having these platforms to understand the differences in the window managers, etc.

matrss (approved):

ReimarBauer:
workaround until this is solved
https://github.com/prefix-dev/pixi/issues/3382

PR #2698, opened 2025-03-15
fix: link to new feature in about section

### LINKED ISSUE #2685: Link to new features in about incomplete
We have added a link to the about section. The version string is to short used there.

Image

Currently it goes to https://github.com/Open-MSS/MSS/issues?q=is%3Aclosed+milestone%3A10.0.

We should also try to add a test for that feature.

divyaajainn:
**Purpose of PR?**: This PR fixes the milestone link in the About section. Previously, the link was incomplete because the milestone version was incorrectly formatted (for example: "9.2." instead of "9.2.0"). This fix ensures that the correct milestone version is used in the link.

Bug fix:
Fixes #2685

### DIFF
--- a/tests/_test_msui/test_msui.py
+++ b/tests/_test_msui/test_msui.py
@@ -87,7 +87,7 @@ class Test_MSS_AboutDialog:
     def test_milestone_url(self):
         with urlopen(self.window.milestone_url) as f:
             text = f.read()
-        pattern = f'value="is:closed milestone:{__version__[:-1]}"'
+        pattern = f'value="is:closed milestone:{__version__}"'
         assert pattern in text.decode('utf-8')
 
 
--- a/tests/_test_msui/test_msui.py
+++ b/tests/_test_msui/test_msui.py
@@ -41,6 +41,7 @@ from tests.utils import ExceptionMock
 from mslib.utils.config import read_config_file
 import re
 
+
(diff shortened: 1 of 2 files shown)


ReimarBauer (changes requested):
please fix also the broken test

tests/_test_msui/test_msui.py
```diff
@@ -87,7 +87,7 @@ def setup(self, qtbot):
     def test_milestone_url(self):
         with urlopen(self.window.milestone_url) as f:
             text = f.read()
-        pattern = f'value="is:closed milestone:{__version__[:-1]}"'
+        pattern = f'value="is:closed milestone:{__version__}"'
         assert pattern in text.decode('utf-8')
```
ReimarBauer: the test maybe become later improved. It had not failed with having the bug. But it should.

It would have been better to write a test first which also fails for the bug.

ReimarBauer: the test should. not just verify what is on the page

Bildschirmfoto 2025-03-16 um 20 24 00

so the tests needs to become improved.

@divyaajainn is that enough information?

divyaajainn: yes thank you for the info, i got it

ReimarBauer (approved):
Thx

PR #2706, opened 2025-03-16
storing mscolab_server_url in users settings

### LINKED ISSUE #2705: mscolab_server_url is not updated in the msui_settings.json
once succeesful logged into a mscolab server we have also to store the recent url in the json config msui_settings.json on the var mscolab_server_url

This var is mandatory to become set for the autoplot docking widget

ReimarBauer:
**Purpose of PR?**:

Fixes #2705 

While working on this, I realized that we need to update parts of the in-memory configuration based on changes made to the file for the next startup.

The mscolab_server_url is required by the AutoPlot docking widget. However, self.config_for_gui may be initialized before the MSColab server has updated its data, leading to discrepancies between what's stored and what's used.

Moving the initialization of self.config_for_gui into create_view increases the likelihood that the MSColab URL will have been updated. 

the modify_config_file changes needs to be reapplied

So we update always the by modify_config_file changed values from the stored configuration by the config_loader.

### DIFF
--- a/mslib/msui/mscolab.py
+++ b/mslib/msui/mscolab.py
@@ -297,7 +297,9 @@ class MSColab_ConnectDialog(QDialog, ui_conn.Ui_MSColabConnectDialog):
                         QMessageBox.Yes | QMessageBox.No, QMessageBox.No)
                     if ret == QMessageBox.Yes:
                         url_list = [self.mscolab_server_url] + url_list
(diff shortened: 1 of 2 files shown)


ReimarBauer:
I am not sure if we can do this following up in a minor or patch level release
https://github.com/Open-MSS/MSS/issues/2707

joernu76 (dismissed):

joernu76 (approved):

PR #2708, opened 2025-03-16
fix: pinning of fastkml<1 

### LINKED ISSUE #2687: pin fastkml < 1
FastKML has a changed API

for example "With version 1.0, .from_string() is a class method that returns a new object."

We have to pin fastkml <1  in 10.0.2 or when possible without an API change of our functions we can change their calls in a 10.1.0.
I am currently unsure about this. pinning is the faster solution. And the refactoring can go into a different issue.

divyaajainn:
**Purpose of PR?**: pinning of fastkml <1 in pixi.toml and removing pyegiof.

Bug fix:
Fixes #2687


divyaajainn:
pygeiof is not used by us in the codebas so I removed its pinning from pixi.toml. Do we need to remove the packages from pixi.lock as well? e.g. https://github.com/Open-MSS/MSS/blob/develop/pixi.lock#L238

ReimarBauer:
for the lockfile we have a bot.

pixi.toml
```diff
@@ -15,7 +15,7 @@ chameleon = "*"
 defusedxml = "*"
 email_validator = "*"
 execnet = "*"
-fastkml = ">=0.11"
+fastkml = "<1"
```
ReimarBauer: I think I would prefer a version string <1.0.0

matrss:
> pygeiof is not used by us in the codebas so I removed its pinning from pixi.toml. Do we need to remove the packages from pixi.lock as well? e.g. https://github.com/Open-MSS/MSS/blob/develop/pixi.lock#L238

Ideally you should use `pixi remove`, then it will be removed from the lock file automatically if necessary.

In this case it shouldn't make a difference though: fastkml pulls in pygeoif as a dependency, so it is required anyway.

pixi.toml
matrss: There is no difference between <1 and <1.0.0.

matrss (approved):

PR #2712, opened 2025-03-19
Fixing Issue (data_dir configuration have to start with ~ #2702)

### LINKED ISSUE #2702: data_dir configuration have to start with ~
We have to document that the configuration of data_dir have to start with a **~**
or another FS url. 

https://mss.readthedocs.io/en/stable/usage.html

And that the new value is used on restart of the whole application.

Anindhiths:
data_dir configuration have to start with ~ or a filesystem URL #2702

**Purpose of PR?**:
Fixes #2702

No

- Setting data_dir with valid prefix (~ or filesystem URL) works correctly
- Setting data_dir with invalid prefix shows a warning and falls back to default
- Verifying that changes to data_dir take effect after application restart

This PR adds validation to ensure the data_dir configuration starts with a tilde (~) or a filesystem URL as documented in the MSS docs. When an invalid format is detected, a warning is logged and the default value is used.

No, the requirement is already documented in https://mss.readthedocs.io/en/stable/usage.html

**Checklist:**
- [x] Bug fix. Fixes #2702
- [x] PR Title follows the convention of `type: description`
- [x] Commit has unit tests

### DIFF
--- a/mslib/utils/config.py
+++ b/mslib/utils/config.py
@@ -587,6 +587,28 @@ def load_settings_qsettings(tag, default_settings=None):
     return default_settings
 
 
+def validate_data_dir(path):
+    """
+    Validates that the data_dir path starts with a tilde (~) or a filesystem URL.
+    
+    Args:
+        path: The data_dir path to validate
+        
+    Returns:
+        bool: True if valid, False otherwise
+    """
+    # Check if path starts with ~ (tilde)
+    if path.startswith('~'):
+        return True
+    
+    # Check if path is a valid filesystem URL (e.g., file://, s3://, etc.)
+    # Using fs.path.isurl method to check for valid FS URLs
+    if fs.path.isurl(path):
+        return True
+        
+    return False
+
+
 def merge_dict(existing_dict, new_dict):
     """
     Merge two dictionaries by comparing all the options from
@@ -641,6 +663,11 @@ def merge_dict(existing_dict, new_dict):
             data, match = compare_data(existing_dict[key], new_dict[key])
             if match:
                 existing_dict[key] = data
+                
+                # Add special validation for data_dir
+                if key == "data_dir" and not validate_data_dir(data):
+                    logging.warning(f"Invalid data_dir format: {data}. Must start with '~' or a filesystem URL.")
+                    existing_dict[key] = existing_dict.get("data_dir", "~/mssdata")
 
     # add filepicker default to import and export plugins if missing
     for plugin_type in ["import_plugins", "export_plugins"]:


ReimarBauer (changes requested):
That is not correct. I wanted to have a documentation of the behaviour because any FS url will work. But it is not obviuos that someone can not use an absolute path in the current definition. 

when it goes to develop, there may be we can think on an input validator of the editor, so that it can't store a value.

ReimarBauer:
> That is not correct. I wanted to have a documentation of the behaviour because any FS url will work. But it is not obviuos that someone can not use an absolute path in the current definition.
> 
> When you want to validate it you must validate it to be a valid fs url, see pyfilesystem2 documentation.

I should have readed all, ignore.

ReimarBauer:
additional we need a documentation entry, and look on the tests.

mslib/utils/config.py
```diff
@@ -641,6 +663,11 @@ def merge_dict(existing_dict, new_dict):
             data, match = compare_data(existing_dict[key], new_dict[key])
             if match:
                 existing_dict[key] = data
+                
+                # Add special validation for data_dir
+                if key == "data_dir" and not validate_data_dir(data):
+                    logging.warning(f"Invalid data_dir format: {data}. Must start with '~' or a filesystem URL.")
```
ReimarBauer, line 669: when we don't have an fsurl it will crash, so that should be an error.

mslib/utils/config.py
```diff
@@ -641,6 +663,11 @@ def merge_dict(existing_dict, new_dict):
...
+                    existing_dict[key] = existing_dict.get("data_dir", "~/mssdata")
```
ReimarBauer, line 670: this will overwrite a user definition. I think we should not do that, but give the hint what is wrong is good.

This is the editors path, someone can just use vim too. 

https://github.com/user-attachments/assets/d4a24421-b814-4296-861b-b360aa347693

I also dislike the crash, so yeah your solution is a good idea. But I am currently unsure if it is good to change an input.

ReimarBauer:
Let's see how @joernu76 looks on that point.

mslib/utils/config.py
ReimarBauer: The problem with this approach is, someone could have already a different fsurl as default and then did a change which will remove his dir and reset to our default. 

`existing_dict[key] = existing_dict.get("data_dir", "~/mssdata")`

ReimarBauer:
I should have looked what it stores. It stores the wrong dir

https://github.com/user-attachments/assets/f2ed71c0-0bc6-4e67-a8de-3d3e0eaaec28

the autoplot dockingwidget reads the variable from the file.

ReimarBauer:
The discussion showed that the issue description and the behavior with data_dir isn't correct in relation to PyFilesystem2. 

We need a different bug fix to the stable branch.

Anindhiths:
Alright Got it

PR #2722, opened 2025-03-20
added feature to copy track/operation into other new track/operation …

### LINKED ISSUE #2699: feature request: "simpler" way of moving flight tracks between local flighttrack/operations
The "work asynchronously" option is not being adopted. People want a "simple" way to copy a flighttrack from the operations pane to the "local flight track pane" and back.

In the operation->local direction a simple copy with the same name (and potentially a windows like (1)/(2) tag at the end) would do. In the local->operation direction one/two menu entries with "copy into new operation" "copy into existing operation" would do. The latter would require a dialogue for selecting an operation.

A bigger redesign could allow the selection of track on both sides with arrow buttons or similar, but I prefer the solution described above.

This would allow the removal of the "work asynchronously" button, which allows a very similar feature, but is not being adopted/understood by users on site.

joernu76:
…without using files

**Purpose of PR?**:

Fixes #2699

no

manually checked all combinations

**Checklist:**
- [ ] Commit has unit tests


joernu76:
I see no "easy" way to import waypoints from a non-active operation? Or am I mistaken?

joernu76:
@ReimarBauer

ReimarBauer:
See the multiple flightpath dockingwidget.
You can fetch without activation.

ReimarBauer:
See hhe multiple flightpath dockingwidget.
You can fetch without activation.

ReimarBauer:
And the autoplot cli can fetch too.

ReimarBauer:
When I do here Copy from selected, what should happen?
 
Bildschirmfoto 2025-03-20 um 21 31 37

I get "ERROR: cannot copy into oneself"

In that case maybe the entry should be disabled

ReimarBauer:
It is meant, one is activated and the other is selected to "send" the data in this case to new revision
Bildschirmfoto 2025-03-20 um 21 39 23

In the revision that is not "named" but we should do that. 

This enables to destinguish regular edits.

mslib/msui/mscolab.py
```diff
@@ -1030,8 +1047,9 @@ def add_operation(self):
         data = {"path": path,
                 "description": description,
                 "category": category}
-        if self.add_proj_dialog.f_content is not None:
-            data["content"] = self.add_proj_dialog.f_content
+        if f_content is not None:
+            data["content"] = f_content
+        print(data)
```
ReimarBauer: this could become a logging or removed?

ReimarBauer:
maybe longer "Copy from selected to activated"

ReimarBauer:
when an flightpath get sended to an mscolab OP we maybe should create a new named revision. 
When multiple people send, those can be listed filtered because of the name.

We should maybe also define a different service message for the chat.

It would be more important to have a different service message in the chat. Because that is the log what happens in the operation. Names can be removed the log is persistent.

mslib/msui/mscolab.py
```diff
@@ -985,29 +990,41 @@ def browse():
...
+            self.add_proj_dialog.f_content = xml
 
         # sets types from defined import menu
         import_menu = self.ui.menuImportFlightTrack
         for im_action in import_menu.actions():
             self.add_proj_dialog.cb_ImportType.addItem(im_action.text())
         self.proj_diag.show()
 
+    def add_operation_from_new_dialogue(self):
+        logging.debug("add_operation_from_dialogue")
```
ReimarBauer: the name is difficult later to understand or to map to the origin. maybe use same writing (add_operation_from_new_dialog)

self.add_proj_dialog.buttonBox.accepted.connect(self.add_operation_from_new_dialogue)

mslib/msui/mscolab.py
```diff
@@ -952,8 +954,11 @@ def delete_own_account(self, _=None):
             if response.status_code == 200 and response.json()["success"] is True:
                 self.logout()
 
-    @verify_user_token
     def add_operation_handler(self, _=None):
+        self.add_operation_dialogue()
+
+    @verify_user_token
+    def add_operation_dialogue(self, name=None, description=None, xml=None):
```
ReimarBauer: we wrote always dialog

that can become later a wording refactoring task

ReimarBauer (changes requested):
a few things to change

and I don't see a problem to have it in a 10.1.0 when you send it to stable. In that case I would skip 10.0.2 and go directly to 10.1.0

mslib/msui/msui_mainwindow.py
```diff
@@ -735,6 +738,85 @@ def handle_export_local(self, extension, function, pickertype):
...
+            name = self.mscolab.active_operation_name
+            template_copy = copy.deepcopy(self.mscolab.waypoints_model.all_waypoint_data())
+
+        # Create a new flight track from the waypoints' template.
+        self.new_flight_track_counter += 1
+        waypoints_model = ft.WaypointsTableModel(
+            name=name + f" (copy {self.new_flight_track_counter:d})")
+        # Make a copy of the template. Otherwise, all new flight tracks would
+        # use the same data structure in memory.
+        waypoints_model.insertRows(0, rows=len(template_copy), waypoints=template_copy)
```
ReimarBauer, line 747: here we could add a flag which writes then in the chat a message, e.g. "copied data". The chat system messages can be seen as a log what happens to the operation 

https://github.com/Open-MSS/MSS/blob/develop/mslib/msui/flighttrack.py#L439

mslib/msui/mscolab.py
joernu76: should be removed.

joernu76:
> maybe longer "Copy from selected to activated"

I am not happy with the terminology either. for practical reasons, one end-point need be the "active" flight track, so we must copy from selected to activated. I adopt your proposal. One could also make this an "import" feature...?

EDIT: I moved it into "Import From". Tell me what you think about it.

joernu76:
> a few things to change
> 
> and I don't see a problem to have it in a 10.1.0 when you send it to stable. In that case I would skip 10.0.2 and go directly to 10.1.0

I do not remember our reasoning here. Shouldn't new features go to develop and be made into a release from there?
Ie. develop -> x+1.0.0 or x.y+1.0 (depending on API breakage) and stable -> x.y.z+1?

joernu76:
> When I do here Copy from selected, what should happen?
> Bildschirmfoto 2025-03-20 um 21 31 37
> 
> I get "ERROR: cannot copy into oneself"
> 
> In that case maybe the entry should be disabled

I decided an the "easy" error message to get it done instead of an easy-to-get-wrong activating/deactivating. I will look into how difficult this really is. We currently have no triggers at all on stuff being simply "selected". An here it is not only being selected, but what is selected...

mslib/msui/mscolab.py
joernu76: done.

joernu76:
> It is meant, one is activated and the other is selected to "send" the data in this case to new revision Bildschirmfoto 2025-03-20 um 21 39 23
> 
> In the revision that is not "named" but we should do that.
> 
> This enables to destinguish regular edits.

I do not follow, what you request here? The current behaviour is similar to an "import flightrack" for operations and thus consistent. Adding a revision/chat message might be a good idea, but should be consistently applied.

joernu76:
I fixed the simple things. (dialogue, moved copy-from-selected to "import" menu.
1) wrt to chat-message/named revision, I propose to move this into an "easy for beginners issue".
2) wrt to disabling the entry when the active-flighttrack/operation is selected, I propose to keep it as it is. Doing this difficult to get right and adds a lot of unnecessary code. I think people will manage. Otherwise -> new issue

I thought develop is the proper branch for a non-bug-fix change, but this can be merged into stable as well.

ReimarBauer:
> > a few things to change
> > and I don't see a problem to have it in a 10.1.0 when you send it to stable. In that case I would skip 10.0.2 and go directly to 10.1.0
> 
> I do not remember our reasoning here. Shouldn't new features go to develop and be made into a release from there? Ie. develop -> x+1.0.0 or x.y+1.0 (depending on API breakage) and stable -> x.y.z+1?

when features don't break existing code we can have a minor release. When the API changes it must be a major.

ReimarBauer (changes requested):
besides that chat message everything fine.

I get only:
Bildschirmfoto 2025-03-24 um 21 06 07

please have a look on that

when that needs more changes we can do it later.

joernu76:
Same here. So merge?

ReimarBauer:
> Same here. So merge?

yes, we can. I am just waiting until at least the ubuntu test succeeds.

joernu76:
Wit the Mac-Tests failing I didn't notice the failing Ubuntu test. Might work now.

ReimarBauer (approved):
Thx

PR #2727, opened 2025-03-24
requests calls always need a timeout definition

### LINKED ISSUE #2726: problems by requests without timeout
Using request methods without setting a timeout can cause calls to hang indefinitely.

ReimarBauer:
**Purpose of PR?**:
... or they can indefinite wait

Also set per module a global MSCOLAB_TIMEOUT to reduce calls of config_loader

Fixes #2726


mslib/msui/mscolab_chat.py
```diff
@@ -532,7 +534,7 @@ def setup_image_message_box(self):
                               self.attachment_path.replace('\\', '/').split('colabdata')[1])
         else:
             img_url = urljoin(self.chat_window.mscolab_server_url, self.attachment_path)
-        data = requests.get(img_url, timeout=tuple(config_loader(dataset="MSCOLAB_timeout"))).content
+        data = requests.get(img_url, timeout=MSCOLAB_TIMEOUT).content
```
matrss: Unrelated issue: this accesses the responses content without first checking for success.

mslib/utils/mssautoplot.py
```diff
@@ -109,7 +112,7 @@ def load_from_operation(op_name, msc_url, msc_auth_password, username, password)
     session.auth = msc_auth
     session.headers.update({'x-test': 'true'})
     # ToDp fix config_loader it gets a list of two times the entry
-    response = session.get(urljoin(msc_url, 'status'), timeout=tuple(config_loader(dataset="MSCOLAB_timeout")[0]))
+    response = session.get(urljoin(msc_url, 'status'), timeout=MSCOLAB_TIMEOUT[0])
```
matrss: Why does it only take the first element here, unlike in the other places?

mslib/msui/mscolab.py
```diff
@@ -75,6 +75,9 @@
 from mslib.utils.config import config_loader, modify_config_file
 
 
+MSCOLAB_TIMEOUT = tuple(config_loader(dataset="MSCOLAB_timeout"))
```
matrss: Does this actually make a measurable difference?

mslib/msui/mscolab_chat.py
ReimarBauer: yes, first step I wanted to solve only timeout configurations. There are likly more like that.

mslib/utils/mssautoplot.py
ReimarBauer: I have no idea, I had that once changed to the common, but decided to keep the value until that question is looked up.

mslib/msui/mscolab.py
ReimarBauer: less characters on other places and the option to try different values for a module.

matrss: The downside is that this could lead to using an outdated value when the configuration is changed while the application is running. I _think_ this value wouldn't be updated if `MSCOLAB_timeout` is changed in the configuration editor, even with the application restart that it does.

ReimarBauer: the editor has another bug -> new issue coming. 
Bildschirmfoto 2025-03-24 um 20 05 19

and there is another problem by that var or on the MBP
Bildschirmfoto 2025-03-24 um 20 17 03

That was not expected.

ReimarBauer: Anyway it is more important to have always a timeout there then a short cut module global. It is an easy refactoring.

matrss (changes requested):
There are two locations in docs/samples/automation/retriever.py where there is no timeout specified yet.

ReimarBauer:
> There are two locations in docs/samples/automation/retriever.py where there is no timeout specified yet.

That was removed because it is outdated

mslib/utils/mssautoplot.py
ReimarBauer: This is a bug https://github.com/Open-MSS/MSS/issues/2730 the ToDo was a hint. 

I try to write a test for that first.

matrss (approved):

PR #2733, opened 2025-03-25
remove outdated retriever.py

### LINKED ISSUE #2732: remove the incompatible retriever.py sample
the retriever.py was the initial idea for mssautoplot.py 

It is not compatible with the released version. So I think we can remove it and links to the official version.

ReimarBauer:
**Purpose of PR?**:

Fixes #2732

When someone wants to use that, then it needs a different MSS. version likly before v9 or older.

The new module in the library is mssautoplot


ReimarBauer:
ops that was not intended

PR #2737, opened 2025-03-27
Fix: Resolved match_type issue to support all PyFileSystem2 path defi…

### LINKED ISSUE #2718: data_dir=/tmp  from config editor cannot become stored and used as expected
Using /tmp for the data_dir in the config editor will store this into the msui_settings.json 

"data_dir": "/tmp",

but display it as ~/mssdata

Image

The bug is located in the comparison in https://github.com/Open-MSS/MSS/blob/develop/mslib/utils/config.py#L672 

we recognized match_type has a bug 
 type(match_type("~/mss")) == StrType while type(c.match_type("/mss")) == FilepathType

This limits currently the path to have a startwith tilde or an fs url definition. 

We need a comparison which allows all PyFileSystem2 path definitions.  What we see in the editor should be the value which we stored in the file.

levi178u:
Fix: Ensure match_type supports all PyFileSystem2 path definitions

Purpose of PR?:
Fixes #2718

No

Verified that file paths are correctly stored and displayed in msui_settings.json

Ensured that paths starting with ~, /, and fs URLs are properly handled

Ran unit tests to confirm no regressions

Additional information for the reviewer?:
This PR addresses an issue where match_type incorrectly identified ~/mss as StrType and /mss as FilepathType. The updated logic ensures that all PyFileSystem2 path definitions are supported, aligning the stored and displayed values.

Does this PR result in some documentation changes?
No

Checklist:

 Bug fix. Fixes #2718

 New feature (Non-API breaking changes that add functionality)

 PR Title follows the convention of <type>: <subject>

 Commit has unit tests

### DIFF
--- a/mslib/utils/config.py
+++ b/mslib/utils/config.py
@@ -483,15 +483,18 @@ def modify_config_file(data, path=constants.MSUI_SETTINGS):
     path = path.replace("\\", "/")
     dir_name, file_name = fs.path.split(path)
     json_file_data = {}
+    print(f"DEBUG: Modifying config file at {path}")
     with fs.open_fs(dir_name) as _fs:
         if _fs.exists(file_name):
             try:
                 file_content = _fs.readtext(file_name)
+                print(f"DEBUG: Original file content: {file_content}")
                 json_file_data = json.loads(file_content, object_pairs_hook=dict_raise_on_duplicates_empty)
                 json_file_data_copy = copy.deepcopy(json_file_data)
                 for key in data:
                     if key not in json_file_data:
                         json_file_data_copy[key] = config_loader(dataset=key, default=True)
+                print(f"DEBUG: Before merging: {json_file_data_copy}")
                 modified_data = merge_dict(json_file_data_copy, data)
                 logging.debug("Merged default and user settings")
                 _fs.writetext(file_name, json.dumps(modified_data, indent=4))
@@ -669,7 +672,7 @@ def compare_data(default, user_data):
             user_data = float(default)
         if isinstance(match_type(default), UrlType) and isinstance(match_type(user_data), StrType):
             return user_data, True
-        if isinstance(match_type(default), type(match_type(user_data))):
+        if isinstance(match_type(default), (StrType, fs.base.FS)) and isinstance(match_type(user_data), (StrType,fs.base.FS)):
             return user_data, True
         else:
             return default, False
--- a/mslib/utils/config.py
+++ b/mslib/utils/config.py
@@ -483,18 +483,15 @@ def modify_config_file(data, path=constants.MSUI_SETTINGS):
     path = path.replace("\\", "/")
     dir_name, file_name = fs.path.split(path)
     json_file_data = {}
-    print(f"DEBUG: Modifying config file at {path}")
     with fs.open_fs(dir_name) as _fs:
         if _fs.exists(file_name):
             try:
                 file_content = _fs.readtext(file_name)
-                print(f"DEBUG: Original file content: {file_content}")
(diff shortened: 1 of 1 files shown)


levi178u:
@ReimarBauer pls review this PR , I had checked the changes made with test cases.  pls suggest what shud I do to fix this ?

levi178u:
I had committed a single change on line 672 , the debug lines have been included in it too (which I used for verifying results), I'll be raising a new PR with one change.

mslib/utils/config.py
```diff
@@ -483,15 +483,18 @@ def modify_config_file(data, path=constants.MSUI_SETTINGS):
     path = path.replace("\\", "/")
     dir_name, file_name = fs.path.split(path)
     json_file_data = {}
+    print(f"DEBUG: Modifying config file at {path}")
```
ReimarBauer: when this should be logging, use loggin

ReimarBauer:
Do you see a possibility to test this by a unit test?

ReimarBauer:
> I had committed a single change on line 672 , the debug lines have been included in it too (which I used for verifying results), I'll be raising a new PR with one change.

you can just update the PR, see also the failing test

mslib/utils/config.py
levi178u: Ok

levi178u:
> Do you see a possibility to test this by a unit test?

@ReimarBauer , as per the issue basically the match_type was incorrectly identifying `~/mss `as **StrType** and` /mss` as **FilepathType**. I just updated the comparison logic to allow all PyFileSystem2 path definitions, ensuring that the displayed path in the editor matches the stored value in **msui_settings.json**. Now it's able to read the path and fetch values which solves this issue but it's failing on 5 cases and passing on 13 cases which i think needs to be addressed in another issue  as to why certain values are fetched incorrectly.
Screenshot 2025-03-27 202718

ReimarBauer:
It must be done in same PR. We don't merge when something is broken.

E.g. the test about modifying was able to do this with that design failure.

ReimarBauer (changes requested):
There is an important test setup missing before doing such a refactoring.
Unfortunately noone had written that test before. So the one starting the refactoring has to do that first.

A test for one case can look like:

```
from mslib.utils.config import compare_data


def test_compare_data():
    assert compare_data({"new_flighttrack_flightlevel": 250},
                        {"new_flighttrack_flightlevel": 25}) == ({"new_flighttrack_flightlevel": 25}, True)

```

You have to figure all types used first with the stable branch, and write also a testcase which fails because of the bug. Then when you do the fix none of the tests should fail.
By this you know then that you did not added a new problem.

Sometimes more problems can be found by writing tests, but then we also have a discussion base, valid test - valid fix.

levi178u:
@ReimarBauer Ok I'll make sure to keep that in check, I had also create a separate pytest file to check for additional test cases, their fetching of data is correct, but why is this issue occurring needs fix. I'll work on this issue simultaneously with other ones too

mslib/utils/config.py
```diff
@@ -669,7 +669,7 @@ def compare_data(default, user_data):
             user_data = float(default)
         if isinstance(match_type(default), UrlType) and isinstance(match_type(user_data), StrType):
             return user_data, True
-        if isinstance(match_type(default), type(match_type(user_data))):
+        if isinstance(match_type(default), (StrType, fs.base.FS)) and isinstance(match_type(user_data), (StrType,fs.base.FS)):
```
matrss, line 672: This does not address the issue, as the bug is incorrect type recognition in `match_type` and therefore must be fixed somewhere in the `match_type` machinery.

matrss (changes requested):

PR #2729, opened 2025-03-24
Investigating server timeouts

ReimarBauer:
**Purpose of PR?**

Fixes #2626

investigating then fixing maybe

```
ERROR tests/_test_msui/test_sideview.py::Test_SideViewWMS::test_server_getmap - Failed: Server: "mslib.mswms.app" did not become ready in time
ERROR tests/_test_msui/test_linearview.py::Test_LinearViewWMS::test_server_getmap - Failed: Server: "mslib.mswms.app" did not become ready in time
ERROR tests/_test_msui/test_topview.py::Test_TopViewWMS::test_server_getmap - Failed: Server: "mslib.mswms.app" did not become ready in time
ERROR tests/_test_msui/test_wms_control.py::Test_VSecWMSControlWidget::test_server_getmap - Failed: Server: "mslib.mswms.app" did not become ready in time
ERROR tests/_test_msui/test_wms_control.py::Test_VSecWMSControlWidget::test_multilayer_drawing - Failed: Server: "mslib.mswms.app" did not become ready in time
ERROR tests/_test_msui/test_wms_control.py::Test_HSecWMSControlWidget::test_filter_handling - Failed: Server: "mslib.mswms.app" did not become ready in time
ERROR tests/_test_msui/test_wms_control.py::Test_HSecWMSControlWidget::test_server_no_thread - Failed: Server: "mslib.mswms.app" did not become ready in time
ERROR tests/_test_msui/test_wms_control.py::Test_HSecWMSControlWidget::test_singlelayer_handling - Failed: Server: "mslib.mswms.app" did not become ready in time
ERROR tests/_test_msui/test_wms_control.py::Test_HSecWMSControlWidget::test_no_server - Failed: Server: "mslib.mswms.app" did not become ready in time
ERROR tests/_test_msui/test_wms_control.py::Test_HSecWMSControlWidget::test_server_getmap_cached - Failed: Server: "mslib.mswms.app" did not become ready in time
ERROR tests/_test_msui/test_wms_control.py::Test_HSecWMSControlWidget::test_no_schema - Failed: Server: "mslib.mswms.app" did not become ready in time
ERROR tests/_test_msui/test_wms_control.py::Test_HSecWMSControlWidget::test_server_service_cache - Failed: Server: "mslib.mswms.app" did not become ready in time
ERROR tests/_test_msui/test_wms_control.py::Test_HSecWMSControlWidget::test_invalid_url - Failed: Server: "mslib.mswms.app" did not become ready in time
[...]

PR #2739, opened 2025-03-28
Use MSUIDefaultConfig.new_flighttrack_template Instead of STUB_CODE

### LINKED ISSUE #2474: unify mscolab_settings.STUB_CODE with MSUIDefaultConfig.new_flighttrack_template
Due to the separate development of the flight path and operations, we now have different templates for their defaults.

We should change the mscolab to be able to use the user defined MSUIDefaultConfig.new_flighttrack_template and replace the server defined mscolab_settings.STUB_CODE. 

This way a user configuration can control the template.

annapurna-gupta:
Fixes #2474


ReimarBauer:
look at the 90 failed, 390 passed, 11 skipped, 44 warnings, 90 errors in 470.02s (0:07:50)

mslib/mscolab/server.py
```diff
@@ -555,22 +557,30 @@ def error413(error):
 @APP.route('/create_operation', methods=["POST"])
 @verify_user
 def create_operation():
+    global _client_default_content
```
ReimarBauer: don't introduce a global

mslib/mscolab/server.py
```diff
@@ -555,22 +557,30 @@ def error413(error):
...
     path = request.form['path']
-    content = request.form.get('content', None)
+    content = request.form.get('content')
+    default_content = request.form.get('default_content')
```
ReimarBauer: is needed to have two times the same "content" get function?

annapurna-gupta: content is for the user upload content and default_content is to request MSUIDefaultConfig.new_flighttrack_template

mslib/mscolab/server.py
annapurna-gupta: okk

mslib/mscolab/server.py
ReimarBauer: But why do you need both?

annapurna-gupta: content was already there i just request for default content form the client ..is it not right??

ReimarBauer: I mean content is many times used while your new get parameter is only used  (or should only be used) on create of an operation. Why can't you use content for that?

annapurna-gupta: how can i do that? any suggestion

annapurna-gupta: okkk

annapurna-gupta:
@ReimarBauer I got the error **TypeError: contents must be unicode,** but I don't understand what 'contents' means.

mslib/mscolab/seed.py
```diff
@@ -63,7 +64,7 @@ def add_all_users_default_operation(path='TEMPLATE', description="Operation to k
             with fs.open_fs(mscolab_settings.OPERATIONS_DATA) as file_dir:
                 if not file_dir.exists(path):
                     file_dir.makedir(path)
-                    file_dir.writetext(f'{path}/main.ftml', mscolab_settings.STUB_CODE)
+                    file_dir.writetext(f'{path}/main.ftml', get_content)
```
ReimarBauer: name it content

mslib/mscolab/file_manager.py
```diff
@@ -57,7 +57,8 @@ def _get_operation_lock(self, op_id):
                 self.operation_locks[op_id] = threading.Lock()
                 return self.operation_locks[op_id]
 
-    def create_operation(self, path, description, user, last_used=None, content=None, category="default", active=True):
+    def create_operation(self, path, description, user, last_used=None,
```
ReimarBauer: I assume the lines are the same, so different formatting?

mslib/mscolab/server.py
```diff
@@ -556,21 +557,25 @@ def error413(error):
...
+    r = str(fm.create_operation(path, description, user, last_used, content=content, category=category, active=active))
     if r == "True":
         token = request.args.get('token', request.form.get('token', False))
         json_config = {"token": token}
         sockio.sm.update_operation_list(json_config)
     return r
 
 
+def get_content():
+    return session.get('get_content')
```
ReimarBauer: why is that needed?

mslib/mscolab/seed.py
ReimarBauer: and add its content to the function.

mslib/mscolab/server.py
```diff
@@ -556,21 +557,25 @@ def error413(error):
 @verify_user
 def create_operation():
     path = request.form['path']
-    content = request.form.get('content', None)
+    content = request.form.get('content', request.form.get('default_content'))
```
ReimarBauer: we need only content

mslib/msui/mscolab.py
```diff
@@ -1027,9 +1027,16 @@ def add_operation(self):
             self.error_dialog.showMessage('Path can\'t contain spaces or special characters')
             return
 
+        user_template = MSUIDefaultConfig.new_flighttrack_template
```
ReimarBauer: take this only from the config_loader

e.g.

waypoints = config_loader(dataset="new_flighttrack_template")

mslib/mscolab/seed.py
```diff
@@ -32,6 +32,7 @@
 from mslib.mscolab.conf import mscolab_settings
 from mslib.mscolab.models import User, db, Permission, Operation
 from mslib.mscolab.server import APP as app
+from mslib.mscolab.server import get_content
```
ReimarBauer: seed is needed for the tests, it needs therefore a constant content.

mslib/msui/mscolab.py
```diff
@@ -1027,9 +1027,16 @@ def add_operation(self):
...
+        xml_waypoints = "".join(
```
ReimarBauer: that goes wrong. we need the coordinates of the waypoints.

the same problem is solved for the local flighttrack.

ReimarBauer (changes requested):
see comments

annapurna-gupta:
> see comments

okay.

mslib/msui/mscolab.py
ReimarBauer: This did not work. we need the coordinates, you can do this similiar to 

https://github.com/annapurna-gupta/MSS/blob/develop/mslib/msui/mscolab.py#L1162

you get by waypoints_model.get_xml_content() your needed xmlcontent.

ReimarBauer (changes requested):
see comments

tests/_test_mscolab/test_server.py
```diff
@@ -504,8 +504,8 @@ def test_import_permissions(self):
             # creator is not listed
             assert data["success"] is True
 
-    def _create_operation(self, test_client, userdata=None, path="firstflight", description="simple test", active=True,
-                          content=None):
+    def _create_operation(self, test_client, userdata=None, path="firstflight", description="simple test",
+                          active=True, content):
```
ReimarBauer: when you don't have a default value for content, then you break the order of arguments

SyntaxError: non-default argument follows default argument

That is the reason for many errors in the tests

ReimarBauer: My IDE PyCharm shows that problem, like

Bildschirmfoto 2025-04-05 um 09 53 16

ReimarBauer: when you remove None here, you need to refactor all the tests where content now is unfilled too

ReimarBauer (changes requested):
see comment

and look on the CI tests too

annapurna-gupta:
> see comment
> 
> and look on the CI tests too

Yes, I just wanted to show you whether the content handled in **@verify_user def create_operation()** is okay or not.

mslib/mscolab/file_manager.py
```diff
@@ -100,10 +100,7 @@ def create_operation(self, path, description, user, last_used=None, content=None
             data = fs.open_fs(self.data_dir)
             data.makedir(operation.path)
             operation_file = data.open(fs.path.combine(operation.path, 'main.ftml'), 'w')
-            if content is not None:
```
ReimarBauer, line 103: that check is important

mslib/mscolab/file_manager.py
```diff
@@ -100,10 +100,7 @@ def create_operation(self, path, description, user, last_used=None, content=None
...
-                operation_file.write(content)
-            else:
-                operation_file.write(mscolab_settings.STUB_CODE)
+            operation_file.write(content)
             operation_path = fs.path.combine(self.data_dir, operation.path)
```
ReimarBauer, line 106: with your change you need to indent this under the if content is not None: condition

mslib/mscolab/server.py
```diff
@@ -556,21 +557,25 @@ def error413(error):
 @verify_user
 def create_operation():
     path = request.form['path']
-    content = request.form.get('content', None)
+    content = request.form.get('default_content')
```
ReimarBauer: content = request.form.get('content', None)

mslib/mscolab/server.py
```diff
@@ -556,21 +557,25 @@ def error413(error):
...
+    session['content'] = content
```
ReimarBauer: Why do we need this in session?

mslib/mscolab/server.py
```diff
@@ -556,21 +557,25 @@ def error413(error):
...
-                                content=content, category=category, active=active))
+    r = str(fm.create_operation(path, description, user, last_used, content=content, category=category, active=active))
     if r == "True":
         token = request.args.get('token', request.form.get('token', False))
         json_config = {"token": token}
         sockio.sm.update_operation_list(json_config)
     return r
 
 
+def content():
```
ReimarBauer: this seems not used, do we need it?

mslib/mscolab/seed.py
```diff
@@ -32,6 +32,7 @@
 from mslib.mscolab.conf import mscolab_settings
 from mslib.mscolab.models import User, db, Permission, Operation
 from mslib.mscolab.server import APP as app
+from mslib.mscolab.server import content
```
ReimarBauer: you need here as content the STUB_CODE  maybe move it to here

And you need that also as default in the tests for test_server.py

tests/_test_mscolab/test_files.py
```diff
@@ -220,7 +220,7 @@ def _example_data(self):
           </ListOfWaypoints>
       </FlightTrack>"""
 
-    def _create_operation(self, flight_path="firstflight", user=None, content=None):
```
ReimarBauer, line 223: that is correct

but you need for the case that it is None use the xml from the stub_code because now to have to initialize an operation by sending data. Which was in the past done on the server.

tests/_test_mscolab/test_files_api.py
```diff
@@ -203,7 +203,7 @@ def test_set_version_name(self):
             version_name = all_changes[-1]["version_name"]
             assert version_name == "berlin"
 
-    def _create_operation(self, flight_path="firstflight", description="example", user=None, content=None):
```
ReimarBauer, line 206: see above

tests/_test_mscolab/test_server.py
```diff
@@ -504,8 +504,8 @@ def test_import_permissions(self):
             # creator is not listed
             assert data["success"] is True
 
-    def _create_operation(self, test_client, userdata=None, path="firstflight", description="simple test", active=True,
```
ReimarBauer, line 507: see above

tests/_test_mscolab/test_server.py
ReimarBauer: I won't do that. 

But when it is None, that just means you need to send the data which was in the past in the stub_code. Because now you intitalize the server oepration with data.

ReimarBauer (changes requested):
also when you find tests having lines self.fm.create_operation(.... 

these have also now to send a default content which was before the stub_code

mslib/msui/mscolab.py
```diff
@@ -1026,10 +1027,16 @@ def add_operation(self):
...
+        waypoints_model = ft.WaypointsTableModel(waypoints=[
+            ft.Waypoint(location=loc, lat=0.0, lon=0.0) 
+            for loc in waypoints
+        ])
+        default_content = waypoints_model.get_xml_content()
         data = {"path": path,
                 "description": description,
-                "category": category}
+                "category": category,
+                "default_content": default_content}
```
ReimarBauer: content

we want always only content

annapurna-gupta:
Hii @ReimarBauer , Look, here's what I was thinking when I made these changes. I needed to pass MSUIDefaultConfig.new_flighttrack_template to the server. But since this template is initialized on the client side, I couldn’t access it directly on the server. That’s why I passed it using the API. On the server side, I needed a way to receive new_flighttrack_template from the client, so I added a new variable called default_content in the create_operation() function. This way, the new_flighttrack_template can be stored under that variable. Am I missing something?

ReimarBauer:
It works without that. Grok it from the direction what the client provides by sending content and what the server has to do.

ReimarBauer (changes requested):
Good progress. :)

Please remove all occurrences of STUB_CODE.

You’ll find this in seed as well. The old values should be preserved there and in the tests. Perhaps you could add an XML_CONTENT_INIT and then have the tests do the same thing as the rest of the code. Some tests already have an XML_CONTENT variable, so use one that won't cause confusion.

annapurna-gupta:
> It works without that. Grok it from the direction what the client provides by sending content and what the server has to do.



> Good progress. :)
> 
> Please remove all occurrences of STUB_CODE.
> 
> You’ll find this in seed as well. The old values should be preserved there and in the tests. Perhaps you could add an XML_CONTENT_INIT and then have the tests do the same thing as the rest of the code. Some tests already have an XML_CONTENT variable, so use one that won't cause confusion.

Thank you
Should I initialize XML_CONTENT_INIT like STUB_CODE with different  waypoints, and have the tests use it?

annapurna-gupta:
> Good progress. :)
> 
> Please remove all occurrences of STUB_CODE.
> 
> You’ll find this in seed as well. The old values should be preserved there and in the tests. Perhaps you could add an XML_CONTENT_INIT and then have the tests do the same thing as the rest of the code. Some tests already have an XML_CONTENT variable, so use one that won't cause confusion.

But the waypoints of XML_CONTENT are different? So will it be okay to still use them.

ReimarBauer:
I am not sure if I know what you mean. The tests have some different content. You will see when you look into them. 
The coverage of the tests have to be the same after you solved it.

For the start I would use an Initial new xmlcontent which is based on the STUB_CODE values. And I would import it where it is needed.

mslib/mscolab/conf.py
```diff
@@ -103,6 +103,19 @@ class default_mscolab_settings:
     </FlightTrack>
     """
 
+    XML_CONTENT_INIT = """<?xml version="1.0" encoding="utf-8"?>
```
ReimarBauer: we need this only for tests, or? at least we should only need it there.
We don't expect a user to change this on the server, right?

so we can import it from anywhere, it has currently a strong relation to seed, also define it there. When it is only related to tests it would be tests.constants.

mslib/mscolab/file_manager.py
```diff
@@ -103,7 +103,7 @@ def create_operation(self, path, description, user, last_used=None, content=None
             if content is not None:
                 operation_file.write(content)
             else:
-                operation_file.write(mscolab_settings.STUB_CODE)
+                operation_file.write(mscolab_settings.XML_CONTENT_INIT)
```
ReimarBauer: this looks wrong

The file_manager should only write what it gets from the client per content parameter.

May be add a condition at the beginning a condition requesting content is None: return False

mslib/mscolab/seed.py
```diff
@@ -63,7 +63,7 @@ def add_all_users_default_operation(path='TEMPLATE', description="Operation to k
             with fs.open_fs(mscolab_settings.OPERATIONS_DATA) as file_dir:
                 if not file_dir.exists(path):
                     file_dir.makedir(path)
-                    file_dir.writetext(f'{path}/main.ftml', mscolab_settings.STUB_CODE)
+                    file_dir.writetext(f'{path}/main.ftml', mscolab_settings.XML_CONTENT_INIT)
```
ReimarBauer: use it from seed

mslib/mscolab/server.py
```diff
@@ -562,8 +562,7 @@ def create_operation():
     active = (request.form.get('active', "True") == "True")
     last_used = datetime.datetime.now(tz=datetime.timezone.utc)
     user = g.user
-    r = str(fm.create_operation(path, description, user, last_used,
-                                content=content, category=category, active=active))
+    r = str(fm.create_operation(path, description, user, last_used, content=content, category=category, active=active))
```
ReimarBauer, line 565: this is only formatting

tests/_test_mscolab/test_file_manager.py
```diff
@@ -560,7 +560,8 @@ def _example_data(self):
     def _create_operation(self, flight_path="firstflight", user=None, content=None, category="default"):
         if user is None:
             user = self.user
-        self.fm.create_operation(flight_path, f"info about {flight_path}", user, content=content, category=category)
+        self.fm.create_operation(flight_path, f"info about {flight_path}", user,
```
ReimarBauer: only formatting

tests/_test_mscolab/test_files.py
```diff
@@ -125,12 +125,12 @@ def test_cant_save(self):
...
             assert self.fm.save_file(operation.id, incomplete, self.user) is False
 
-    def test_stub_data(self):
+    def test_xml_content_data(self):
         with self.app.test_client():
             flight_path, operation = self._create_operation(flight_path="operationstub")
             content = self.fm.get_file(operation.id, self.user)
             assert flight_path == "operationstub"
-            assert content == mscolab_settings.STUB_CODE
+            assert content == mscolab_settings.XML_CONTENT_INIT
```
ReimarBauer, line 135: seed.XML_CONTENT_INIT

ReimarBauer (changes requested):
Moin @annapurna-gupta 
a few changes.  Please remove STUB_CODE also from the sample configuration files in the docs and conftest.py.

annapurna-gupta:
@ReimarBauer now **XML_CONTENT_INIT** is in correct place?

conftest.py
```diff
@@ -162,18 +162,6 @@ def generate_initial_config():
 MAX_UPLOAD_SIZE = 2 * 1024 * 1024  # 2MB
 
 # text to be written in new mscolab based ftml files.
```
ReimarBauer: the coment is now obsolete

ReimarBauer:
Hi, I am currently reading.
I always look first on new files added. You should do that yourself too.

.coverage was added and we don't need this. -> remove

docs/samples/config/mscolab/mscolab_settings.py.sample
```diff
@@ -83,18 +83,6 @@ SQLALCHEMY_ECHO = False
 enable_basic_http_authentication = False
 
 # text to be written in new mscolab based ftml files.
```
ReimarBauer: obsolete comment

ReimarBauer:
new flighttrack template is used 
Bildschirmfoto 2025-04-17 um 21 46 39

ReimarBauer (changes requested):
look on the failing tests, locally they fail on my MBP too
5 failed, 564 passed, 12 skipped, 44 warnings, 1 error in 500.78s (0:08:20)

pytest -k test_get_recent_oid
pytest -k test_save_overwrite_to_server
pytest -k test_save_keep_server_points
pytest -k test_fetch_from_server
pytest -k test_save_merge_points

The error I have hopefully addressed in a PR.

There is likly in theire _create_operation the testclient not sending XML_CONTENT_INIT.

annapurna-gupta:
> new flighttrack template is used Bildschirmfoto 2025-04-17 um 21 46 39

How you did this? Because I checked it using the UI by creating an operation and saw that the default waypoints were Delhi and Nagpur, but what you shared is different.

ReimarBauer:
> > new flighttrack template is used Bildschirmfoto 2025-04-17 um 21 46 39
> 
> How you did this? Because I checked it using the UI by creating an operation and saw that the default waypoints were Delhi and Nagpur, but what you shared is different.

See File - Configuration 

in locations you have the names you can use new_flighttrack_template

annapurna-gupta:
> > > new flighttrack template is used Bildschirmfoto 2025-04-17 um 21 46 39
> > 
> > 
> > How you did this? Because I checked it using the UI by creating an operation and saw that the default waypoints were Delhi and Nagpur, but what you shared is different.
> 
> See File - Configuration
> 
> in locations you have the names you can use new_flighttrack_template

Screenshot from 2025-04-18 20-26-51
this is what i got..what does it mean?

ReimarBauer:
A flightpath has minimum 2 waypoints.

It is a name which is defined in locations.

Click on such a name and write e.g. EDMO

Or click Add and add a third name

ReimarBauer:
please remove your version of .coverage

ReimarBauer:
have a look on this one
https://github.com/annapurna-gupta/MSS/blob/feature/mscolab-templete/tests/_test_mscolab/test_utils.py#L90

annapurna-gupta:
> please remove your version of .coverage

Done

annapurna-gupta:
@ReimarBauer When I ran the tests locally on macOS -13, macOS -14, and Ubuntu, there were no errors. However, the macOS -13 or macOS -14 test failed on GitHub?

ReimarBauer:
Yes known. There is an open PR by me, trying to track that down.

annapurna-gupta:
@ReimarBauerIn **commit 21**, the tests failed on **macOS-13 and macOS-14,** but they were passing in **commit** **20**. Only the linter was failing in commit 20, and after fixing the lint issues, **the tests started failing**. what should i do?

ReimarBauer:
Ignore macos13/14 wms tests. They are not related to your change.

annapurna-gupta:
> Ignore macos13/14 wms tests. They are not related to your change.

Okay, then I guess this issue is fixed? Because whenever I create a new operation, new waypoints are used. Or am I missing something?

ReimarBauer:
I review it later when there is a bit time. We have eastern this weekend.

annapurna-gupta:
> I review it later when there is a bit time. We have eastern this weekend.

okk

.gitignore
```diff
@@ -20,3 +20,4 @@ tutorials/recordings
 tutorials/cursor_image.png
 # pixi environments
 .pixi
+.coverage
```
ReimarBauer, line 23: nice

mslib/mscolab/server.py
```diff
@@ -556,14 +556,13 @@ def error413(error):
 @verify_user
 def create_operation():
     path = request.form['path']
-    content = request.form.get('content', None)
+    content = request.form.get('content')
```
ReimarBauer: , None

the filemanager reacts on None with False. 

We should keep this here

tests/_test_mscolab/test_file_manager.py
```diff
@@ -558,6 +560,8 @@ def _example_data(self):
   </FlightTrack>"""
 
     def _create_operation(self, flight_path="firstflight", user=None, content=None, category="default"):
+        if content is None:
```
ReimarBauer: you have to set it as default content=XML_CONTENT_INIT

and use then the parameter.

tests/_test_mscolab/test_file_manager.py
```diff
@@ -567,7 +571,7 @@ def _create_operation(self, flight_path="firstflight", user=None, content=None,
     def _create_operation_with_users(self, flight_path="firstflight", user=None, content=None):
```
ReimarBauer: use here instead of None XML_CONTENT_INIT as default and then always content.

tests/_test_mscolab/test_file_manager.py
ReimarBauer: you use in the tests a global var, but the tests should use the parameter call.

tests/_test_mscolab/test_file_manager.py
```diff
@@ -567,7 +571,7 @@ def _create_operation(self, flight_path="firstflight", user=None, content=None,
...
         if user is None:
             user = self.user
-        self.fm.create_operation(flight_path, f"info about {flight_path}", user, content=content)
+        self.fm.create_operation(flight_path, f"info about {flight_path}", user, content=XML_CONTENT_INIT)
```
ReimarBauer: use content, initialize it in the signature with XML_CONTENT_INIT

tests/_test_mscolab/test_file_manager.py
```diff
@@ -576,7 +580,7 @@ def _create_operation_with_users(self, flight_path="firstflight", user=None, con
     def _create_operation_with_opposite_permissions(self, flight_path="firstflight", user=None, content=None):
         if user is None:
             user = self.user
-        self.fm.create_operation(flight_path, f"info about {flight_path}", user, content=content)
+        self.fm.create_operation(flight_path, f"info about {flight_path}", user, content=XML_CONTENT_INIT)
```
ReimarBauer: same

tests/utils.py
```diff
@@ -192,10 +193,13 @@ def mscolab_delete_all_operations(app, msc_url, email, password, username, fulln
         response = app.test_client().post(url, data=data)
 
 
-def mscolab_create_operation(app, msc_url, response, path='f', description='description'):
+def mscolab_create_operation(app, msc_url, response, path='f', description='description', content=None):
+    if content is None:
```
ReimarBauer: you can initialize content by XML_CONTENT_INIT

ReimarBauer (changes requested):
few changes in tests

annapurna-gupta:
@ReimarBauer i am getting **Timeout** error, should i increase the timeout duration?

ReimarBauer:
> @ReimarBauer i am getting **Timeout** error, should i increase the timeout duration?

Please don't. There is a race condition, which we have to find. longer timeout just let us wait longer.
It is not easy to track down. 
https://github.com/Open-MSS/MSS/pull/2729

annapurna-gupta:
> > @ReimarBauer i am getting **Timeout** error, should i increase the timeout duration?
> 
> Please don't. There is a race condition, which we have to find. longer timeout just let us wait longer. It is not easy to track down. #2729

Okay, then is there any change I should make in this PR?

ReimarBauer (approved):
Thx :)

PR #2752, opened 2025-03-31
fix: replace mutable default args with None

### LINKED ISSUE #2743: default argument value is mutable in mpl_map
Default argument values are evaluated only once, at function definition. Consequently, any modification to that value will affect all subsequent calls.

```
def set_draw_airports(self, value, port_type=["small_airport"], reload=True):
def set_draw_airspaces(self, value, airspaces=[], range_km=None, reload=True):
def draw_airspaces(self, countries=[], range_km=None):
```

a solution is to define it None and set the value because it is None.

e.g.
```
def set_draw_airports(self, value, port_type=None, reload=True):
    if port_type is None:
        port_type=["small_airport"]
```

divyaajainn:
**Purpose of PR?**:This PR fixes issues with mutable default argument values by replacing them with None and initializing new lists inside the functions.

Bug Fix
Fixes #2743

### DIFF
--- a/mslib/msui/mpl_map.py
+++ b/mslib/msui/mpl_map.py
@@ -351,11 +351,11 @@ class MapCanvas(basemap.Basemap):
             self.ax.figure.canvas.draw()
 
     def set_draw_airports(self, value, port_type=None, reload=True):
-        if port_type is None:
-            port_type = ["small_airport"]
         """
         Sets airports to visible or not visible
         """
+        if port_type is None:
+            port_type = ["small_airport"]
         if (reload or not value or len(port_type) == 0) and self.airports:
             self.update_info_text(ourairports="")
             self.airports.remove()
@@ -367,11 +367,11 @@ class MapCanvas(basemap.Basemap):
             self.draw_airports(port_type)
 
     def set_draw_airspaces(self, value, airspaces=None, range_km=None, reload=True):
-        if airspaces is None:
-            airspaces = []
         """
         Sets airspaces to visible or not visible
         """
+        if airspaces is None:
+            airspaces = []
         if (reload or not value or len(airspaces) == 0) and self.airspaces:
             self.update_info_text(openaip="")
             self.airspaces.remove()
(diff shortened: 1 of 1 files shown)


mslib/msui/mpl_map.py
```diff
@@ -350,7 +350,9 @@ def set_graticule_visible(self, visible=True):
             # Update the figure canvas.
             self.ax.figure.canvas.draw()
 
-    def set_draw_airports(self, value, port_type=["small_airport"], reload=True):
+    def set_draw_airports(self, value, port_type=None, reload=True):
+        if port_type is None:
```
ReimarBauer: move this below the doc string

mslib/msui/mpl_map.py
```diff
@@ -364,7 +366,9 @@ def set_draw_airports(self, value, port_type=["small_airport"], reload=True):
         if value and len(port_type) > 0:
             self.draw_airports(port_type)
 
-    def set_draw_airspaces(self, value, airspaces=[], range_km=None, reload=True):
+    def set_draw_airspaces(self, value, airspaces=None, range_km=None, reload=True):
+        if airspaces is None:
+            airspaces = []
```
ReimarBauer: please move it below the doc string

mslib/msui/mpl_map.py
```diff
@@ -379,7 +383,9 @@ def set_draw_airspaces(self, value, airspaces=[], range_km=None, reload=True):
             country_codes = [airspace.split(" ")[-1] for airspace in airspaces]
             self.draw_airspaces(country_codes, range_km)
 
-    def draw_airspaces(self, countries=[], range_km=None):
+    def draw_airspaces(self, countries=None, range_km=None):
+        if countries is None:
+            countries = []
```
ReimarBauer: same here

ReimarBauer (changes requested):
see comments

ReimarBauer (approved):
Thx

PR #2753, opened 2025-03-31
fix: update abstract base class to include missing parameter

### LINKED ISSUE #2745: unexpected argument in dataaccess
the signature of def _determine_filename(self, variable, vartype, init_time, valid_time) is different to its usage in dataaccess.have_data with variable, vartype, init_time, valid_time, reload=False)

The functionality is not directly covered by a test. A test needs to become added too.

divyaajainn:
**Purpose of PR?**: Updates the abstract base class signature of _determine_filename to include a default reload parameter, aligning it with its implementations and usage in have_data.

Bug fix
Fixes #2745

### DIFF
--- a/mslib/mswms/dataaccess.py
+++ b/mslib/mswms/dataaccess.py
@@ -118,7 +118,7 @@ class NWPDataAccess(metaclass=ABCMeta):
         pass
 
     @abstractmethod
-    def _determine_filename(self, variable, vartype, init_time, valid_time, reload= False):
+    def _determine_filename(self, variable, vartype, init_time, valid_time, reload=False):
         """
         Must be overwritten in subclass. Determines the filename
         (without path) of the variable <variable> at the forecast
--- a/mslib/mswms/dataaccess.py
+++ b/mslib/mswms/dataaccess.py
@@ -118,7 +118,7 @@ class NWPDataAccess(metaclass=ABCMeta):
         pass
 
     @abstractmethod
-    def _determine_filename(self, variable, vartype, init_time, valid_time, reload=False):
+    def _determine_filename(self, variable, vartype, init_time, valid_time, reload=True):
         """
         Must be overwritten in subclass. Determines the filename
         (without path) of the variable <variable> at the forecast
--- a/mslib/mswms/dataaccess.py
+++ b/mslib/mswms/dataaccess.py
@@ -118,7 +118,7 @@ class NWPDataAccess(metaclass=ABCMeta):
         pass
 
     @abstractmethod
-    def _determine_filename(self, variable, vartype, init_time, valid_time):
+    def _determine_filename(self, variable, vartype, init_time, valid_time, reload= False):
         """
         Must be overwritten in subclass. Determines the filename
         (without path) of the variable <variable> at the forecast
--- a/tests/_test_mswms/test_dataaccess.py
+++ b/tests/_test_mswms/test_dataaccess.py
@@ -90,6 +90,22 @@ class Test_DefaultDataAccess:
                     datetime(2012, 10, 18, 12, 0),
                     datetime(2012, 10, 19, 0, 0)])
 
+    def test_determine_filename_reload(self):
+        variable = "air_pressure"
+        vartype = "ml"
+        init_time = datetime(2012, 10, 17, 12, 0)
+        valid_time = datetime(2012, 10, 17, 18, 0)
+
+        self.dut._filetree = {vartype: {init_time: {variable: {}}}}
+
+        def fake_setup():
+            self.dut._filetree[vartype][init_time][variable][valid_time] = "expected_filename.nc"
+
+        self.dut.setup = mock.MagicMock(side_effect=fake_setup)
+
+        filename = self.dut._determine_filename(variable, vartype, init_time, valid_time, reload=True)
+        assert filename == "expected_filename.nc"
+
 
 class Test_CachedDataAccess(Test_DefaultDataAccess):
     """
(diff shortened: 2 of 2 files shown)


ReimarBauer:
see the 2nd task in the issue

mslib/mswms/dataaccess.py
```diff
@@ -118,7 +118,7 @@ def is_reload_required(self, filenames):
         pass
 
     @abstractmethod
-    def _determine_filename(self, variable, vartype, init_time, valid_time):
+    def _determine_filename(self, variable, vartype, init_time, valid_time, reload=False):
```
matrss: The existing implementations of this method have True as the default value.

ReimarBauer (changes requested):
see comment

divyaajainn:
i will look into it today itself. i was busy with an event.

tests/_test_mswms/test_dataaccess.py
```diff
@@ -90,6 +90,22 @@ def test_get_all_valid_times(self):
...
+        self.dut._filetree = {vartype: {init_time: {variable: {}}}}
+
+        def fake_setup():
+            self.dut._filetree[vartype][init_time][variable][valid_time] = "expected_filename.nc"
+
+        self.dut.setup = mock.MagicMock(side_effect=fake_setup)
+
+        filename = self.dut._determine_filename(variable, vartype, init_time, valid_time, reload=True)
+        assert filename == "expected_filename.nc"
+
```
ReimarBauer, line 129: good :)

please also check the False state so that you can assert the ValueError

divyaajainn: okay.

divyaajainn:
Does this look good?

tests/_test_mswms/test_dataaccess.py
```diff
@@ -90,6 +90,32 @@ def test_get_all_valid_times(self):
...
+        filename = self.dut._determine_filename(variable, vartype, init_time, valid_time, reload=True)
+        assert filename == "expected_filename.nc"
+        self.dut.setup.assert_called_once()
+
+        self.dut._filetree = {vartype: {init_time: {variable: {}}}}
+        exception_was_raised = False
+        try:
+            self.dut._determine_filename(variable, vartype, init_time, valid_time, reload=False)
+        except ValueError:
```
ReimarBauer: you want to use 
with pytest.raises(ValueError): 
see https://pytest.org/en/latest/getting-started.html#assert-that-a-certain-exception-is-raised

ReimarBauer (changes requested):
see hint

tests/_test_mswms/test_dataaccess.py
divyaajainn: Okay, I was going for that initially, but didn’t import pytest at the time so I just quickly checked the behavior with a try-except. I’ll update it now.

divyaajainn:
I was going for that initially, but didn’t import pytest at the time so I just quickly checked the behavior with a try-except. I have updated it, does this look good?

tests/_test_mswms/test_dataaccess.py
```diff
@@ -90,6 +91,27 @@ def test_get_all_valid_times(self):
...
+        vartype = "ml"
+        init_time = datetime(2012, 10, 17, 12, 0)
+        valid_time = datetime(2012, 10, 17, 18, 0)
+
+        self.dut._filetree = {vartype: {init_time: {variable: {}}}}
+
+        def fake_setup():
+            self.dut._filetree[vartype][init_time][variable][valid_time] = "expected_filename.nc"
+
+        self.dut.setup = mock.MagicMock(side_effect=fake_setup)
```
matrss: This kind of mocking and modification of internal data structures is an absolute last resort approach. Everything that is mocked no longer tests the real application code, and modifying internal data structures ties the test to a specific implementation, instead of just testing for correct behavior.

I think this test can instead be written against the existing test data and the real setup implementation. E.g.:
- call with reload=False, observe an error
- call with reload=True, observe test data being loaded
- remove a file, call with reload=False, observe it is still being returned
- call with reload=True, observe the removed file no longer being returned
- add a file, call with reload=False, observe it not being returned
- call with reload=True, observe the added file being returned
- possibly more

matrss (changes requested):

tests/_test_mswms/test_dataaccess.py
divyaajainn: thanks for the feedback. i will update the tests.

divyaajainn:
I think the error is caused when I clear _filetree (by setting it to {}), the expected ValueError isn’t raised maybe DUT’s setup or a caching mechanism might be preventing it from being empty. 
i am thinking of adding an assertion to verify the internal state before calling _determine_filename and use debug logging to see if the except clause is reached.
That way if the implementation needs to be adjusted.

matrss:
> I think the error is caused when I clear _filetree (by setting it to {}), the expected ValueError isn’t raised maybe DUT’s setup or a caching mechanism might be preventing it from being empty.

The test failure you are seeing is not caused by the first `with pytest.raises(ValueError):`, but by the second one. Congratulations, you have found a new bug: `_determine_filename` returns an invalid/outdated filename with `reload=True` because it blindly trusts the cached filenames in `_filetree`. @ReimarBauer I think this needs some clarification on how this method is supposed to work, the reload parameter suggests there have been performance concerns in the past that prevented an implementation that tries to always be consistent with the filesystem?

In an attempted fix of this bug I noticed that just appending .bak to a filename in the test is not enough, as the next call to `.setup` will then pick up the variable from the .bak file.

divyaajainn:
Thanks. I hadn’t realized the failure was from the second ValueError. I’ll hold off on changes for now and wait for clarification on the expected behavior of reload. I’ll also update the test to avoid .bak files being picked up by .setup.

PR #2755, opened 2025-03-31
fix:  _icon call args to match parent signature 

### LINKED ISSUE #2747: unexpected call argument in mpl_qtwidget
In msui.mpl_qtwidget we have return super()._icon(name, *args) but the signature of _icon has not 2 arguments

divyaajainn:
**Purpose of PR?**: Removes extra arguments from the super()._icon call to match the parent method's expected signature.

Bug Fix
Fixes #2747


ReimarBauer (approved):
Thx

PR #2757, opened 2025-03-31
fix: add missing super().__init__ calls

### LINKED ISSUE #2750: Missed call to __init__ of the super class
mslib.utils.netCDF4tools:
`class MFDatasetCommonDims(netCDF4.MFDataset)` missed call to `__init__` of the super class 

mslib.utils.ogcwms
class WebMapService(wms111.WebMapService_1_1_1) missed call to `__init__` of the super class

divyaajainn:
**Purpose of PR?**: This PR adds missing super().__init__() calls to ensure proper parent class initialization in MFDatasetCommonDims and WebMapService.

Bug Fix
Fixes #2750

### DIFF
--- a/mslib/utils/netCDF4tools.py
+++ b/mslib/utils/netCDF4tools.py
@@ -238,7 +238,7 @@ class MFDatasetCommonDims(netCDF4.MFDataset):
         # Open the master file in the base class, so that the CDFMF instance
         # can be used like a CDF instance.
 
-        super.__init__(files)
+        super().__init__(files)
         exclude = exclude or []
         skip_dim_check = skip_dim_check or []
         if isinstance(files, str):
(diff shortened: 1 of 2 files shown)


ReimarBauer (changes requested):
106 failed, 463 passed, 12 skipped, 19 warnings, 13 errors in 479.55s (0:07:59)

matrss (changes requested):
Both classes seem to have at least parts of their constructors copied over from the classes they inherit, so just calling the super classes constructor is not enough, it has to be checked what that constructor does and what ours do so far, what the differences are, and how to achieve the same result with a cleaner refactoring of these classes.

divyaajainn:
got it. i'll look into it.

PR #2769, opened 2025-04-01
Fixed issues with call to Super class

levi178u:
## fix: add missing super().__init__() calls  

This PR adds the missing `super().__init__()` calls in `MFDatasetCommonDims` and  
`WebMapService` to ensure proper parent class initialization.  
Ensured proper initialization of the parent class `netCDF4.MFDataset`, improving compatibility  
and following rules of OOPs (object-oriented principles).

### DIFF
--- a/.github/workflows/build_docs_gallery.yml
+++ b/.github/workflows/build_docs_gallery.yml
@@ -8,7 +8,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v4
-    - uses: prefix-dev/setup-pixi@v0.8.3
+    - uses: prefix-dev/setup-pixi@v0.8.4
       with:
         pixi-version: latest
         cache: true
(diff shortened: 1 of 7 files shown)


levi178u:
@ReimarBauer  this PR just adds the missing `super().__init__()` call in `WebMapService` and  `MFDatasetCommonDims` to ensure that the parent class constructor is properly invoked (following Inheritance ). May I know the issue since it failed to pass some CI Test cases.

levi178u:
Also I had an intuition to apply this ```
super().__init__(files, mask_and_scale=False, decode_times=False,decode_cf=False, format="NETCDF3_CLASSIC")
``` instead of this `super().__init__(files) `, will that cause any difference ? Also how to fix the same here..

Also I referred this while mentioning the 'format' part : 

> ''' Datasets must be in C{NETCDF4_CLASSIC, NETCDF3_CLASSIC or NETCDF3_64BIT} format (C{NETCDF4} Datasets won't work).'''

Like if I use **NETCDF3_64BIT** instead of **NETCDF3_CLASSIC** , will it enable better results.

PR #2770, opened 2025-04-02
Fix : Renamed route functions to avoid naming conflicts

### LINKED ISSUE #2758: rename pages which render in index.py
All functions which do a `return render_template(...` should become unified with a prefix, e.g. render_help, render_about.

This solves also a naming conflict with the help() and it makes the functions distinguishable from the others there.

levi178u:
**Purpose of PR?**:

**Fix: Renamed Built-in Python Functions to Avoid Naming Conflicts**  

This PR renames functions in `index.py` to prevent conflicts with built-in Python functions.  

Fixes #2758  

**Changes:**  
- Renamed conflicting functions  
- Ensured clarity and consistency  

**Checklist:**  
- [x] Function names updated  
- [x] Verified all affected routes work correctly

### DIFF
--- a/.github/workflows/build_docs_gallery.yml
+++ b/.github/workflows/build_docs_gallery.yml
@@ -8,7 +8,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v4
-    - uses: prefix-dev/setup-pixi@v0.8.3
+    - uses: prefix-dev/setup-pixi@v0.8.4
       with:
         pixi-version: latest
         cache: true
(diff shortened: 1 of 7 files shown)


levi178u:
@ReimarBauer and @matrss , I have made some changes in the codebase regarding the fixing of the naming issue for the functions based on the route they serve like :  `help()` to  `mss_help()` and `about()` to  `mss_about() `to prevent conflicts with built-in Python functions and provide clarity .Since the 'enhancement' tag was removed, i thought to push it in stable to check if it works.  
Pls review the PR

matrss:
Clearly this doesn't work, see the failing tests. Granted, the timeouts are irritating. There are still references to the old function names that break the tests, keyword `url_for`.

ReimarBauer (changes requested):
API change can only go to develop

PR #2771, opened 2025-04-02
Update setup.py for Python 3.12 compatibility

### LINKED ISSUE #2738: py3.12
We can build MSS for py3.12 so we should verify if we can change the upper limit for python

levi178u:
**Purpose of PR?**:

Fixes #2738 

**Enhancement: Ensure Python 3.12 Compatibility**  

This PR verifies and updates MSS to support Python 3.12.  
- Tested MSS on Python 3.12.  
- Updated `setup.py` to include Python 3.12 version.
- Ran tests to ensure compatibility.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -48,7 +48,7 @@ setup(
     version=__version__,  # noqa
     description="MSS - Mission Support System",
     long_description=long_description,
-    classifiers="Development Status :: 5 - Production/Stable",
+    classifiers=["Development Status :: 5 - Production/Stable","Programming Language :: Python :: 3.12",],
     keywords="mslib",
     maintainer="Reimar Bauer",
     maintainer_email="rb.proj@gmail.com",


setup.py
```diff
@@ -48,7 +48,7 @@
     version=__version__,  # noqa
     description="MSS - Mission Support System",
     long_description=long_description,
-    classifiers="Development Status :: 5 - Production/Stable",
+    classifiers=["Development Status :: 5 - Production/Stable","Programming Language :: Python :: 3.12",],
```
matrss, line 51: This is only the classifier shown on PyPI and does not affect the python version used in any way.

matrss (changes requested):

ReimarBauer (changes requested):
where are the updates for running tests?

setup.py
levi178u: Thnx for the clarification, I included the classifier to indicate compatibility with Python 3.12, but I understand it doesn't enforce the version. Let me know if you suggest any changes

levi178u:
> where are the updates for running tests?

@ReimarBauer I’ve made updates for running tests, maybe they haven’t been pushed yet. I’ll check and update soon

ReimarBauer:
one issue one task, please don't mix changes

setup.py
levi178u: Also, may i know what other changes should I make in `setup.py` or other files to make it compatible with the Pypi 3.12 version .

levi178u:
> one issue one task, please don't mix changes

@ReimarBauer Sure I'll keep it in mind, I just raised a PR for fixing the `classifiers` list (issue #2772 ), so I went on to the changes here too, but I'll make sure and treat each task separately

PR #2784, opened 2025-04-06
Fix mismatched arguments in dataaccess._determine_filename (#2745)

### LINKED ISSUE #2745: unexpected argument in dataaccess

praharsh777:
Purpose of PR?:
Fixes #2745
This PR resolves a mismatch in arguments passed to the _determine_filename() method in dataaccess. The function was being called with an unexpected reload parameter, which is not accepted by its signature.

❌ No breaking changes. The fix aligns the method call with the defined function signature.

Verified that the method _determine_filename() is now called with the correct number of arguments.

Ensured no other part of the codebase passes reload to this method.

Additional information for reviewer?:
This PR is a direct fix for an issue discussed in #2745. The functionality is not yet covered by tests — test coverage will be handled in a follow-up PR.

Does this PR result in some Documentation changes?
❌ No documentation changes required.

Checklist:

 Bug fix. Fixes #2745

 New feature (Non-API breaking changes that adds functionality)

 PR Title follows the convention of fix: mismatched arguments in dataaccess._determine_filename

 Commit has unit tests (To be added in a separate PR)

### DIFF
--- a/mslib/mswms/dataaccess.py
+++ b/mslib/mswms/dataaccess.py
@@ -69,7 +69,8 @@ class NWPDataAccess(metaclass=ABCMeta):
         """
         pass
 
-    def have_data(self, variable, vartype, init_time, valid_time):
+    def have_data(self, variable, vartype, init_time, valid_time, reload=False):
+
         """
         Checks whether a file with data for the specified variable,
         type and times is known. This does not trigger a search for
--- a/tests/_test_mswms/test_dataaccess.py
+++ b/tests/_test_mswms/test_dataaccess.py
@@ -89,6 +89,14 @@ class Test_DefaultDataAccess:
                     datetime(2012, 10, 17, 18, 0),
                     datetime(2012, 10, 18, 12, 0),
                     datetime(2012, 10, 19, 0, 0)])
+    def test_have_data_with_reload(self):
+        result = self.dut.have_data(
+            "air_pressure", "ml",
+            datetime(2012, 10, 17, 12, 0),
+            datetime(2012, 10, 17, 18, 0),
+            reload=False
+        )
+        assert isinstance(result, bool)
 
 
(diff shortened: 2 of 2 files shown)


mslib/mswms/dataaccess.py
```diff
@@ -69,7 +69,8 @@ def setup(self):
         """
         pass
 
-    def have_data(self, variable, vartype, init_time, valid_time):
+    def have_data(self, variable, vartype, init_time, valid_time, reload=False):
```
matrss, line 72: You are changing the wrong method signature here.

tests/_test_mswms/test_dataaccess.py
```diff
@@ -89,6 +89,14 @@ def test_get_all_valid_times(self):
                     datetime(2012, 10, 17, 18, 0),
                     datetime(2012, 10, 18, 12, 0),
                     datetime(2012, 10, 19, 0, 0)])
+    def test_have_data_with_reload(self):
```
matrss, line 92: Formatting.

matrss (changes requested):
This PR does not address what is described in the issue nor does the change make sense at all. The issue is also already taken by someone else.

tests/_test_mswms/test_dataaccess.py
praharsh777: Thanks for the clarification, I realize now that I misunderstood the issue and made changes to the wrong method. I appreciate the feedback and will make sure to review the context and issues more carefully next time. I'm still learning and truly welcome guidance like this to help me improve. Looking forward to contributing more meaningfully!

ReimarBauer:
If the work is continued, we can reopen it.

Issue #2716, opened 2025-03-19
mscolab_chat, timeout tuple problem


ReimarBauer:
That happens very rare. 
```
Traceback (most recent call last):
  File "/Users/reimarbauer/PycharmProjects/MSS/mslib/msui/mscolab.py", line 1068, in operation_options_handler
    self.open_chat_window()
  File "/Users/reimarbauer/PycharmProjects/MSS/mslib/msui/mscolab.py", line 92, in wrapper
    result = func(self, *args, **vargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/reimarbauer/PycharmProjects/MSS/mslib/msui/mscolab.py", line 1087, in open_chat_window
    self.chat_window = mc.MSColabChatWindow(
                       ^^^^^^^^^^^^^^^^^^^^^
  File "/Users/reimarbauer/PycharmProjects/MSS/mslib/msui/mscolab_chat.py", line 126, in __init__
    self.load_users()
  File "/Users/reimarbauer/PycharmProjects/MSS/mslib/msui/mscolab_chat.py", line 340, in load_users
    users_response = requests.get(users_url, data=data, timeout=tuple(config_loader(dataset="MSCOLAB_timeout")))
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/reimarbauer/PycharmProjects/MSS/.pixi/envs/dev/lib/python3.11/site-packages/requests/api.py", line 73, in get
    return request("get", url, params=params, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/reimarbauer/PycharmProjects/MSS/.pixi/envs/dev/lib/python3.11/site-packages/requests/api.py", line 59, in request
    return session.request(method=method, url=url, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/reimarbauer/PycharmProjects/MSS/.pixi/envs/dev/lib/python3.11/site-packages/requests/sessions.py", line 589, in request
    resp = self.send(prep, **send_kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #2790, opened 2025-04-09
fix: mscolab_chat, timeout tuple problem

levi178u:
fixes: #2716

What are the changes ?

- Updated default config to provide timeout as a tuple instead of a list.
- Modified validation logic to expect tuples for timeout.
- Enhanced merge_dict() to convert user-provided lists to tuples during merge.
- Ensured compatibility with requests.get() by enforcing correct format at config level.


levi178u:
Mentors plz review the PR

ReimarBauer (changes requested):
please look on what you pushed.

which changes are related to the issue?

Could you also add a unit test which shows the problem happening and that your suggestion does fix it.

levi178u:
> please look on what you pushed.
> 
> which changes are related to the issue?
> 
> Could you also add a unit test which shows the problem happening and that your suggestion does fix it.

Basically the `requests.get()` funct is getting `([2, 10], [2, 10])` but it expects it to be `([2,10]) ` i.e `(connect_timeout, read_timeout)`. So in the `config.py` I just made sure the above is done and I tested it via `_test_utils `'s `test_config` pytest , where all cases passed and `pytest.log` was auto-generated.

levi178u:
I also wanna know what other changes should I do so that this timeout config. issue is resolved.

ReimarBauer:
the idea for getting knowledge if it is complete is to write first a test which shows the problem. Then provide the fix, which does not anymore have a failing test

ReimarBauer:
look on the changes url yourselft. What is irrelevant for the issue?

Bildschirmfoto 2025-04-10 um 10 49 45

levi178u:
> look on the changes url yourselft. What is irrelevant for the issue?
> 
> Bildschirmfoto 2025-04-10 um 10 49 45

Yes in the changes section the `"MSCOLAB_timeout": ([2, 10])` and     `MSCOLAB_timeout = (2, 10)` is only relevent as I'm solving the tuple issue and i had done changes in the `config.py` under `utils`, also i added `fs` in `pixi.toml` coz I ran the pytest for this using this: 
```
 pixi install
pixi run -e dev pytest tests/_test_utils
```

image
Here's the test log, once review this @ReimarBauer

levi178u:
> I also wanna know what other changes should I do so that this timeout config. issue is resolved.

Ok I got it, I'll write some test cases against the changes,sure.

pixi.toml
```diff
@@ -26,6 +26,7 @@ flask-socketio = ">=5.1.0"
 flask-sqlalchemy = ">=3.0.0"
 flask-wtf = "*"
 fs_filepicker = "*"
+fs = "*"
```
ReimarBauer: this is inherited by fs_filepicker, what is the reason to add it?

pytest.log
```diff
@@ -0,0 +1,141 @@
+2025-04-09 19:09:01 DEBUG Failed to pyproj from mpl_toolkits.basemap
```
ReimarBauer: this file is not wanted to be committed

pixi.toml
levi178u: Yes your right, I'll remove it, since I just added it so as to check the pytest suite as the file had `import fs`, no need for it to be committed as such

pytest.log
levi178u: Yes i'll make changes

ReimarBauer (changes requested):
The test case does not accurately reflect the scenario related to this issue. The proposed fix breaks the config editor. Test module names must start with "test_".

With only 70% test coverage, we need to exercise caution when modifying code outside of known issue areas. The best practice is to create a failing test case that demonstrates the problem before implementing a fix.  This approach will also contribute to increasing our overall test coverage.

https://github.com/user-attachments/assets/18c68c3a-be33-44a9-8ebb-b369c09d5324

levi178u:
Ok @ReimarBauer  I have prepared a test case file covering more cases and edge cases too, can I push it for review. Also I have asked few doubts in slack thread, do check them. Cases such as negative nums, other ds besides tuple, empty ds, float nums, int nums, etc.. i have added cases considering them too

ReimarBauer:
> #2716

The issue is not assigned to you. But when you want to work on it reproduce first by a test the issue description.

levi178u:
> > #2716
> 
> The issue is not assigned to you. But when you want to work on it reproduce first by a test the issue description.

Ok , here I have added the test file (`test_mscolab_timeout.py`) , pls review the same, also what changes should I do in the test file pls suggest

ReimarBauer (commented):
A solution has to succeed this test:

https://github.com/Open-MSS/MSS/blob/stable/tests/_test_msui/test_mscolab.py#L301

ReimarBauer:
solved by https://github.com/Open-MSS/MSS/pull/2815

PR #2792, opened 2025-04-11
sets a tmp dir for wms_cache

### LINKED ISSUE #2546: test_activate_operation_updates_active_users needs a review
I see this sometimes on my local tests
```
[gw5] [100%] PASSED tests/_test_msui/test_mscolab.py::Test_Mscolab::test_activate_operation_updates_active_users 

Exception ignored in: <function WMSControlWidget.__del__ at 0x718a67ac9120>
  File "/home/reimar/MAIN/MSS/mslib/msui/wms_control.py", line 550, in __del__
    self.service_cache()
  File "/home/reimar/MAIN/MSS/mslib/msui/wms_control.py", line 1455, in service_cache
    files = [os.path.join(self.wms_cache, f) for f in os.listdir(self.wms_cache)]
                                                      ^^^^^^^^^^^^^^^^^^^^^^^^^^
FileNotFoundError: [Errno 2] Datei oder Verzeichnis nicht gefunden: '/tmp/tmpnodr4uwv/msui/wms_cache/'
```

### LINKED ISSUE #2793: testing wms_cache is not used
> We have to set wms_cache
> https://github.com/Open-MSS/MSS/blob/stable/tests/_test_msui/test_wms_control.py#L500
> 
> similiar as here
> https://github.com/Open-MSS/MSS/blob/stable/tests/_test_msui/test_wms_control.py#L69
> 
> On the MBP I got today
> ```
> Failed: An unhandled message box popped up during your test!
> E       PyQt5.QtWidgets.QMessageBox.critical(): [call(<mslib.msui.wms_control.HSecWMSControlWidget object at 0x34be62950>, 'Web Map Service', "ERROR: Cannot delete file '/Users/reimarbauer/Library/Caches/msui/wms_cache/6a69a349092204e79ecfb3fabe12998a.png'. (<class 'FileNotFoundError'>: [Errno 2] No such file or directory: '/Users/reimarbauer/Library/Caches/msui/wms_cache/6a69a349092204e79ecfb3fabe12998a.png')")]
> ```
> 
> that looks like a test does not use a tmp dir. It is not clear yet if it is the same problem as in #2546 

 _Originally posted by @ReimarBauer in [#2546](https://github.com/Open-MSS/MSS/issues/2546#issuecomment-2773392885)_

ReimarBauer:
**Purpose of PR?**:

Fixes #2793

To ensure tests don't write to the user space, a wms_cache must be configured.

### DIFF
--- a/tests/_test_msui/test_wms_control.py
+++ b/tests/_test_msui/test_wms_control.py
@@ -496,8 +496,12 @@ class TestWMSControlWidgetSetupSimple:
 
     @pytest.fixture(autouse=True)
     def setup(self, qtbot):
+        wc.WMS_SERVICE_CACHE = {}
         self.view = HSecViewMockup()
-        self.window = wc.HSecWMSControlWidget(view=self.view)
+        self.tempdir = tempfile.mkdtemp()
+        if not os.path.exists(self.tempdir):
+            os.mkdir(self.tempdir)
+        self.window = wc.HSecWMSControlWidget(view=self.view, wms_cache=self.tempdir)
         self.window.show()
 
         # Remove all previous cached URLs
@@ -507,6 +511,10 @@ class TestWMSControlWidgetSetupSimple:
 
         yield
         self.window.hide()
+
+    def test_wms_cache_tmp(self):
+        assert self.window.wms_cache is not None
+        assert "tmp" in self.window.wms_cache
 
     def test_xml(self):
         testxml = self.xml.format("", self.srs_base, self.dimext_time + self.dimext_inittime + self.dimext_elevation)
--- a/tests/_test_msui/test_wms_control.py
+++ b/tests/_test_msui/test_wms_control.py
@@ -495,10 +495,10 @@ class TestWMSControlWidgetSetupSimple:
         <Extent name="ELEVATION" default="900.0"> 500.0,600.0,700.0,900.0 </Extent>"""
 
     @pytest.fixture(autouse=True)
-    def setup(self, tmp_path_factory, qtbot):
+    def setup(self, tmp_path, qtbot):
         wc.WMS_SERVICE_CACHE = {}
         self.view = HSecViewMockup()
-        self.tempdir = tmp_path_factory.mktemp("TMP_WMS")
+        self.tempdir = tmp_path
         self.window = wc.HSecWMSControlWidget(view=self.view, wms_cache=self.tempdir)
         self.window.show()
 
--- a/tests/_test_msui/test_wms_control.py
+++ b/tests/_test_msui/test_wms_control.py
@@ -512,7 +512,8 @@ class TestWMSControlWidgetSetupSimple:
 
     def test_wms_cache_tmp(self):
         assert self.window.wms_cache is not None
-        assert "TMP_WMS" in self.window.wms_cache
+        # wms_cache has an ending path separator
+        assert str(self.tempdir) == self.window.wms_cache[:-1]
 
     def test_xml(self):
         testxml = self.xml.format("", self.srs_base, self.dimext_time + self.dimext_inittime + self.dimext_elevation)
--- a/tests/_test_msui/test_wms_control.py
+++ b/tests/_test_msui/test_wms_control.py
@@ -495,12 +495,10 @@ class TestWMSControlWidgetSetupSimple:
         <Extent name="ELEVATION" default="900.0"> 500.0,600.0,700.0,900.0 </Extent>"""
 
     @pytest.fixture(autouse=True)
-    def setup(self, qtbot):
+    def setup(self, tmp_path_factory, qtbot):
         wc.WMS_SERVICE_CACHE = {}
         self.view = HSecViewMockup()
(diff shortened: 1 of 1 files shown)


tests/_test_msui/test_wms_control.py
```diff
@@ -496,8 +496,12 @@ class TestWMSControlWidgetSetupSimple:
 
     @pytest.fixture(autouse=True)
     def setup(self, qtbot):
+        wc.WMS_SERVICE_CACHE = {}
         self.view = HSecViewMockup()
-        self.window = wc.HSecWMSControlWidget(view=self.view)
+        self.tempdir = tempfile.mkdtemp()
+        if not os.path.exists(self.tempdir):
+            os.mkdir(self.tempdir)
```
matrss: This is unnecessary.

tests/_test_msui/test_wms_control.py
```diff
@@ -496,8 +496,12 @@ class TestWMSControlWidgetSetupSimple:
 
     @pytest.fixture(autouse=True)
     def setup(self, qtbot):
+        wc.WMS_SERVICE_CACHE = {}
         self.view = HSecViewMockup()
-        self.window = wc.HSecWMSControlWidget(view=self.view)
+        self.tempdir = tempfile.mkdtemp()
```
matrss: Cleanup is missing. I'd prefer we'd just use pytest's builtin temp_path fixture instead, though.

matrss (changes requested):

tests/_test_msui/test_wms_control.py
ReimarBauer: much better

the other similiiar usages can be refactored afterwads by a new issue.

tests/_test_msui/test_wms_control.py
```diff
@@ -495,12 +495,10 @@ class TestWMSControlWidgetSetupSimple:
         <Extent name="ELEVATION" default="900.0"> 500.0,600.0,700.0,900.0 </Extent>"""
 
     @pytest.fixture(autouse=True)
-    def setup(self, qtbot):
+    def setup(self, tmp_path_factory, qtbot):
```
matrss: Why `tmp_path_factory` instead of `tmp_path`?

ReimarBauer: I found this problem by investigating timeouts 

https://pytest-with-eric.com/pytest-best-practices/pytest-tmp-path/#How-Many-Test-Runs-Does-Pytest-keep-in-tmp-path

reading about that the tmp_path will be retained I had the impression that tmp_path_factory may be better.
I liked also that I have not to deal with path separators for adding a name I can lookup.

matrss: > reading about that the tmp_path will be retained I had the impression that tmp_path_factory may be better.

I'd expect both to be the same in this regard. Only the last 3 temporary directories are retained by default.

> I liked also that I have not to deal with path separators for adding a name I can lookup.

If you want to check that the wms_cache is set to the temporary directory, then check that they are equal. Checking for a known part expected in the path is just a poor proxy for that.

ReimarBauer: after this is fixed i have to lookup the md5_filename again. This is also based on the dir of wms_cache and gets a wrong path currently

matrss (commented):
I'd still prefer using the simpler tmp_path.

ReimarBauer:
For me it is to simple currently. Maybe we can postpone which it becomes later. There are some more parts where tmpdirs not used. With a name I can destinguish if some problems solved in a way I have not expected.

matrss:
Sorry, I don't understand. If you are concerned about being able to associate a temporary directory with a test function, the paths created by the `tmp_path` factory are named after the test function which requested the fixture. This is strictly better than naming it ourselves since there can't really be a name collision. With what you are currently doing the association isn't as obvious and collisions could happen.

ReimarBauer:
> Sorry, I don't understand. If you are concerned about being able to associate a temporary directory with a test function, the paths created by the `tmp_path` factory are named after the test function which requested the fixture. This is strictly better than naming it ourselves since there can't really be a name collision. With what you are currently doing the association isn't as obvious and collisions could happen.

it creates dirs like 
/private/var/folders/4c/f6cfpyvn4_xb9ln54y6vnh5w0000gn/T/pytest-of-reimarbauer/pytest-3/test_xml_time_init_period0

Ok I can go over the test name, and improve some of the names later.

tests/_test_msui/test_wms_control.py
```diff
@@ -495,9 +495,11 @@ class TestWMSControlWidgetSetupSimple:
         <Extent name="ELEVATION" default="900.0"> 500.0,600.0,700.0,900.0 </Extent>"""
 
     @pytest.fixture(autouse=True)
-    def setup(self, qtbot):
+    def setup(self, tmp_path, qtbot):
+        wc.WMS_SERVICE_CACHE = {}
```
matrss, line 499: Having to reset this here is a good indicator that this variable shouldn't be global.

matrss (approved):

PR #2794, opened 2025-04-11
test description of the issue for wrong config values

ReimarBauer:
**Purpose of PR?**:

part of #2716

this is a test to describe the problem, with xfail until it is worked on.

### DIFF
--- a/tests/_test_msui/test_mscolab.py
+++ b/tests/_test_msui/test_mscolab.py
@@ -298,8 +298,8 @@ class Test_Mscolab:
         # close all hanging operation option windows
         self.window.mscolab.close_external_windows()
 
+    @pytest.mark.xfail(reason="https://github.com/Open-MSS/MSS/issues/2716", strict=True)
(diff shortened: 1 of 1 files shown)


tests/_test_msui/test_mscolab.py
```diff
@@ -298,6 +298,14 @@ def setup(self, qtbot, mscolab_app, mscolab_server):
         # close all hanging operation option windows
         self.window.mscolab.close_external_windows()
 
+    def test_modify_mscolab_timeout(self, qtbot):
+        pytest.xfail("https://github.com/Open-MSS/MSS/issues/2716")
```
matrss: This is different from using the xfail mark decorator in that it doesn't run the rest of the test. But the test should run so that we can catch if it starts to pass. The `strict` option for xfail also needs to be set so that we will notice when it starts to pass for some reason.

ReimarBauer: thx for that hint and changed :)

matrss (approved):

Issue #2736, opened 2025-03-26
Use immutable references for external GitHub Actions


matrss:
Referencing third-party actions by version tag can be dangerous if that actions repository is compromised, since git tags are mutable and an attacker can therefore change what gets executed. See e.g. this for a recent example of this kind of attack: https://alexwlchan.net/2025/github-actions-audit/.

We should replace all mutable references with immutable commit shas. We can still use dependabot for updates as it recognizes this approach using a syntax like this:
```
       uses: actions/checkout@5a4ac9002d0be2fb38bd78e4b4dbde5606d7042f # v2.3.4
```

levi178u:
@matrss i got the commit SHA key for the `actions/checkout@v4` but how to obtain the same for `prefix-dev/setup-pixi@v0.8.4`, can u pls say from where I can get the same.

levi178u:
@matrss I have changed the mutable `actions/checkout@v4` and `prefix-dev/setup-pixi@v.0.8.4` with their respective immutable commit SHA keys. Now how should I test them and what other changes should I do to resolve this issue

ReimarBauer:
What have we done that we don't pin by hash?
Seems also using hash pinning can write the version string as comment 

https://github.com/dependabot/dependabot-core/issues/4691

ReimarBauer:
This needs a review:
https://stacklok.com/blog/automating-security-for-github-actions-in-minder

ReimarBauer:
looks good to me
https://github.com/stacklok/frizbee#installation

We maybe can prepare an action based on this

ReimarBauer:
link to:
https://github.com/prefix-dev/pixi/pull/3369

ReimarBauer:
Because this action receives frequent updates, it often invalidates manually added checksums
https://github.com/marketplace/actions/setup-pixi

This issue gets only closed after we automated checksum updates

matrss:
> This issue gets only closed after we automated checksum updates

AFAIU they are already automated by dependabot.

ReimarBauer:
When I use frizbee locally it updates to the commit hash. I tried to add that.  
https://github.com/stacklok/frizbee-action

I don't want to keep manually added PR for having an immutable reference. An automated PR is fine, or dependabot does it similiar in one shot.

matrss:
[...]

PR #2802, opened 2025-04-15
Ensure SHA Pinned Actions

ReimarBauer:
**Purpose of PR?**:

Part of #2736 

seen that it is used by prefix-dev initially now with a dry-run
https://github.com/marketplace/actions/ensure-sha-pinned-actions


ReimarBauer:
related: https://github.com/prefix-dev/pixi/pull/3369

.github/workflows/lint.yml
```diff
@@ -18,7 +18,7 @@ jobs:
     timeout-minutes: 10
     steps:
     - uses: actions/checkout@v4
```
ReimarBauer: I excluded that to enable testing.

matrss (dismissed):

.github/workflows/lint.yml
```diff
@@ -68,3 +68,14 @@ jobs:
...
+  ensure-sha-pinned-actions:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4
+      - name: Ensure SHA pinned actions
+        uses: zgosalvez/github-actions-ensure-sha-pinned-actions@4830be28ce81da52ec70d65c552a7403821d98d4 # v3
+        with:
+          allowlist: |
+            Open-MSS/
+          dry_run: 'false'
```
matrss:
```suggestion
```

matrss: false is the default already

matrss: Actually, I am surprised this isn't a footgun: 'false' is a string, and non-empty strings are often evaluated to true in logical expressions. In this case the string 'false' is properly interpreted as false though.

matrss (approved):

.github/workflows/lint.yml
ReimarBauer: Also for true it was with 'true' correct interpreted, but that would been expected behaviour. 

https://github.com/Open-MSS/MSS/pull/2802/commits/0532a4a2a9ecddb2d25220b3469f7094039e24e9

PR #2806, opened 2025-04-18
Refactor contour styling to fix MatplotlibDeprecationWarning

### LINKED ISSUE #2728: MatplotlibDeprecationWarning for matplotlib-3.10.0
mpl_vsec_styles.py:91: MatplotlibDeprecationWarning: The **collections** attribute was deprecated in Matplotlib 3.8 and will be removed in 3.10.

This needs a review for other changes and a refactoring.

levi178u:
**Purpose of PR?**:

Fixes #2728 
Changes done:
- Fixed deprecated use of `.collections` in contour plotting for matplotlib>=3.8 to prevent `MatplotlibDeprecationWarning`.
- Updated `_plot_style()` to use `ContourSet.set_path_effects()` directly.
- updated contour styling files of styles and plots.
- Verified existing plot behavior remains unchanged by test cases passing.

### DIFF
--- a/docs/samples/config/mswms/mss_chem_plots.py
+++ b/docs/samples/config/mswms/mss_chem_plots.py
@@ -93,9 +93,10 @@ class HS_MSSChemStyle(MPLBasemapHorizontalSectionStyle):
                                colors=cont_colour, linestyles=cont_style, linewidths=cont_lw)
             cs_pv_lab = ax.clabel(cs_pv, colors=cont_label_colour, fmt='%i')
             if pe:
-                plt.setp(cs_pv.collections, path_effects=[patheffects.withStroke(linewidth=cont_lw + 2,
-                                                                                 foreground="w")])
-                plt.setp(cs_pv_lab, path_effects=[patheffects.withStroke(linewidth=1, foreground="w")])
+                for i in cs_pv.collections:
+                    i.set_path_effects([patheffects.withStroke(linewidth=cont_lw + 2, foreground="w")])
+                for label in cs_pv_lab:
+                    label.set_path_effects([patheffects.withStroke(linewidth=1, foreground="w")])
 
         # define position of the colorbar and the orientation of the ticks
         if self.crs.lower() == "epsg:77774020":
--- a/docs/samples/config/mswms/mss_chem_plots.py
+++ b/docs/samples/config/mswms/mss_chem_plots.py
@@ -93,8 +93,7 @@ class HS_MSSChemStyle(MPLBasemapHorizontalSectionStyle):
                                colors=cont_colour, linestyles=cont_style, linewidths=cont_lw)
             cs_pv_lab = ax.clabel(cs_pv, colors=cont_label_colour, fmt='%i')
             if pe:
-                for i in cs_pv.collections:
-                    i.set_path_effects([patheffects.withStroke(linewidth=cont_lw + 2, foreground="w")])
+                cs_pv.set_path_effects([patheffects.withStroke(linewidth=cont_lw + 2, foreground="w")])
                 for label in cs_pv_lab:
                     label.set_path_effects([patheffects.withStroke(linewidth=1, foreground="w")])
 
--- a/docs/samples/config/mswms/mss_chem_plots.py
+++ b/docs/samples/config/mswms/mss_chem_plots.py
@@ -253,11 +253,11 @@ class VS_MSSChemStyle(AbstractVerticalSectionStyle):
             else:
                 cs_pv = ax.contour(curtain_lat, curtain_p, self.data[cont_data], cont_levels,
                                    colors=cont_colour, linestyles=cont_style, linewidths=cont_lw)
-                plt.setp(cs_pv.collections,
-                         path_effects=[patheffects.withStroke(linewidth=cont_lw + 2, foreground="w")])
                 cs_pv_lab = ax.clabel(cs_pv, colors=cont_label_colour, fontsize=8, fmt='%i')
-                plt.setp(cs_pv_lab, path_effects=[patheffects.withStroke(linewidth=1, foreground="w")])
-
+                if pe:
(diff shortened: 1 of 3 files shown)


levi178u:
@ReimarBauer pls review this PR

ReimarBauer:
By which pytest do you verify existing plot behavior remains unchanged?

levi178u:
> By which pytest do you verify existing plot behavior remains unchanged?

By `test_mss_plot_driver.py` in `_test_mswms` I check if it passed all the test cases for the cases of plots in that way i verified as it passed all test cases

matrss:
Your proposed changes still use the collections attribute, which as stated in the issue is deprecated. How is this supposed to fix the issue?

levi178u:
> Your proposed changes still use the collections attribute, which as stated in the issue is deprecated. How is this supposed to fix the issue?

yes u r right, my aim was to suppress the deprecation warning by reducing the dependence on `.collections` , here i have applied styling using `path_effects` at plot creation time. i had read the docs and found that currently matplotlib doesnt support passing` path_effects` directly to `plt.contour()` without accessing `.collections`, so i did this

levi178u:
Mentors, we can follow an alternate method in which we don't rely on `.collections` completely. Here i referred that https://matplotlib.org/stable/api/collections_api.html#matplotlib.collections.PathCollection , we can use `Linecollection` or `Pathcollection` but it has a limitation that we can’t pass `path_effects` directly to `ax.contour()` or `plt.contour()` in this scenario.

matrss:
I don't understand why you focus on not being able to pass `path_effects` to `.contour`, there is no need to do that. Please read the documentation for this API change in matplotlib: https://matplotlib.org/stable/api/prev_api_changes/api_changes_3.8.0.html#contourset-is-now-a-single-collection. Following this it should now be possible to call the `.set_path_effects` method on the `ContourSet` object directly.

levi178u:
> I don't understand why you focus on not being able to pass `path_effects` to `.contour`, there is no need to do that. Please read the documentation for this API change in matplotlib: https://matplotlib.org/stable/api/prev_api_changes/api_changes_3.8.0.html#contourset-is-now-a-single-collection. Following this it should now be possible to call the `.set_path_effects` method on the `ContourSet` object directly.

Thnx @matrss, I missed this part , I'll make the changes for the same by direct calling instead of passing,

levi178u:
@matrss pls review the changes

levi178u:
@matrss pls review the changes

docs/samples/config/mswms/mss_chem_plots.py
```diff
@@ -253,11 +253,11 @@ def _plot_style(self):
             else:
                 cs_pv = ax.contour(curtain_lat, curtain_p, self.data[cont_data], cont_levels,
                                    colors=cont_colour, linestyles=cont_style, linewidths=cont_lw)
-                plt.setp(cs_pv.collections,
-                         path_effects=[patheffects.withStroke(linewidth=cont_lw + 2, foreground="w")])
                 cs_pv_lab = ax.clabel(cs_pv, colors=cont_label_colour, fontsize=8, fmt='%i')
-                plt.setp(cs_pv_lab, path_effects=[patheffects.withStroke(linewidth=1, foreground="w")])
-
+                if pe:
```
matrss, line 256: This if wasn't here before.

docs/samples/config/mswms/mss_chem_plots.py
```diff
@@ -93,9 +93,9 @@ def _plot_style(self):
                                colors=cont_colour, linestyles=cont_style, linewidths=cont_lw)
             cs_pv_lab = ax.clabel(cs_pv, colors=cont_label_colour, fmt='%i')
             if pe:
-                plt.setp(cs_pv.collections, path_effects=[patheffects.withStroke(linewidth=cont_lw + 2,
-                                                                                 foreground="w")])
-                plt.setp(cs_pv_lab, path_effects=[patheffects.withStroke(linewidth=1, foreground="w")])
```
matrss, line 98: Why are you changing these clabel related lines as well? They do not touch the deprecated collections attribute.

levi178u: True, i added them to make it concise but yes it's not required here . I'll revert the change

docs/samples/config/mswms/mss_chem_plots.py
levi178u: Yes this `if pe: `condition wasn't there, I added it to apply the `set_path_effects()` only on those contours or labels where p`e= true`, also I saw examples from other parts of the code so applied.   Should I revert this ?

matrss: Hmm, right, I initially didn't see where `pe` was coming from. I agree that it makes sense to add, if it is also correct in those other places.

docs/samples/config/mswms/mss_chem_plots.py
levi178u: Also here can u check if this part needs to be reverted or not as I iterated the label in `cs_pv_lab` to apply the path_effects in each

matrss: As documented [here](https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.setp.html#:~:text=When%20setting%20properties%2C%20all%20artists%20are%20affected), setp also assigns the provided properties on all artists if its first argument is a list. They are functionally identical.

levi178u: @matrss I have done the change pls review

levi178u:
@matrss pls review the changes

levi178u:
@matrss ,@joernu76 pls review

levi178u:
@Mentors need review for this PR pls

matrss (approved):

PR #2811, opened 2025-04-24
Reintroduce coverage upload to coveralls

### LINKED ISSUE #2780: coveralls not updating
seems we don't get updates

matrss:
**Purpose of PR?**:

Fixes #2780.

**Checklist:**
- [ ] Commit has unit tests

### DIFF
--- a/.github/workflows/testing-all-oses.yml
+++ b/.github/workflows/testing-all-oses.yml
@@ -29,3 +29,21 @@ jobs:
     - name: Run tests
       timeout-minutes: 40
       run: pixi run -e dev env QT_QPA_PLATFORM=offscreen pytest -v -n logical --durations=20 --cov=mslib tests
+    - run: pixi run -e dev coverage xml
+    - name: Send coverage to Coveralls (parallel)
+      uses: coverallsapp/github-action@648a8eb78e6d50909eff900e4ec85cab4524a45b # v2.3.6
+      with:
+        debug: true
+        parallel: true
+        format: cobertura
+        flag-name: ${{ join(matrix.*, ' - ') }}
+
+  finish:
+    needs: test
+    runs-on: ubuntu-latest
+    if: ${{ always() }}
+    steps:
+      - name: Close parallel build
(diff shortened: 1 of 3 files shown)


.github/workflows/testing-all-oses.yml
```diff
@@ -29,3 +29,21 @@ jobs:
     - name: Run tests
       timeout-minutes: 40
       run: pixi run -e dev env QT_QPA_PLATFORM=offscreen pytest -v -n logical --durations=20 --cov=mslib tests
+    - run: pixi run -e dev coverage xml
+    - name: Send coverage to Coveralls (parallel)
+      uses: coverallsapp/github-action@648a8eb78e6d50909eff900e4ec85cab4524a45b # v2.3.6
+      with:
+        debug: true
+        parallel: true
```
ReimarBauer, line 37: do we need to take care for a combine too?

https://coverage.readthedocs.io/en/latest/cmd.html#combining-data-files-coverage-combine

setup.cfg
```diff
@@ -28,6 +28,7 @@ test = pytest
 
 [coverage:run]
 branch = True
+relative_files = True
```
ReimarBauer, line 31: makes sense. I just don't know if we have a bunch of .coverage.id files which we have to combine

ReimarBauer (commented):
This looks principle better than what we had before. A question added.

.github/workflows/testing-all-oses.yml
matrss: I don't think so, coveralls should be responsible to visualize the coverage from the different test runs in an appropriate way. I'd say we should wait and see how it works out when the macOS tests get reliable, and if there is anything we would like to have changed about how coverage is reported then.

setup.cfg
matrss: This was a  config option I read about while debugging the issue that coveralls said the coverage was empty. I don't this is strictly necessary for it to work, but I felt like it generally makes more sense if coverage just used relative paths.

ReimarBauer (approved):
thx

PR #2814, opened 2025-05-06
Add config & CLI support for .ftml path in mssautoplot

### LINKED ISSUE #2714: autoplot ftml files need the path and file stored
currently the autoplot requires the ftml file in the dir where the mssautoplot was called.
We have to change this. the mssautoplot.json has to keep the path + file 

Additional the cli mssautoplot can have an option to overewrite the path.

annapurna-gupta:
Fixes #2714


mslib/msui/constants.py
```diff
@@ -67,6 +67,7 @@
 
 # ToDo refactor to a function
 MSS_AUTOPLOT = os.getenv('MSS_AUTOPLOT', os.path.join(MSUI_CONFIG_PATH, "mssautoplot.json"))
+DEFAULT_FTML_PATH = os.getenv('MSS_FTML_PATH', os.path.join(MSUI_CONFIG_PATH, "example_flight.ftml"))
```
ReimarBauer: please use the users config var mss_dir, and don't create a new constant

or investigate how it stores the first used dir.
e.g. when you store in a ftml dir, it uses that dir for e.g. an import

it should not behave different to the common user interface.

look for self.last_save_directory

mslib/utils/mssautoplot.py
```diff
@@ -234,15 +235,14 @@ def __init__(self, cpath, msc_url=None, msc_auth_password=None, username=None, p
         if filename != "" and filename == flight:
             self.read_operation(flight, msc_url, msc_auth_password, username, password)
         elif filename != "":
-            # Todo add the dir to the file in the mssautoplot.json
-            dirpath = "./"
-            file_path = os.path.join(dirpath, filename)
+            file_path = DEFAULT_FTML_PATH
```
ReimarBauer: use the users config of mss_dir

ReimarBauer (changes requested):
see comment

ReimarBauer:
Let’s also make sure we save the file path in mssautoplot.json. We can just use self.last_save_directory for this; no need to create a new default.

ReimarBauer:
the autoplot dockingwidget still does not save a path into the json configuration for the filenames, see
https://github.com/Open-MSS/MSS/blob/develop/mslib/msui/autoplot_dockwidget.py#L318
The filenames can be stored on different directories. In the json file we need the file location, not only the names.

annapurna-gupta:
> the autoplot dockingwidget still does not save a path into the json configuration for the filenames, see
> https://github.com/Open-MSS/MSS/blob/develop/mslib/msui/autoplot_dockwidget.py#L318
> The filenames can be stored on different directories. In the json file we need the file location, not only the names.

 I'm currently working on the CLI version of autoplot (not the GUI-based autoplot_dockwidget). My changes are focused on command-line behavior, so they won't affect or fix the GUI behavior related to JSON path saving? or am i missing something?

annapurna-gupta:
Hii @ReimarBauer
I’ve added the path for the .ftml file like this:

`if filename and not os.path.isabs(filename): default_dir = os.path.join(os.path.expanduser("~"), ".config", "msui") filename = os.path.join(default_dir, filename)`

But I think it’s not working as expected — the file still isn’t being found. Could you please help me check if I’m missing something?

ReimarBauer:
> Hii @ReimarBauer I’ve added the path for the .ftml file like this:
> 
> `if filename and not os.path.isabs(filename): default_dir = os.path.join(os.path.expanduser("~"), ".config", "msui") filename = os.path.join(default_dir, filename)`
> 
> But I think it’s not working as expected — the file still isn’t being found. Could you please help me check if I’m missing something?

When configuring the new mssautoplot.json, we should mirror the behavior of existing configurations regarding file paths. Specifically, the path used to store an FTML file should be relative to the user's designated storage location. For example, if a user saves an FTML file as /Users/reimarbauer/FTML/example.ftml, the new configuration should reflect this same path.  This consistency is important because sharing my mssautoplot.json necessitates also sharing the corresponding FTML file – and knowing the storage location ensures I can easily identify the correct one among potentially many examples.

Please implement the functionality for the autoplot docking widget https://github.com/Open-MSS/MSS/blob/develop/mslib/msui/autoplot_dockwidget.py to save the FTML path/file before proceeding with CLI improvements.

mslib/msui/autoplot_dockwidget.py
```diff
@@ -302,7 +302,9 @@ def add_to_treewidget(self, parent, parent2, config_settings, treewidget, flight
                 flight = ""
             else:
                 if filename != parent2.mscolab.active_operation_name:
-                    filename += ".ftml"
+                    filename = os.path.join(const.MSUI_CONFIG_PATH, filename + ".ftml")
```
ReimarBauer: when you do "save as" the used directory is stored at 
https://github.com/Open-MSS/MSS/blob/develop/mslib/msui/msui_mainwindow.py#L946

self.last_save_directory

You need to enable access to the var and use it then. We need the dir where the file is stored by File - Save_As

ReimarBauer: you can pass it into the topview and from there to the dockingwidget, see https://github.com/Open-MSS/MSS/blob/develop/mslib/msui/msui_mainwindow.py#L1029

ReimarBauer (changes requested):
see comment

annapurna-gupta:
Hii @ReimarBauer 
I noticed that running:
**mssautoplot --cpath ~/.config/msui/mssautoplot.json**
_gives resource 'example_flight.ftml' not found_
but this works:
**mssautoplot --cpath ~/.config/msui/mssautoplot.json --ftml-path /home/anuu/.config/msui/example_flight.ftml**

ReimarBauer:
> Hii @ReimarBauer I noticed that running: **mssautoplot --cpath ~/.config/msui/mssautoplot.json** _gives resource 'example_flight.ftml' not found_ but this works: **mssautoplot --cpath ~/.config/msui/mssautoplot.json --ftml-path /home/anuu/.config/msui/example_flight.ftml**

Please share the section of your mssautoplot.json  for
```
"automated_plotting_flights": [],
"automated_plotting_hsecs": [],
"automated_plotting_vsecs": [],
"automated_plotting_lsecs": []
```

PR #2815, opened 2025-05-11
fixed MSCOLAB_timeout definition by list_operation_structure

### LINKED ISSUE #2716: mscolab_chat, timeout tuple problem

### LINKED ISSUE #2819: modify_config_file needs to have the configurations/editor logic
To ensure consistency, we need to update `modify_config_file` to use `MSUIDefaultConfig.fixed_dict_options` and `MSUIDefaultConfig.fixed_list_options`, mirroring the logic used in the config editor UI.

We can solidify this logic for this module by adding tests.

ReimarBauer:
**Purpose of PR?**:

Fixes #2716 #2819

no

xfailed test enabled

**Checklist:**
- [X] Bug fix. Fixes #2790 
- [X] Commit has unit tests

### DIFF
--- a/mslib/utils/config.py
+++ b/mslib/utils/config.py
@@ -309,6 +309,8 @@ class MSUIDefaultConfig:
 
     # Dictionary options with fixed key/value pairs
     fixed_dict_options = ["layout", "wms_prefetch", "topview", "sideview", "linearview"]
+    # List options with fixed length
+    fixed_list_options = ["MSCOLAB_timeout", ]
 
     # Fixed key/value pair options
     key_value_options = [
@@ -366,7 +368,7 @@ class MSUIDefaultConfig:
         "new_flighttrack_template": ["new-location"],
         "gravatar_ids": ["example@email.com"],
         "WMS_preload": ["https://wms-preload-url.com"],
-        "MSCOLAB_timeout": [int(), int()],
+        "MSCOLAB_timeout": [0, 0],
         "automated_plotting_flights": [["", "", "", "", "", ""]],
         "automated_plotting_hsecs": [["http://www.your-wms-server.de", "", "", ""]],
         "automated_plotting_vsecs": [["http://www.your-wms-server.de", "", "", ""]],
@@ -418,6 +420,7 @@ for key in [
     "__dict__",
     "__weakref__",
     "fixed_dict_options",
+    "fixed_list_options",
     "dict_option_structure",
     "list_option_structure",
     "key_value_options",
--- a/mslib/utils/config.py
+++ b/mslib/utils/config.py
@@ -600,7 +600,7 @@ def merge_dict(existing_dict, new_dict):
     new_dict -- Dict with new values
     """
     # Check if dictionary options with fixed key/value pairs match data types from default
-    for key in MSUIDefaultConfig.fixed_dict_options:
+    for key in MSUIDefaultConfig.fixed_dict_options + MSUIDefaultConfig.fixed_list_options:
         if key in new_dict:
             existing_dict[key] = compare_data(
                 existing_dict[key], new_dict[key]
@@ -618,6 +618,8 @@ def merge_dict(existing_dict, new_dict):
             for option_key in new_dict[key]:
                 for dos_key_key in dos[key]:
                     data, match = compare_data(dos[key][dos_key_key], new_dict[key][option_key])
+                    if key in MSUIDefaultConfig.fixed_list_options:
+                        match = len(dos[key][dos_key_key]) == len(new_dict[key][option_key])
                     if match:
                         temp_data[option_key] = new_dict[key][option_key]
                         break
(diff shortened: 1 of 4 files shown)


tests/_test_msui/test_msui.py
```diff
@@ -89,7 +89,7 @@ def test_milestone_url(self):
         with urlopen(self.window.milestone_url) as f:
             text = f.read().decode("utf-8")
         expected_version = __version__
-        pattern = rf'value="is:closed milestone:{re.escape(expected_version)}"'
+        pattern = rf'value="is:closed milestone:{re.escape(expected_version)} "'
```
ReimarBauer, line 92: seems gh has changed the content. 

The test is then very fragile and needs a better solution

ReimarBauer: https://github.com/Open-MSS/MSS/issues/2816

mslib/utils/config.py
```diff
@@ -366,7 +366,7 @@ class MSUIDefaultConfig:
         "new_flighttrack_template": ["new-location"],
         "gravatar_ids": ["example@email.com"],
         "WMS_preload": ["https://wms-preload-url.com"],
-        "MSCOLAB_timeout": [[2, 10]],
+        "MSCOLAB_timeout": [int(), int()],
```
matrss: `int() == 0`, what's the reasoning for changing it to that? The main difference seems to be removing the inner list, but is that correct? Afterall, the timeout setting can be a tuple of two values, or something equivalent.

ReimarBauer: Although it can be zero. These entries are in place to guarantee that new values have the same data type as existing ones. I’m hesitant to keep both default values because we used only placeholders elsewhere.

The inner list definition is wrong, it should only have 2 values of type integer.

matrss: > The inner list definition is wrong, it should only have 2 values of type integer.

Are you sure about that? The other definitions all have one element in the outer list, which defines a placeholder value for the config option. And in the case of MSCOLAB_timeout that config option is supposed to be a sequence of two elements, right?

ReimarBauer:
Yes, it is the same as with gravatar_ids, new_flightrack_template and some
more.
The whole needs a refactoring for readability. It is also not that clear
what shows an edit failure on the config_editor ui.

Matthias Riße ***@***.***> schrieb am Mo., 12. Mai 2025,
16:36:

mslib/utils/config.py
ReimarBauer: Yes, it is the same as with gravatar_ids, new_flightrack_template and some
more.

The list_option_structure is used by the config_editor, to create a new row for e.g. automated_plotting_flights `[[]] `with 6 elements.

Bildschirmfoto 2025-05-13 um 13 08 55

using `[]` creates for the gravatar_id a new row with 1 element.
Bildschirmfoto 2025-05-13 um 15 49 30


Because MSCOLAB_timeout is implemented as a list allowing additions and removals, this introduces potential problems in the UI. We need to ensure the UI behaves consistently with how it handles non-list values – for example

Bildschirmfoto 2025-05-13 um 13 16 56

ReimarBauer:
The editor does it now correct
Bildschirmfoto 2025-05-13 um 15 16 38

ReimarBauer:
To maintain a unified approach, the logic needs to be consistently applied across the modify_config_file module, its tests, the configuration settings, and the user interface. That can be done with #2819 

I am looking currently in that issue and write more tests :)

joernu76 (approved):
The whole configuration scheme has gotten pretty complex to enforce the proper form of parameters... This should be simpler.

PR #2830, opened 2025-06-25
added a special handling for path strings

### LINKED ISSUE #2718: data_dir=/tmp  from config editor cannot become stored and used as expected

ReimarBauer:
**Purpose of PR?**:

Fixes #2718

### DIFF
--- a/mslib/support/qt_json_view/datatypes.py
+++ b/mslib/support/qt_json_view/datatypes.py
@@ -88,7 +88,7 @@ class NoneType(DataType):
     """None"""
 
     def matches(self, data):
-        return data is None
+        return data is None or data is ""
 
     def value_item(self, value, model, key=None):
         item = super(NoneType, self).value_item(value, model, key)
--- a/mslib/support/qt_json_view/datatypes.py
+++ b/mslib/support/qt_json_view/datatypes.py
@@ -88,7 +88,7 @@ class NoneType(DataType):
     """None"""
 
     def matches(self, data):
-        return data is None or data is ""
+        return data is None
 
     def value_item(self, value, model, key=None):
         item = super(NoneType, self).value_item(value, model, key)
@@ -366,7 +366,24 @@ class UrlType(DataType):
 class FilepathType(DataType):
     """Files and paths can be opened."""
 
-    REGEX = re.compile(r'(\/.*)|([A-Z]:\\.*)')
+    REGEX = re.compile(r'''
+            ^(?:
+                # Windows-path (C:\path\to\file) - at least one separator and a drive letter
+                [A-Za-z]:[\\\/](?:[^<>:"|?*\n\r]*[\\\/])*[^<>:"|?*\n\r]*
+                |
+                # Unix/Linux absolute path (/path/to/file) - has to start with a slash /
+                \/(?:[^\/\0\n\r]*\/)*[^\/\0\n\r]*
+                |
+                # Unix/Linux relative path with explicite ./ or ../
+                \.{1,2}\/(?:[^\/\0\n\r]*\/)*[^\/\0\n\r]*
+                |
+                # Unix/Linux relative path with at least one / (path/to/file)
+                [^\/\0\n\r]+\/(?:[^\/\0\n\r]*\/)*[^\/\0\n\r]*
+                |
+                # UNC-Path (\\server\share\path)
+                \\\\[^\\\/\n\r]+\\[^\\\/\n\r]+(?:\\[^\\\/\n\r]*)*
+            )$
+        ''', re.VERBOSE | re.IGNORECASE)
 
     def matches(self, data):
         # (mss)
--- a/mslib/support/qt_json_view/datatypes.py
+++ b/mslib/support/qt_json_view/datatypes.py
@@ -374,7 +374,7 @@ class FilepathType(DataType):
                 # Unix/Linux absolute path (/path/to/file) - has to start with a slash /
                 \/(?:[^\/\0\n\r]*\/)*[^\/\0\n\r]*
                 |
-                # Unix/Linux relative path with explicite ./ or ../
+                # Unix/Linux relative path with explicitly ./ or ../
                 \.{1,2}\/(?:[^\/\0\n\r]*\/)*[^\/\0\n\r]*
                 |
                 # Unix/Linux relative path with at least one / (path/to/file)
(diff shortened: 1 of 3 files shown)


joernu76:
Alternative, we might remove the FilePathType altogether? Differentiating a String from a FilePathType seems impossible to me. At least, FilePathType should inherit from StrType, which might solve the issue as well.

The whole config-UI editor is way to complex.

ReimarBauer:
> Alternative, we might remove the FilePathType altogether? Differentiating a String from a FilePathType seems impossible to me. At least, FilePathType should inherit from StrType, which might solve the issue as well.
> 


FilePathType is a question of the correct regex, the one I now have is suggested by pycharm + claude sonnet. 
That reminded me https://github.com/moinwiki/moin-1.9/blob/master/MoinMoin/parser/text_moin_wiki.py
I also extended the tests. Not all all strings are valid.


> The whole config-UI editor is way to complex.

yes, a refactoring needs more tests too.

mslib/support/qt_json_view/datatypes.py
```diff
@@ -366,7 +366,24 @@ def actions(self, index):
 class FilepathType(DataType):
     """Files and paths can be opened."""
 
-    REGEX = re.compile(r'(\/.*)|([A-Z]:\\.*)')
+    REGEX = re.compile(r'''
+            ^(?:
+                # Windows-path (C:\path\to\file) - at least one separator and a drive letter
```
matrss, line 371: Huh, I didn't know regex could have comments.

mslib/support/qt_json_view/datatypes.py
```diff
@@ -366,7 +366,24 @@ def actions(self, index):
...
+                # Unix/Linux relative path with explicitly ./ or ../
+                \.{1,2}\/(?:[^\/\0\n\r]*\/)*[^\/\0\n\r]*
+                |
+                # Unix/Linux relative path with at least one / (path/to/file)
+                [^\/\0\n\r]+\/(?:[^\/\0\n\r]*\/)*[^\/\0\n\r]*
+                |
+                # UNC-Path (\\server\share\path)
+                \\\\[^\\\/\n\r]+\\[^\\\/\n\r]+(?:\\[^\\\/\n\r]*)*
+            )$
+        ''', re.VERBOSE | re.IGNORECASE)
```
matrss, line 386: This regex does not match some valid paths, e.g.:
- paths that contain `\n` or `\r`
- relative paths without a `/`

And there is unnecessary redundancy between relative paths starting with `.` or `..` and other relative paths.

tests/_test_msui/test_msui.py
```diff
@@ -215,6 +215,22 @@ def test_open_config(self):
         with mock.patch("PyQt5.QtWidgets.QMessageBox.warning", return_value=QtWidgets.QMessageBox.Yes):
             self.window.config_editor.close()
 
+    @pytest.mark.parametrize("input_value, expected_value", [("~/mssdata", "~/mssdata"),
+                                                             ("/tmp", "/tmp"),
+                                                             ("../tmp", "../tmp"),
+                                                             ("./tmp", "./tmp"),
+                                                             ("name_no_path", "~/mssdata"),
```
matrss, line 222: This test case is wrong, the `input_value` is a valid path.

matrss (changes requested):
I agree with Jörn, this thing is way too complex. Why do we have to reinvent types that are already part of the stdlib (`str`, `int`, `pathlib.Path`, etc.)?

The regex your LLM generated is not correct for paths on Linux. I cannot comment on the Windows part since I have no idea about that OS, but considering that the rules for Linux paths are way simpler I'd expect that to be wrong as well.

Another issues with this is that it allows both Windows and Linux paths on both Windows and Linux. But having a windows path in the configuration file on linux doesn't make much sense.

A simpler approach would be to do
```
pathlib.Path(str).resolve()
```
and if it throws an exception then the path isn't valid, if it resolves successfully it is.

mslib/support/qt_json_view/datatypes.py
joernu76: I just tried to find out what would *not* be a valid path/filename in UNIX and failed to do so.
I think almost everything is a valid path/filename in UNIX.

And every valid filename is *certainly* a String.

It would be much simpler to check for the existance of a file/directory at this point instead of the legality of its name.

matrss: Yes, almost everything goes as a path on linux. There are only two restrictions: paths cannot contain a null byte, and filenames can additionally not contain a forward slash, since that is the path separator.

There are many arguments to be made that it _should_ be more restricted, but it isn't.

I think with some of the configurable paths MSS creates defaults if the file/directory does not exist yet, right? In that case checking for existence wouldn't work.

ReimarBauer:
> I agree with Jörn, this thing is way too complex. Why do we have to reinvent types that are already part of the stdlib (`str`, `int`, `pathlib.Path`, etc.)?
> 
> The regex your LLM generated is not correct for paths on Linux. I cannot comment on the Windows part since I have no idea about that OS, but considering that the rules for Linux paths are way simpler I'd expect that to be wrong as well.
> 
> Another issues with this is that it allows both Windows and Linux paths on both Windows and Linux. But having a windows path in the configuration file on linux doesn't make much sense.
> 
> A simpler approach would be to do
> 
> ```
> pathlib.Path(str).resolve()
> ```
> 
> and if it throws an exception then the path isn't valid, if it resolves successfully it is.

after fs is gone that check is a nice option

mslib/support/qt_json_view/datatypes.py
ReimarBauer: this is since long, py2

tests/_test_msui/test_msui.py
ReimarBauer: oh, pathlib sets it relativ to the current path.

mslib/support/qt_json_view/datatypes.py
ReimarBauer: the pathlib.Path(str).resolve() is currently the best option after we replaced fs by Path.

PR #2839, opened 2025-07-17
fix:fix for airspace access and shapely use

### LINKED ISSUE #2836: refactor mpl_map to use pygeoif
Currently, mpl_map relies on the shapely library. We need to verify through testing that we can successfully replace it with pygeoif. pygeoif is already used by fastkml, and shapely previously served as a fallback within fastkml. If this replacement works, we may be able to eliminate the last remaining use case for shapely and remove it from our dependencies.

joernu76:
**Purpose of PR?**:

Fixes #2836

no

**Checklist:**
- [ ] Commit has unit tests


joernu76:
This PR removes the shapely dependence at minor performance cost.
But it also fixes a longstanding issue, where only part of the available airspaces were selectable.

mslib/msui/mpl_map.py
```diff
@@ -391,11 +391,8 @@ def draw_airspaces(self, countries=[], range_km=None):
 
             for i, airspace in enumerate(airspaces):
                 airspaces[i]["polygon"] = list(zip(*self.projtran(*list(zip(*airspace["polygon"])))))
-            map_polygon = Polygon([(self.llcrnrx, self.llcrnry), (self.urcrnrx, self.llcrnry),
-                                  (self.urcrnrx, self.urcrnry), (self.llcrnrx, self.urcrnry)])
```
ReimarBauer, line 401: Polygon not needed? fine

joernu76: "not needed" is a strong word, but this is a performance feature that we can do without. I do not think it worth to add the complexity of such a big package such as shapely in particular if we want to get rid of it of reasons.

ReimarBauer (approved):
Thx

PR #2856, opened 2025-08-06
moved creation of empty configuration dict

### LINKED ISSUE #2851: develop: on restart of msui - msui_settings.json gets cleared
seems we have for checking what happens to msui_settings.json on a start not a test.

I assume that I introduced the bug with https://github.com/Open-MSS/MSS/pull/2823

currently when changing something on the msui_settings.json that persist only the actual session. With a restart the file is set to default.

A fix for this should add a test, verifying that the file is not cleared on start of the main msui window.

ReimarBauer:
**Purpose of PR?**:

Fixes #2851

solved an old ToDo too, added a test.

### DIFF
--- a/mslib/msui/autoplot_dockwidget.py
+++ b/mslib/msui/autoplot_dockwidget.py
@@ -188,7 +188,7 @@ class AutoplotDockWidget(QWidget, Ui_AutoplotDockWidget):
 
         # Create the configuration path
         config_path = const.MSS_AUTOPLOT.as_posix()
-        if not os.path.exists(const.MSUI_CONFIG_PATH.as_posix()):
+        if not os.path.exists(config_path):
             const.MSS_AUTOPLOT.parent.mkdir(parents=True, exist_ok=True)
         if not os.path.exists(config_path):
             const.MSS_AUTOPLOT.write_text("{}")
--- a/mslib/msui/autoplot_dockwidget.py
+++ b/mslib/msui/autoplot_dockwidget.py
@@ -187,7 +187,11 @@ class AutoplotDockWidget(QWidget, Ui_AutoplotDockWidget):
             view = "linear"
 
         # Create the configuration path
-        config_path = os.path.join(const.MSUI_CONFIG_PATH, "mssautoplot.json")
+        config_path = const.MSS_AUTOPLOT.as_posix()
+        if not os.path.exists(const.MSUI_CONFIG_PATH.as_posix()):
+            const.MSS_AUTOPLOT.parent.mkdir(parents=True, exist_ok=True)
+        if not os.path.exists(config_path):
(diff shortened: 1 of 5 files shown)


mslib/msui/autoplot_dockwidget.py
```diff
@@ -187,7 +187,11 @@ def download_plots_cli(self, config_settings):
             view = "linear"
 
         # Create the configuration path
-        config_path = os.path.join(const.MSUI_CONFIG_PATH, "mssautoplot.json")
+        config_path = const.MSS_AUTOPLOT.as_posix()
+        if not os.path.exists(const.MSUI_CONFIG_PATH.as_posix()):
```
joernu76: MSS_AUTOPLOT may be in another directory if env-variables are used.

joernu76 (changes requested):

joernu76:
I think the check on directories etc. should be centralized. If the value of the variables may change during runtime, I'd prefer a "reboot" of the application. If this is neither possible nor desired, the first point of use is indeed a proper place as well.

ReimarBauer:
> I think the check on directories etc. should be centralized. If the value of the variables may change during runtime, I'd prefer a "reboot" of the application. If this is neither possible nor desired, the first point of use is indeed a proper place as well.

Changes made with the config editor trigger an application restart.
When we reboot the application after using modify_config_file, it will boot up when we add a WMS server or log in to a new MSColab server. I think we need to revisit how and when we populate the JSON file with data. Let's track this as a separate issue. 

The current pattern with the environment variable is necessary for both our tests and tutorials because it allows specifying a path different from the production setup. I'm hoping we can use fixtures to handle this differently within the test suite. Right now, the tests are relying on their own constants and aren’t utilizing this functionality.  Therefore, the directory check should be implemented where it is first required.

mslib/msui/autoplot_dockwidget.py
```diff
@@ -187,7 +187,11 @@ def download_plots_cli(self, config_settings):
             view = "linear"
 
         # Create the configuration path
-        config_path = os.path.join(const.MSUI_CONFIG_PATH, "mssautoplot.json")
+        config_path = const.MSS_AUTOPLOT.as_posix()
+        if not os.path.exists(config_path):
```
joernu76, line 191: ```if not os.path.exists(config_path.parent):``` would be cleaner, but it works as it is.

joernu76 (approved):

mslib/msui/autoplot_dockwidget.py
ReimarBauer: yes, I assume we will have another refactoring for pathlib replacing os whereever it is possible. This module needs more tests related to usages. I was unsure so I wanted to keep the old behaviour.

PR #2867, opened 2025-08-24
use __version__

### LINKED ISSUE #2866: use __version__ in setup.py
we need to remove past

ReimarBauer:
**Purpose of PR?**:

Fixes #2866


matrss:
I think importing the to-be-installed package from setup.py is not a good idea, as it forces runtime dependencies to be build time dependencies AFAIU.

ReimarBauer:
better idea is welcome, we need to get the dependency from past removed soon

matrss:
Alternative proposal in #2868.

matrss:
Closing in favor of #2868.

PR #2868, opened 2025-08-25
Replace setup.py with pyproject.toml

### LINKED ISSUE #2001: execfile has a dependency to the future / past module
at least for building by boa and conda we need future in the host requirement. 
This includes past

We need a different way to get the result of 
execfile('mslib/version.py')
in https://github.com/Open-MSS/MSS/blob/develop/setup.py#L33

### LINKED ISSUE #2866: use __version__ in setup.py

matrss:
This eliminates the need for the future module by setting the mslib version in pyproject.toml and populating __version__ from package metadata. One limitation of pyproject.toml is that it is not possible to define OS-specific entrypoints, so the msidp application now has a runtime check to error out on Windows instead of just not existing there.

The pyproject.toml file is the latest PEP standard for python project metadata. In the future it could also accomodate additional configuration that are currently set in setup.cfg and pytest.ini.

**Purpose of PR?**:

Fixes #2001, fixes #2866.

**Checklist:**
- [ ] Commit has unit tests


pyproject.toml
```diff
@@ -1,6 +1,41 @@
...
+readme = "README.md"
+classifiers = [
+    "Development Status :: 5 - Production/Stable",
+]
+keywords = ["mslib"]
+maintainers = [
+    {name = "Reimar Bauer", email = "rb.proj@gmail.com"},
+]
+authors = [
+    {name = "Marc Rautenhaus", email = "wxmetvis@posteo.de"},
```
ReimarBauer, line 14: We maybe can add here all from the AUTHORS file?

matrss: That would make sense, but I initially didn't want to push too much into this PR. If we put them all here and removed the AUTHORS file entirely we have to update all copyright notices in every file and update some things in the docs.

mslib/msidp/idp.py
```diff
@@ -1092,6 +1092,9 @@ def main():
         '\033[91mWARNING: msidp is solely for development and '
         'testing purposes; do not use in production environments.\033[0m'
     )
+    if os.name == "nt":
```
ReimarBauer, line 1095: I am currently not sure if "nt" will crash because of a missing "xmlsec1" on the `import saml2.xmldsig as ds`line

ReimarBauer: makes no trouble
Bildschirmfoto 2025-08-26 um 15 23 32

ReimarBauer (approved):
Thx

PR #2915, opened 2025-11-22
Fix #2885: Prevent crash on corrupted performance settings

### LINKED ISSUE #2885: without performance data we should not show ERROR
The placeholder for performance_settings needs a different impementation

ERROR: Failed to serialize performance_settings: 'str' object does not support item assignment

Delta17920:
**Purpose of PR?**:

Fixes #2885

This PR adds a type check to the `load_settings` method in `mslib/msui/flighttrack.py`. It ensures that `performance_settings` is always initialized as a dictionary. If the loaded configuration data is corrupted (e.g., loaded as a `str` instead of a `dict`), the application now safely falls back to `DEFAULT_PERFORMANCE` instead of crashing with a `TypeError`.

No

Manually verified by temporarily modifying the code to force `performance_settings` to be a string ("corrupted data"). Confirmed that the application detected the invalid type, reverted to the default dictionary, and launched successfully without crashing.

_None_

No

**Checklist:**
- [x] Bug fix. Fixes #2885
- [ ] Commit has unit tests

### DIFF
--- a/mslib/msui/flighttrack.py
+++ b/mslib/msui/flighttrack.py
@@ -213,7 +213,11 @@ class WaypointsTableModel(QtCore.QAbstractTableModel):
         """
         Load settings from the file self.settingsfile.
         """
-        self.performance_settings = load_settings_qsettings(self.settings_tag, DEFAULT_PERFORMANCE)
+        settings = load_settings_qsettings(self.settings_tag, DEFAULT_PERFORMANCE)
+        # Ensure we have a dictionary. If QSettings returns a string/garbage, reset to default.
+        if not isinstance(settings, dict):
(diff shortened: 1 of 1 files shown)


Delta17920:
It seems the tests failed on `test_milestone_url` with an `AssertionError: Expected milestone format not found`.
This appears to be unrelated to my changes in `flighttrack.py` (likely a network issue or a change in the GitHub HTML response). Could you please re-run the CI?

joernu76:
Please provide a test demonstrating the improved behaviour.
I can neither replicate the issue nor validate whether the fix solves anything.
Eg a test that fails with the old code, but not with the new one.

Issue #2812, opened 2025-04-25
Incorrect heading markup


levi178u:
The heading underline(~) must exactly match the length of the title line as per Sphinx docs. See this in `netCDF4tools.py`, `config.py`, `coordinate.py`. `loggerdef.py` , `airdata.py`,` mslib\mscolab\__init__.py `, `chat_manager.py` , `conf.py`  etc.....

**Samples:**
![Image](https://github.com/user-attachments/assets/8f9e6842-b2ce-48c6-9846-6cd175d9c937)

![Image](https://github.com/user-attachments/assets/f09eec6f-6ded-43ea-b0f9-60a96393e234)

ReimarBauer:
To ensure consistency, we need to add a linter that checks the project's state. Start by setting up the linter and then fix all of the errors it finds.

ReimarBauer:
This can be done when next time the doc strings have to become changed

keerthanam0905-22:
Hello! I’m new to open-source and would like to work on this issue as my first contribution. Could you please assign it to me? 
Thank you!

ReimarBauer:
Hi @keerthanam0905-22 you have it.

We really appreciate your interest in this project! We won’t be participating in GSoC next year, but please don't let that discourage you from contributing. If open source is important to you, this is a great place to get involved and we'd love to have your help.

MohamedSharfan:
hello @ReimarBauer, i would like to work on this issue and i am new to this could you please help me with this

ReimarBauer:
Hi @MohamedSharfan 
There's already a pull request addressing this issue: https://github.com/Open-MSS/MSS/pull/2925/changes. It's missing a unit test which should document the initial state, and then consistently verify the fix.  Later, this unit test should prevent us from accepting a merge request if that 'linting' check fails.

PR #2925, opened 2025-11-26
Fix heading underline length in mslib.utils.config

keerthanam0905-22:
Corrected the underline to match the heading length as required by Sphinx documentation style. This fixes part of issue #2812.


ReimarBauer:
Hi @keerthanam0905-22 
My apologies for not reviewing this earlier. I have been significantly limited on time recently.

Would it be possible to include a unit test, in order to avoid accepting merge requests that don't adhere to this in the future?

PR #2929, opened 2025-11-27
timeout in fixture resolved

### LINKED ISSUE #2928: avoid infinite loop in _running_eventlet_server
We have a while loop without a time limit in the fixture to start the _running_eventlet_server
May be this causes the hangs on the tests sometimes. At least it blocks me currently in pycharm.

ReimarBauer:
**Purpose of PR?**:

Fixes #2928

adds a timeout into a fixture for starting the eventlet_server

**Checklist:**
- [X] Bug fix. Fixes #2928 
- [ ] Commit has unit tests

### DIFF
--- a/tests/fixtures.py
+++ b/tests/fixtures.py
@@ -201,8 +201,11 @@ def _running_eventlet_server(app):
     process = ctx.Process(target=eventlet.wsgi.server, args=(socket, app), daemon=True)
     try:
         process.start()
-        while not is_url_response_ok(urllib.parse.urljoin(url, "index")):
-            time.sleep(0.5)
+        retry_time = 0
+        sleep_time = 0.5
+        while not is_url_response_ok(urllib.parse.urljoin(url, "index")) and retry_time < 5:
+            time.sleep(sleep_time)
+            retry_time += sleep_time
         yield url
     finally:
         process.terminate()
--- a/tests/fixtures.py
+++ b/tests/fixtures.py
@@ -206,6 +206,8 @@ def _running_eventlet_server(app):
         while not is_url_response_ok(urllib.parse.urljoin(url, "index")) and retry_time < 5:
             time.sleep(sleep_time)
             retry_time += sleep_time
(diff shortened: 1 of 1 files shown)


joernu76:
If the server could not be started, shouldn't an error be raised instead?

ReimarBauer:
@joernu76 the server needs a bit time to startup. But I think raising an error now shows which tests get blocked by this.
If we don't reach timeout at 40mins then we have found the culprit.

tests/fixtures.py
```diff
@@ -201,8 +201,13 @@ def _running_eventlet_server(app):
     process = ctx.Process(target=eventlet.wsgi.server, args=(socket, app), daemon=True)
     try:
         process.start()
-        while not is_url_response_ok(urllib.parse.urljoin(url, "index")):
-            time.sleep(0.5)
+        retry_time = 0
+        sleep_time = 0.5
+        while not is_url_response_ok(urllib.parse.urljoin(url, "index")) and retry_time < 5:
+            time.sleep(sleep_time)
+            retry_time += sleep_time
```
matrss: Can this do exponential backoff instead?

tests/fixtures.py
```diff
@@ -201,8 +201,13 @@ def _running_eventlet_server(app):
...
+        if not is_url_response_ok(urllib.parse.urljoin(url, "index")):
+            raise RuntimeError(f"Server did not start within 5 seconds at {url}")
```
matrss:
```suggestion
        start_time = time.time()
        sleep_time = 0.01
        while not is_url_response_ok(urllib.parse.urljoin(url, "index")):
            if (time.time() - start_time) > 5:
                raise RuntimeError(f"Server did not start within 5 seconds at {url}")
            time.sleep(sleep_time)
            sleep_time *= 2
            if sleep_time > 1:
                sleep_time = 1
```

Something like this.

matrss (approved):

PR #2931, opened 2025-12-02
feat: Fix #2930

### LINKED ISSUE #2930: Extend standard atmosphere
For the upcoming WAVEGUIDE campaign, we need to visualise model data above 70km, which currently crashes due to thermolib only supporting conversions above 71km.

The ISA standard atmosphere provides gradients up to 81km, but above a constant gradient of 0K would be accurate enough up to 100km altitude, which is sufficient for our purposes.

Standard atmospheres providing values above 80 km seem to be all latitude dependent and do not use the concept of constant temperature gradients, which is underlying the computations of thermolib.

joernu76:
Implement a standard atmosphere up to 100km altitude, raising the ceiling from the current 71km.

**Purpose of PR?**:

Fixes #2930

no

tests are provided for new altitude regionfrom 71km to 100km.

**Checklist:**
- [x] Commit has unit tests

### DIFF
--- a/mslib/utils/thermolib.py
+++ b/mslib/utils/thermolib.py
@@ -126,6 +126,7 @@ def omega_to_w(omega, p, t):
 
 # Values according to the 1976 U.S. Standard atmosphere [NOAA1976]_.
 # List of tuples (height, temperature, pressure, temperature gradient)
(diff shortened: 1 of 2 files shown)


ReimarBauer (approved):
thx

PR #2936, opened 2025-12-03
feature:extend sideview vertical range to 99km / Fix #2935

### LINKED ISSUE #2935: Extend vertical SideView range
The vertical range of the SideView is currently limited to about 65km altitude. For WAVEGUIDE, this also needs to extend to 99km.

joernu76:
Extend range of SideView to 99km.

**Purpose of PR?**:

Fixes #2935

no
See also #2930

**Checklist:**
- [x] Commit has unit tests

### DIFF
--- a/tests/_test_utils/test_thermolib.py
+++ b/tests/_test_utils/test_thermolib.py
@@ -33,7 +33,7 @@ from mslib.utils import thermolib
 
 
 def test_flightlevel2pressure2flightlevel():
-    fs = (np.arange(1, 95000, 1000.) * units.m).to(units.hft)
+    fs = (np.arange(1, 99900, 100.) * units.m).to(units.hft)
     ps = thermolib.flightlevel2pressure(fs)
     fs_p = thermolib.pressure2flightlevel(ps).magnitude
     assert fs.magnitude == pytest.approx(fs_p)
@@ -41,6 +41,10 @@ def test_flightlevel2pressure2flightlevel():
 
 def test_pressure2flightlevel2pressure():
     ps = np.arange(1, 105000, 1.)[::-1] * units.Pa
+    fs = thermolib.pressure2flightlevel(ps)
+    ps_p = thermolib.flightlevel2pressure(fs).magnitude
+    assert ps.magnitude == pytest.approx(ps_p)
+    ps = np.arange(0.024, 1, 0.001)[::-1] * units.Pa
     fs = thermolib.pressure2flightlevel(ps)
     ps_p = thermolib.flightlevel2pressure(fs).magnitude
     assert ps.magnitude == pytest.approx(ps_p)
@@ -54,7 +58,7 @@ def test_flightlevel2pressure():
     assert thermolib.flightlevel2pressure(1626.8966 * units.hft).magnitude == pytest.approx(80)
     assert thermolib.flightlevel2pressure(1804.2727 * units.hft).magnitude == pytest.approx(40)
     with pytest.raises(ValueError):
-        thermolib.flightlevel2pressure(102000 * units.m)
+        thermolib.flightlevel2pressure(100000 * units.m)
 
 
 def test_pressure2flightlevel():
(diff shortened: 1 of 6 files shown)


tests/_test_utils/test_thermolib.py
```diff
@@ -81,14 +81,15 @@ def test_isa_temperature():
...
-        thermolib.isa_temperature(100001 * units.m)
+        thermolib.isa_temperature(100000 * units.m)
     assert thermolib.isa_temperature(11000 * units.m).magnitude == pytest.approx(216.65)
     assert thermolib.isa_temperature(20000 * units.m).magnitude == pytest.approx(216.65)
     assert thermolib.isa_temperature(32000 * units.m).magnitude == pytest.approx(228.65)
     assert thermolib.isa_temperature(47000 * units.m).magnitude == pytest.approx(270.65)
     assert thermolib.isa_temperature(51000 * units.m).magnitude == pytest.approx(270.65)
     assert thermolib.isa_temperature(71000 * units.m).magnitude == pytest.approx(214.65)
     assert thermolib.isa_temperature(84852 * units.m).magnitude == pytest.approx(186.95)
+    assert thermolib.isa_temperature(99999 * units.m).magnitude == pytest.approx(186.95)
```
ReimarBauer, line 94: why is this the same value pytest.approx(186.95) as the line before? Seeing that other checks show this pattern too. it looks like a range which gives the same result.

Is that 99999 a 99.999 with fractions? (I can't enter that in the UI)

ReimarBauer (commented):
It looks good so far, but how does I know it's limited to 0-99 km? For example, if I want to change 98.5 to 99.5, it doesn't allow me to until I remove the decimal places.

Maybe all larger values should become replaced by its upper limit.

tests/_test_utils/test_thermolib.py
joernu76: The conversion routine works from 0 to 99999m/99.999km (effectively for any number < 100km).
I changed the order of the tests to make this more apparent.

The SideView goes only to 99km as we add internally 1km to the selected top height to define the grid.
But this can be improved upon. I just changed the code to allow values up to 99.9km and changed the precision to be dependent on the type, ie for pressure_altitude, a single digit is allowed. 4 digits is only allowed for pressure, where this is necessary for the large altitudes.

I also adapted the code to allow smaller pressures, as previously above 99km the pressure could not be converted back to km.

ReimarBauer:
Setting the largest value and toggling through the vertical axis options shows a traceback. 


https://github.com/user-attachments/assets/50006125-0eb0-4c5d-b0fe-19018572bfb7




```
Fatal error in MSS 10.2.0 on macOS-26.1-arm64-arm-64bit
Python 3.11.11 | packaged by conda-forge | (main, Dec  5 2024, 14:21:42) [Clang 18.1.8 ]

  File "/Users/reimarbauer/MAIN/MSS/mslib/msui/sideview.py", line 276, in verticalunitsclicked
    thermolib.pressure2flightlevel(sb.value() * units.hPa).magnitude, "hft", new_unit))
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/reimarbauer/MAIN/MSS/.pixi/envs/dev/lib/python3.11/site-packages/metpy/xarray.py", line 1330, in wrapper
    result = func(*bound_args.args, **bound_args.kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/reimarbauer/MAIN/MSS/.pixi/envs/dev/lib/python3.11/site-packages/metpy/units.py", line 333, in wrapper
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/Users/reimarbauer/MAIN/MSS/mslib/utils/thermolib.py", line 240, in pressure2flightlevel
    raise ValueError("pressure to flight level conversion not implemented for p <= 0.0238725468 Pa")
ValueError: pressure to flight level conversion not implemented for p <= 0.0238725468 Pa
```

ReimarBauer (commented):
besides the linter failure I pushed last version, it looks now good, see PR https://github.com/Open-MSS/MSS/pull/2961

ReimarBauer (approved):
thx

Issue #2878, opened 2025-09-02
numpy problem with mss for Macs


ReimarBauer:
tableview can show a problem

```
Fatal error in MSS 10.1.0 on macOS-15.6.1-arm64-arm-64bit
Python 3.11.10 | packaged by conda-forge | (main, Oct 16 2024, 01:26:25) [Clang 17.0.6 ]

Please report bugs in MSS to https://github.com/Open-MSS/MSS

Information about the fatal error:

Traceback (most recent call last):
  File "/Users/reimarbauer/Programs/msss2/.pixi/envs/default/lib/python3.11/site-packages/mslib/msui/msui_mainwindow.py", line 997, in create_view_handler
    self.create_view(_type, self.active_flight_track)
  File "/Users/reimarbauer/Programs/msss2/.pixi/envs/default/lib/python3.11/site-packages/mslib/msui/msui_mainwindow.py", line 1048, in create_view
    view_window = tableview.MSUITableViewWindow(model=model, tutorial_mode=self.tutorial_mode)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/reimarbauer/Programs/msss2/.pixi/envs/default/lib/python3.11/site-packages/mslib/msui/tableview.py", line 69, in __init__
    self.setFlightTrackModel(model)
  File "/Users/reimarbauer/Programs/msss2/.pixi/envs/default/lib/python3.11/site-packages/mslib/msui/tableview.py", line 281, in setFlightTrackModel
    self.update_roundtrip_enabled()
  File "/Users/reimarbauer/Programs/msss2/.pixi/envs/default/lib/python3.11/site-packages/mslib/msui/tableview.py", line 266, in update_roundtrip_enabled
    self.btRoundtrip.setEnabled(self.is_roundtrip_possible())
TypeError: setEnabled(self, a0: bool): argument 1 has unexpected type 'numpy.bool'
```

ReimarBauer:
2.3.2 was 5 days ago released https://anaconda.org/conda-forge/numpy

2.3.1 works

2.3.2 works now too

ReimarBauer:
this sounds a bit similiar to https://github.com/conda-forge/netcdf4-feedstock/issues/172

the problems are gone with cycling the installation

ReimarBauer:
We have this also in tests

```
[...]

PR #2940, opened 2025-12-04
removed macos-13, added macos-15

### LINKED ISSUE #2908: macOS 13 runner image will be retired by December 4th, 2025.
The macOS 13 runner image will be retired by December 4th, 2025. To raise awareness of the upcoming removal, jobs using macOS 13 will temporarily fail during the scheduled brownout time periods defined below:
* November 4, 14:00 UTC - November 5, 00:00 UTC
* November 11, 14:00 UTC - November 12, 00:00 UTC
* November 18, 14:00 UTC - November 19, 00:00 UTC
* November 25, 14:00 UTC - November 26, 00:00 UTC

This deprecation includes the following labels:
* macos-13
* macos-13-large
* macos-13-xlarge
## What you need to do
(Recommended) If your workflow is architecture agnostic, you can migrate to any of our arm64 labels:
* macos-15 or macos-latest
* macos-14
* macos-14-xlarge
* macos-latest-xlarge or macos-15-xlarge

For users that require the x86_64 (Intel) architecture, jobs can be migrated to one of the following labels:
* macos-15-intel (new)
* macos-14-large
* macos-latest-large or macos-15-large

You can always get up-to-date information on our tools by reading about the software in the runner images repository. Please contact GitHub Support if you run into any problems or need help.

## Notice of macOS x86_64 (Intel) architecture deprecation
Apple has discontinued support for the x86_64 (Intel) architecture going forward. GitHub will no longer support this architecture on macOS after the macOS 15 runner image is retired in Fall 2027. You should begin migrating your workloads to arm64-based (Apple Silicon) runners as soon as possible to prepare for this eventual deprecation.

Best,
The GitHub Team

ReimarBauer:
**Purpose of PR?**:

Fixes #2908 #2878


pixi.toml
```diff
@@ -33,14 +33,15 @@ hdf4 = "*"
...
+libgfortran5 = ">=14"
 libtiff = "*"
 lxml = "*"
 markdown = "*"
 matplotlib = ">=3.5.3"
 metpy = "*"
 multidict = "*"
 netcdf4 = "*"
-numpy = ">=2"
+numpy = ">=2.2.4,<3"
```
matrss: Numpy >=3 does not exist yet. Why do you presume it to be incompatible?

ReimarBauer: that was the suggestion here, 
https://github.com/prefix-dev/pixi/issues/3479

I read it all  between 2.2.4 and lower 3 including 2.2.4
We need to start with 2.2.4 and not below

ReimarBauer: we can also wait until 3 exists, but yeah lower limit in our case would work

matrss: If we only need a minimum version, then there should only be a lower bound. Given that a version 3 does not exist I see no reason why we would assume that it won't work with MSS.

joernu76 (approved):

PR #2957, opened 2025-12-17
improved tests by waiting asserts

### LINKED ISSUE #2956: test_multiple_times_save_filename and other tests did not wait
We have the test only in develop. We need to qwait until it is saved.

and other tests also did not wait for a button clickable. Many times qtbot.waitUntil seems missing

ReimarBauer:
**Purpose of PR?**:

Fixes #2956


ReimarBauer:
lets see how this performs after we merged stable into develop

ReimarBauer:
it looks to me that we now see more failures on github by file I/O related tests.

mslib/msui/mpl_qtwidget.py
```diff
@@ -720,8 +721,8 @@ def redraw_xaxis(self, lats, lons):
         for line in self.vertical_lines:
             try:
                 line.remove()
-            except ValueError as e:
-                logging.debug("Vertical line was somehow already removed:\n%s", e)
+            except (ValueError, NotImplementedError) as e:
```
matrss: When is this encountered? AFAICS self.vertical_lines is only appended to with axvline, which produces a matplotlib.lines.Line2D object, which definitely has a remove method. It might be a bug that NotImplementedError can come up here at all.

tests/_test_msui/test_mscolab_merge_waypoints.py
```diff
@@ -103,6 +103,23 @@ def _select_waypoints(self, table):
         for row in range(table.model().rowCount()):
             table.selectRow(row)
 
+    def _sync_local_from_server(self):
+        server_xml = self.window.mscolab.request_wps_from_server()
+        self.window.mscolab.waypoints_model = ft.WaypointsTableModel(xml_content=server_xml)
```
matrss: Why are you assigning to attributes of the window instance? The application code itself is supposed to do this, otherwise it is broken because it doesn't update the model properly.

tests/_test_msui/test_msui.py
```diff
@@ -76,13 +76,18 @@ def test_multiple_times_save_filename(qtbot, tmp_path):
...
     msui.save_flight_track(filename)
     assert os.path.exists(filename)
-    first_timestamp = os.path.getmtime(filename)
-    msui.save_handler()
-    second_timestamp = os.path.getmtime(filename)
+    first_timestamp = os.stat(filename).st_mtime_ns
+
+    # on github saving is too fast
+    def assert_():
+        msui.save_handler()
```
matrss: I think this should only be called once?

tests/_test_msui/test_sideview.py
```diff
@@ -207,17 +207,29 @@ def setup(self, qtbot, mswms_server, tmp_path):
...
+    def query_server(self, qtbot, url, timeout=5000):
+        self.wms_control.multilayers.cbWMS_URL.clear()
+        self.wms_control.multilayers.cbWMS_URL.setEditText(url)
+        self.wms_control.multilayers.cbWMS_URL.lineEdit().setText(url)
+
+        QtTest.QTest.mouseClick(
+            self.wms_control.multilayers.btGetCapabilities,
+            QtCore.Qt.LeftButton)
+
+        qtbot.waitUntil(
```
matrss: In other places we have consistently used the more pythonic method name with an underscore (`qtbot.wait_{until,signal,exposed}`).

tests/_test_msui/test_sideview.py
```diff
@@ -207,17 +207,29 @@ def setup(self, qtbot, mswms_server, tmp_path):
...
+            lambda: getattr(self.wms_control, "cpdlg", None) is not None,
```
matrss: I think cpdlg is a progress dialog that is closed automatically when it finishes. Is the attribute not None when the dialog finishes? Otherwise it might happen that this wait_until does not see the dialog if it happens too fast.

tests/_test_msui/test_topview.py
```diff
@@ -284,10 +284,26 @@ def setup(self, qtbot, mswms_server, tmp_path):
...
+        QtTest.QTest.mouseClick(
+            self.wms_control.multilayers.btGetCapabilities,
+            QtCore.Qt.LeftButton)
+
+        qtbot.waitUntil(
+            lambda: getattr(self.wms_control, "cpdlg", None) is not None,
+            timeout=timeout)
+
+        qtbot.waitUntil(
+            lambda: getattr(self.wms_control, "cpdlg", None) is not None and not self.wms_control.cpdlg.isVisible(),
```
matrss: This is different from the wait in test_sideview.py. Why?

tests/_test_msui/test_wms_control.py
```diff
@@ -105,39 +114,39 @@ def test_no_server(self, qtbot):
         """
         with mock.patch("PyQt5.QtWidgets.QMessageBox.critical") as mock_critical:
             self.query_server(qtbot, f"{self.scheme}://{self.host}:{self.port - 1}")
-            mock_critical.assert_called_once()
+            qtbot.waitUntil(lambda: mock_critical.call_count == 1, timeout=3000)
```
matrss: Couldn't this just be `qtbot.wait_until(mock_critical.assert_called_once)`?

tests/_test_msui/test_wms_control.py
```diff
@@ -177,8 +186,11 @@ def test_server_getmap(self, qtbot):
         """
         self.query_server(qtbot, self.url)
 
-        with qtbot.wait_signal(self.window.image_displayed):
-            QtTest.QTest.mouseClick(self.window.btGetMap, QtCore.Qt.LeftButton)
+        try:
+            with qtbot.wait_signal(self.window.image_displayed, timeout=5000):
```
matrss: `timeout=5000` is the default anyway.

tests/_test_msui/test_wms_control.py
```diff
@@ -177,8 +186,11 @@ def test_server_getmap(self, qtbot):
...
+                QtTest.QTest.mouseClick(self.window.btGetMap, QtCore.Qt.LeftButton)
+        except TimeoutError:
+            pytest.fail("Timeout: image_displayed was not emitted.")
```
matrss: No need to catch this, if it's thrown and pytest sees it the test will fail in effectively the same way.

tests/_test_msui/test_wms_control.py
```diff
@@ -391,12 +404,42 @@ def test_server_no_thread(self, mockthread, qtbot):
...
+        self.window.metadata_displayed.connect(self.view.draw_metadata)
+
+        # Ensure legend drawing is triggered when fetch_legend is called
+        def _emit_legend_displayed(*args, **kwargs):
+            # only emit when we simulate a non-cached fetch
+            if kwargs.get("use_cache") is False:
+                self.window.legend_displayed.emit()
+
+        self.window.legend_displayed.connect(self.view.draw_legend)
+        self.window.fetcher.fetch_legend.side_effect = _emit_legend_displayed
```
matrss: All of this is re-implementing behavior that the actual application code has to exhibit to function properly, I think. If this is necessary for the test to pass, then there must be something broken in the application or in what the test tries to assert.

tests/_test_msui/test_wms_control.py
```diff
@@ -391,12 +404,42 @@ def test_server_no_thread(self, mockthread, qtbot):
         server.child(0).setCheckState(0, 2)
         server.child(1).setCheckState(0, 2)
 
-        with qtbot.wait_signal(self.window.image_displayed):
+        # Mock the actual fetching so that no real HTTP call happens.
+        self.window.fetcher.fetch_map = mock.MagicMock()
+        self.window.fetcher.fetch_legend = mock.MagicMock()
```
matrss: Why should it not talk to mswms?

matrss (changes requested):

mslib/msui/mpl_qtwidget.py
ReimarBauer: we recheck this. I wondered that this comes up.

tests/_test_msui/test_mscolab_merge_waypoints.py
ReimarBauer: We should assume there's an error in the code. Unfortunately, it's not easy to pinpoint because it doesn't fail consistently – but it fails very often. We need a test that can force the error to occur. Right now, I’ve only been able to demonstrate the opposite: that the test itself doesn't have an error. If it passes like this, the problem must lie within the actual code.

tests/_test_msui/test_sideview.py
ReimarBauer: ok

tests/_test_msui/test_wms_control.py
ReimarBauer: yes, with the extra timeout value. it seems not to come sometimes to an end.

matrss: The timeout should be 5 seconds by default.

tests/_test_msui/test_mscolab_merge_waypoints.py
matrss: If it's determined that the application code is broken, then we shouldn't change the test in a way in which it hides this error. Rather, mark it as xfail, or fix the real bug if it's simple enough.

tests/_test_msui/test_wms_control.py
ReimarBauer: ok, I thought that it is different.

tests/_test_msui/test_wms_control.py
ReimarBauer: agree

mslib/msui/wms_control.py
```diff
@@ -401,13 +401,15 @@ class WMSControlWidget(QtWidgets.QWidget, ui.Ui_WMSDockWidget):
     signal_disable_cbs = QtCore.pyqtSignal(name="disable_cbs")
     signal_enable_cbs = QtCore.pyqtSignal(name="enable_cbs")
     image_displayed = QtCore.pyqtSignal()
+    legend_displayed = QtCore.pyqtSignal()
```
ReimarBauer, line 404: I wondered a bit that the signals weren't complete. And that this makes random crashes.

tests/_test_msui/test_mscolab_merge_waypoints.py
ReimarBauer: Okay, I think it’s good that we’ve come to that decision. It makes it easier to see that the test code is actually correct and the error lies elsewhere. I’ll mark it as xfail in the original.

tests/_test_msui/test_msui.py
ReimarBauer: I'm seeing strange behavior on GitHub with this. I changed to using st_mtime_ns after noticing GitHub reported identical values, but it’s still happening – GitHub sometimes shows the same value even when using st_mtime_ns.  It looks like qwait might be a better solution.

tests/_test_msui/test_mscolab_merge_waypoints.py
```diff
@@ -110,6 +110,7 @@ def __init__(self, *args, **kwargs):
         self.overwriteBtn.animateClick()
 
 
+@pytest.mark.skip(reason='The test identifies an inaccuracy in the code, which sporadically results in errors.')
```
ReimarBauer, line 113: xfail - does not help. Sometimes some of the skipped tests will succeed.

mslib/msui/flighttrack.py
```diff
@@ -635,6 +635,10 @@ def save_to_ftml(self, filename=None):
         doc = self.get_xml_doc()
         with open(filename, "w") as file_object:
             doc.writexml(file_object, indent="  ", addindent="  ", newl="\n", encoding="utf-8")
+            # the with should ensure that the file is flushed and synced to disk
+            # nevertheless we try if that changes the linux test behavior
```
ReimarBauer: This looks like it helps.

matrss: I wonder why this helps. The flush definitely happens on close / leaving the context manager too, here: https://github.com/python/cpython/blob/b82b12bc80373e2cfd0c0bed273e55e67640cdfd/Modules/_io/iobase.c#L280

The fsync apparently _doesn't_ happen automatically. But can a missing fsync lead to inconsistent data read in another process, or what's happening here?

tests/_test_msui/test_flighttrack.py
```diff
@@ -46,50 +44,43 @@ def test_load_settings_with_corrupted_performance_settings(self):
...
-                "Should fall back to DEFAULT_PERFORMANCE when data is corrupted"
-
-        finally:
-            # Restore original function
-            config_module.read_config_file = original_func
-            # Clean up the temporary file
-            if os.path.exists(temp_file):
-                os.unlink(temp_file)
+        # Verify it was reset to DEFAULT_PERFORMANCE
+        assert model.performance_settings == DEFAULT_PERFORMANCE, \
```
matrss: This assert is a duplicate, it happens on line 78 already.

tests/_test_msui/test_linearview.py
```diff
@@ -103,10 +103,26 @@ def setup(self, qtbot, mswms_server, tmp_path):
...
         self.window.hide()
 
-    def query_server(self, qtbot, url):
-        QtTest.QTest.keyClicks(self.wms_control.multilayers.cbWMS_URL, url)
-        with qtbot.wait_signal(self.wms_control.cpdlg.canceled):
-            QtTest.QTest.mouseClick(self.wms_control.multilayers.btGetCapabilities, QtCore.Qt.LeftButton)
+    def query_server(self, qtbot, url, timeout=5000):
+        self.wms_control.multilayers.cbWMS_URL.clear()
+        self.wms_control.multilayers.cbWMS_URL.setEditText(url)
+        self.wms_control.multilayers.cbWMS_URL.lineEdit().setText(url)
```
matrss: Are these two lines setting different things, or the same thing? If different, what exactly are they setting? A user does not have to enter the URL twice in the same input, right?

tests/_test_msui/test_linearview.py
```diff
@@ -103,10 +103,26 @@ def setup(self, qtbot, mswms_server, tmp_path):
...
+
+        QtTest.QTest.mouseClick(
+            self.wms_control.multilayers.btGetCapabilities,
+            QtCore.Qt.LeftButton)
+
+        qtbot.waitUntil(
```
matrss: Nit: non-pythonic naming

tests/_test_msui/test_linearview.py
```diff
@@ -103,10 +103,26 @@ def setup(self, qtbot, mswms_server, tmp_path):
...
+            lambda: getattr(self.wms_control, "cpdlg", None) is not None,
```
matrss: I think some of my comments have gone missing. Is it ensured that cpdlg is not None even after the dialog is finished? Otherwise this would again be a race condition between the first iteration of the wait_until and the btGetCapabilities' actions.

tests/_test_msui/test_wms_control.py
```diff
@@ -416,8 +425,11 @@ def test_server_getmap(self, qtbot):
         assert that a getmap call to a WMS server displays an image
         """
         self.query_server(qtbot, self.url)
-        with qtbot.wait_signal(self.window.image_displayed):
-            QtTest.QTest.mouseClick(self.window.btGetMap, QtCore.Qt.LeftButton)
+        try:
+            with qtbot.wait_signal(self.window.image_displayed, timeout=10000):
+                QtTest.QTest.mouseClick(self.window.btGetMap, QtCore.Qt.LeftButton)
+        except TimeoutError:
+            pytest.fail("Timeout: image_displayed signal was not emitted (VSec getmap).")
```
matrss: https://github.com/Open-MSS/MSS/pull/2957#discussion_r2704322704

tests/_test_msui/test_wms_control.py
```diff
@@ -430,8 +442,11 @@ def test_multilayer_drawing(self, qtbot):
         self.query_server(qtbot, self.url)
         server = self.window.multilayers.listLayers.findItems(f"{self.url}/",
                                                               QtCore.Qt.MatchFixedString)[0]
-        with qtbot.wait_signal(self.window.image_displayed):
-            server.child(0).draw()
+        try:
+            with qtbot.wait_signal(self.window.image_displayed, timeout=10000):
+                server.child(0).draw()
+        except TimeoutError:
+            pytest.fail("Timeout: image_displayed signal was not emitted (VSec multilayer draw).")
```
matrss: https://github.com/Open-MSS/MSS/pull/2957#discussion_r2704322704

mslib/msui/flighttrack.py
ReimarBauer: If these additions result in a new timestamp now existing, it didn't happen fast enough before.

I later try fsync only

tests/_test_msui/test_linearview.py
ReimarBauer: one appears to be enough. I was hoping to bypass a potential problem related to KeyClicks, but further testing is required.

tests/_test_msui/test_linearview.py
ReimarBauer: ok

mslib/msui/flighttrack.py
ReimarBauer: fsync only does not help, next is to try flush only

matrss: Try moving the fsync out of the with block, then it should be equivalent to both flush and fsync because the automatic flush happens on context exit.

ReimarBauer: After the with block file_object is destroyed and we have a ValueError

ReimarBauer: We maybe should investigate how the Runner VM is configured
https://github.com/actions/runner-images/issues/1939#issuecomment-719903248

matrss: > After the with block file_object is destroyed and we have a ValueError

That is not how variable scoping works in python, even if it would seem reasonable:

```
>>> print(f)
  File "<stdin>", line 1, in <module>
NameError: name 'f' is not defined
>>> with open("test123", mode="w") as f:
...     pass
...
>>> print(f)
<_io.TextIOWrapper name='test123' mode='w' encoding='utf-8'>
```

ReimarBauer: ok, it is not destroyed but closed, I've not used the right wording

```
>       os.fsync(file_object.fileno())
E       ValueError: I/O operation on closed file
```

this then crashes also more tests. 

- ERROR at teardown of Test_Save_Merge_Points.test_save_merge_points
- test_multiple_times_save_filename
- Test_MSSSideViewWindow.test_plugin_saveas[save_file0]
- Test_MSSSideViewWindow.test_flight_track_io
- Test_Save_Merge_Points.test_save_merge_points
- Test_Mscolab.test_work_locally_toggle

5 failed, 570 passed, 19 skipped, 22 warnings, 1 error in 100.46s (0:01:40)

I think we have to do that for stability reasons running tests

ReimarBauer: I've undone this. I am also not sure that it really helps. There was at least 1 run where it doesn't.  I disable the test for the github ci for linux.

ReimarBauer:
on my macbook it errors by repeating the tests seldom
```
=========== 575 passed, 19 skipped, 22 warnings in 102.21s (0:01:42) ===========
=========== 575 passed, 19 skipped, 22 warnings in 103.70s (0:01:43) ===========
=========== 575 passed, 19 skipped, 22 warnings in 98.99s (0:01:38) ============
=========== 575 passed, 19 skipped, 22 warnings in 97.16s (0:01:37) ============
=========== 575 passed, 19 skipped, 22 warnings in 99.06s (0:01:39) ============
=========== 575 passed, 19 skipped, 22 warnings in 101.82s (0:01:41) ===========
=========== 575 passed, 19 skipped, 22 warnings in 100.83s (0:01:40) ===========
=========== 575 passed, 19 skipped, 22 warnings in 98.31s (0:01:38) ============
=========== 575 passed, 19 skipped, 22 warnings in 99.12s (0:01:39) ============
= 1 failed, 574 passed, 19 skipped, 22 warnings, 1 error in 116.97s (0:01:56) ==
=========== 575 passed, 19 skipped, 22 warnings in 100.84s (0:01:40) ===========
=========== 575 passed, 19 skipped, 22 warnings in 97.20s (0:01:37) ============
```

matrss (approved):

PR #2959, opened 2025-12-17
use a fixture for mss_dir to set a different path for tests

### LINKED ISSUE #2958: we need to change the fixture to have a tmp_path for mss_dir
> When this occures, rather seldom, it looks like https://github.com/Open-MSS/MSS/blob/develop/tests/constants.py was not used.
> 
> We maybe should find a different solution than using the python namespace.
> 
> It is not the constants definition, rather the msui_settings.json or its defaults. Only there is a ~/mss.
> 
> On tests that should not be always the same dir. 

 _Originally posted by @ReimarBauer in [#2947](https://github.com/Open-MSS/MSS/issues/2947#issuecomment-3661398470)_

ReimarBauer:
**Purpose of PR?**:

Fixes #2958

### DIFF
--- a/tests/fixtures.py
+++ b/tests/fixtures.py
@@ -71,7 +71,12 @@ def close_remaining_widgets():
 
 
 @pytest.fixture
-def qtbot(qtbot, fail_if_open_message_boxes_left, close_remaining_widgets):
+def msui_configs(tmp_path):
+    modify_config_file({"mss_dir": str(tmp_path)})
+
+
+@pytest.fixture
+def qtbot(qtbot, fail_if_open_message_boxes_left, close_remaining_widgets, msui_configs):
     """Fixture that re-defines the qtbot fixture from pytest-qt with additional checks."""
     yield qtbot
     # Wait for a while after the requesting test has finished. At time of writing this


tests/fixtures.py
```diff
@@ -191,16 +196,44 @@ def _running_eventlet_server(app):
...
     scheme = "http"
     host = "127.0.0.1"
-    socket = eventlet.listen((host, 0))
-    port = socket.getsockname()[1]
-    url = f"{scheme}://{host}:{port}"
-    app.config['URL'] = url
+
     if "fork" not in multiprocessing.get_all_start_methods():
         pytest.skip("requires the multiprocessing start_method 'fork', which is unavailable on this system")
+
```
ReimarBauer: that change enables that I can run tests on my macbook

tests/fixtures.py
```diff
@@ -141,6 +141,11 @@ def reset_mscolab(mscolab_session_app):
         handle_db_reset()
 
 
+@pytest.fixture(scope="session")
+def msui_configs(tmp_path):
+    modify_config_file({"mss_dir": tmp_path.name})
+
```
ReimarBauer: this change alone gives on my macbook a lot errors

tests/fixtures.py
```diff
@@ -141,6 +141,11 @@ def reset_mscolab(mscolab_session_app):
         handle_db_reset()
 
 
+@pytest.fixture(scope="session")
```
matrss: Shouldn't this temporary directory be newly created for every test to avoid implicit dependencies?

tests/fixtures.py
```diff
@@ -191,16 +196,44 @@ def _running_eventlet_server(app):
...
-    url = f"{scheme}://{host}:{port}"
-    app.config['URL'] = url
+
     if "fork" not in multiprocessing.get_all_start_methods():
         pytest.skip("requires the multiprocessing start_method 'fork', which is unavailable on this system")
+
     ctx = multiprocessing.get_context("fork")
-    process = ctx.Process(target=eventlet.wsgi.server, args=(socket, app), daemon=True)
+
+    def _serve_in_child(conn, _app, _host, _scheme):
```
matrss: Starting parameters with an underscore usually indicates that they are unused AFAIU.

tests/fixtures.py
```diff
@@ -191,16 +196,44 @@ def _running_eventlet_server(app):
...
+        # Create the listening socket inside the child to avoid passing GreenSocket/FD across fork.
+        sock = eventlet.listen((_host, 0))
+        port = sock.getsockname()[1]
+        url = f"{_scheme}://{_host}:{port}"
+
+        try:
+            _app.config["URL"] = url
+        except Exception:
+            # If app/config is not writable for some reason, still start server.
+            pass
```
matrss: This is bad because now you can't rely on the correctness of this config value. If it is used somewhere then it must be set properly, if it is not used then we should never set it, IMO.

matrss:
Are you sure the changes in d67169f2f6de6038a36a3d12524169b029b34e78 are necessary? CI passes without them on macOS runners with the usual (un-)reliability.

tests/fixtures.py
ReimarBauer: oh, yes, good catch

ReimarBauer:
> Are you sure the changes in [d67169f](https://github.com/Open-MSS/MSS/commit/d67169f2f6de6038a36a3d12524169b029b34e78) are necessary? CI passes without them on macOS runners with the usual (un-)reliability.

I recheck, I had first tried a PR with the first none optimal change. Lets see.

ReimarBauer:
on my macOS 26.2 system I get currently

many "RuntimeError: Server did not start within 5 seconds"

18 passed, 16 skipped, 67 warnings, 358 errors in 84.93s (0:01:24)
I will retry after a fresh reboot.

ReimarBauer:
```
=============================================================== short test summary info ===============================================================
ERROR tests/_test_msui/test_wms_control.py::Test_HSecWMSControlWidget::test_server_service_cache - RuntimeError: Server did not start within 5 seconds at http://127.0.0.1:52141
ERROR tests/_test_msui/test_wms_control.py::Test_HSecWMSControlWidget::test_multilayer_handling - RuntimeError: Server did not start within 5 seconds at http://127.0.0.1:52141
ERROR tests/_test_msui/test_wms_control.py::Test_HSecWMSControlWidget::test_filter_handling - RuntimeError: Server did not start within 5 seconds at http://127.0.0.1:52141
ERROR tests/_test_msui/test_wms_control.py::Test_HSecWMSControlWidget::test_multilayer_syncing - RuntimeError: Server did not start within 5 seconds at http://127.0.0.1:52141
ERROR tests/_test_msui/test_wms_control.py::Test_HSecWMSControlWidget::test_server_getmap - RuntimeError: Server did not start within 5 seconds at http://127.0.0.1:52141
ERROR tests/_test_msui/test_wms_control.py::Test_HSecWMSControlWidget::test_invalid_url - RuntimeError: Server did not start within 5 seconds at http://127.0.0.1:52141
ERROR tests/_test_msui/test_wms_control.py::Test_HSecWMSControlWidget::test_no_server - RuntimeError: Server did not start within 5 seconds at http://127.0.0.1:52141
ERROR tests/_test_msui/test_wms_control.py::Test_HSecWMSControlWidget::test_invalid_schema - RuntimeError: Server did not start within 5 seconds at http://127.0.0.1:52141
ERROR tests/_test_msui/test_wms_control.py::Test_HSecWMSControlWidget::test_singlelayer_handling - RuntimeError: Server did not start within 5 seconds at http://127.0.0.1:52487
ERROR tests/_test_msui/test_wms_control.py::Test_HSecWMSControlWidget::test_no_schema - RuntimeError: Server did not start within 5 seconds at http://127.0.0.1:52487
ERROR tests/_test_msui/test_wms_control.py::Test_HSecWMSControlWidget::test_server_getmap_cached - RuntimeError: Server did not start within 5 seconds at http://127.0.0.1:52487
ERROR tests/_test_msui/test_wms_control.py::Test_HSecWMSControlWidget::test_server_no_thread - RuntimeError: Server did not start within 5 seconds at http://127.0.0.1:52487
ERROR tests/_test_msui/test_sideview.py::Test_SideViewWMS::test_server_getmap - RuntimeError: Server did not start within 5 seconds at http://127.0.0.1:52141
========================================= 563 passed, 16 skipped, 92 warnings, 13 errors in 110.51s (0:01:50) =========================================
```

Here we see an attempt to use 8 times 52141 and 4 times 52487

ReimarBauer:
Another run 567 passed, 16 skipped, 92 warnings, 9 errors in 113.76s (0:01:53)

```
ERROR tests/_test_msui/test_wms_control.py::Test_HSecWMSControlWidget::test_invalid_url - RuntimeError: Server did not start within 5 seconds at http://127.0.0.1:54919
ERROR tests/_test_msui/test_wms_control.py::Test_HSecWMSControlWidget::test_no_server - RuntimeError: Server did not start within 5 seconds at http://127.0.0.1:54919
ERROR tests/_test_msui/test_linearview.py::Test_LinearViewWMS::test_server_getmap - RuntimeError: Server did not start within 5 seconds at http://127.0.0.1:54977
ERROR tests/_test_msui/test_wms_control.py::Test_VSecWMSControlWidget::test_server_getmap - RuntimeError: Server did not start within 5 seconds at http://127.0.0.1:55162
ERROR tests/_test_msui/test_wms_control.py::Test_VSecWMSControlWidget::test_multilayer_drawing - RuntimeError: Server did not start within 5 seconds at http://127.0.0.1:55162
ERROR tests/_test_msui/test_wms_control.py::Test_HSecWMSControlWidget::test_invalid_schema - RuntimeError: Server did not start within 5 seconds at http://127.0.0.1:55162
ERROR tests/_test_msui/test_wms_control.py::Test_HSecWMSControlWidget::test_filter_handling - RuntimeError: Server did not start within 5 seconds at http://127.0.0.1:55162
ERROR tests/_test_msui/test_sideview.py::Test_SideViewWMS::test_server_getmap - RuntimeError: Server did not start within 5 seconds at http://127.0.0.1:55162
ERROR tests/_test_msui/test_topview.py::Test_TopViewWMS::test_server_getmap - RuntimeError: Server did not start within 5 seconds at http://127.0.0.1:55359
```

Maybe the problem is that not each test has a different port? here we see 2 times 54919, 5 times 55162

matrss:
Port selection is not the issue, the OS selects a free port for us which should be race-free even in the presence of many unrelated processes. The reason why multiple errors show the same port number is because those tests share the same fixture instance for the server. E.g. a mswms server fixture is requested at the class-level here: https://github.com/Open-MSS/MSS/blob/27a2fb1ad46905afc798af8d0a3b07bc58cbb9bd/tests/_test_msui/test_wms_control.py#L52-L56. This server instance is shared between all tests in that class, or rather its subclasses (but not /between/ subclasses, it seems).

matrss:
FWIW I know that it passes in CI because I tried it here: https://github.com/matrss/MSS/actions/runs/20426406249/job/58694840047

It just required a bunch of restarts of the macOS 14 tests, which seem to be the most unreliable. But I am not (yet) convinced that the additional changes made it any better.

ReimarBauer:
> FWIW I know that it passes in CI because I tried it here: https://github.com/matrss/MSS/actions/runs/20426406249/job/58694840047
> 
> It just required a bunch of restarts of the macOS 14 tests, which seem to be the most unreliable. But I am not (yet) convinced that the additional changes made it any better.

On github there is usually not a ~/mss for the account running the tests. But for a user trying the mswms with demodata has some python config files there. 
A user trying mscolab does have there also a colabdata. 

Tests should not have the option to change user data or can get different results because of userdata.

matrss:
I meant the tests pass in CI without the server startup modifications but with the change to the mss_dir configuration, so I don't quite understand your comment...

joernu76 (dismissed):

ReimarBauer:
> I meant the tests pass in CI without the server startup modifications but with the change to the mss_dir configuration, so I don't quite understand your comment...

The tests always pass on GitHub, with or without the change, because we always have a new CI environment there. However, when I run tests locally – which I do frequently – the tests can branch into my local ~/mss directory as observed. In the WayPointsTable and others, we use config_loader(dataset="mss_dir") for the data_dir. If we don't set mss_dir differently, it will use ~/mss.

Let’s treat the problems macOS 26.2 is having with the tests as separate from the related issue. That should be a different issue altogether. It wasn’t a good idea of mine to mix those together here.

matrss:
> The tests always pass on GitHub, with or without the change, because we always have a new CI environment there. However, when I run tests locally – which I do frequently – the tests can branch into my local ~/mss directory as observed. In the WayPointsTable and others, we use config_loader(dataset="mss_dir") for the data_dir. If we don't set mss_dir differently, it will use ~/mss.

Sure, I get that, but how does this relate to what I said? I didn't say we shouldn't set this path to a temporary directory...

> Let’s treat the problems macOS 26.2 is having with the tests as separate from the related issue. That should be a different issue altogether. It wasn’t a good idea of mine to mix those together here.

Are there actually any macOS 26.2 specific issues here? AFAICS there is the long-standing issue that macOS tests in general are highly unreliable irrespective of version, but that is unrelated to this PR (and wasn't solved by the server startup modifications that you reverted in the meantime). FWIW it's an approve from me as soon as I see the macOS tests pass just once.

ReimarBauer:
> > The tests always pass on GitHub, with or without the change, because we always have a new CI environment there. However, when I run tests locally – which I do frequently – the tests can branch into my local ~/mss directory as observed. In the WayPointsTable and others, we use config_loader(dataset="mss_dir") for the data_dir. If we don't set mss_dir differently, it will use ~/mss.
> 
> Sure, I get that, but how does this relate to what I said? I didn't say we shouldn't set this path to a temporary directory...
> 
> > Let’s treat the problems macOS 26.2 is having with the tests as separate from the related issue. That should be a different issue altogether. It wasn’t a good idea of mine to mix those together here.
> 
> Are there actually any macOS 26.2 specific issues here? AFAICS there is the long-standing issue that macOS tests in general are highly unreliable irrespective of version, but that is unrelated to this PR (and wasn't solved by the server startup modifications that you reverted in the meantime). FWIW it's an approve from me as soon as I see the macOS tests pass just once.

My observations are limited to my local environment. After the updates to libgfortran5 and NumPy, I have not observed any differences in daily usage when compared to Ubuntu. Testing no longer hangs. However, it's displaying numerous errors. The amount has changed.

Are you tying this change to the outcome of an unrelated test that's failing intermittently and for unclear reasons?

ReimarBauer:
> FWIW I know that it passes in CI because I tried it here: https://github.com/matrss/MSS/actions/runs/20426406249/job/58694840047
> 
> It just required a bunch of restarts of the macOS 14 tests, which seem to be the most unreliable. But I am not (yet) convinced that the additional changes made it any better.

I've realized it wasn't a good idea after all. It would have been nice to fix it with something along those lines, but that was not the solution.

ReimarBauer:
@matrss Do you see a problem if I define it as (autouse=True)? I'm not sure if it's just another coincidence, but then at least I can run test_open_chat_window without it timing out.

matrss:
Ahhh, the way it is written now this fixture actually doesn't change anything at all because nothing is requesting it. I don't think making it autouse is a good idea though. Since it only affects msui I think you should add it as another argument to the qtbot fixture (like how it's done for fail_if_open_message_boxes_left and close_remaining_widgets).

tests/fixtures.py
```diff
@@ -70,8 +70,13 @@ def close_remaining_widgets():
             pass
 
 
+@pytest.fixture()
+def msui_configs(tmp_path):
+    modify_config_file({"mss_dir": tmp_path.name})
```
matrss: Just noticed another issue: `tmp_path.name` only contains the directory name of the temporary directory, not the full path. To actually use the created temporary directory this should be `str(tmp_path)` instead.

ReimarBauer: ok.

matrss (approved):

PR #2965, opened 2025-12-28
show revision id in view and plot titles

### LINKED ISSUE #2026: show on the Views which revision is used
Currently we see only the operation name on the various views. We maybe should also show an information from the revision used.

If a revision is named, that name should be shown. On the plot too :)

That would help to see, what last was checked out or if something changed.

spiel_op

aman-coder03:
### summary
always show the revision ID in view window titles and plot titles.
if a revision name exists, it is appended.

### motivation
improves traceability when switching revisions or comparing views.

Fixes #2026

### DIFF
--- a/mslib/msui/flighttrack.py
+++ b/mslib/msui/flighttrack.py
@@ -62,11 +62,13 @@ xml.dom.minidom.Element.writexml = writexml  # nosec, we take care of writing co
 LOCATION, LAT, LON, FLIGHTLEVEL, PRESSURE = list(range(5))
 TIME_UTC = 9
 
+
 class Revision:
     """
     Represents a revision of a flight track.
     ID is mandatory, name is optional.
     """
+
     def __init__(self, revision_id, name=None):
         self.id = revision_id
         self.name = name
@@ -193,7 +195,7 @@ class WaypointsTableModel(QtCore.QAbstractTableModel):
 
     def __init__(self, name="", filename=None, waypoints=None, mscolab_mode=False,
                  data_dir=config_loader(dataset="mss_dir"),
-                 xml_content=None): 
+                 xml_content=None):
         super().__init__()
         self.revision = None
         self.name = name  # a name for this flight track
@@ -218,12 +220,12 @@ class WaypointsTableModel(QtCore.QAbstractTableModel):
 
         if waypoints:
             self.replace_waypoints(waypoints)
-            
+
         if self.revision is None:
             self.revision = Revision(
                 revision_id=int(datetime.datetime.utcnow().timestamp()),
                 name=None
-            )    
+            )
 
     def load_settings(self):
         """
@@ -686,7 +688,6 @@ class WaypointsTableModel(QtCore.QAbstractTableModel):
 
         return doc
 
-
     def get_xml_content(self):
         doc = self.get_xml_doc()
         return doc.toprettyxml(indent="  ", newl="\n")
@@ -729,7 +730,6 @@ class WaypointsTableModel(QtCore.QAbstractTableModel):
             self.replace_waypoints(_waypoints_list)
         else:
             raise SyntaxError(f"Invalid flight track filename: {name}")
-
 
 
 #
--- a/mslib/msui/flighttrack.py
+++ b/mslib/msui/flighttrack.py
@@ -733,7 +733,7 @@ class WaypointsTableModel(QtCore.QAbstractTableModel):
 
     def get_filename(self):
         return self.filename
-    
+
 #
 # CLASS  WaypointDelegate
 #
--- a/mslib/msui/flighttrack.py
+++ b/mslib/msui/flighttrack.py
@@ -142,12 +142,17 @@ class Waypoint:
 
     def __init__(self, lat=0., lon=0., flightlevel=0., location="", comments=""):
         self.location = location
-        locations = config_loader(dataset='locations')
-        if location in locations:
-            self.lat, self.lon = locations[location]
-        else:
-            self.lat = lat
-            self.lon = lon
+
+        # Always trust explicitly provided coordinates first
+        self.lat = lat
+        self.lon = lon
+
+        # Only resolve location name if coordinates are not provided
+        if (lat == 0.0 and lon == 0.0) and location:
+            locations = config_loader(dataset='locations')
+            if location in locations:
+                self.lat, self.lon = locations[location]
+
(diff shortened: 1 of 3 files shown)


mslib/mscolab/mscolab.py
```diff
@@ -58,7 +58,10 @@ def handle_start(args=None):
     start_server(APP, sockio, cm, fm)
 
 
-def confirm_action(confirmation_prompt):
+def confirm_action(confirmation_prompt, assume_yes=False):
+    if assume_yes:
```
ReimarBauer, line 62: please separate this. You have this in a different PR.

mslib/mscolab/mscolab.py
```diff
@@ -375,17 +375,27 @@ def main():
...
-    database_parser.add_argument("--users_by_file", type=argparse.FileType('r'),
-                                 help="adds users into database, fileformat: suggested_username  name   <email>")
-    database_parser.add_argument("--delete_users_by_file", type=argparse.FileType('r'),
-                                 help="removes users from the database, fileformat: email")
-    database_parser.add_argument("--default_operation", help="adds all users into a default TEMPLATE operation",
-                                 action="store_true")
-    database_parser.add_argument("--add_all_to_all_operation", help="adds all users into all other operations",
-                                 action="store_true")
+
+    db_actions = database_parser.add_mutually_exclusive_group(required=True)
```
ReimarBauer, line 382: This is not part of the issue. 

We split tasks and also PRs do handle only 1 issue.

aman-coder03:
@ReimarBauer i have removed the `mscolab.py` from this PR
i hope this is more clear now!

ReimarBauer:
Please show the ID, this will always work, because a name can be shifted over IDs. If there is a name at an ID then show it too.

The feature is interesting for those comparing images, it gives the possibility to see that there was a change.

ReimarBauer:
When I currently try it I can't see the result, have you a different image?

Bildschirmfoto 2026-01-02 um 10 33 20

ReimarBauer:
While looking up the servers retrieved data I recognized that we don't have revision information yet implemented. So that is not an easy task.

aman-coder03:
thanks for checking this and for the clarification.
my intention with this PR was to prepare the UI side to display the revision ID.
but I see now that without server-side support the change is not visible yet.
i would be happy to work on that part(server or data flow), so the UI can be fitted properly.

ReimarBauer:
Please keep in mind that we only merge when a feature is completely integrated. Otherwise, someone during review might remove code that’s dead (incomplete).

If you want to start working on the server-side code, begin with everything that uses ftml code. ftml is the default format for flighttracks and mscolab. Extend it so that it also works for flight tracks without mscolab. For example, `load_from_xml_data`, `save_to_ftml`, `load_from_xml_data`.  For mscolab, you'll then need to adjust the `seed` for demo and tests.

There’s an example at `samples/examples.ftml` which is used by `test_msui`. The tests also require adaptation. All changes must be accompanied by tests so we don’t lose any existing functionality.  This is quite extensive, but probably not difficult to implement.

aman-coder03:
thanks for the detailed guidance @ReimarBauer 
i will extend the FTML format to include revision information (ID mandatory, name optional), update the loaders/savers so it works both with and without `MSColab`, adapt the demo seed and tests (including `samples/examples.ftml` and `test_msui`), and then apply the UI changes on top.
i will keep everything in this PR and update it step by step with accompanying tests.

ReimarBauer (changes requested):
We are leaving the pull request open for further updates, which will be posted here. Therefore, I am requesting changes.

aman-coder03:
@ReimarBauer i have now fully implemented revision support at the FTML level
The FTML format was extended with a `<Revision id="…" name="…"/>` element (ID mandatory, name optional)
The loader (`load_from_xml_data`) and saver (`get_xml_doc`) both support this with backward compatibility for legacy FTML
i have added tests that cover FTML with revision and legacy FTML without revision
please let me know if you would like any adjustments or additional coverage

mslib/msui/flighttrack.py
```diff
@@ -671,15 +702,35 @@ def load_from_ftml(self, filename):
...
+
+        # Existing waypoint loading
         if verify_waypoint_data(xml_content):
             _waypoints_list = load_from_xml_data(xml_content, name)
             self.replace_waypoints(_waypoints_list)
         else:
             raise SyntaxError(f"Invalid flight track filename: {name}")
 
-    def get_filename(self):
-        return self.filename
```
ReimarBauer, line 679: why was this removed?

ReimarBauer (changes requested):
see tests and comment

aman-coder03:
@ReimarBauer can you please specify what should be changed in tests?

ReimarBauer:
> @ReimarBauer can you please specify what should be changed in tests?

When you remove a function you can likly also see by a test that it was needed too ;)
here is also a
[syntax error
](https://github.com/Open-MSS/MSS/actions/runs/20918590131/job/60098155390?pr=2965#step:4:1506)

Which model have you used for support?

aman-coder03:
@ReimarBauer can you help me out regarding this?

aman-coder03:
@ReimarBauer it would be really helpful if you could help me out!

ReimarBauer:
Please give me a hint what I should look at.
When you mean the randomly timeout of tests. That is something I do lookup already.

aman-coder03:
the macos-15 test is getting passed but macos-14 test is not getting passed

aman-coder03:
@ReimarBauer now all the tests are successful!
you can have a look!

ReimarBauer:
> @ReimarBauer now all the tests are successful! you can have a look!

Thx, I do soon.

mslib/msui/flighttrack.py
```diff
@@ -143,12 +143,17 @@ class Waypoint:
     def __init__(self, lat=0., lon=0., flightlevel=0., location="", comments=""):
         self.location = location
 
-        locations = config_loader(dataset='locations')
-        if location in locations:
-            self.lat, self.lon = locations[location]
-        else:
+        # Prefer explicitly provided coordinates (e.g., from server/XML)
+        if lat != 0.0 or lon != 0.0:
```
ReimarBauer: the tests fails for other reasons. When they  get a timeout then an update of the data can not be done.

We should not have a test based solution which we don't have using the software.


the configured values should be the only truth to start here.

PR #2978, opened 2026-01-14
update basemap in our toml

### LINKED ISSUE #2977: passlib has a dependency to crypt and crypt slated for removal in Python 3.13
/home/runner/work/MSS/MSS/.pixi/envs/dev/lib/python3.11/site-packages/passlib/utils/__init__.py:854: DeprecationWarning: 'crypt' is deprecated and slated for removal in Python 3.13
    from crypt import crypt as _crypt

We need to look at our pinning or find a different solution

ReimarBauer:
**Purpose of PR?**:

Fixes #2977 

update our stack


ReimarBauer:
When I install in a newer py version it is updated. I think we don't need to change that.

PR #2983, opened 2026-01-17
using the mss_dir in tests with verification

### LINKED ISSUE #2958: we need to change the fixture to have a tmp_path for mss_dir

ReimarBauer:
**Purpose of PR?**:

Fixes #2958

### DIFF
--- a/mslib/support/qt_json_view/datatypes.py
+++ b/mslib/support/qt_json_view/datatypes.py
@@ -366,7 +366,9 @@ class UrlType(DataType):
 class FilepathType(DataType):
     """Files and paths can be opened."""
 
-    REGEX = re.compile(r'(\/.*)|([A-Z]:\\.*)')
+    # REGEX = re.compile(r'(\/.*)|([A-Z]:\\.*)')
+    # (mss)
+    REGEX = re.compile(r'(~\/.*)|(\/.*)|([A-Z]:\\.*)')
 
     def matches(self, data):
         # (mss)
--- a/tests/_test_msui/test_msui.py
+++ b/tests/_test_msui/test_msui.py
@@ -51,8 +51,9 @@ def test_main():
 
 
 def test_keep_config_file(qtbot):
-    # in conftest we set always the mss_dir in the config file
+    # in fixtures we set the mss_dir in the config file
     mss_dir = config_loader(dataset="mss_dir")
+    assert "~" not in mss_dir
     _config = MSUI_CONFIG_FILE_PATH.read_text()
     assert _config == f'''{{
     "mss_dir": "{mss_dir}"
--- a/tests/fixtures.py
+++ b/tests/fixtures.py
@@ -23,6 +23,8 @@
     See the License for the specific language governing permissions and
     limitations under the License.
 """
+import os
+
 import pytest
 import mock
 import multiprocessing
@@ -72,6 +74,7 @@ def close_remaining_widgets():
 
 @pytest.fixture
 def msui_configs(tmp_path):
+    os.environ['MSUI_SETTINGS'] = os.path.join(tmp_path, "msui_settings.json")
     modify_config_file({"mss_dir": str(tmp_path)})
 
 


ReimarBauer:
This needs first the merge stable to develop

mslib/support/qt_json_view/datatypes.py
```diff
@@ -366,7 +366,9 @@ def actions(self, index):
 class FilepathType(DataType):
     """Files and paths can be opened."""
 
-    REGEX = re.compile(r'(\/.*)|([A-Z]:\\.*)')
+    # REGEX = re.compile(r'(\/.*)|([A-Z]:\\.*)')
+    # (mss)
+    REGEX = re.compile(r'(~\/.*)|(\/.*)|([A-Z]:\\.*)')
```
ReimarBauer, line 372: without this change the used value gets ignored and the default is then used

tests/_test_msui/test_msui.py
```diff
@@ -51,8 +51,9 @@ def test_main():
 
 
 def test_keep_config_file(qtbot):
-    # in conftest we set always the mss_dir in the config file
+    # in fixtures we set the mss_dir in the config file
     mss_dir = config_loader(dataset="mss_dir")
+    assert "~" not in mss_dir
```
ReimarBauer, line 56: safety check, that the value was changed

tests/fixtures.py
```diff
@@ -72,6 +74,7 @@ def close_remaining_widgets():
 
 @pytest.fixture
 def msui_configs(tmp_path):
+    os.environ['MSUI_SETTINGS'] = os.path.join(tmp_path, "msui_settings.json")
```
ReimarBauer, line 77: msui imports from here

https://github.com/Open-MSS/MSS/blob/develop/mslib/msui/constants.py#L43

The environment variable adds (un?)necessary complexity and could be overlooked.

ReimarBauer:
@matrss 
Ubuntu is experiencing recurring failures, which will likely result in an xfail on https://github.com/Open-MSS/MSS/pull/2957 or the change supports finding the problem.
The macOS tests don't appear to be having any issues.

tests/fixtures.py
matrss: This does not work as intended:

```
os.environ['MSUI_SETTINGS']='/tmp/pytest-of-icg149/pytest-4/test_keep_config_file0/msui_settings.json'
mslib.msui.constants.MSUI_SETTINGS=PosixPath('/tmp/tmp659x2mo4/msui/msui_settings.json')
mss_dir='/tmp/pytest-of-icg149/pytest-4/test_keep_config_file0'
```

(diff:
```
diff --git a/tests/_test_msui/test_msui.py b/tests/_test_msui/test_msui.py
index c1505023..f558f8b4 100644
--- a/tests/_test_msui/test_msui.py
+++ b/tests/_test_msui/test_msui.py
@@ -53,6 +53,9 @@ def test_main():
 def test_keep_config_file(qtbot):
     # in fixtures we set the mss_dir in the config file
     mss_dir = config_loader(dataset="mss_dir")
+    import sys
+    import mslib.msui.constants
+    print(f"{os.environ['MSUI_SETTINGS']=}\n{mslib.msui.constants.MSUI_SETTINGS=}\n{mss_dir=}", file=sys.stderr)
     assert "~" not in mss_dir
     _config = MSUI_CONFIG_FILE_PATH.read_text()
     assert _config == f'''{{
```
)

matrss: I assume this is because the constants module is imported somewhere before the fixture is executed. IMO these constants modules are a mistake.

ReimarBauer:
When refactoring the server code to use blueprints and we use an application factory https://flask.palletsprojects.com/en/stable/tutorial/factory/#the-application-factory we likly have also fixed this problem https://github.com/Open-MSS/MSS/issues/2958

PR #3000, opened 2026-02-18
Merge develop to stable - this will become 11.0.0

ReimarBauer:
**Purpose of PR?**:

This is the base for the next stable release


docs/environment.yml
```diff
@@ -2,8 +2,10 @@ name: rtd
 channels:
   - conda-forge
 dependencies:
-  - python
-  - setuptools <81
+  - python<=3.12
```
ReimarBauer, line 5: just seen that readthedocs runs now on 3.14

We are not compatible to that

verified with a local setup similiar as readthedocs did

joernu76 (approved):
This is too much to properly review. every detail, in particular without the associated issues.
The changes look largely familiar and the merged code runs locally, so I would say we can proceed. Earlier smaller merges would be preferable.

matrss:
> This is too much to properly review. every detail, in particular without the associated issues. The changes look largely familiar and the merged code runs locally, so I would say we can proceed. Earlier smaller merges would be preferable.

Agreed. IMO having two development branches is still the fundamentally wrong approach, reviewing with one development branch and backport PRs would be much easier.

The develop to stable merge can be reviewed completely mechanically though: the result should contain the same content as the develop branch. In this case:

```diff
$ git diff develop
diff --git a/docs/environment.yml b/docs/environment.yml
index ccbe100d..7c155a13 100644
--- a/docs/environment.yml
+++ b/docs/environment.yml
@@ -2,8 +2,10 @@ name: rtd
 channels:
   - conda-forge
 dependencies:
-  - python
-  - setuptools <81
+  - python<=3.12
+  - setuptools<81
+  - xstatic-jquery
+  - xstatic-bootstrap
   - pip
   - pip:
       - flask
@@ -15,8 +17,6 @@ dependencies:
       - metpy
       - markdown
       - xstatic>=1.0.3
-      - xstatic-jquery
-      - xstatic-bootstrap
       - defusedxml
       - sphinx_rtd_theme
       - sphinxcontrib-video>=0.2.1
```

I think these differences are intentional and make sense?

The stable to develop merges are much worse, since they can contain multiple unrelated changes and each of them has to be reviewed for if they still apply the intended change on top of the changed base branch without having the context of each of those PRs and without easily seeing which line belonged to which PR.

ReimarBauer:
> > This is too much to properly review. every detail, in particular without the associated issues. The changes look largely familiar and the merged code runs locally, so I would say we can proceed. Earlier smaller merges would be preferable.
> 
> Agreed. IMO having two development branches is still the fundamentally wrong approach, reviewing with one development branch and backport PRs would be much easier.

This means we keep the stable branch and the only difference is that we do backports, instead of merges? 
When the API is different for the backport, we won't do a backport?, so the bug stays forever?
I think that would be a good approach for minor or patch updates. But how about major releases – how are those handled?
> 
> The develop to stable merge can be reviewed completely mechanically though: the result should contain the same content as the develop branch. In this case:
> 
> ```diff
> $ git diff develop
> diff --git a/docs/environment.yml b/docs/environment.yml
> index ccbe100d..7c155a13 100644
> --- a/docs/environment.yml
> +++ b/docs/environment.yml
> @@ -2,8 +2,10 @@ name: rtd
>  channels:
>    - conda-forge
>  dependencies:
> -  - python
> -  - setuptools <81
> +  - python<=3.12
> +  - setuptools<81
> +  - xstatic-jquery
> +  - xstatic-bootstrap
>    - pip
>    - pip:
>        - flask
> @@ -15,8 +17,6 @@ dependencies:
>        - metpy
>        - markdown
>        - xstatic>=1.0.3
> -      - xstatic-jquery
> -      - xstatic-bootstrap
>        - defusedxml
>        - sphinx_rtd_theme
>        - sphinxcontrib-video>=0.2.1
> ```
> 
> I think these differences are intentional and make sense?

This is an intermediate step, and it's along the lines of your backport suggestion. I can implement this in current stable, merge it into develop, and then bring it all together. 
We need to find out why the new PyPI package isn’t playing nicely.  I also want to understand the problem first before I update the package on conda-forge. That explains why using conda-forge is working for now. 

We also need a [CI Script](https://github.com/Open-MSS/MSS/issues/3001) which tries these dependencies for RTFD. 

> 
> The stable to develop merges are much worse, since they can contain multiple unrelated changes and each of them has to be reviewed for if they still apply the intended change on top of the changed base branch without having the context of each of those PRs and without easily seeing which line belonged to which PR.

joernu76:
As all individual changes were reviewed already, it is also up for discussion whether a content review is strictly necessary.

I agree with the analysis of the current state of affairs.
I see *all* development taking place in develop and *only* major/critical-bug/security hotfixes in stable, *ideally* as backport or with an accompanying MR to develop dealing with the same issue.
Thus there is a clear directional arrow.

Then, new non-minor releases would be effectively a force push from develop to stable.

ReimarBauer:
Then let’s try that after version 11. My concern is that we might end up doing more major releases and fewer patch or minor releases as a result of this change. If we take into account the effort required for server operators or users, we could probably achieve a similarly good outcome.

Do I have to do something different to get into that flow?
1. merge this one
1. add in develop version and CHANGES
1. backport or merge this to stable, tag stable

matrss:
> This means we keep the stable branch and the only difference is that we do backports, instead of merges?

Everything would go to develop first. A PR to develop can opt in to being backported (`git cherry-pick -x` on the stable branch). This backport PR can be automatically created if there is no conflict, otherwise it has to be created (and conflicts resolved) manually.

> When the API is different for the backport, we won't do a backport?, so the bug stays forever?

No, the patch just has to be different for both branches then, just like now. Except that both patches can be reviewed on their own, instead of being just a part of the stable to develop merge.

> But how about major releases – how are those handled?

`git branch release/vXY`. That branch is then the stable branch for this major version. Alternatively, if you really want it to be named "stable", it would require a force push, as Jörn said.

> As all individual changes were reviewed already, it is also up for discussion whether a content review is strictly necessary.

With the current workflow it is not necessary for the develop to stable merge, that's just a diff that should be empty. For the stable to develop merges a review is necessary though: as a simple example imagine a PR that updates the copyright years on stable. This is then part of the next stable to develop merge. Develop could have files that weren't present on stable, so their years wouldn't be updated. This is a mistake, because the merges job is to apply the intended change on the new target branch, i.e. update all copyright notices. A content review is necessary to catch this, especially in more complex cases.

ReimarBauer:
> > This means we keep the stable branch and the only difference is that we do backports, instead of merges?
> 
> Everything would go to develop first. A PR to develop can opt in to being backported (`git cherry-pick -x` on the stable branch). This backport PR can be automatically created if there is no conflict, otherwise it has to be created (and conflicts resolved) manually.

ack
> 
> > When the API is different for the backport, we won't do a backport?, so the bug stays forever?
> 
> No, the patch just has to be different for both branches then, just like now. Except that both patches can be reviewed on their own, instead of being just a part of the stable to develop merge.

ack

> 
> > But how about major releases – how are those handled?
> 
> `git branch release/vXY`. That branch is then the stable branch for this major version. Alternatively, if you really want it to be named "stable", it would require a force push, as Jörn said.
> 
I see the benefits of the workflow change as well. I think I’d like to keep the stable branch for now. We can discuss replacing it again with a later release.
So now we need to figure out how to perform this force push without requiring an admin override on the branch (on github).  Are there alternatives to merging, or replacing the stable branch entirely with a copy of develop?
Hmm, maybe we do it like debian, rename stable to oldstable, create a new branch stable from develop.


> > As all individual changes were reviewed already, it is also up for discussion whether a content review is strictly necessary.
> 
> With the current workflow it is not necessary for the develop to stable merge, that's just a diff that should be empty. For the stable to develop merges a review is necessary though: as a simple example imagine a PR that updates the copyright years on stable. This is then part of the next stable to develop merge. Develop could have files that weren't present on stable, so their years wouldn't be updated. This is a mistake, because the merges job is to apply the intended change on the new target branch, i.e. update all copyright notices. A content review is necessary to catch this, especially in more complex cases.

matrss:
> So now we need to figure out how to perform this force push without requiring an admin override on the branch (on github). Are there alternatives to merging, or replacing the stable branch entirely with a copy of develop?
> Hmm, maybe we do it like debian, rename stable to oldstable, create a new branch stable from develop.

Delete + recreate should work too, or a rename as you say.

matrss:
> Do I have to do something different to get into that flow?
> 
>     1. merge this one
> 
>     2. add in develop version and CHANGES
> 
>     3. backport or merge this to stable, tag stable

No, I think that is fine. For now we would essentially just have to stop merging anything from stable to develop, preferentially target all PRs against develop, and start cherry-picking the things that should go to stable too. Automating the backports isn't strictly necessary, that would just be a convenience upgrade.

ReimarBauer:
> > Do I have to do something different to get into that flow?
> > ```
> > 1. merge this one
> > 
> > 2. add in develop version and CHANGES
> > 
> > 3. backport or merge this to stable, tag stable
> > ```
> 
> No, I think that is fine. For now we would essentially just have to stop merging anything from stable to develop, preferentially target all PRs against develop, and start cherry-picking the things that should go to stable too. Automating the backports isn't strictly necessary, that would just be a convenience upgrade.

My plan is to first fix the RTFD issue in develop (PR #3002) and then create a new PR with updated documentation, versioning, and changes. After both are merged, I'll try renaming/recreating the stable branch. If that works, I’ll tag the commit and move forward with the release. This pull request doesn't cover everything needed for the full release. So lets close it :)

PR #3002, opened 2026-02-20
CI for checking pixi env on RTFD will succeed

### LINKED ISSUE #3001: add a test ci script which covers a documentation build on readthedocs
These are the commands I do see in the log of the build. We need a CI verifying that our setup in principle can succeed on RTFD. 

```
git clone --depth 1 https://github.com/Open-MSS/MSS .
git fetch origin --force --prune --prune-tags --depth 50 refs/heads/stable:refs/remotes/origin/stable
git checkout --force origin/stable
cat .readthedocs.yaml
asdf global python mambaforge-4.10.3-10
cat docs/environment.yml
mamba env create --quiet --name stable --file docs/environment.yml
```

ReimarBauer:
**Purpose of PR?**:
This is currently the test setup for showing the problem which happens on RTFD

Fixes #3001


matrss:
It is possible to use a pixi environment on readthedocs: <https://docs.readthedocs.com/platform/stable/build-customization.html#install-dependencies-with-pixi>.

IMO it would be better to use that instead and get rid of conda/mamba/environment.yml entirely.

ReimarBauer:
> It is possible to use a pixi environment on readthedocs: https://docs.readthedocs.com/platform/stable/build-customization.html#install-dependencies-with-pixi.
> 
> IMO it would be better to use that instead and get rid of conda/mamba/environment.yml entirely.

seems to be possible.

joernu76 (approved):

matrss:
The new docs environment duplicates a lot of dependencies unnecessarily. The default for environments is to include the named features as well as the default feature, so all dependencies of the default feature are already included. Some of the dependencies now request different versions. Also, I think all dependencies necessary for the docs feature should be part of the dev environment, so instead of duplicating them between dev and docs the dev environment could instead include the docs feature too.

PR #3004, opened 2026-02-20
prepare v11.0.0

ReimarBauer:
**Purpose of PR?**:

preparation for the next major release


docs/development.rst
```diff
@@ -484,7 +494,7 @@ Creating a new release
 * check version number of upcoming release in pyproject.toml
 * verify that pyproject.toml, MANIFEST.in are complete
 * Update the version in the lock file by `pixi lock`
-* for a new stable release merge from develop to stable
+* for a new major release rename the existing stable branch to oldstable and recreate it from the develop branch
```
joernu76: I am not sure what purpose the "oldstable" branch serves. Instead:
* Ensure that the latest commit of stable is released and tag'ged accordingly.
?

docs/development.rst
```diff
@@ -456,24 +457,33 @@ After filling the template completely click on Pull request
 
 
 
-Merging stable into develop
----------------------------
+Creating a new minor, patch level release
+-----------------------------------------
 
-Bug fixes we have done in stable we need to merge regularly into develop too::
+The stable branch will only receive critical bug fixes, and security patches – ideally with corresponding
```
joernu76: -> ideally as a backport from a related fix to the develop branch.

docs/development.rst
```diff
@@ -456,24 +457,33 @@ After filling the template completely click on Pull request
...
+  * Mailing list
+  * Mastodon (follow https://fosstodon.org/@MSS)
 
-Then create the proposed merge request. The merge request must *not* be squashed or rebased.
-To allow the merging, the requirement for a linear-history must be disabled *temporarily*
-for the develop branch and one needs to ensure that the merge request is accepted with a
-regular merge with merge commit. Remove the merge_stable_to_develop branch if still present.
 
 
 Creating a new release
```
joernu76: This list seems less complete than the list of AIs for the minor release above?

docs/development.rst
ReimarBauer: I’m not entirely sure how our branch protection will behave until I test it with our setup, or whether any adjustments are needed.  I'd like to avoid reconfiguring it repeatedly. So, I can rename branches – for instance, renaming 'stable' to 'oldstable'. If something goes wrong, I still have the original ‘stable’ branch and its configuration as a backup since it was just renamed. That allows me to try another approach. 'oldstable' is a temporary placeholder until 'stable' is working correctly with releases.

Once that’s done, I can document the process for inclusion in our next patch level release.

docs/development.rst
ReimarBauer: Let’s avoid making it exclusive. Develop is going to have API changes, so it might not be ready for release while we need a critical fix in production stable.  Releasing a new major when we can't do a backport would be preferable.

docs/development.rst
ReimarBauer: After making minor or patch-level changes, we can release directly from the stable branch without needing to rename it. (tag, release)

For a new major release I need to rename the existing stable on github and create it from the develop branch (git branch stable). The major difference is the rename.

docs/development.rst
matrss: > If something goes wrong, I still have the original ‘stable’ branch and its configuration as a backup since it was just renamed.

There is nothing special about a branch compared to a tag, except for the general convention that tags are treated as immutable (though technically they are not). If the latest commit of stable is released and tagged you can always recreate it from that.

docs/development.rst
matrss:
```suggestion
The stable branch will only receive critical bug fixes, and security patches. These patches must either address a bug that is only present on stable, be a backport of a patch that was already merged into develop, or there must simultaneously be a PR addressing the same issue on develop.
```

I want to avoid a situation in which a "quick hotfix" addresses an issue on stable, but no one feels responsible to fix the same issue on develop.

docs/development.rst
ReimarBauer: I know. but it is a step more. 
It’ll probably all work fine if I rename the branch rules branch to ‘oldstable’, and then rename or delete ‘stable’ before creating a “new” stable version. After that, I'll rename the rules back.  I'm just a little carefully of it right now, which isn't a bad thing.

docs/development.rst
```diff
@@ -456,24 +457,33 @@ After filling the template completely click on Pull request
...
 
 
-Merging stable into develop
----------------------------
+Creating a new minor, patch level release
+-----------------------------------------
 
-Bug fixes we have done in stable we need to merge regularly into develop too::
+The stable branch will only receive critical bug fixes, and security patches. These patches must either address a bug that is only present on stable, be a backport of a patch that was already merged into develop, or there must simultaneously be a PR addressing the same issue on develop.
+updates to the develop branch.
```
ReimarBauer: This sentence doesn't seem finished?

docs/development.rst
```diff
@@ -456,24 +457,33 @@ After filling the template completely click on Pull request
 
 
 
-Merging stable into develop
----------------------------
+Creating a new minor, patch level release
+-----------------------------------------
 
-Bug fixes we have done in stable we need to merge regularly into develop too::
+The stable branch will only receive critical bug fixes, and security patches. These patches must either address a bug that is only present on stable, be a backport of a patch that was already merged into develop, or there must simultaneously be a PR addressing the same issue on develop.
```
ReimarBauer: It’s also possible that the code is no longer present in the develop branch.  In other words, the bug is exclusive to stable.
It needs to be clarified in that situation. Ideas how to add that rule?

docs/development.rst
matrss: Oops, I must have missed deleting that part in my suggestion.

docs/development.rst
matrss: It is already in there:

> These patches must either address a bug that is only present on stable, [...]

joernu76 (approved):

PR #3010, opened 2026-02-25
CI: enforce newline at EOF and fix existing violations

arsh118:
**Purpose of PR?**:

Fixes #

**Checklist:**
- [ ] Commit has unit tests

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -55,6 +55,7 @@ jobs:
     timeout-minutes: 10
     steps:
     - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v4
+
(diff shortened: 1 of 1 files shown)


matrss (changes requested):
Hey! Thanks for the PR.

Unfortunately it doesn't address the issue sufficiently. We have some binary files in the repository that this check also finds, while the rule should only apply to text files.

Could you please explain why the check for the newline works? I get that you are reading out the last byte of each file, but I don't understand why `read -r _` exits with non-zero when that byte is not a newline. I couldn't find documentation for this behavior.

The first `-f` check seems redundant, since `git ls-files` should already only output file paths. The first read would also mangle some (admittedly very weird) file names containing e.g. backslashes.

PR #3012, opened 2026-02-25
avoid disruption of workflows when coveralls has an API outage

### LINKED ISSUE #3011: coverage outage
because of outage of coveralls.io add the suggested fail-on-error: false see https://status.coveralls.io/

ReimarBauer:
**Purpose of PR?**:

Fixes #3011


matrss (changes requested):
The fail-on-error missing in the finish job, so that one still fails.

matrss (approved):

PR #3016, opened 2026-02-27
Migrate app's settings to a conventional configuration method.

### LINKED ISSUE #2506: Use the flask-native global configuration store instead of rolling our own
Every Flask app has a `.config` attribute to store configuration. Additionally, for MSColab we have `mscolab_settings`. I think this is redundant, we could just use `app.config` for everything that should be made globally available in the application in a key-value fashion.

This would also become important once there is not just a singular global APP instance, but the possibility to instantiate more then one Flask app, as each app needs its own configuration. This is something that `mscolab_settings` is not set up for ATM.

This came up in https://github.com/Open-MSS/MSS/pull/2499#discussion_r1740502760.

JanEisermann:
**Purpose of PR?**:

Fixes #2506

rename of variables, using app.config instead of mscolab_settings and mswms_settings, new way of using index.py methods
tests succeed, not manually verified 

documentation is also changed

**Checklist:**
- [x] Commit has unit tests


ReimarBauer (changes requested):
Welcome @JanEisermann thanks for taking this one :)

I do want to release 11 in the next days. And this change should be verified also as an human user.  

I decided therefore that it is on hold until 11 is released

mslib/mscolab/app/__init__.py
```diff
@@ -29,44 +29,163 @@
...
+        return render_template("/index.html")
+
+    @APP.route("/mss/about")
+    @APP.route("/mss")
+    def about():
+        _file = os.path.join(DOCS_SERVER_PATH, 'static', 'docs', 'about.md')
+        img_url = url_for('overview')
+        md_overrides = ('![image](/mss/overview.png)', f'![image]({img_url})')
+
+        html_overrrides = ('<img alt="image" src="/mss/overview.png" />',
```
ReimarBauer: html_overrides

mslib/mscolab/app/__init__.py
```diff
@@ -29,44 +29,163 @@
...
+
+# Keep backward compatibility with callers/tests expecting an app attribute.
+APP.xstatic = _xstatic
+
+
+def create_app(name="", imprint=None, gdpr=None):
+    imprint_file = imprint
+    gdpr_file = gdpr
+
+#    if "mscolab.server" in name:
```
ReimarBauer: if that is not needed remove it

mslib/mscolab/app/__init__.py
```diff
@@ -29,44 +29,163 @@
...
+    @APP.route("/mss/about")
+    @APP.route("/mss")
+    def about():
+        _file = os.path.join(DOCS_SERVER_PATH, 'static', 'docs', 'about.md')
+        img_url = url_for('overview')
+        md_overrides = ('![image](/mss/overview.png)', f'![image]({img_url})')
+
+        html_overrrides = ('<img alt="image" src="/mss/overview.png" />',
+                           '<img class="mx-auto d-block img-fluid" alt="image" src="/mss/overview.png" />')
+        content = get_content(_file, md_overrides=md_overrides, html_overrides=html_overrrides)
```
ReimarBauer: html_overrides

mslib/mscolab/app/__init__.py
```diff
@@ -29,44 +29,163 @@
...
+        base_path = _xstatic(name)
+        if base_path is None:
+            abort(404)
+        if not filename:
+            abort(404)
+        return send_from_directory(base_path, filename)
+
+    @APP.route('/mss_theme/img/<path:filename>')
+    def mss_theme(filename):
+        base_path = os.path.join(DOCS_SERVER_PATH, 'static', 'img')
```
ReimarBauer: simplify the repeated os.path.join by creating constants, e.g.

DOCS_STATIC_DIR = os.path.join(DOCS_SERVER_PATH, "static")
DOCS_IMG_DIR = os.path.join(DOCS_STATIC_DIR, "img")

...

ReimarBauer (approved):
looks good to me

PR #3048, opened 2026-03-30
use a fixture to always clear _shared_cache

### LINKED ISSUE #2805: WMS_SERVICE_CACHE should probably not be global
Having to reset this here is a good indicator that this variable shouldn't be global.

_Originally posted by @matrss in https://github.com/Open-MSS/MSS/pull/2792#discussion_r2050596325_

ReimarBauer:
**Purpose of PR?**:

Fixes #2805

No, changes in tests


tests/_test_msui/conftest.py
```diff
@@ -0,0 +1,35 @@
...
+"""
+import pytest
+from mslib.utils.service_manager import WMSServiceManager
+
+
+@pytest.fixture(autouse=True)
+def _reset_wms_service_manager():
+    WMSServiceManager._shared_cache.clear()
+    yield
+    WMSServiceManager._shared_cache.clear()
```
joernu76, line 35: maybe
WMSServiceManager._instance = None
to nuke the singleton for a *really* fresh clear?

joernu76 (approved):

tests/_test_msui/conftest.py
joernu76: And, obviously, use an instance of WMSServiceManager here instead of the private member call?

what about

@pytest.fixture(autouse=True) def _reset_wms_service_manager():
    mgr = WMSServiceManager()
    mgr.clear_cache()
    yield
    mgr.reset()

and add a reset() function to nuke the instance?

joernu76 (approved):

PR #3057, opened 2026-04-02
bbox_values reorder  independent of top_view

### LINKED ISSUE #3056: wms_control take early version 1.3.0 axisorder_yx
currently we do only for top view changing the bbox values based on epsg_code in axisorder_yx for version 1.3.0 
I want to do that earlier.

ReimarBauer:
**Purpose of PR?**:

Fixes #3056

### DIFF
--- a/mslib/msui/wms_control.py
+++ b/mslib/msui/wms_control.py
@@ -122,9 +122,10 @@ class MSUIWebMapService(ogcwms.WebMapService):
 
         """
         base_url = self.get_redirect_url(method)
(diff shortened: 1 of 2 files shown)


joernu76 (approved):

joernu76:
Does this solve an actual problem?

ReimarBauer:
> Does this solve an actual problem?

I remember discussions about why the order is wrong. I don't have a specific problem in mind right now, but I think this solution is better

PR #3060, opened 2026-04-02
replaced QtCore.QCoreApplication.processEvents()

### LINKED ISSUE #3058: eliminate QtCore.QCoreApplication.processEvents()
we have various QtCore.QCoreApplication.processEvents() in the library code, these should use the common functions

ReimarBauer:
**Purpose of PR?**:

Fixes #3058

yes

tests succeeds

### DIFF
--- a/mslib/msui/mpl_qtwidget.py
+++ b/mslib/msui/mpl_qtwidget.py
@@ -29,768 +29,26 @@
 # Parts of the code have been adapted from Chapter 6 of Sandro Tosi,
 # 'Matplotlib for Python Developers'.
 
-from datetime import datetime
 import enum
 import os
 import logging
 import numpy as np
-from matplotlib import cbook, figure
+from matplotlib import cbook
 from matplotlib.backends.backend_qt5agg import NavigationToolbar2QT, FigureCanvasQTAgg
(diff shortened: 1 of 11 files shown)


joernu76:
I do not understand the purpose of this humongous change. Looks like AI work (which is not necessarily bad, but style-wise it differs). Please provide a high-level description of the problem to be solved/intent of the refactoring.
The processEvents() calls and similar things were likely not good QT code, but it kind-of-worked. Adding threads might be better but certainly adds complexity. I'd like to understand the trade-offs and design-guidelines better.

ReimarBauer:
The QtWidgets.QApplication.processEvents() in mscolab_connect_dialog.py and mslib/msui/mscolab.py should be replaced by the proper calls. Maybe the functions were just not known. 
processEvents() handles all the events waiting in the event loop. I'm trying to reduce them as much as I can since we still have tests that fail sporadic.

The viewplotter can also be extracted from mpl_qtwidget in a subsequent refactoring task.

I am fine with only replacing those for which we have our own functions for now. Regarding the PyCharm-supported ones, there is already some complexity in the original code that we can probably address differently.

joernu76 (approved):

PR #3065, opened 2026-04-20
subprocess instead of multiprocessing, refactoring testsetup, enable windows tests

### LINKED ISSUE #2233: Replace multiprocessing usage in test fixtures with subprocess

ReimarBauer:
**Purpose of PR?**:

Fixes #2233


tests/constants.py
```diff
@@ -55,6 +55,9 @@
 os.environ["MSUI_CONFIG_PATH"] = str(MSUI_CONFIG_PATH.resolve())
 MSUI_CONFIG_FILE_PATH = MSUI_CONFIG_PATH / "msui_settings.json"
 
+if not MSUI_CONFIG_PATH.exists():
```
ReimarBauer, line 60: this problem was found by claude in pycharm ai

tests/_server_runner.py
```diff
@@ -0,0 +1,48 @@
...
+    # allowing the subprocess to find settings modules (e.g. mscolab_settings).
+    for extra_path in sys.argv[4:]:
+        if extra_path not in sys.path:
+            sys.path.insert(0, extra_path)
+    module = importlib.import_module(app_module)
+    app = getattr(module, app_attr)
+    srv = make_server(host, 0, app, threaded=True)
+    port = srv.server_address[1]
+    # Signal the chosen port to the parent process via stdout
+    print(port, flush=True)
```
ReimarBauer: this idea stems from claude in pycharm ai

tests/fixtures.py
```diff
@@ -110,21 +112,15 @@ def mscolab_session_managers(mscolab_session_app):
...
 
 
-# TODO: Having this fixture be autouse is a crutch. It seems like if it is not autouse some tests can bring the pytest
-# processes objects into a state in which the MSColab server will have trouble starting the Flask-SocketIO server once
-# it is forked. With autouse the fork happens first, before any test runs. After that, the pytest process can no longer
-# affect the now-running server, thus mitigating the issue. This is my understanding at time of writing.
-#
-# This issue would also be avoided if the background server process wasn't started with multiprocessing and a fork, but
-# with a real subprocess, which would solve some other issues (e.g. testing on Windows) as well.
 @pytest.fixture(scope="session", autouse=True)
```
matrss: Please do not remove the comment without also removing the `autouse=True`.

mslib/utils/config.py
```diff
@@ -221,7 +218,7 @@ class MSUIDefaultConfig:
     MSS_auth = {}
 
     # timeout of Url request
-    WMS_request_timeout = 30
+    WMS_request_timeout = 60
```
ReimarBauer, line 221: this is at least for CI tests needed

mslib/utils/config.py
```diff
@@ -56,9 +56,6 @@ class MSUIDefaultConfig:
     Do not change any value for good reasons.
     Your values can be set in your personal msui_settings.json file
     """
-    # this skips the verification of the user token on each mscolab request
-    mscolab_skip_verify_user_token = True
-
```
ReimarBauer, line 61: this was only client site for debug purposes.

mslib/msui/mscolab.py
```diff
@@ -70,33 +67,6 @@
 from mslib.utils.config import config_loader
 
 
-def verify_user_token(func):
-    if not hasattr(verify_user_token, "depth"):
```
ReimarBauer, line 74: removing this improved stability of mscolab tests

tests/_test_msui/test_mscolab.py
```diff
@@ -31,12 +31,12 @@
...
 import pytest
-
 from PIL import Image
-
-from tests.constants import ROOT_DIR, MSCOLAB_DATA_DIR
-from tests import constants
 import mslib.utils.auth
+
+
+_ROOT_DIR = Path(os.environ["MSUI_CONFIG_PATH"]).parent
```
ReimarBauer: root_dir, mscolab_data_dir is a fixture tool, can be simplified

conftest.py
```diff
@@ -209,5 +268,50 @@ def reset_config():
     read_config_file()
 
 
+@pytest.fixture(scope="session")
```
ReimarBauer: move this block to fixtures

mslib/mscolab/seed.py
```diff
@@ -182,6 +183,7 @@ def add_operation(operation_name, description):
                 main_file_git.write_text(XML_CONTENT_INIT, encoding='utf-8')
                 r.index.add(['main.ftml'])
                 r.index.commit("initial commit")
+                r.close()
```
joernu76: I am rather certain that one should be able to use the with git.Repo.init as r... statement

mslib/mscolab/file_manager.py
joernu76: I am not sure this makes sense...

joernu76 (commented):
this is a pretty large change that should have been split into smaller parts

joernu76:
Why has all the verify-token stuff been removed?

ReimarBauer:
> Why has all the verify-token stuff been removed?

It was added in the early days for debugging the client. On default it is disabled.

All tests in the same worker process share the same depth counter. If a test triggers Qt events that call decorated methods asynchronously (after the test "ends" but before the event loop is drained), depth may be non-zero when the next test starts. This causes the error-handling branch (depth > 1) to behave incorrectly for unrelated tests.

Without that also the stability of the tests increased.

mslib/mscolab/file_manager.py
ReimarBauer: you are right.
It got removed. That was maybe a leftover from pyfilesystem2 times

ReimarBauer:
> this is a pretty large change that should have been split into smaller parts

Now that I know how to adjust our code, we could split this up if needed. But to avoid making the PR even bigger, I’d rather not add anything else yet; I think we should merge this and the blueprints first. We'll likely have more refactoring ideas once that's done.

matrss:
My understanding is that significant parts of this are LLM generated, is that correct? If yes, did you fully review it yourself already?

ReimarBauer:
> My understanding is that significant parts of this are LLM generated, is that correct? If yes, did you fully review it yourself already?

My Workflow was

   1.  Isolated machine/account, no GitHub access.
   2.  Setup a skill and task.
   3.  Get the task done.
   4.  Create a patch file.
   5.  Import the patch file on my system.
   6.  Review, verify, improve the result.
   7.  Commit.
   
I wanted to use an ASCII patch file so that I could review the contents again before proceeding with the next steps. I'm not perfect—sometimes I miss things that only get fixed in the next stage.
Many of the issues identified could have been spotted in earlier merge requests. It was a valuable learning experience that reminded me of how I felt during my very first refactoring of MSS.
Some refactoring steps were performed in PyCharm without explicitly requesting AI assistance.

mslib/mscolab/seed.py
ReimarBauer: done

.github/workflows/testing-all-oses.yml
```diff
@@ -18,17 +18,28 @@ jobs:
...
+        echo "Open files: $(ulimit -n 2>/dev/null || echo 'N/A')"
     - name: Run tests
       timeout-minutes: 40
-      run: pixi run -e dev env QT_QPA_PLATFORM=offscreen pytest -v -n logical --durations=20 --cov=mslib tests
+      shell: bash
+      run: |
+        # Set stricter resource limits for tests on Linux
+        if [ "$(uname)" = "Linux" ]; then
+          ulimit -n 16384 2>/dev/null || true
+        fi
```
matrss: What's the reason for this?

mslib/mscolab/events.py
```diff
@@ -0,0 +1,74 @@
...
+"""
+    mslib.mscolab.events
+    ~~~~~~~~~~~~~~~~~~~~
+
+    This module defines all socket event names used in the mscolab module.
+
+    This file is part of MSS.
+
+    :copyright: Copyright 2019 Shivashis Padhi
+    :copyright: Copyright 2019-2026 by the MSS team, see AUTHORS.
```
matrss, line 11: Isn't this a new file?

mslib/mscolab/events.py
```diff
@@ -0,0 +1,74 @@
...
+
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+"""
+
+
+class SocketEvents:
```
matrss, line 28: This seems like a good use-case for StrEnum, you wouldn't have to implement get_all_events yourself then.

mslib/mscolab/file_manager.py
```diff
@@ -313,12 +308,13 @@ def upload_file(self, file, subfolder=None, identifier=None, include_prefix=Fals
         with file_path.open(mode="wb") as f:
             file.save(f)
 
-        # Relative File path
+        # Relative File path — always use forward slashes so the path can be
+        # used as a URL segment on both Windows and Linux.
         if include_prefix:  # ToDo: add a namespace for the chat attachments, similar as for profile images
             static_dir = Path(upload_folder).name
-            static_file_path = str(Path(static_dir) / str(subfolder) / file_name)
+            static_file_path = '/'.join([static_dir, str(subfolder), file_name])
```
matrss: `urljoin` / `safe_join`? Or `as_posix()` like below? The term "path" in the variable name seems a bit ambiguous, is it a filesystem path or a URL path?

mslib/mscolab/file_manager.py
```diff
@@ -373,12 +369,7 @@ def update_operation(self, op_id, attribute, value, user):
             if new_path.exists():
                 return False
 
-            new_path.mkdir(parents=True, exist_ok=True)
-
-            try:
-                old_path.rename(new_path)
-            except OSError:
-                shutil.move(str(old_path), str(new_path))
+            shutil.move(str(old_path), str(new_path))
```
matrss: `shutil.move` should accept path-like objects, so no need to cast to str.

mslib/mscolab/file_manager.py
```diff
@@ -543,9 +534,9 @@ def get_change_content(self, ch_id, user):
             return False
         operation = Operation.query.filter_by(id=change.op_id).first()
         operation_path = Path(self.data_dir) / operation.path
-        repo = git.Repo(operation_path)
-        change_content = repo.git.show(f'{change.commit_hash}:main.ftml')
-        return change_content
+        with git.Repo(operation_path) as gr:
+            change_content = gr.git.show(f'{change.commit_hash}:main.ftml')
+        return change_content.replace('\r\n', '\n')
```
matrss, line 538: What's the issue this is solving? MSColab server on Windows reading the content back with CRLF and sending it to a Linux client that expects LF?

mslib/mscolab/mscolab.py
```diff
@@ -46,15 +46,22 @@
 from mslib.utils import setup_logging
 
 
-def handle_start(args=None):
-    from mslib.mscolab.server import APP, sockio, cm, fm, start_server
-    if args is not None:
-        setup_logging(args)
+def handle_server_start(app_module='mslib.mscolab.server', app_attr='APP',
```
matrss: AFAICT app_module and app_attr are unused.

mslib/mscolab/mscolab.py
```diff
@@ -46,15 +46,22 @@
...
+                        host='127.0.0.1', port=8083):
+    import importlib
+    from werkzeug.serving import make_server
     logging.info("MSS Version: %s", __version__)
     logging.info("Python Version: %s", sys.version)
     logging.info("Platform: %s (%s)", platform.platform(), platform.architecture())
     logging.info("Launching MSColab Server")
-    start_server(APP, sockio, cm, fm)
+    module = importlib.import_module(app_module)
```
matrss: Why importlib instead of just importing?

mslib/mscolab/seed.py
```diff
@@ -83,12 +83,13 @@ def add_all_users_default_operation(path='TEMPLATE', description="Operation to k
...
                 git_repo_path = Path(APP.config['OPERATIONS_DATA']) / path
                 git_repo_path.mkdir(parents=True, exist_ok=True)
-                r = git.Repo.init(str(git_repo_path))
-                r.git.clear_cache()
-                main_file_git = git_repo_path / "main.ftml"
-                main_file_git.write_text(XML_CONTENT_INIT, encoding='utf-8')
-                r.index.add(['main.ftml'])
-                r.index.commit("initial commit")
+                #  Todo this should be done by the file_manager
+                with git.Repo.init(str(git_repo_path)) as gr:
```
matrss: git.Repo.init also accepts Path-like's, so casting to str shouldn't be necessary.

mslib/mscolab/server.py
```diff
@@ -360,6 +374,31 @@ def hello():
         })
 
 
+if os.environ.get("MSCOLAB_TEST_MODE") == "1":
```
matrss: I don't like this, this means that the tested code is not the same as the production code. Especially the first function implies that there are tests that leave unclosed sockets lying around, which sounds like a bug.

mslib/msui/flighttrack.py
```diff
@@ -607,7 +607,11 @@ def invert_direction(self):
             if len(wp_comm) == 9 and wp_comm.startswith("Hexagon "):
                 wp_comm = f"Hexagon {(8 - int(wp_comm[-1])):d}"
                 self.waypoints[i].comments = wp_comm
+        # Block signals during update_distances() so its internal dataChanged emission
+        # doesn't trigger a redundant save; we emit a single dataChanged below instead.
```
matrss, line 611: Why block the signal instead of just not emitting it twice?

mslib/msui/mscolab_version_history.py
```diff
@@ -128,172 +123,144 @@ def load_all_changes(self):
...
-                    self.load_current_waypoints()
-                    self.load_all_changes()
+    def handle_delete_version_name(self):
+        selected_item = self.changes.currentItem()
+        res = self.request_set_version_name(None, selected_item.id)
+        if res.text != "False":
+            res = res.json()
+            if res["success"] is True:
+                # Remove item if the filter is set to Named version only
+                if self.versionFilterCB.currentIndex() == 1:
```
matrss, line 228: What's the reason for changing from 0 to 1?

mslib/msui/socket_control.py
```diff
@@ -160,63 +153,50 @@ def handle_operation_list_update(self):
...
-                          "version_name": version_name,
-                          "messageText": messageText})
-        else:
-            # this triggers disconnect
-            self.signal_reload.emit(op_id)
+        self.sio.emit(SocketEvents.OPERATION_SELECTED, {'token': self.token, 'op_id': op_id})
+
+    def save_file(self, op_id, content, comment=None, version_name=None, messageText=""):
+        # Token validity is already checked on server-side. The extra HTTP round-trip here
+        # only adds load under parallel saves and can cause the decorator to trigger logout.
```
matrss: This comment seems to be about the verify_user_token decorator that is no longer present?

mslib/utils/__init__.py
```diff
@@ -37,7 +37,7 @@ def __init__(self, error_string):
 def setup_logging(args):
     logger = logging.getLogger()
     # this is necessary as "someone" has already initialized logging, preventing basicConfig from doing stuff
-    for ch in logger.handlers:
+    for ch in logger.handlers[:]:
```
matrss, line 41: How does this change anything?

mslib/utils/__init__.py
```diff
@@ -37,7 +37,7 @@ def __init__(self, error_string):
 def setup_logging(args):
     logger = logging.getLogger()
     # this is necessary as "someone" has already initialized logging, preventing basicConfig from doing stuff
```
matrss: Maybe that's obsolete with the removal of the basicConfig call in mslib/mswms/wms.py?

mslib/utils/ogcwms.py
```diff
@@ -372,11 +372,16 @@ def read(self, service_url,
...
-        """Parse a WMS capabilities document, returning an elementtree instance
-                string should be an XML capabilities document
-                """
+        """Parse a WMS capabilities document, returning an elementtree instance.
+        Accepts bytes or str; stores the raw document in capabilities_document.
+        """
+        # (mss) handle bytes
+        if isinstance(st, bytes):
+            self.capabilities_document = st
+            return etree.fromstring(st)
```
matrss: fromstring takes a bytes x)

mslib/utils/ogcwms.py
```diff
@@ -372,11 +372,16 @@ def read(self, service_url,
...
         if not isinstance(st, str):
-            raise ValueError("String must be of type string, not %s" % type(st))
+            raise ValueError("String must be of type string or bytes, not %s" % type(st))
+        self.capabilities_document = st.encode('utf-8')
         return etree.fromstring(st)
```
matrss: Is this change needed? fromstring seems to accept both bytes and str, is it using the wrong encoding with str on Windows?

matrss (commented):
Partial review, it's getting late.

.github/workflows/testing-all-oses.yml
ReimarBauer: After making the change, I noticed that file handles were not being closed. Reason: Only the Linux runner could be limited to 1024 (macOS adjusts automatically; Windows is not affected). I didn’t want to run into problems due to differing limits.

ToDo: try with a revert

mslib/mscolab/events.py
ReimarBauer: the file is new, but the content was introduced in 2019

mslib/mscolab/events.py
ReimarBauer: yes, but then we have to change the python limit > 3.10, < 3.12
https://docs.python.org/3/library/enum.html#enum.StrEnum
was introduced in 3.11

At the moment we do the refactoring for more modern python versions, we have also an updated enum and this can become replaced.

mslib/utils/__init__.py
ReimarBauer: We don't want to modify the list during the iteration. Otherwise, we might end up omitting some values. It advances the iterator index without using a copy on each remove.

When we have 
[<StreamHandler <_io.FileIO name=11 mode='rb+' closefd=True> (NOTSET)>, <LogCaptureHandler (NOTSET)>, <LogCaptureHandler (NOTSET)>]

it skips every second, in this case [<LogCaptureHandler (NOTSET)>] is left over.

mslib/utils/ogcwms.py
ReimarBauer: windows surprised with utf-16

utf-8 is a configuration with beta state since many years. Usually it is not set.

There is maybe an overseen problem. Sometimes the WMS server returned raw bytes.
That was skipped  by `if not isinstance(st, str):`

mslib/mscolab/file_manager.py
ReimarBauer: safe_join :)

mslib/mscolab/file_manager.py
ReimarBauer: yes

mslib/utils/__init__.py
ReimarBauer: Importing mslib.mswms.wms triggers logging.warning(...) calls at module
  level, which auto-initialises the root logger with a default StreamHandler.
  We will have a second handler which prints twice
  
 on mscolab it looks like 
```
mscolab start
WARNING:root:Couldn't import setup_saml2_backend (ImportError:'No module named 'setup_saml2_backend''), using dummy config.
INFO:root:MSS Version: 11.0.0
INFO: MSS Version: 11.0.0
INFO:root:Python Version: 3.11.14 | packaged by conda-forge | (main, Jan 27 2026, 00:00:54) [Clang 19.1.7 ]
INFO: Python Version: 3.11.14 | packaged by conda-forge | (main, Jan 27 2026, 00:00:54) [Clang 19.1.7 ]
INFO:root:Platform: macOS-26.5-x86_64-i386-64bit (('64bit', ''))
INFO: Platform: macOS-26.5-x86_64-i386-64bit (('64bit', ''))
```

ReimarBauer: improved the comment

mslib/mscolab/file_manager.py
ReimarBauer: we maybe should also create a .gitattributes with 
`*.ftml text eol=lf`
windows can handle lf files. I got to this because some tests compare with linefeed. At all I don't want CRLF in files which can be exchanged to others.

mslib/mscolab/mscolab.py
ReimarBauer: leftover

mslib/mscolab/mscolab.py
ReimarBauer: no names needed, so we can exchange it.

mslib/mscolab/seed.py
ReimarBauer: lets try

mslib/msui/mscolab_version_history.py
ReimarBauer: I changed a year ago the order, that All Changes is on top and Named is index 1. Changing this was missed.
https://github.com/Open-MSS/MSS/blame/develop/mslib/msui/mscolab_version_history.py#L137

https://github.com/Open-MSS/MSS/commit/f434dd686a8724dfe59871b831c679aec2c2388a

mslib/msui/flighttrack.py
ReimarBauer: That's an unnecessary duplicate network/file save. It would send the entire XML content of a Flighttrack to the server twice. Depending on the distance to the server, that could take a long time.

ReimarBauer:
seems timeouts are back.

mslib/mscolab/server.py
ReimarBauer: I don't like it either; I'll see if I can figure it out.
I'm not sure if it's still valid, since a lot has changed since then.

mslib/mscolab/mscolab.py
ReimarBauer: Back on this state. import seem to trigger timeouts on gh. No idea yet.

mslib/utils/ogcwms.py
ReimarBauer: simplified that,

https://github.com/Open-MSS/MSS/blame/develop/mslib/mswms/wms.py#L612

the change was needed for my change at https://github.com/Open-MSS/MSS/pull/3065/changes#diff-b11c87da16bf89f1f1586f2a7d9cba04ca5a6914c2472066cfa0f010bae58218R999

I changed this because of timeings.

ReimarBauer: undone my last update change, I run out of time. have to look on that again.

ReimarBauer: The change is needed, I updated the comments and the names.

.github/workflows/testing-all-oses.yml
```diff
@@ -18,16 +18,22 @@ jobs:
...
     - uses: prefix-dev/setup-pixi@5185adfbffb4bd703da3010310260805d89ebb11 # v0.9.6
       with:
         pixi-version: latest
         cache: true
         environments: dev
+    - name: Check CI resource limits
+      shell: bash
+      run: |
+        echo "=== Resource Limits ==="
+        echo "Open files: $(ulimit -n 2>/dev/null || echo 'N/A')"
```
matrss, line 33: This seems more like debug printing?

.github/workflows/testing-all-oses.yml
```diff
@@ -18,16 +18,22 @@ jobs:
...
     - name: Run tests
       timeout-minutes: 40
+      shell: bash
```
matrss, line 36: I wonder why this is needed here but not for the next run step?

mslib/mscolab/file_manager.py
matrss: There is also core.autocrlf.

mslib/mscolab/file_manager.py
```diff
@@ -578,22 +568,22 @@ def undo_changes(self, ch_id, user):
...
-                repo.index.add(['main.ftml'])
-                cm = repo.index.commit(f"checkout to {ch.commit_hash}")
-                change = Change(ch.op_id, user.id, cm.hexsha)
-                db.session.add(change)
-                db.session.commit()
-                return True
-            except Exception as ex:
-                logging.debug(ex)
-                return False
+            with git.Repo(str(operation_path)) as gr:
```
matrss, line 571: I think all occurrences of this could drop the `str()`?

mslib/mscolab/sockets_manager.py
```diff
@@ -59,6 +60,15 @@ def __init__(self, chat_manager, file_manager):
     def handle_connect(self):
         logging.debug(request.sid)
 
+    def clear_state(self):
+        """Drop all in-memory socket bookkeeping.
+
+        Called by handle_db_reset so a database reset also discards these registries,
+        which would otherwise reference user/operation rows that no longer exist.
+        """
+        self.sockets[:] = []
```
matrss, line 69: Do those sockets have to be closed somehow first?

mslib/msui/wms_control.py
```diff
@@ -728,25 +735,49 @@ def leftrow_is_selected(self, vtime):
...
+        wms_wait_ms = (config_loader(dataset="WMS_request_timeout") + 5) * 1000
+        try:
+            if hasattr(self, 'thread_prefetch') and self.thread_prefetch is not None:
+                if self.thread_prefetch.isRunning():
+                    self.thread_prefetch.quit()
+                    if not self.thread_prefetch.wait(wms_wait_ms):
+                        self.thread_prefetch.terminate()
+                        self.thread_prefetch.wait(1000)
+        except Exception:
+            pass
```
matrss, line 754: This is usually bad practice, why is it needed here?

mslib/msui/wms_control.py
```diff
@@ -728,25 +735,49 @@ def leftrow_is_selected(self, vtime):
...
+        # matplotlib, which holds the GIL; terminating the thread early leaves the server
+        # GIL-busy and blocks subsequent GetCapabilities requests in later tests.
+        wms_wait_ms = (config_loader(dataset="WMS_request_timeout") + 5) * 1000
+        try:
+            if hasattr(self, 'thread_prefetch') and self.thread_prefetch is not None:
+                if self.thread_prefetch.isRunning():
+                    self.thread_prefetch.quit()
+                    if not self.thread_prefetch.wait(wms_wait_ms):
+                        self.thread_prefetch.terminate()
+                        self.thread_prefetch.wait(1000)
```
matrss, line 752: I wish we were using QRunnable's and the global thread pool instead of spawning our own threads, I am pretty sure Qt handles cleanup better by default already...

mslib/mswms/mss_plot_driver.py
```diff
@@ -69,12 +69,23 @@ def __init__(self, data_access_object):
         self.plot_object = None
         self.filenames = []
 
-    def __del__(self):
+    def close(self):
```
matrss, line 72: This screams for a context manager.

mslib/mswms/wms.py
```diff
@@ -945,8 +955,13 @@ def application():
...
         else:
-            logging.debug("Request type '%s' is not valid.", request)
-            raise RuntimeError("Request type is not valid.")
+            logging.debug("Request type '%s' is not valid.", request_type)
+            error_message = "RuntimeError: Request type is not valid.\n"
+            response_headers = [('Content-type', 'text/plain'), ('Content-Length', str(len(error_message)))]
+            res = make_response(error_message, 404)
+            for response_header in response_headers:
+                res.headers[response_header[0]] = response_header[1]
+            return res
```
matrss, line 964: Why is this copying the exception handler that is already in place below?

mslib/utils/__init__.py
matrss: Ah, I missed that it is modifying the list inside the loop. I still think this shouldn't be necessary at all if logging was set up properly...

tests/fixtures.py
```diff
@@ -23,21 +23,31 @@
...
-from werkzeug.serving import make_server
+import mslib.mswms.wms
+import mslib.mswms.gallery_builder
 
 from PyQt5 import QtWidgets
 from contextlib import contextmanager
+
+from conftest import MSCOLAB_SERVER_CONFIG_DIR, generate_initial_config, ROOT_DIR, \
+    MSWMS_SERVER_CONFIG_DIR, MSWMS_DATA_DIR, MSWMS_SERVER_CONFIG_FILE_PATH, MSCOLAB_DATA_DIR, \
+    MSCOLAB_SERVER_CONFIG_FILE_PATH, MSUI_CONFIG_PATH, MSUI_CONFIG_FILE_PATH
```
matrss, line 46: I don't know how I feel about this. I don't think this and the fixtures are an improvement over the constants module, it is still globally defined constants on the module level instead of parameters at object instantiation time.

tests/_test_msui/test_msui.py
```diff
@@ -28,20 +28,22 @@
...
-from tests.constants import ROOT_DIR, MSUI_CONFIG_PATH, MSUI_CONFIG_FILE_PATH
 from mslib.msui import msui
 from mslib.msui import msui_mainwindow as msui_mw
 from tests.utils import ExceptionMock
 from mslib.utils.config import read_config_file, config_loader
 import re
 
+ROOT_DIR = Path(os.environ["MSUI_CONFIG_PATH"]).parent
+MSUI_CONFIG_PATH = Path(os.environ["MSUI_CONFIG_PATH"])
+MSUI_CONFIG_FILE_PATH = MSUI_CONFIG_PATH / "msui_settings.json"
```
matrss, line 45: Why repeat these here?

tests/_test_msui/test_wms_control.py
```diff
@@ -80,16 +82,35 @@ def _setup(self, widget_type, tmp_path):
...
 
     def query_server(self, qtbot, url):
         while len(self.window.multilayers.cbWMS_URL.currentText()) > 0:
             QtTest.QTest.keyClick(self.window.multilayers.cbWMS_URL, QtCore.Qt.Key_Backspace)
         QtTest.QTest.keyClicks(self.window.multilayers.cbWMS_URL, url)
+        # Use a generous timeout: parallel xdist execution puts all workers'
+        # WMS servers under simultaneous CPU load, making the two-step HTTP
+        # chain (requests.get + MSUIWebMapService) occasionally exceed the
+        # default 5 s limit.
         with qtbot.wait_signal(self.window.cpdlg.canceled):
```
matrss, line 109: The comment is there, but the timeout is not changed?

tests/_test_msui/test_wms_control.py
```diff
@@ -160,15 +184,18 @@ def test_forward_backward_clicks(self, qtbot):
...
-        with qtbot.wait_signal(self.window.image_displayed):
-            QtTest.QTest.mouseClick(self.window.btGetMap, QtCore.Qt.LeftButton)
-            QtTest.QTest.keyClick(self.window.pdlg, QtCore.Qt.Key_Enter)
+        QtTest.QTest.mouseClick(self.window.btGetMap, QtCore.Qt.LeftButton)
+        # started_request is emitted before the blocking HTTP call, so pdlg.isVisible()
+        # becomes True while the request is still in-flight — giving us time to cancel.
+        qtbot.wait_until(self.window.pdlg.isVisible)
+        self.window.pdlg.cancel()
+        # Let the fetch complete; continue_retrieve_image will see wasCanceled=True.
+        QtTest.QTest.qWait(500)
```
matrss, line 198: [Time is not a synchronization primitive](https://xeiaso.net/blog/nosleep/)

tests/_test_mswms/test_dataaccess.py
```diff
@@ -28,11 +28,13 @@
...
 import os
 from datetime import datetime
+from pathlib import Path
 
 import mock
 
 from mslib.mswms.dataaccess import DefaultDataAccess, CachedDataAccess
-from tests.constants import MSWMS_DATA_DIR
+
+MSWMS_DATA_DIR = Path(os.environ["MSUI_CONFIG_PATH"]).parent / "mswms" / "testdata"
```
matrss, line 37: This is already defined in conftest.py

tests/_test_mswms/test_mplhsec.py
```diff
@@ -27,14 +27,17 @@
 """
 
 import importlib
+import os
+from pathlib import Path
 
 from mslib.mswms.mpl_hsec import MPLBasemapHorizontalSectionStyle
-from tests import constants
+
+_MSWMS_SERVER_CONFIG_FILE_PATH = Path(os.environ["MSUI_CONFIG_PATH"]).parent / "mswms" / "mswms_settings.py"
```
matrss, line 35: Dito

tests/_test_mswms/test_mswms.py
```diff
@@ -25,31 +25,227 @@
...
+        refresh=False,
+        levels="",
+        itimes="",
+        vtimes="",
+        show_code=False,
+        url_prefix="",
+        plot_types=None,
+    )
+    defaults.update(kwargs)
+    return argparse.Namespace(**defaults)
```
matrss, line 70: What's the purpose of this instead of only passing argparse.Namespace with those options that are to be set? Aren't the defaults already defined within the application?

tests/_test_mswms/test_mswms.py
```diff
@@ -25,31 +25,227 @@
...
-                        return_value=argparse.Namespace(plot_types=None, version=True)):
+    def test_version_prints_mss_header(self, mocker, capsys):
+        mocker.patch("mslib.mswms.mswms.argparse.ArgumentParser.parse_args",
+                     return_value=_args(version=True))
+        with pytest.raises(SystemExit):
             mswms.main()
-        assert pytest_wrapped_e.typename == "SystemExit"
+        out = capsys.readouterr().out
+        assert "Mission Support System" in out
+        assert "Version" in out
```
matrss, line 110: test_version_exits is a subset of test_version_prints_mss_header, IMO the latter would be enough.

tests/_test_mswms/test_mswms.py
```diff
@@ -25,31 +25,227 @@
...
+# ---------------------------------------------------------------------------
+
+class TestMainSeed:
+    def test_seed_calls_create_server_config_and_data(self, mocker, tmp_path,
+                                                      mock_wms, mock_make_server):
+        mock_examples = mocker.MagicMock()
+        mocker.patch("mslib.mswms.mswms.argparse.ArgumentParser.parse_args",
+                     return_value=_args(seed=True))
+        mocker.patch("mslib.mswms.mswms.DataFiles", return_value=mock_examples)
+        mocker.patch("mslib.mswms.mswms.Path.home", return_value=tmp_path)
```
matrss, line 159: Why mock this?

tests/fixtures.py
```diff
@@ -56,33 +66,38 @@ def fail_if_open_message_boxes_left():
             pytest.fail(f"An unhandled message box popped up during your test!\n{summary}")
 
 
-@pytest.fixture
-def close_remaining_widgets():
```
matrss, line 60: Why is this inlined?

tests/fixtures.py
```diff
@@ -56,33 +66,38 @@ def fail_if_open_message_boxes_left():
...
-def qtbot(qtbot, fail_if_open_message_boxes_left, close_remaining_widgets, msui_configs):
+def qtbot(qtbot, fail_if_open_message_boxes_left, msui_configs):
     """Fixture that re-defines the qtbot fixture from pytest-qt with additional checks."""
     yield qtbot
-    # Wait for a while after the requesting test has finished. At time of writing this
-    # is required to (mostly) stabilize the coverage reports, because tests don't
-    # properly close their Qt-related stuff and therefore there is no guarantee about
-    # what the Qt event loop has or hasn't done yet. Waiting just gives it a bit more
-    # time to converge on the same result every time the tests are executed. This is a
-    # band-aid fix, the proper fix is to make sure each test cleans up after itself.
```
matrss, line 85: How is this comment addressed? If it isn't why is it deleted?

tests/fixtures.py
```diff
@@ -110,21 +125,17 @@ def mscolab_session_managers(mscolab_session_app):
...
+@pytest.fixture(scope="session")
+def mscolab_session_server(mscolab_session_app, mscolab_session_managers, mscolab_server_config_dir):
     """Session-scoped fixture that provides a running MSColab server.
 
     This fixture should not be used in tests. Instead use :func:`mscolab_server`, which
     handles per-test cleanup as well.
     """
-    with _running_server(mscolab_session_app) as url:
+    # Use port 0 to let OS assign available port - early failure if unavailable
+    cmd = [sys.executable, '-m', 'mslib.mscolab.mscolab', 'start', '--host', '127.0.0.1', '--port', '0']
```
matrss, line 136:
```suggestion
    cmd = ['mscolab', 'start', '--host', '127.0.0.1', '--port', '0']
```

tests/fixtures.py
```diff
@@ -163,11 +174,13 @@ def mscolab_managers(mscolab_session_managers, reset_mscolab):
...
 
 @pytest.fixture
-def mscolab_server(mscolab_session_server, reset_mscolab):
+def mscolab_server(mscolab_session_server, mscolab_session_app):
     """Fixture that provides a running MSColab server and does cleanup actions.
 
     :returns: The URL where the server is running.
     """
+    with mscolab_session_app.app_context():
+        handle_db_reset(verbose=False)
```
matrss, line 183: With a subprocess mscolab_session_app is entirely unrelated to the running mscolab server, so I don't know if this is a good idea.

tests/fixtures.py
```diff
@@ -176,75 +189,249 @@ def mscolab_server(mscolab_session_server, reset_mscolab):
 @pytest.fixture(scope="session")
 def mswms_app():
     """Fixture that provides the MSWMS WSGI app instance."""
-    return mslib.mswms.mswms.application
+    yield mslib.mswms.mswms.application
+    # Close all open NetCDF4 datasets to release file handles
+    from mslib.mswms import wms
+    wms.server.close_datasets()
```
matrss, line 195: Since mswms is running in a subprocess any opened datasets at this point seem to be a bug in itself.

tests/fixtures.py
```diff
@@ -176,75 +189,249 @@ def mscolab_server(mscolab_session_server, reset_mscolab):
...
+        port = int(port_line.strip().decode(errors='replace').rsplit(":", 1)[-1])
+
+        # Drain stdout and stderr in daemon threads so the subprocess's pipe
+        # buffers never fill up.  Werkzeug logs each request to stderr; after
+        # ~12 GetCapabilities + GetMap calls the accumulated output exceeds the
+        # 64 KB pipe buffer on macOS/Linux, causing logging.info() in the
+        # request-handler thread to block indefinitely and producing ReadTimeout
+        # errors in the next test that tries to contact the server.
+        threading.Thread(target=_drain_pipe, args=(process.stdout,), daemon=True).start()
+        threading.Thread(target=_drain_pipe, args=(process.stderr,), daemon=True).start()
```
matrss, line 276: Why write to a pipe at all if it is just discarded? Why not /dev/null or proper logging?

tests/fixtures.py
```diff
@@ -176,75 +189,249 @@ def mscolab_server(mscolab_session_server, reset_mscolab):
...
+    """Fixture to reset WMS module-level globals that can affect test isolation."""
+    static_location = mslib.mswms.wms.STATIC_LOCATION
+    docs_location = mslib.mswms.wms.DOCS_LOCATION
+    gallery_static = mslib.mswms.gallery_builder.STATIC_LOCATION
+    gallery_docs = mslib.mswms.gallery_builder.DOCS_LOCATION
+    yield
+    mslib.mswms.wms.STATIC_LOCATION = static_location
+    mslib.mswms.wms.DOCS_LOCATION = docs_location
+    mslib.mswms.gallery_builder.STATIC_LOCATION = gallery_static
+    mslib.mswms.gallery_builder.DOCS_LOCATION = gallery_docs
```
matrss, line 326: Am I understanding it right that this is a band-aid fix to reset the environment when some tests fail to clean up their changes after themselves?

tests/fixtures.py
```diff
@@ -176,75 +189,249 @@ def mscolab_server(mscolab_session_server, reset_mscolab):
...
+
+
+@pytest.fixture
+def reset_gallery_builders():
+    """Fixture to reset gallery_builder module-level mutable state."""
```
matrss, line 331: Module-level mutable state is the original sin.

tests/_test_msui/test_wms_control.py
```diff
@@ -80,16 +82,35 @@ def _setup(self, widget_type, tmp_path):
...
+        # initialise_wms) so they don't hold the WMS server busy when the next
+        # test starts.  Under parallel xdist execution this prevents the new
+        # test's request from queuing behind the previous one, which would push
+        # the total capabilities-loading time past wait_signal's timeout.
+        # Use isRunning() rather than truthiness of Worker.workers: the set
+        # always contains the un-started capabilities_worker placeholder so a
+        # plain bool check would busy-wait for the full 10 s on every teardown.
+        deadline = time.time() + 10
+        while any(w.isRunning() for w in list(Worker.workers)) and time.time() < deadline:
+            QtTest.QTest.qWait(100)
```
matrss, line 99: qtbot.wait_until

tests/test_meta.py
```diff
@@ -42,9 +42,13 @@ def test_processEvents_is_not_used_in_tests(request):
 def test_qWait_is_not_used_in_tests(request):
     """Check that no test is calling PyQt5.QtTest.QTest.qWait explicitly."""
     tests_path = pathlib.Path(request.config.rootdir) / "tests"
+    excluded = {str(test_file) for test_file in [
+        request.fspath,
+        tests_path / "_test_msui" / "test_wms_control.py",
+    ]}
```
matrss, line 48: No, the qWait in that other file aren't necessary.

.gitattributes
```diff
@@ -1,3 +1,4 @@
 * text=auto
+*.py text eol=lf
```
matrss, line 2: This would cause checked out .py files on windows to use LF instead of CRLF, which seems wrong. What is this supposed to do?

pixi.toml
```diff
@@ -99,6 +99,7 @@ py = "*"
 pynco = "*"
 pytest = "*"
 pytest-cov = "*"
+pytest-mock = "*"
```
matrss, line 102: We have been using mocking without pytest-mock just fine already, what advantage does this package bring?

matrss:
> this is a pretty large change that should have been split into smaller parts

I have to agree with this. This PR contains a lot of changes that seem to be unrelated to the issue it is supposed to fix, and some seem to not be improvements at all.

I also don't really want to review this behemoth another time. IMO this PR should be split into smaller ones, or reduced to the minimum required to get tests running on windows with subprocess.

ReimarBauer:
lets close it, at least we know that tests can run without failures

ReimarBauer:
> > this is a pretty large change that should have been split into smaller parts
> 
> I have to agree with this. This PR contains a lot of changes that seem to be unrelated to the issue it is supposed to fix, and some seem to not be improvements at all.
> 
> I also don't really want to review this behemoth another time. IMO this PR should be split into smaller ones, or reduced to the minimum required to get tests running on windows with subprocess.

After the blueprints PR has been merged, I am considering extracting all the fixes that were essentially required for the tests. I expect that these could then be executed using multiprocessing; this would be one possible splitting strategy. My previous approach was quite the opposite: I initially used subprocesses, but since that did not improve things, it led to an increasing number of changes.
Afterwards change to subprocess. This step should then also have none flaky tests.
Then adding windows testing.

PR #3111, opened 2026-07-07
add confidence to all location based functions

### LINKED ISSUE #3110: tutorials location functions should use confidence
We need a threshold to to hit pysical pixels based on screenshots

ReimarBauer:
**Purpose of PR?**:

Fixes #3110

We can use confidence to ensure. we can hit some pixels

### DIFF
--- a/tutorials/utils/__init__.py
+++ b/tutorials/utils/__init__.py
@@ -76,12 +76,7 @@ def locate_center_on_screen(pic, region=None, confidence=None):
       there; None means an exact match.
     :return: The (x, y) center in logical points, or None if not found.
     """
-    kwargs = {}
-    if region is not None:
-        kwargs['region'] = region
-    if confidence is not None:
-        kwargs['confidence'] = confidence
-    location = pag.locateCenterOnScreen(pic, **kwargs)
+    location = pag.locateCenterOnScreen(pic, region=region, confidence=confidence)
     if location is None:
         return None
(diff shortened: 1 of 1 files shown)


tutorials/utils/__init__.py
```diff
@@ -63,19 +63,25 @@ def screen_scale():
...
-    if region is None:
-        location = pag.locateCenterOnScreen(pic)
-    else:
-        location = pag.locateCenterOnScreen(pic, region=region)
+    kwargs = {}
+    if region is not None:
+        kwargs['region'] = region
+    if confidence is not None:
+        kwargs['confidence'] = confidence
+    location = pag.locateCenterOnScreen(pic, **kwargs)
```
joernu76: pag.locateCenterOnScreen(pic, region=region, confidence=confidence)

joernu76 (changes requested):

tutorials/utils/__init__.py
ReimarBauer: ok

joernu76 (approved):

tutorials/utils/__init__.py
ReimarBauer: Initializing with None resulted in the default value being applied. I must re-initialize all of them with 1.0 or the value subsequently assigned, likely 0.9.
Definitly it was not a good idea to use None.

PR #3129, opened 2026-07-15
increase WMS_request_timeout to 60 seconds

### LINKED ISSUE #3128: default WMS_request_timeout
we should extend this to 60 seconds

ReimarBauer:
**Purpose of PR?**:

Fixes #3128

some of our servers need more than 30 second

### DIFF
--- a/mslib/utils/config.py
+++ b/mslib/utils/config.py
@@ -218,7 +218,7 @@ class MSUIDefaultConfig:
     MSS_auth = {}
 
     # timeout of Url request
-    WMS_request_timeout = 30
+    WMS_request_timeout = 60
 
     WMS_preload = []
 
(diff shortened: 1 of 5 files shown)


mslib/utils/config.py
```diff
@@ -218,7 +218,7 @@ class MSUIDefaultConfig:
     MSS_auth = {}
 
     # timeout of Url request
-    WMS_request_timeout = 30
+    WMS_request_timeout = 60
```
ReimarBauer, line 221: I will use this as (WMS_request_timeout  +1)  * 1000 in the tests

joernu76:
This is generally a problem after a fresh server start, where the first WMS request triggers a lengthy re-loading of all data. Incredibly annoying and happens always after a server hasn't been used for a long time, probably due to reboots?

joernu76 (approved):

ReimarBauer:
> This is generally a problem after a fresh server start, where the first WMS request triggers a lengthy re-loading of all data. Incredibly annoying and happens always after a server hasn't been used for a long time, probably due to reboots?

On my MAC after this change I run by pytest-repeat a count=10 of all tests without any failure. 
When the CI still have timeouts on these tests I add another PR to get informations about the reason for the timeout.

PR #3133, opened 2026-07-15
removed duplicated dataChanged.emit

### LINKED ISSUE #3132: duplicate dataChanged in flightrack.py
This is the inner https://github.com/Open-MSS/MSS/blob/develop/mslib/msui/flighttrack.py#L598 execution 
for https://github.com/Open-MSS/MSS/blob/develop/mslib/msui/flighttrack.py#L610 which sends then again in https://github.com/Open-MSS/MSS/blob/develop/mslib/msui/flighttrack.py#L614

We need only once the dataChanged.

ReimarBauer:
**Purpose of PR?**:

Fixes #3132


joernu76:
I concur with your reasoning. I am aware of certain things being emitted twice *under certain circumstances*. Let's how this goes, with AI we have better debugging options for this stuff.

joernu76 (approved):

PR #3145, opened 2026-07-18
added fork_mscolab_server_before_any_test looking on the request object

### LINKED ISSUE #3144: test should run msolab server very early when it is needed
not any test requres a running mscolab server. 
Depending on test ordering, a request for the mscolab server can happen after many Qt-based tests have already executed, leaving the pytest process in a state that causes the fork()-based server startup to fail or behave incorrectly with Flask-SocketIO.
We should start it by looking at the request object.

ReimarBauer:
**Purpose of PR?**:

Fixes #3144

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -210,5 +210,18 @@ def reset_config():
     read_config_file()
 
 
+def pytest_collection_modifyitems(config, items):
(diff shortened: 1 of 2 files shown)


ReimarBauer:
did not see a difference to the existing solution

PR #3161, opened 2026-08-04
fix: handle missing selected user ids in bulk permissions

xeseing:
## Description

Fixes the default value used when parsing `selected_userids` in bulk permission routes.

## Changes

- Updated `add_bulk_permissions` to default missing `selected_userids` to `"[]"`
- Updated `modify_bulk_permissions` to default missing `selected_userids` to `"[]"`
- Updated `delete_bulk_permissions` to default missing `selected_userids` to `"[]"`

## Why

`json.loads()` expects a string, bytes, or bytearray. The previous default value was a Python list (`[]`), which can raise a `TypeError` if `selected_userids` is missing from the submitted form data.

Using `"[]"` keeps the intended empty-list behavior while passing a valid JSON string to `json.loads()`.

## Testing

- Ran `python -m py_compile mslib/mscolab/blueprints/operation/__init__.py`
- Reviewed the three bulk permission routes locally

### DIFF
--- a/mslib/mscolab/blueprints/operation/__init__.py
+++ b/mslib/mscolab/blueprints/operation/__init__.py
@@ -252,7 +252,7 @@ def get_users_with_permission():
 def add_bulk_permissions():
     fm = current_app.extensions['fm']
     op_id = int(request.form.get('op_id'))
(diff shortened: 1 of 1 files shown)


ReimarBauer (approved):
Thx for fixing this

PR #3169, opened 2026-08-24
linux-aarch64 added

### LINKED ISSUE #3024: Run tests on linux-aarch64
A quick test on my new MacBook Air with Asahi showed that after modifying pixi.toml to include the linux-aarch64 platform at all, all but one test pass just fine. I think we should add it as a platform and run CI against it. GitHub provides hosted runners for Ubuntu on ARM: https://docs.github.com/en/actions/reference/runners/github-hosted-runners#standard-github-hosted-runners-for-public-repositories:~:text=%28public%20preview%29%2C%20windows%2D2022-,Linux,ubuntu%2D24%2E04%2Darm%2C%20ubuntu%2D22%2E04%2Darm,-Windows.

ReimarBauer:
**Purpose of PR?**:

Fixes #3024

adding a new platform linux-aarch64

adjusted the CI workflows_

### DIFF
--- a/.github/workflows/testing-all-oses.yml
+++ b/.github/workflows/testing-all-oses.yml
@@ -18,7 +18,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: ["macos-15", "macos-26", "ubuntu-latest", "ubuntu-24.04-arm"]
+        os: ["macos-14", "macos-15", "macos-26", "ubuntu-latest", "ubuntu-24.04-arm"]
     steps:
     - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v4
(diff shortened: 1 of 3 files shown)


.github/workflows/testing-all-oses.yml
```diff
@@ -18,7 +18,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: ["macos-14", "macos-15", "ubuntu-latest"]
+        os: ["macos-15", "macos-26", "ubuntu-latest", "ubuntu-24.04-arm"]
```
matrss: According to this page: https://endoflife.date/macos macOS 14 is not end-of-life yet and still receives updates. I think we shouldn't drop it from the tests as long as it is still available and supported by Apple.

joernu76 (approved):

PR #3170, opened 2026-08-25
overlay for vibepod with respecting the users .pixi dir

### LINKED ISSUE #3168: setup vibepod
When we use AI-Agents, we should do this in a protected configuration

We can setup for vibepod a Docker overlay which has an ARCHITECTURE pixi installation and uses 
PIXI_CACHE_DETACHED_ENVIRONMENTS_DIR to configure an architecture related /workspace/.vp_pixi-linux-ARCHITECTURE

By this the binaries for the operation system in the docker container did not mixup our host installation.

The workspace on linux has a compatible dev container. But on all platforms we have not a pixi installation.

ReimarBauer:
**Purpose of PR?**:

Fixes #3168

the PR enables that a model uses its own dev container by e.g. vp run (pi, claude, etc)
Tried with pi and poolside/laguna-s-2.1 creates a new folder (patched in https://github.com/Open-MSS/MSS/pull/3169) at  `.pixi-linux-aarch64` instead overwriting the common `.pixi` folder

verified on a Mac M* processor with pi and laguna-s-2.1, needs a verification on an AMD64

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,4 +24,5 @@ tutorials/recordings
 tutorials/cursor_image.png
 # pixi environments
 .pixi
+.pixi-*
 .coverage
--- a/.vibepod/overlay/Dockerfile
+++ b/.vibepod/overlay/Dockerfile
@@ -1,6 +1,6 @@
 # overlay: pixi for different architectures
 # initial version https://www.fz-juelich.de/de/blogs/programmiere/lokale-ki-gestuetzte-pdf-verarbeitung-vibepod-pi-lm-studio-und-pixi
-# 
+#
 # FROM-less Dockerfile fragment; VibePod prepends the base image's `FROM <image>`
 
 ENV PIXI_HOME=/opt/pixi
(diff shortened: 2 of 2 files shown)


ReimarBauer:
on my Mac this works with pi, but I have problems with claude
https://github.com/VibePod/vibepod-cli/issues/157

ReimarBauer:
Now it works also on the Mac with Claude.  

598 passed, 14 skipped, 2 failed in 371.77s (~6:12)

The 2 failures:
- tests/_test_msui/test_topview.py::Test_TopViewWMS::test_server_getmap
- tests/_test_msui/test_sideview.py::Test_SideViewWMS::test_server_getmap

Claude seems to need more RAM when running the tests than pi with my local models.
I recreated my images based on 16GB RAM looked what we have on https://docs.github.com/en/actions/reference/runners/github-hosted-runners
and up to 8 Cores. 
Running the tests used at some point 80% of the defined RAM


I had before only 2GB RAM.

joernu76 (approved):

JanEisermann (commented):
601 test passed, 14 skipped on linux (2nd try). 1st Try: one test failed. Claude says following about it: The cause is the test’s first assertion, `self.operation.id not in self.sm.active_users_per_operation` (tests/_test_mscolab/test_sockets_manager.py:97): it assumes that no previous test in the same xdist worker has left this operation ID in the SocketsManager state. Since pytest randomly shuffles the order for each run, this only applies to certain seeds. A real fix would be to reset `active_users_per_operation` between tests.

PR #3176, opened 2026-09-01
wms_service_cache should use complete url

### LINKED ISSUE #2841: wms_service_cache should use complete url
We maybe should use a slugify version of the full url instead of the base_url in the wms_service_cache

This would allow also to cache urls with parameters for the get_capabilties requests

JanEisermann:
**Purpose of PR?**:

Fixes #2841 

 use of a slugify version of the full url instead of the base_url in the wms_service_cache

tests succeed, not manually verified

**Checklist:**
- [x] Commit has unit tests

### DIFF
--- a/mslib/utils/service_manager.py
+++ b/mslib/utils/service_manager.py
@@ -25,9 +25,6 @@
 """
 import urllib.parse
 
-# query parameters which describe a single request instead of the service itself
-OGC_REQUEST_PARAMS = ("service", "request")
-
 
 def strip_request_params(url):
     """Remove the OGC request parameters (service, request) from an url.
@@ -37,7 +34,7 @@ def strip_request_params(url):
     """
     scheme, netloc, path, params, query, fragment = urllib.parse.urlparse(url)
     kept = [(key, value) for key, value in urllib.parse.parse_qsl(query)
-            if key.lower() not in OGC_REQUEST_PARAMS]
+            if key.lower() not in ("service", "request")]
     return urllib.parse.urlunparse(
         (scheme, netloc, path, params, urllib.parse.urlencode(kept), fragment))
 
--- a/mslib/utils/service_manager.py
+++ b/mslib/utils/service_manager.py
@@ -25,8 +25,6 @@
 """
 import urllib.parse
 
-from slugify import slugify
-
 # query parameters which describe a single request instead of the service itself
 OGC_REQUEST_PARAMS = ("service", "request")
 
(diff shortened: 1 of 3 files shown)


mslib/utils/service_manager.py
```diff
@@ -23,6 +23,44 @@
...
+    The parameters of the GetCapabilities request itself are dropped and the
+    remaining ones are sorted, so that the very same service is found again no
+    matter how the request was spelled. Scheme and host are lower cased because
+    they are case insensitive, the rest of the url is not.
+    """
+    scheme, netloc, path, params, query, fragment = urllib.parse.urlparse(strip_request_params(url))
+    query = urllib.parse.urlencode(sorted(urllib.parse.parse_qsl(query)))
+    normalised = urllib.parse.urlunparse(
+        (scheme.lower(), netloc.lower(), path, params, query, fragment))
+    return slugify(normalised, lowercase=False)
```
ReimarBauer: slugify can make problems here. 
it isn't a normalizer for Urls. It creates strings which can be saved and used when they have special signs in the name. Any none alphabethic character gets replaced into a "-"

By this we can get identical cache keys, e.g. 

http://a.com:1/wms
http://a.com/1/wms

gives both "http-a-com-1-wms"

Maybe use the normalized Url?

Add also a test for those cases, also with params

ReimarBauer (changes requested):
see comment

mslib/utils/service_manager.py
```diff
@@ -23,6 +23,46 @@
     See the License for the specific language governing permissions and
     limitations under the License.
 """
+import urllib.parse
+
+# query parameters which describe a single request instead of the service itself
+OGC_REQUEST_PARAMS = ("service", "request")
```
ReimarBauer: How often is that imported? Or why do you think that it is good to have it defined as a constant?

tests/_test_utils/test_service_manager.py
```diff
@@ -0,0 +1,124 @@
...
+    tests._test_utils.test_service_manager
+    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+    This module provides pytest functions to tests mslib.utils.service_manager
+
+    This file is part of MSS.
+
+    :copyright: Copyright 2025 Reimar Bauer
+    :copyright: Copyright 2025-2026 by the MSS team, see AUTHORS.
+    :license: APACHE-2.0, see LICENSE for details.
```
ReimarBauer, line 12: this looks like a new test module, 

:copyright: Copyright 2026 Jan Eisermann

and there are currently no other contributions? by the team, or?

ReimarBauer (changes requested):
this looks strange, you have added changes which are not yours

JanEisermann:
It happened while I rebased. I don't know why it looks strange like that. I try a new rebase.

mslib/utils/service_manager.py
```diff
@@ -23,6 +23,43 @@
     See the License for the specific language governing permissions and
     limitations under the License.
 """
+import urllib.parse
+
+
+def strip_request_params(url):
+    """Remove the OGC request parameters (service, request) from an url.
```
ReimarBauer, line 30:
```
strip_request_params('http://example.com/wms?dataset=&service=WMS')
# -> 'http://example.com/wms'   (dataset= vanished, not just service=WMS)
strip_request_params('http://example.com/wms?debug&service=WMS')
# -> 'http://example.com/wms'   (bare `debug` flag vanished)
```

This directly contradicts the docstring's guarantee ("All other query parameters are kept in their original order") 


Maybe fix the docstring to what it is doing. Recheck what is needed.

ReimarBauer (changes requested):

ReimarBauer (approved):
thx

PR #3181, opened 2026-09-09
Missed call to __init__ of the super class

### LINKED ISSUE #2750: Missed call to __init__ of the super class

JanEisermann:
**Purpose of PR?**:

Fixes #2750

added superclass __init__

test succeded, not manually verified

**Checklist:**
- [x] Commit has unit tests

with help by Claude

### DIFF
--- a/mslib/utils/netCDF4tools.py
+++ b/mslib/utils/netCDF4tools.py
@@ -185,6 +185,31 @@ def get_latlon_data(ncfile, autoreverse=True):
     return lat_data, lon_data, lat_order
 
 
+def _identify_aggregation_dim(dataset, exclude):
+    """
+    Identify a dimension of an open netCDF4.Dataset that can serve as
+    aggregation dimension of a netCDF4.MFDataset.
+
+    netCDF4.MFDataset insists on a dimension along which the variables of the
+    files of the set are concatenated, and this dimension has to be the
+    leftmost dimension of at least one variable. MFDatasetCommonDims does not
+    concatenate anything (all its files share the same dimensions), it
+    nevertheless has to name such a dimension when calling the constructor of
+    its base class. An unlimited dimension is preferred, as this is the
+    dimension the base class would choose itself.
+
+    Returns None if no suitable dimension exists, leaving the error reporting
+    to the base class.
+    """
+    for dim_name, dim in dataset.dimensions.items():
+        if dim.isunlimited():
+            return dim_name
+    for var_name, variable in dataset.variables.items():
+        if var_name not in exclude and len(variable.dimensions) > 0:
+            return variable.dimensions[0]
+    return None
+
+
 class MFDatasetCommonDims(netCDF4.MFDataset):
     """MFDatasetCommonDims(self, files, exclude=[], require_dim_num=False)
 
@@ -235,9 +260,6 @@ class MFDatasetCommonDims(netCDF4.MFDataset):
         GRIB1/2 files. (mr 03Aug2012)
         @param require_dim_num: see above.
         """
-        # Open the master file in the base class, so that the CDFMF instance
-        # can be used like a CDF instance.
-
         exclude = exclude or []
         skip_dim_check = skip_dim_check or []
         if isinstance(files, str):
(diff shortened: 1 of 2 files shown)


mslib/utils/netCDF4tools.py
```diff
@@ -185,6 +185,31 @@ def get_latlon_data(ncfile, autoreverse=True):
     return lat_data, lon_data, lat_order
 
 
+def _identify_aggregation_dim(dataset, exclude):
```
matrss, line 188: Why is this function necessary? Are there real word cases where selecting the unlimited dimension (i.e. `aggdim=None` in the `super().__init__` call) is not enough?

mslib/utils/netCDF4tools.py
```diff
@@ -235,21 +260,27 @@ def __init__(self, files, exclude=None, skip_dim_check=None,
...
+        # and the file format information of this instance.
+        # Only the master is handed over, because netCDF4.MFDataset joins the
+        # *equally* named variables of all files of the set along an
+        # aggregation dimension, while this class joins the *differently* named
+        # variables of files that all share the same dimensions. The
+        # aggregation dimension required by the base class is looked up in
+        # advance, as the master is not open yet at this point.
+        with netCDF4.Dataset(master) as cdf_master:
+            aggdim = _identify_aggregation_dim(cdf_master, exclude)
+        super().__init__([master], exclude=exclude, aggdim=aggdim)
```
matrss, line 282: ~The constructor of MFDataset expects all files that comprise the MFDataset as its first argument, not just the first/master file.~ Now that I've read the comment above, I am of the opinion that subclassing from MFDataset is just wrong if this class is not supposed to concatenate the files along a dimension.

mslib/utils/netCDF4tools.py
```diff
@@ -296,26 +332,27 @@ def __init__(self, files, exclude=None, skip_dim_check=None,
                     raise IOError("number of dimensions not consistent in master "
                                   f"'{master}' and '{f}'")
 
+            if part.file_format == "NETCDF4":
+                raise ValueError("MFNetCDF4 only works with NETCDF3_CLASSIC, "
+                                 "NETCDF3_64BIT and NETCDF4_CLASSIC "
+                                 "formatted files, not NETCDF4")
```
matrss, line 338: It does not seem like a good thing to not support the latest version of netCDF.

mslib/utils/ogcwms.py
```diff
@@ -178,45 +178,37 @@ def __init__(self, url, version=None, xml=None, username=None, password=None,
...
         # (mss) Store capabilities document.
         self.capabilities_document = reader.capabilities_document
-        # (mss)
 
-        # build metadata objects
-        self._buildMetadata(parse_remote_metadata)
+        if isinstance(xml, str):
+            # (mss) owslib parses with lxml, which rejects str input carrying an
+            # encoding declaration.
+            xml = xml.encode("utf-8")
```
matrss, line 208: Parts of this are redundant with the superclasses constructor (compare with https://github.com/geopython/OWSLib/blob/9c94121ca2f6cedf9d50b218a3882c06698c783e/owslib/map/wms111.py#L54, I assume this was copied over from some old version and never updated). The subclass should not repeat what the superclass already does.

matrss (commented):
I do not feel too confident reviewing this, as I have no experience with what this part of MSS is supposed to do and why it does what it does. Nevertheless, some thoughts.

matrss:
What would it take to replace MFDatasetCommonDims entirely with `xarray.open_mfdataset` or `xarray.combine_by_coords`?

mslib/utils/netCDF4tools.py
joernu76: Disagree. NetCDF4 introduces a *lot* of complicated features. We describe a pretty reasonable format for MSS supported files, which does not use any NetCDF4 features. Every model file I have ever seen uses only NETCDF4_CLASSIC features.

*Disallowing* NETCDF4 is not strictly necessary, though, as you can store a "NETCDF4_CLASSIC" file inadvertently as "NETCDF4".

Summarising: Do not support NETCDF4 (-only features), get rid of this check here.

joernu76:
I do not understand this MR. It neither removes the copy-pasting of classes nor exploits it xarray to remove a lot of code.

JanEisermann:
The issue #2750 this PR belongs to is now divided in 2 new issues (#3184 and #3185). See reason here: https://github.com/Open-MSS/MSS/issues/2750#issuecomment-5604409663.

PR #3183, opened 2026-09-09
use a rubber band selection for moving or deleting multiple waypoints

### LINKED ISSUE #1576: move or delete several waypoints in one step
it would be useful to be able to delete or graphically move several waypoints simultaneoulsy

ReimarBauer:
**Purpose of PR?**:

Fixes #1576

https://github.com/user-attachments/assets/d749ef92-d42e-40e9-9ae2-8ef7d155ae85

with support from claude

### DIFF
--- a/mslib/msui/mpl_pathinteractor.py
+++ b/mslib/msui/mpl_pathinteractor.py
@@ -695,17 +695,21 @@ class PathV_Plotter(PathPlotter):
         self.numintpoints = numintpoints
         self.redraw_xaxis = redraw_xaxis
         self.clear_figure = clear_figure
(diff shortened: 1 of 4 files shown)


joernu76:
Nice, but this introduces 329 lines-of-code in an already questionable  part of the code (this broke a lot in the past) for making it slightly more comfortable to work with (I admit, there are some use cases). In light of keeping things manageable I propose to reject both the MR and the issue.

ReimarBauer:
We can reopen this issue when we plan to refactor that section of the path interactor. I previously attempted a manual implementation, but we should aim to avoid a situation where a model becomes a necessity for this specific area of the code.

PR #2432, opened 2024-07-08
Fix flask_migrate setup

matrss:
**Purpose of PR?**:

Fixes #2416.

**Does this PR introduce a breaking change?**

**If the changes in this PR are manually verified, list down the scenarios covered:**:

**Additional information for reviewer?** :
_Mention if this PR is part of any design or a continuation of previous PRs_

**Does this PR results in some Documentation changes?**
_If yes, include the list of Documentation changes_

**Checklist:**
- [x] Bug fix. Fixes #<issue number>
- [x] New feature (Non-API breaking changes that adds functionality)
- [ ] PR Title follows the convention of  `<type>: <subject>`
- [x] Commit has unit tests

PR #3186, opened 2026-09-10
Refactor MSColab away from a bunch of global variables

### LINKED ISSUE #2442: Refactor MSColab away from a bunch of global variables

JanEisermann:
**Purpose of PR?**:

Fixes #2442 

Commit 6779420 replaced the global APP with an AppFactory. Without a global app, there’s no longer a way to access the config or DB session from anywhere—both are now handled via `current_app` and `db.session`, respectively, and these only work within an application context. As a result, the context shifts from the called functions to the caller.

Previously, virtually every function in `seed.py` had its own mini-context:

def add_user(...):
    APP.config[‘SQLALCHEMY_DATABASE_URI’]=APP.config[‘SQLALCHEMY_DATABASE_URI’] 
    with APP.app_context():
        ...

This only worked because there was exactly one app that the module could pin itself to upon import. With the factory, it is no longer up to the function to decide in which app it runs—only the caller knows that.

test succeeded, not manually verified 

**Checklist:**
- [x] Commit has unit tests

small corrections by claude

### DIFF
--- a/mslib/mscolab/models.py
+++ b/mslib/mscolab/models.py
@@ -103,19 +103,18 @@ class User(db.Model):
 
     def generate_auth_token(self, expiration=None):
         # Importing conf here to avoid loading settings on opening chat window
-        from mslib.mscolab.app import APP
-        expiration = APP.__dict__.get('EXPIRATION', expiration)
+        from mslib.mscolab.conf import mscolab_settings
         if expiration is None:
-            expiration = 864000
-            token = jwt.encode(
-                {
-                    "id": self.id,
-                    "exp": datetime.datetime.now(tz=datetime.timezone.utc) + datetime.timedelta(seconds=expiration)
-                },
-                APP.config['SECRET_KEY'],
-                algorithm="HS256"
-            )
-            return token
+            expiration = getattr(mscolab_settings, 'EXPIRATION', 864000)
+        token = jwt.encode(
+            {
+                "id": self.id,
+                "exp": datetime.datetime.now(tz=datetime.timezone.utc) + datetime.timedelta(seconds=expiration)
+            },
+            mscolab_settings.SECRET_KEY,
+            algorithm="HS256"
+        )
+        return token
 
     @staticmethod
     def verify_auth_token(token):
--- a/mslib/mscolab/seed.py
+++ b/mslib/mscolab/seed.py
@@ -6,6 +6,9 @@
 
     Seeder utility for database
 
+    All functions of this module access the database and the app configuration, so
+    they have to be called within an application context.
+
     This file is part of MSS.
 
     :copyright: Copyright 2019 Shivashis Padhi
@@ -29,7 +32,7 @@ import git
 from pathlib import Path
 from sqlalchemy.exc import IntegrityError
 
-from mslib.mscolab.app import APP
+from flask import current_app
 from mslib.mscolab.models import User, db, Permission, Operation
 
 
@@ -50,12 +53,8 @@ XML_CONTENT_INIT = """<?xml version="1.0" encoding="utf-8"?>
 # Todo: refactor move to mscolab.utils
 def add_all_users_to_all_operations(access_level='collaborator'):
     """ on db level we add all users as collaborator to all operations """
-    APP.config['SQLALCHEMY_DATABASE_URI'] = APP.config['SQLALCHEMY_DATABASE_URI']
-    APP.config['SQLALCHEMY_TRACK_MODIFICATIONS'] = False
-    with APP.app_context():
-        all_operations = Operation.query.all()
-        all_path = [operation.path for operation in all_operations]
-        db.session.close()
+    all_operations = Operation.query.all()
+    all_path = [operation.path for operation in all_operations]
     for path in all_path:
         if path == "TEMPLATE":
             access_level = 'admin'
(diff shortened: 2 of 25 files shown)


joernu76 (changes requested):
Nice work. But I do not understand this *massive* MR (without looking stuff up and KI help).

Replacing APP by current_app seems sensible as some other related changes. But there are several section that added a "with [...]app_context():" or moved lines within such context. Please elaborate in you MR description the intent and reasoning behind the general changes (both for documentation and making my life easier).

This seems to remove the APP global variable; does this *fully* or *partially* adress #2442 ?

Some minor comments in the files.

mslib/mscolab/models.py
```diff
@@ -103,19 +103,18 @@ def verify_password(self, password_):
...
-                algorithm="HS256"
-            )
-            return token
+            expiration = getattr(mscolab_settings, 'EXPIRATION', 864000)
+        token = jwt.encode(
+            {
+                "id": self.id,
+                "exp": datetime.datetime.now(tz=datetime.timezone.utc) + datetime.timedelta(seconds=expiration)
+            },
+            mscolab_settings.SECRET_KEY,
```
joernu76: not current_app.config['SECRET_KEY']  ?

mslib/mscolab/seed.py
```diff
@@ -64,184 +63,152 @@ def add_all_users_to_all_operations(access_level='collaborator'):
...
+    if operation:
+        user = User.query.filter_by(emailid=emailid).first()
+        if user:
+            perm = Permission.query.filter_by(u_id=user.id, op_id=operation.id).first()
+            if perm is None:
+                return False
+            elif perm.access_level not in ["admin", "creator"]:
+                return False
+            operation.active = False
+            db.session.commit()
```
joernu76, line 211: Unrelated to this MR, but isn't here a
```
        return True
return False
``` missing to be consistent with the other methods?

joernu76 (changes requested):

JanEisermann:
@joernu76 I updated the description

mslib/mscolab/seed.py
JanEisermann: It is the same like before the change. I only removed with "APP.app_context()"

joernu76: I know, one could either fix it en passant, or, more cleanly in a separate PR.

joernu76 (approved):

ReimarBauer (changes requested):
in the issue stated
"""
It would also make it possible to construct differently configured Flask app instances for different tests. This would mainly come in handy to make tests/_test_mscolab/test_server_auth_required.py run as part of an ordinary run of the test suite.
"""
can you address this?

ReimarBauer:
Feedback from Claude

This PR doesn't fully solve issue #2442.

What it does well
- mslib/mscolab/app/__init__.py implements a real, idiomatic create_app() factory (unbound db/migrate/mail extensions bound via .init_app(), blueprints registered per-instance).
- Globals are cleanly removed from seed.py, chat_manager.py, file_manager.py, sockets_manager.py, models.py, mscolab.py — all switched to current_app/db.session/passed-in app.
- It concretely fixes the motivating pain point behind #2432: test_migrations.py::test_upgrade_from now builds a genuine second app instance via create_app(source_settings) instead of a hand-rolled bare Flask("whatever").

Where it falls short of #2442's stated goals
1. Goal #3 is untouched: tests/_test_mscolab/test_server_auth_required.py still has its pytest.skip(..., allow_module_level=True) import guard and isn't part of the PR's changed files at all — even though the refactor incidentally removed the module-level side effects (APP = create_app(...), CORS(...) at import time) that used to make importing server.py unsafe, nobody wired the test back into the normal suite.
2. Goal #2 (independent per-test DB) is only a one-off, not structural: tests/fixtures.py's mscolab_session_app is still session-scoped, i.e. still one shared app/DB across the whole test session.
3. CI is red: 5 of 6 test-matrix jobs fail (ubuntu-latest, ubuntu-24.04-arm, macos-14/15/26) — only lint/codespell/docs/import-contracts pass. The most plausible cause: seed.py's per-function with APP.app_context(): blocks were removed and pushed onto callers, and that's a wide, easy-to-miss migration (the author's own PR description flags this exact risk and says "not manually verified"). Couldn't fetch the actual CI logs (403, needs repo admin rights) to confirm the traceback, so that's a strong hypothesis, not a confirmed root cause.

Verdict: solid factory implementation, real progress on the underlying architecture, but as currently open it shouldn't be considered done/mergeable — CI is failing on nearly every platform, and one of the three concretely-named goals in #2442 (the auth-required test) was never actually acted on.

JanEisermann:
Some commits are duplicated. No worries, it happened because I had to do a `git pull` and resolve some merge conflicts. The duplicated commits contain the same changes as the original ones.

ReimarBauer (approved):
Thx

PR #3187, opened 2026-09-11
 use of xarray instead of the MFDatasetCommonDims class

### LINKED ISSUE #3185: We should use xarray instead of the MFDatasetCommonDims class
I quote @matrss to describe the issue.

"[The MFDatasetCommonDims] class seems to have been copied from MFDataset in the past and has diverged from the super classes implementation in the meantime. It's goal seems to be to merge datasets with different variables in addition to or instead of concatenating along one dimension. Today xarray exists, and this goal could be accomplished in a single line with xarray.open_mfdataset or xarray.combine_by_coords. I think it would improve the code the most if we could refactor it to use xarray instead and rely on a mature library for this purpose."

This issue originated from the following issue: https://github.com/Open-MSS/MSS/issues/2750

JanEisermann:
**Purpose of PR?**:

Fixes #3185 

 use of xarray instead of the MFDatasetCommonDims class, which is removed within this PR

tests succeeded, not manually verified 

**Checklist:**
- [x] Commit has unit tests

with help by Claude


ReimarBauer (changes requested):
until Joern looked on that I hold it blocked. 

@joernu76 already was against thi change in a different PR

mslib/mswms/mss_plot_driver.py
```diff
@@ -156,14 +156,14 @@ def _set_time(self, init_time, fc_time):
 
         # Open NetCDF files as one dataset with common dimensions.
         logging.debug("opening datasets.")
-        dsKWargs = self.data_access.mfDatasetArgs()
-        dataset = netCDF4tools.MFDatasetCommonDims(self.filenames, **dsKWargs)
+        dataset = netCDF4tools.open_mfdataset_commondims(
```
joernu76, line 159: can xarray.open_mfdataset be used here? *That* would really remove code?

JanEisermann: I let run claude over it. Claude said following: 
I tried it, and the short answer is that it would cost us more
than the ~25 lines it saves. Three findings (xarray 2026.2.0, measured today):

**1. `open_mfdataset` always requires dask.** 
It hard-codes `open_kwargs = dict(engine=engine, chunks=chunks or {}, **kwargs)`
(`xarray/backends/api.py`), so there is no way to opt out of the chunk manager.
Without dask installed it fails with
`ImportError: chunk manager 'dask' is not available`.
That would add dask + toolz + cloudpickle + partd + locket + fsspec to the MSS
runtime dependencies, which we otherwise do not need anywhere.

**2. With dask it reads whole fields instead of slices.** 
Our NetCDF files have no internal chunking, so the default `chunks={}` gives exactly one chunk per
variable, and dask does not fuse the subsequent slicing into the read. On a
synthetic `(time=10, lev=60, lat=200, lon=400)` float32 field (192 MB), reading
the single 2-D slice that `HorizontalSectionDriver` asks for
(`var[3, 20, ::-1, :]`):

| variant                                                                | time    | peak RSS delta |
|--------------------------------------------------------------|-----------|-----------------------|
| `xr.open_dataset` (what this PR does)          | 0.001 s| 0 MB                  |
| `open_mfdataset`, default `chunks={}`        | 0.039 s| +186 MB            |
| `open_mfdataset`, `chunks="auto"`              | 0.066 s| +122 MB           |  
| `open_mfdataset`, `chunks={time:1, lev:1}` | 0.004 s| 0 MB                 |

So the whole 4-D field is pulled from disk for every single horizontal section.
For a WMS server serving one level per request that is a real regression, and
real ECMWF fields are a lot bigger than 192 MB. The last row shows it *can* be
fixed, but only with an explicit per-dimension `chunks` dict — which needs the
lat/lon dimension names, so we would have to open the first file and run
`identify_CF_lonlat()` before the actual open. That is more code, not less.

**3. `skip_dim_check` would not go away.** 
It would have to be reimplemented as a `preprocess` hook 
that recognises the master file via `ds.encoding["source"]`,
so the special-case code stays either way.

The net result would be six new dependencies, whole-field reads, and roughly the
same line count, so I kept the explicit `xr.merge` of lazily opened datasets —
it is effectively a dask-free `open_mfdataset` with `concat_dim=None`.

What I did change in response to your comment: the helper is now noticeably
smaller and explains itself.

- `contextlib.ExitStack` replaces the manual `close_all()` plus the two `except`
  branches; `stack.pop_all().close` becomes the merged dataset's close callback,
  and the failure path still closes every file.
- `dataset.drop_dims(skip_dim_check, errors="ignore")` replaces the comprehension
  over all variables and their `dims`.
- The docstring now states *why* `xarray.open_mfdataset()` is not used, with the
  two reasons above, so the next reader does not have to re-derive this.

PR #3190, opened 2026-09-11
enabled to copy as CSV from table view

### LINKED ISSUE #3188: add a c/p feature to table view
We need a c/p feature in table view to save its content as CSV to the clipboard and use it somewhere else

We have to reuse the icon defined for the other views.

A complete empty column should not be copied to the CSV file.

ReimarBauer:
**Purpose of PR?**:

Fixes #3188

this makes it possible to copy the data from the table view in a csv format to another application.

assised by claude


joernu76 (changes requested):
I am missing a basic test? Claude should be able to construct one.

ReimarBauer:
> I am missing a basic test? Claude should be able to construct one.

More basic than this one? https://github.com/ReimarBauer/MSS/blob/05d1f1b1c9d9a11adadbabf33575a59aa1e1a64d/tests/_test_msui/test_tableview.py#L489

joernu76 (approved):
Was this there 1 hour ago? Nice.

PR #3209, opened 2026-09-15
Add typed REST contract layer for mscolab (mslib/mscolab/api/)

### LINKED ISSUE #3208: Add a typed REST request/response contract for mscolab (mslib/mscolab/api/)
REST/socket payloads between mslib/mscolab (server) and mslib/msui/mscolab.py (client) are ad-hoc dicts with no machine-checkable link between a client call site and its server route. A rename or shape change on one side currently has no compiler/linter/test signal on the other

- This task is a wire-format migration only, behavior must not change.
- Create mslib/mscolab/api/ package: schemas.py (`stdlib @dataclass` request/response shapes per REST route — no new runtime dependency), endpoints.py (named endpoint registry, replacing bare strings like "create_operation").
- Migrate blueprint handlers and their client call sites route-by-route to build from / parse into these dataclasses, for every route that has a real msui caller. Convert any duplicate ad-hoc call sites onto the same schema as they're found.
- Add tests/_test_api/ with a round-trip test per schema (serialize → parse → compare), wired into test-fast so it stays a fast tripwire.
- Rework the server-isolation import-linter contract in setup.cfg so mslib.msui can reach mslib.mscolab.api without a growing list of per-symbol exceptions.
- Move mslib/mscolab/events.py and message_type.py into mslib/mscolab/api/ alongside the new modules — one clearly-named location for "this is the contract" instead of scattered files.
- Mark pre-existing bugs when you find them through the refactoring. by "NOTE: pre-existing bug"

ReimarBauer:
Introduce mslib/mscolab/api/ (schemas.py + endpoints.py) as a typed request/response contract between mslib/msui and the mscolab server, replacing ad-hoc dicts and bare-string endpoint names for every route with a real client caller (28 of 49 routes, 5 known duplicate call sites unified). Move events.py/message_type.py into api/ alongside it, and collapse the server-isolation import-linter contract to a single mslib.msui.* -> mslib.mscolab.api.* rule.

Add tests/_test_api/ with 124 round-trip tests (serialize -> parse -> compare), wired into test-fast as a fast tripwire.

No behavior changes. Several pre-existing bugs found while migrating are flagged with `# NOTE` comments rather than fixed here (tracked in restructuring-todo.md for follow-up).

assisted by claude

**Purpose of PR?**:

Fixes #3208

Yes, API Changes

**Checklist:**
- [X] Commit has unit tests


ReimarBauer:
This needs a rebase before we can merge it

ReimarBauer:
This is also a great step forward to be able to do a package split

## Layout

```
MSS/
  pyproject.toml            # "open-mss" umbrella distribution (current file, trimmed)
  pixi.toml                 # umbrella dev env — unchanged workflow: `pixi shell -e dev`
  packages/
    mss-utils/    {pyproject.toml, pixi.toml, src/mslib/utils/...}
    mss-mswms/    {pyproject.toml, pixi.toml, src/mslib/mswms/...}
    mss-mscolab/  {pyproject.toml, pixi.toml, src/mslib/mscolab/...}
    mss-msui/     {pyproject.toml, pixi.toml, src/mslib/{msui,plugins,support,autoplot}/...}
    mss-msidp/    {pyproject.toml, pixi.toml, src/mslib/msidp/...}
  tests/            # unchanged — _test_msui, _test_mscolab, _test_mswms, _test_utils, _test_plugins already mirror this
```

`mslib` becomes a PEP 420 namespace package (drop the current
`mslib/__init__.py`'s package-owning role — its only real content,
`__version__ = importlib.metadata.version("open-mss")`, needs to move
somewhere that still works once `mslib` has no single owner; simplest is to
drop the namespace-level `__version__` and read it per-distribution, or keep
a thin non-namespace `mslib/__init__.py` shipped by `mslib-utils` since
everything depends on it anyway).

Each `packages/*/pixi.toml` is a minimal, self-contained env for isolated
dev/test of that one package (its own deps only — this is what makes
isolation verifiable, unlike PR #2390's guards):

joernu76:
Opus5.5:

● Review: typed REST contract layer for mscolab (2 commits by Reimar Bauer → origin/develop)

  Verdict: good direction, but it needs another round before merge. The approach fits the repo's constraints: stdlib dataclasses, no new dependencies, and it preserves the wire format. The
  migration is careful, and pre-existing quirks are left alone on purpose rather than quietly changed. What holds it back: one client-side regression, one known bug that's commented instead
  of fixed, and a test story that promises more than the tests check.

  What I ran

  - tests/_test_api: 124 passed.
  - tests/_test_mscolab (server suite): 149 passed, 1 skipped.
  - GUI tests (test_mscolab*, test_multiple_flightpath_dockwidget, and related): not run. They fail at import because of a broken package in the local dev env (metpy → importlib_metadata
    returns None). The MR didn't cause this, but the client half is untested by me.
  - flake8 crashes in that same env, and lint-imports isn't installed there, so neither ran. Your pixi is also too old for the v7 pixi.lock; updating pixi should fix all of this.

  Must fix

  1. Delete-account regression (mslib/msui/mscolab.py:518).
     - Before: status_code == 200 and response.json()[...] only parsed the body on a 200.
     - Now: DeleteOwnAccountResponse.from_text(response.text) always runs first. A non-200 response with a non-JSON body (an HTML 500, or a 401 from basic auth) now raises JSONDecodeError
       instead of simply not logging out.
     - Fix: check the status code before parsing.
  2. The unarchive bug is commented, not fixed (mscolab_archive_browser.py:63-69).
     - The comment itself confirms that request_post() takes no timeout argument, so the Unarchive button always raises TypeError. I checked socket_control.py:219; that's correct.
     - The fix is deleting one keyword argument. Shipping a feature we know is broken, with a paragraph explaining why, is worse than a one-line fix.
  3. The round-trip tests can't catch the drift they're said to catch.
     - schemas.py's docstring and ARCHITECTURE.md say the tests/_test_api round-trips are "the tripwire that keeps a schema and its route's actual wire format in sync". They also say that
       for a migrated route you "only change the dataclass; run test-api".
     - In fact every test in test_schemas.py round-trips a schema against itself. None of them touches a Flask route or real request.form string values.
     - Fix: add a few contract tests using the Flask test client and the mscolab_app fixture (one per response shape: plain text, json.dumps, jsonify, bare "False"). Or tone the claim down
       in the docs.
  4. The server now round-trips manager output through dataclasses for no gain (chat/__init__.py:57, operation/__init__.py:150,169).
     - X.from_dict(m) followed by .to_dict() adds a new KeyError → 500 failure mode.
     - It also silently drops any field a manager adds later. The same thing already happens with get_all_changes, which passes plain dicts through.
     - Fix: either have the managers build the dataclasses directly, or pass the dicts through as get_all_changes does. Pick one and use it everywhere.
  5. setup.cfg:88 references restructuring-todo.md, which doesn't exist in the repo.

  Should fix / discuss

  - Nothing keeps mscolab.api a leaf module.
    - The ignore mslib.msui.* -> mslib.mscolab.api.* removes that edge from the import graph. If api/ ever imports models or Flask, the GUI pulls in the server stack and no contract fails.
    - Fix: add a forbidden contract with source_modules = mslib.mscolab.api and forbidden_modules = the rest of mslib.mscolab, plus flask/sqlalchemy.
  - "Rename is a single-point change" isn't true yet.
    - Server routes still use bare strings (@OPERATION_BP.route('/create_operation')); endpoints.py is only used by the client.
    - Using the constants in the route decorators would make the claim true.
  - Too much boilerplate (1246 lines for about 28 routes).
    - About 6 classes have the same {success, message} shape: DeleteOperation, DeleteBulkPermissions, BulkPermissions, SetVersionName, and others.
    - dataclasses.asdict plus one generic from_dict(cls, data) built from fields() would replace most hand-written to_dict/from_dict pairs. That removes the exact field-name duplication this
      layer is meant to get rid of.
  - Inconsistent API across schemas. Some have to_text, some to_dict, some both; some have from_form, others from_args_and_form.
    - IdpLoginAuthResponse.to_dict() is a trap: it returns {"success": ...} even when success=True, dropping token and user.
  - About 6 "pre-existing bug, preserved as-is" comments (get_all_changes never returning its error, show_error_popup not existing, users_response != "False" always being true).
    - Documenting them is fine for a no-behaviour-change refactor, but each should become an issue. Comments like these go stale fast.
    - One now leaves a dead line in the code: jsonify(GetAllChangesResponse(success=False, ...)) builds a response and throws it away.
  - The types are annotations only. op_id: object appears in three places and nothing is validated at runtime. That's fine, but "typed" oversells it a little.

  Small behaviour changes (all fine, but worth knowing)

  - message_attachment now parses int(message_type) before the membership check. A non-member request with no message_type now gets a 500 instead of "False".
  - set_label_text dropped the leading slash ("/creator_of_operation" → relative). That's actually a fix for servers deployed under a sub-path.
  - undo_changes on the client now treats a response as success only if it is exactly "True". Before, anything other than "False" counted, including an HTML error page.

  Nits

  - schemas.py:1099 still says mscolab.events.SocketEvents; it should be api.events.
  - The new file headers say "Copyright 2026 Reimar Bauer", while the moved files use "by the MSS team".
  - The client still checks raw response.text == "False" at mscolab.py:802 (leave operation) even though that route has a schema.

  I can draft fixes for items 1, 2, 5, the leaf contract, and a first Flask test-client contract test if you want them. Items 4 and the boilerplate cleanup are design choices better settled
  with Reimar first.

ReimarBauer:
@joernu76 when this is merged I do a schema cleanup #3217. I looked already also into dropping support for older servers. A fix for #3220 is also needed. But that are all quite different task to those I included.

PR #3214, opened 2026-09-18
autoplot ftml files need the path and file stored

### LINKED ISSUE #2714: autoplot ftml files need the path and file stored

JanEisermann:
**Purpose of PR?**:

Fixes #2714

The configuration now stores the flight track with its path and `--fpath` overrides the
directory. This fixes the two cases the GUI still got wrong:

- a configuration selected in the dockwidget which names the flight track without a path
  was looked up in the working directory msui happens to be started in, not next to the
  configuration file

- a row whose file is not there killed the download through `SystemExit`, with a path the
  user never typed and no hint where it came from. For example: 

 The flight track file of 'flight1' does not exist:
   /home/mss/example.ftml

   The configuration /home/mss/mssautoplot.json names it 'example.ftml', without a
   directory, so it is looked up next to that file.
   Correct it there, or open the flight track in the MSUI, save it and add the row again.

or 

The flight track file of 'flight1' does not exist:
   /home/mss/flight1.ftml

   The flight track was never saved, only its name is stored, so the file is looked up in
   the working directory /home/mss.
   Save the flight track and add the row again.

tests succeeded, manually verified: Download Plots button works correctly and hints are shown when it fails (see above)

docs/autoplot_dock_widget.rst: how a selected configuration resolves a relative flight
track, and what the Download Plots button does with a file which is not there.

**Checklist:**
- [x] Bug fix. Fixes #2714 
- [x] Commit has unit tests

coded with help by Claude


mslib/utils/config.py
```diff
@@ -694,7 +665,10 @@ def compare_data(default, user_data):
     if not isinstance(default, dict) and not isinstance(default, list):
         if isinstance(default, float) and isinstance(user_data, int):
             user_data = float(default)
-        if isinstance(match_type(default), UrlType) and isinstance(match_type(user_data), StrType):
+        if isinstance(default, str) and isinstance(user_data, str):
```
ReimarBauer: This weakens validation for other filepath-typed defaults like data_dir and wms_cache.


this needs a careful review. Have you looked what happens on the other server configs?

ReimarBauer (changes requested):
1. mslib/utils/config.py:668 — tighten the string-match fix (correctness, most severe)

What's there now: compare_data() was changed to accept any string as a match for any string-typed default — this was meant to fix comparison for the automated_plotting_flights path field, but it's not scoped to that field.
Fix: Scope the relaxed check to the specific field/type this PR is about (e.g. keep using match_type()'s Url/Filepath/Str classification for other string keys like data_dir, wms_cache, and only relax it for the path-list field being changed).

2. mslib/msui/autoplot_dockwidget.py:384 — wrong error message for legacy configs (correctness)

What's there now: missing_flighttrack_message() assumes flighttrack_sources.get(entry) being None + no parent dir means "never saved." But flighttrack_sources is only populated inside configure_from_path() — if the config loaded as the default startup config, it's empty even for legitimately-saved legacy (bare-filename) entries.
Fix: Either populate flighttrack_sources on default-config load too, or don't conclude "never saved" purely from the dict being empty — check whether a config file is actually the source before choosing that message.

3. mslib/msui/autoplot_dockwidget.py:307 — path-collision overwrites (correctness)

What's there now: The dict comprehension building flighttrack_sources keys by resolved absolute path, so two rows that resolve to the same file (one written as a bare filename, one as an absolute path) collide and the later one silently wins.
Fix: Decide desired behavior on collision (keep first, warn, or merge) instead of silent last-write-wins — at minimum a comment explaining why last-write-wins is acceptable, if it is.

4. mslib/msui/autoplot_dockwidget.py:358 — redundant resolution (efficiency, minor)

What's there now: missing_flighttrack() calls resolve_ftml_path() on every row again, even though resolve_flights_paths() already resolved them at load time.
Fix: Reuse the already-resolved paths (e.g. from flighttrack_sources keys) instead of re-resolving, or factor resolution into one shared call site.

Items 1–3 are real correctness bugs worth fixing before merge; item 4 is a nice-to-have cleanup.

tests/_test_msui/test_autoplot_dockwidget.py
```diff
@@ -0,0 +1,223 @@
...
+"""
+
+    tests._test_msui.test_autoplot_dockwidget
+    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+    This module provides pytest functions to tests msui.autoplot_dockwidget
+
+    This file is part of MSS.
+
+    :copyright: Copyright 2026 by the MSS team, see AUTHORS.
```
ReimarBauer: this is a new module?

so author is only Jan :)

when tests are moved from other tests then likly the year needs an adjustment

mslib/msui/wms_control.py
```diff
@@ -718,11 +718,17 @@ def style_changed(layer):
                     break
 
     def leftrow_is_selected(self, vtime):
+        layer = self.multilayers.get_current_layer()
+        if layer is None:
+            # A row of the autoplot dockwidget can be selected before this widget is
```
ReimarBauer, line 723: this also drops the cbValidTime.setCurrentText(vtime); valid_time_changed()

usually we csn only change these when a layer was choosen. So that should be ok.

docs/mssautoplot.rst
```diff
@@ -61,6 +69,12 @@ This command stores the data of topview as PNG for overlays without axis, titles
 This could be used in systems like PLANET. You may want to set large values in the layout
 of topview in the mssautoplot.json.
 
+5. ``mssautoplot --cpath mssautoplot.json --fpath ~/flights/campaign2``
+
```
ReimarBauer, line 74: on recent windows systems "~" is not working as before.

mslib/autoplot/__init__.py
```diff
@@ -573,14 +611,16 @@ def draw(self, flight, section, vertical, filename, init_time, time, url, layer,
 @click.option('--cpath', default=constants.MSS_AUTOPLOT, help='Path of the configuration file.')
 @click.option('--view', default="top", help='View of the plot (top/side/linear).')
 @click.option('--ftrack', default="", help='Flight track.')
+@click.option('--fpath', default="", help='Directory of the flight track files. Overrides the directory stored '
```
ReimarBauer: help='Directory of the flight track files. Replaces the directory stored with the file name in the configuration file, any subdirectory of that stored path is discarded.')

docs/mssautoplot.rst
```diff
@@ -15,6 +15,9 @@ The CLI tool has the following parameters:
 +--------------+-------+----------------------------------------------------------------------+
 | ``--ftrack`` | TEXT  |   Flight track.                                                      |
 +--------------+-------+----------------------------------------------------------------------+
+| ``--fpath``  | TEXT  |   Directory of the flight track files. Overrides the directory       |
```
ReimarBauer: Directory of the flight track files. Replaces the directory stored with the file name in the config file; only the base name of the entry is kept.

docs/mssautoplot.rst
```diff
@@ -61,6 +69,12 @@ This command stores the data of topview as PNG for overlays without axis, titles
 This could be used in systems like PLANET. You may want to set large values in the layout
 of topview in the mssautoplot.json.
 
+5. ``mssautoplot --cpath mssautoplot.json --fpath ~/flights/campaign2``
+
+This command reads the flight tracks from the given directory instead of the directory
+stored in mss_autoplot.json. The file names themselves stay as configured, so the same
+configuration can be used for the flight tracks of several campaigns.
+
```
ReimarBauer, line 81: This command reads the flight tracks from the given directory instead of the directory stored in mss_autoplot.json. Only the base name of each stored flight track is kept, any subdirectory stored with it is discarded, so the "/home/mss/campaign1/example.ftml" of the configuration is loaded from "~/flights/campaign2/example.ftml". This lets the same configuration be used for the flight tracks of several campaigns, provided they share the file names.

mslib/utils/config.py
```diff
@@ -694,6 +694,15 @@ def compare_data(default, user_data):
...
             user_data = float(default)
+        if isinstance(match_type(default), StrType) and isinstance(user_data, str):
+            # A default which is a plain string takes any string, a path included.
+            # match_type() maps absolute paths to FilepathType and urls to UrlType, so
+            # comparing the types would drop e.g. the flight track path of an
+            # "automated_plotting_flights" entry, whose default is "". Defaults which
+            # are a path ("data_dir", "wms_cache", "mscolab_local_data_dir") or a url
+            # ("mscolab_server_url", "default_WMS") keep their own type below and still
+            # require a path, respectively a url.
+            return user_data, True
```
ReimarBauer, line 705: **Impact:** Allows ANY string through if default is string (simpler but may be too permissive)

**develop version (commit `eff4ce837`)** — More targeted fix:
```python
if isinstance(match_type(default), UrlType) and isinstance(match_type(user_data), StrType):
    return user_data, True
```
**Impact:** Only allows string→string when the typed match marks it as a URL (more restrictive, closer to reviewer's intent)

**Reviewer's Intent (1):** Type checks should only relax for StrType defaults, not FilepathType or UrlType. The develop version appears to implement this more precisely by checking the typed result.

ReimarBauer (changes requested):
see comments

ReimarBauer (changes requested):
1. One missing flight-track file now stops the whole command-line run (medium, mslib/autoplot/__init__.py:285)
- What changed: read_ftml() now goes through ftml_file(), which calls SystemExit when the file is missing. draw() catches errors with except Exception, which doesn't catch SystemExit, so the whole run ends.
- Before the PR: a missing file later in the list raised FileNotFoundError, which draw() caught. It printed the error and went on to the next file.
- Example: automated_plotting_flights lists [A.ftml, B.ftml, C.ftml] and B.ftml has been moved. Plots for C.ftml are never made, and a GUI run gets no SUCCESS/FAILURE summary.
- Where it shows up: MSUI checks for missing files up front, which hides this there. It still affects the standalone mssautoplot command, and files deleted between that check and the download.
- Fix: raise a normal exception in ftml_file() instead of calling SystemExit, or catch it per file in draw().

2. The missing-file dialog can give the wrong reason (low, mslib/msui/autoplot_dockwidget.py:379)
- What happens: missing_flighttrack_message tells the user the configuration named the file "without a directory, so it is looked up next to that file" whenever the stored name differs from the resolved path.
- Why that's wrong: resolve_flights_paths also changes names that do include a directory. It expands ~, joins relative paths such as sub/x.ftml onto the config's directory, and follows symlinks and ...
- Example: the config names /data/x.ftml, where /data is a symlink to /mnt/data. The dialog says the file was named without a directory, which is false and points the user to the wrong place.
- Fix: only use that wording when the stored name has no directory part.

mslib/utils/config.py
```diff
@@ -694,6 +694,15 @@ def compare_data(default, user_data):
     if not isinstance(default, dict) and not isinstance(default, list):
         if isinstance(default, float) and isinstance(user_data, int):
             user_data = float(default)
+        if isinstance(match_type(default), StrType) and isinstance(user_data, str):
```
ReimarBauer: Only automated_plotting_flights needs to accept a path, but the PR relaxes compare_data for every option whose default is a string. The fix is to make that exception explicit, using the same pattern the file already uses for fixed_list_options.

1. List the options that accept any string (mslib/utils/config.py, near line 308):
    `# List options with fixed length`
    fixed_list_options = ["MSCOLAB_timeout", ]
    `# List options whose string entries take any string, e.g. a flight track path`
    free_string_list_options = ["automated_plotting_flights", ]
Also add "free_string_list_options" to the del default_options[key] list (around line 414). Otherwise it would show up as a user option.

2. Pass a flag only from the list-structure loop in merge_dict (around line 657):
data, match = compare_data(
    los_key_item, new_dict[key][i],
    any_string=key in MSUIDefaultConfig.free_string_list_options)

3. Replace the broad StrType rule in compare_data with the flag:
def compare_data(default, user_data, any_string=False):
    ...
    if not isinstance(default, dict) and not isinstance(default, list):
        if isinstance(default, float) and isinstance(user_data, int):
            user_data = float(default)
        if any_string and isinstance(default, str) and isinstance(user_data, str):
            return user_data, True
        # ... existing UrlType / type comparison unchanged
The flag also has to be passed on in both recursive calls (list and dict branches), because the flight row is a list nested inside a list.

Why this is better:
- Options like filepicker_default, MSCOLAB_category, new_flighttradule/function entries and CRS go back to their old checks. Nothingoutside autoplot changes.
- Relative names such as example.ftml still work in the flight rowe PR. They count as StrType, not FilepathType.
- The long comment explaining which defaults are paths or URLs can go, because the rule no longer reaches them.
- The two tests the PR added (test_user_option_with_path, test_patuld still pass. Add one test showing that a string option outsidethe list, e.g. filepicker_default: "/tmp/x", still falls back to its default.

ReimarBauer (changes requested):
see comment

